Issue #84, opened 2026-04-23
[feat] Cycle generation endpoints — port legacy start-cycle and update-training-maxes to NestJS API


brownm09:
## Overview

Port the deterministic cycle generation logic from the legacy Google Sheets app to `apps/api`. The core pure functions already exist in `packages/core`; this issue wires them into NestJS endpoints with proper port interfaces and Sheets adapter implementations.

## Background

The legacy app exposes two user-facing actions:
- **Start New Cycle** — computes the next cycle dashboard, updates training maxes from the completed workout, generates a new workout plan, and persists everything
- **Update Training Maxes** — re-runs the max update logic against recent lift records without starting a new cycle

The NestJS API has `ICycleDashboardRepository` but is missing port interfaces and adapters for program specs and training maxes, and has no service or endpoint layer for cycle generation.

## Acceptance Criteria

- [ ] `ILiftingProgramSpecRepository` port interface defined in `apps/api/src/ports/`
- [ ] `ITrainingMaxRepository` port interface defined in `apps/api/src/ports/`
- [ ] Google Sheets adapter implementations for both ports (mirrors legacy `LiftingProgramSpecRepository` and `TrainingMaxRepository`)
- [ ] `POST /programs/:program/cycles` — starts a new cycle: runs `updateCycle` → `updateMaxes` → persists updated dashboard and training maxes; returns new `CycleDashboardResponse`
- [ ] `POST /programs/:program/training-maxes/recalculate` — re-runs `updateMaxes` against recent lift records and persists; returns updated `TrainingMaxResponse[]`
- [ ] No dependency on `createGridV2` — the NestJS API returns structured JSON, not a `SpreadsheetCell[][]` grid
- [ ] Unit tests for the NestJS service layer (port mocks acceptable here)
- [ ] Integration test for at least one full cycle start turn

## Key Files

| Component | Path |
|---|---|
| Cycle update logic | `packages/core/src/services/dashboard/updateCycle.ts` |
| Training max update | `packages/core/src/services/maxes/updateMaxes.ts` |
| Existing cycle port | `apps/api/src/ports/ICycleDashboardRepository.ts` |
| Legacy program spec repo | `src/api/repositories/LiftingProgramSpecRepository.ts` |
| Legacy training max repo | `src/api/repositories/TrainingMaxRepository.ts` |
| Legacy start cycle action | `src/api/controllers/actions/StartNewCycleAction.ts` |

## Out of Scope

[...]

PR #86, opened 2026-04-23
[feat] Core read endpoints with in-memory adapters

brownm09:
## Summary

- Implements the five GET endpoints under `/programs/:program` (`cycles/current`, `workouts/:workoutNum`, `training-maxes`, `lift-records`, `spec`) wired through port-token DI to in-memory adapters.
- Seeds in-memory adapters with one cycle of fixture data — sufficient to render the J1/J2 "today's workout" UI path end-to-end. Real Google Sheets adapters and per-user auth land in a follow-up alongside `IRepositoryFactory` wiring.
- Adds `LiftingProgramSpecResponse` to `@lifting-logbook/types` (referenced by the issue but missing).

## Scope decision

The original AC asked for Google Sheets adapter implementations. The legacy `src/api/repositories/` adapters use Google Apps Script's `SpreadsheetApp` global and won't run in Node — building real adapters would require adding `googleapis`, OAuth credentials, per-user spreadsheet ID storage, and an auth provider, none of which exist yet. That's pulled out into a follow-up so this PR can stay focused on the API surface, controller wiring, and DI seam (which is what unblocks #84 and the first web client iteration).

## Acceptance criteria

- [x] All five endpoints implemented in `apps/api`
- [x] Each endpoint backed by its corresponding port interface — controllers `@Inject` port tokens, no direct adapter imports
- [ ] Google Sheets adapter implementations — **deferred to follow-up** (in-memory adapters used instead; see scope decision)
- [x] Unit tests for each controller (port mocks)
- [x] Integration test exercising all five endpoints + a 404 path through Fastify (`programs.e2e.spec.ts`)

## Test plan

- [ ] `cd apps/api && npm test` — all 14 tests pass
- [ ] `npm run build` from repo root — full Turbo build succeeds
- [ ] Hit each endpoint locally against `npm run dev --workspace @lifting-logbook/api` and verify the JSON shape matches the contracts in `packages/types/src/api.ts`

## Follow-ups (separate issues)

- Real Google Sheets adapters + `IRepositoryFactory.forUser` wiring + auth provider
- `CycleDashboardResponse.weeks[]` composition (currently empty — needs a core use case combining dashboard + workouts + records)
- Remove the `@src/core` path alias inside `packages/core` so the workaround in `apps/api/jest.config.js` and `tsconfig.json` (`isolatedModules`) can be dropped

### DIFF
--- a/apps/api/jest.config.js
+++ b/apps/api/jest.config.js
@@ -7,4 +7,10 @@ module.exports = {
     '^.+\\.tsx?$': ['ts-jest', { tsconfig: '<rootDir>/tsconfig.json' }],
   },
   transformIgnorePatterns: ['/node_modules/'],
+  moduleNameMapper: {
+    // @src/core is a pre-existing path alias inside @lifting-logbook/core sources.
+    // Map it so jest can resolve transitive imports when a test loads core.
+    '^@src/core$': '<rootDir>/../../packages/core/src/index.ts',
+    '^@src/core/(.*)$': '<rootDir>/../../packages/core/src/$1',
+  },
 };
--- a/apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts
+++ b/apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts
@@ -1,0 +1,25 @@
+import { Injectable, NotFoundException } from '@nestjs/common';
+import { CycleDashboard } from '@lifting-logbook/core';
+import { ICycleDashboardRepository } from '../../ports/ICycleDashboardRepository';
+import { SEED_PROGRAM, seedCycleDashboard } from './fixtures';
+
+@Injectable()
+export class InMemoryCycleDashboardRepository
+  implements ICycleDashboardRepository
+{
+  private dashboards = new Map<string, CycleDashboard>([
+    [SEED_PROGRAM, seedCycleDashboard()],
+  ]);
+
+  async getCycleDashboard(program: string): Promise<CycleDashboard> {
+    const dashboard = this.dashboards.get(program);
+    if (!dashboard) {
+      throw new NotFoundException(`Program '${program}' not found`);
+    }
+    return dashboard;
+  }
+
+  async saveCycleDashboard(dashboard: CycleDashboard): Promise<void> {
+    this.dashboards.set(dashboard.program, dashboard);
+  }
+}
--- a/apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts
+++ b/apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts
@@ -1,6 +1,7 @@
-import { Injectable, NotFoundException } from '@nestjs/common';
+import { Injectable } from '@nestjs/common';
 import { CycleDashboard } from '@lifting-logbook/core';
 import { ICycleDashboardRepository } from '../../ports/ICycleDashboardRepository';
+import { ProgramNotFoundError } from '../../ports/errors';
 import { SEED_PROGRAM, seedCycleDashboard } from './fixtures';
 
 @Injectable()
@@ -14,7 +15,7 @@ export class InMemoryCycleDashboardRepository
   async getCycleDashboard(program: string): Promise<CycleDashboard> {
     const dashboard = this.dashboards.get(program);
     if (!dashboard) {
-      throw new NotFoundException(`Program '${program}' not found`);
+      throw new ProgramNotFoundError(program);
     }
     return dashboard;
   }
--- a/apps/api/src/adapters/in-memory/fixtures.ts
+++ b/apps/api/src/adapters/in-memory/fixtures.ts
@@ -1,0 +1,94 @@
+import {
+  CycleDashboard,
+  LiftRecord,
+  LiftingProgramSpec,
+  TrainingMax,
+  Weekday,
+} from '@lifting-logbook/core';
+
+/**
+ * Seed data for the in-memory adapters used in v0.2 to wire the API end-to-end
+ * before real persistence (Google Sheets adapters + auth) lands. One program,
+ * one cycle, enough records to render the "today's workout" UI path.
+ */
+
+export const SEED_PROGRAM = '5-3-1';
+
+export const seedCycleDashboard = (): CycleDashboard => ({
+  program: SEED_PROGRAM,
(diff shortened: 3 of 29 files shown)


brownm09:
## Automated review

### Summary
Merge with revisions. The shape of the PR is right — port-token DI, in-memory seam, scope decision well-argued — but five blocking findings cluster around the seam itself (mapper helpers and the adapter↔HTTP coupling). Highest-severity: `apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts:14` throws `NotFoundException` from `@nestjs/common`, which couples the port adapter to the HTTP framework and undermines the hexagonal seam this PR is meant to establish. Fixing these now is much cheaper than fixing them after they're copied into the Sheets adapter.

---

### Blocking Findings

**[correctness]** `apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts:14` (and `workout.adapter.ts:21`)
Adapters throw `NotFoundException` from `@nestjs/common`.
The whole point of this PR is the port/adapter seam — but `@nestjs/common` is an HTTP-framework dependency. Whatever shape these adapters take will be copied into the Sheets adapter, the queue worker, and any test harness, and they'll all carry NestJS as a dependency they shouldn't have.
**Fix:** Define a domain error (e.g. `ProgramNotFoundError` in `packages/core` or a small `apps/api/src/ports/errors.ts`), throw that from the adapter, and translate to `NotFoundException` in the controller or via an exception filter.

**[correctness]** `apps/api/src/programs/programs.module.ts:32`
All five repositories are registered as Nest singletons (default scope) and the `InMemory*` adapters hold mutable `Map` state.
The PR commits to the seam being sound for when real adapters land. Today's wiring is fine — single tenant, no auth — but swapping `useClass` for a per-user Sheets adapter without changing scope will silently leak one user's data into another's request.
**Fix:** Either set `scope: Scope.REQUEST` on the providers now, or add a skipped test asserting per-request isolation that will force the wiring decision when auth lands.

**[correctness]** `apps/api/src/programs/mappers.ts:13`
`isoDate` formats via `d.toISOString().slice(0, 10)` — UTC.
Seed data uses `T00:00:00.000Z` so tests pass, but Sheets cells parsed in server-local TZ will silently shift days for users west of UTC. This helper is the single funnel for every date in the API contract.
**Fix:** Either accept dates as `YYYY-MM-DD` strings from the domain layer (no `Date` objects in the contract) or document and standardize on UTC with an explicit `Intl.DateTimeFormat`. Lock the contract before any client depends on it.

**[correctness]** `apps/api/src/programs/mappers.ts:75`
`week` derived as `Math.ceil(workoutNum / 2)` clamped to `[1, 4]`.
The clamp silently maps `workoutNum >= 9` to week 4, returning a fabricated value behind a typed contract. Real programs vary workouts-per-week.
**Fix:** Reject out-of-range `workoutNum` with `BadRequestException`, or source the week from the program spec / a core use case. Don't fabricate.

**[correctness]** `apps/api/src/programs/mappers.ts:24` + `lift-records.controller.ts:28`
`toLiftRecordResponse(r, index)` builds `id` from a positional `index` that the controller passes implicitly via `Array.prototype.map`.
The `id` then changes if the underlying ordering or filter changes, breaking client-side keying/caching (React lists, dedup). The implicit-`map`-index dependency is also a footgun for the next caller.
**Fix:** Drop `index` from `id`; the natural composite `${program}-${cycleNum}-${workoutNum}-${lift}-${setNum}` is unique per the domain's set semantics. Drop the second parameter from the function signature too.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/jest.config.js:11-14` and `apps/api/tsconfig.json:8`
`moduleNameMapper` for `@src/core` and `isolatedModules: true` paper over a path alias inside `packages/core` that breaks ts-jest.
The workaround couples `apps/api` test config to `packages/core`'s internal layout — renaming `src/` or removing the alias upstream silently breaks api tests.
**Suggestion:** File a follow-up to remove the `@src/core` self-alias inside `packages/core` (use relative imports), and reference its issue number in the workaround's comment so the exit ramp is visible.

**[reliability]** `apps/api/src/programs/workouts.controller.ts:30-37` and `lift-records.controller.ts:24-29`
`dashboard.cycleNum` is consumed without bounds-checking.
If a future dashboard adapter returns `0` or negative, you get a 404 from a missing workoutKey lookup instead of a clearer 5xx. Two controllers already share this pattern.
**Suggestion:** Single helper on the controller side (or invariant-check inside the dashboard adapter) so the assumption is enforced once.

---

### Questions for Author

- `CycleDashboardResponse.weeks: []` is intentional and called out as a follow-up — but the controller returns it on every request. Worth a `Deprecated`/`x-incomplete` header or response-side comment so the web client doesn't wire to an empty array as if it were authoritative?
- `LiftingProgramSpecResponse` was added to `packages/types` without GraphQL/SDL counterparts — is that deferred too, or expected to remain REST-only?

brownm09:
Addressed the five blocking findings in 50f7ac6:

1. **Adapters no longer throw `NotFoundException`** — new `ProgramNotFoundError` / `WorkoutNotFoundError` in `apps/api/src/ports/errors.ts`; `DomainNotFoundFilter` (registered globally in `main.ts` and in the e2e setup) translates to HTTP 404. Adapters keep only `@Injectable()` from `@nestjs/common`.
2. **Module scope forcing function** — added a skipped e2e test `isolates adapter state per request` and pointed the `ProgramsModule` doc at it. Opted for the skipped-test route over `Scope.REQUEST` now to avoid tearing down seed state per request before auth lands.
3. **`isoDate` UTC contract** — documented in `mappers.ts` that API dates are emitted as UTC `YYYY-MM-DD`; adapters own TZ normalization at their boundary.
4. **`week` clamp removed** — `MAX_WORKOUT_NUM = 8` and `isValidWorkoutNum` exported from the mapper; workouts controller rejects out-of-range `workoutNum` with `BadRequestException` instead of silently mapping to week 4.
5. **`toLiftRecordResponse` drops `index`** — id is now the natural composite `${program}-${cycleNum}-${workoutNum}-${lift}-${setNum}`; `records.map(toLiftRecordResponse)` still works since extra args are ignored.

Non-blocking findings and open questions deferred — happy to file follow-ups if you want separate tickets for the `@src/core` self-alias or the `CycleDashboardResponse.weeks` incomplete-response signal.

PR #87, opened 2026-04-25
[feat] Cycle generation write endpoints

brownm09:
## Summary

- Adds `CycleGenerationService` that orchestrates `updateCycle` + `updateMaxes` via the four existing port interfaces (`ICycleDashboardRepository`, `ILiftingProgramSpecRepository`, `ITrainingMaxRepository`, `ILiftRecordRepository`)
- Adds `CycleGenerationController` with two POST endpoints:
  - `POST /programs/:program/cycles` (201) — advances cycle counter, updates training maxes from current-cycle lift records, persists both, returns `CycleDashboardResponse`
  - `POST /programs/:program/training-maxes/recalculate` (200) — re-runs max calculation without advancing cycle, returns `TrainingMaxResponse[]`
- Registers controller + service in `ProgramsModule`; no new adapters needed (in-memory adapters already implement all required ports)

## Acceptance Criteria

- [x] `ILiftingProgramSpecRepository` port — already existed, no changes needed
- [x] `ITrainingMaxRepository` port — already existed, no changes needed
- [x] `POST /programs/:program/cycles` endpoint
- [x] `POST /programs/:program/training-maxes/recalculate` endpoint
- [x] No dependency on `createGridV2`
- [x] Unit tests for service layer (6 cases across `startNewCycle` and `recalculateMaxes`)
- [x] Integration test for full cycle start turn (e2e spec)

## Test plan

- [ ] `npm run test --workspace=apps/api` — all 23 tests pass (1 pre-existing skipped)
- [ ] `POST /programs/5-3-1/cycles` returns 201 with `cycleNum: 2` and advances persisted state
- [ ] `GET /programs/5-3-1/cycles/current` after POST returns `cycleNum: 2`
- [ ] `POST /programs/5-3-1/training-maxes/recalculate` returns 200 with full maxes array
- [ ] `POST /programs/unknown/cycles` returns 404

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -15,6 +15,8 @@
     "@nestjs/common": "^10.0.0",
     "@nestjs/core": "^10.0.0",
     "@nestjs/platform-fastify": "^10.0.0",
+    "class-transformer": "^0.5.1",
+    "class-validator": "^0.15.1",
     "fastify": "^4.0.0",
     "reflect-metadata": "^0.2.0",
     "rxjs": "^7.8.0"
--- a/apps/api/src/adapters/in-memory/fixtures.ts
+++ b/apps/api/src/adapters/in-memory/fixtures.ts
@@ -24,10 +24,10 @@ export const seedCycleDashboard = (): CycleDashboard => ({
 });
 
 export const seedTrainingMaxes = (): TrainingMax[] => [
-  { lift: 'Squat', weight: 315, dateUpdated: new Date('2026-04-20T00:00:00.000Z') },
-  { lift: 'Bench Press', weight: 225, dateUpdated: new Date('2026-04-20T00:00:00.000Z') },
-  { lift: 'Deadlift', weight: 405, dateUpdated: new Date('2026-04-20T00:00:00.000Z') },
-  { lift: 'Overhead Press', weight: 145, dateUpdated: new Date('2026-04-20T00:00:00.000Z') },
+  { lift: 'Squat', weight: 315, dateUpdated: new Date('2026-04-13T00:00:00.000Z') },
+  { lift: 'Bench Press', weight: 225, dateUpdated: new Date('2026-04-13T00:00:00.000Z') },
+  { lift: 'Deadlift', weight: 405, dateUpdated: new Date('2026-04-13T00:00:00.000Z') },
+  { lift: 'Overhead Press', weight: 145, dateUpdated: new Date('2026-04-13T00:00:00.000Z') },
 ];
 
 export const seedLiftRecords = (): LiftRecord[] => [
--- a/apps/api/src/main.ts
+++ b/apps/api/src/main.ts
@@ -1,5 +1,6 @@
 import 'reflect-metadata';
 import { NestFactory } from '@nestjs/core';
+import { ValidationPipe } from '@nestjs/common';
 import { FastifyAdapter, NestFastifyApplication } from '@nestjs/platform-fastify';
 import { AppModule } from './app.module';
 import { DomainNotFoundFilter } from './programs/not-found.filter';
@@ -10,6 +11,7 @@ async function bootstrap() {
     new FastifyAdapter(),
   );
   app.useGlobalFilters(new DomainNotFoundFilter());
+  app.useGlobalPipes(new ValidationPipe({ whitelist: true }));
   await app.listen(3000, '0.0.0.0');
 }
 
--- a/apps/api/src/programs/cycle-generation.controller.spec.ts
+++ b/apps/api/src/programs/cycle-generation.controller.spec.ts
@@ -1,0 +1,91 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { Weekday } from '@lifting-logbook/core';
+import { CycleGenerationController } from './cycle-generation.controller';
+import { CycleGenerationService } from './cycle-generation.service';
+
+const PROGRAM = '5-3-1';
+
+const stubCycleDashboard = () => ({
+  program: PROGRAM,
+  cycleUnit: 'week' as const,
+  cycleNum: 2,
+  cycleDate: new Date('2026-04-27T00:00:00.000Z'),
+  sheetName: '5-3-1_Cycle_2_20260427',
+  cycleStartWeekday: Weekday.Monday,
+});
+
+describe('CycleGenerationController', () => {
+  let controller: CycleGenerationController;
+  let service: jest.Mocked<Pick<CycleGenerationService, 'startNewCycle' | 'recalculateMaxes'>>;
+
+  beforeEach(async () => {
+    service = {
+      startNewCycle: jest.fn(),
+      recalculateMaxes: jest.fn(),
+    };
+
+    const module: TestingModule = await Test.createTestingModule({
+      controllers: [CycleGenerationController],
+      providers: [{ provide: CycleGenerationService, useValue: service }],
+    }).compile();
+
+    controller = module.get(CycleGenerationController);
+  });
+
+  describe('startNewCycle', () => {
+    it('calls service with program and dto, returns mapped response', async () => {
+      service.startNewCycle.mockResolvedValue(stubCycleDashboard());
+
+      const result = await controller.startNewCycle(PROGRAM, {});
+
+      expect(service.startNewCycle).toHaveBeenCalledWith(PROGRAM, {});
+      expect(result).toEqual({
+        program: PROGRAM,
+        cycleNum: 2,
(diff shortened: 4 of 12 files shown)


brownm09:
## Review: [feat] Cycle generation write endpoints

### Summary

Ready to merge with one reliability concern addressed first. The service orchestration is clean — thin controller, pure core functions untouched, all four ports wired correctly. The main issue is that `startNewCycle` uses `Promise.all` for the two final writes, which can leave the cycle dashboard and training maxes in divergent state if one write fails after the other succeeds. With in-memory adapters this is academic, but the pattern established here is what will be copied when Sheets adapters land. Everything else is minor.

---

### Blocking Findings

**[reliability]** `cycle-generation.service.ts:47-50`

`startNewCycle` fires `saveCycleDashboard` and `saveTrainingMaxes` in parallel via `Promise.all`. If `saveCycleDashboard` resolves but `saveTrainingMaxes` rejects, the cycle counter is permanently incremented with stale training maxes. The caller receives a 500, may retry, and advances the cycle counter a second time — double-advancement with mismatched maxes. The inverse failure leaves correct maxes but the dashboard still on the old cycle.

The in-memory adapters can't fail today, so this is currently harmless, but the `Promise.all` pattern is what will be carried forward when Sheets adapters land (issue #84 is explicitly "port legacy logic to NestJS"). Sequential writes (`await save1; await save2`) bound the blast radius: the second write failing is recoverable; the first failing means nothing changed.

**Fix:** Write sequentially — `await this.cycleDashboardRepo.saveCycleDashboard(newCycle)` then `await this.trainingMaxRepo.saveTrainingMaxes(program, newMaxes)`. Add a comment noting that true atomicity requires a transaction primitive when real adapters land.

---

### Non-Blocking Findings

**[maintainability]** `cycle-generation.controller.ts:24`

`@HttpCode(HttpStatus.CREATED)` on `startNewCycle` is redundant — `@Post()` already returns 201 by default in NestJS. It adds a decorator that looks meaningful but does nothing, which will confuse a future reader who wonders what it was overriding.

**Suggestion:** Remove `@HttpCode(HttpStatus.CREATED)` from `startNewCycle`.

---

**[reliability]** `programs.e2e.spec.ts:74-121`

The write-operation tests rely silently on document order: the `startNewCycle` test advances cycleNum to 2 (and mutates adapter state), which means the `recalculateMaxes` test immediately after it runs against a cycle that has zero lift records — so it verifies shape but not that maxes actually change. If test order ever shifts (e.g., `--randomize`, a reorder during refactor), the existing GET test at line ~35 that asserts `cycleNum === 1` will fail.

**Suggestion:** Wrap the write tests in a `describe('write operations — order-sensitive, mutates adapter state', ...)` block and add a one-line comment on `recalculateMaxes` explaining that it runs on cycle 2 (no records) intentionally.

---

**[maintainability]** `apps/api/src/programs/` (missing file)

Every other controller in this directory has a paired `.spec.ts` (`cycle-dashboard.controller.spec.ts`, `training-maxes.controller.spec.ts`, etc.). `CycleGenerationController` doesn't have one. The controller is thin, but the missing spec breaks the established pattern and leaves the mapper call and 404 propagation path untested at the unit level.

**Suggestion:** Add `cycle-generation.controller.spec.ts` matching the pattern in `cycle-dashboard.controller.spec.ts`.

---

### Questions for Author

- `startNewCycle` fetches records via `getLiftRecords(program, dashboard.cycleNum)` — i.e., the cycle being *completed*. The legacy `StartNewCycleAction` used the active Sheets tab as source of truth, not a lift-record store. Is the in-memory lift record store always the canonical source here, or should this endpoint also accept an optional `cycleNum` override for cases where the caller wants to backfill?

- Is the `recalculateMaxes` e2e test coverage intentional (shape-only, no delta assertion)? Given that it runs after `startNewCycle` bumped the cycle to 2 with no records, it can't verify a max actually changed. If the intent is to test the full update path, a dedicated fixture setup (or running recalculate before start-cycle) would be needed.

brownm09:
## Review: [feat] Cycle generation write endpoints

### Summary
This PR is well-structured — the service cleanly orchestrates existing port interfaces, controller mapping is correct, and the test coverage is solid (6 unit cases + 6 e2e cases). It is close to mergeable but has two issues that should be addressed first: the DTO has no validation decorators (allowing corrupt data to reach the service), and the `program` path parameter is unvalidated at the API boundary. The write-ordering torn-state risk is acknowledged in a comment but leaves a silent compounding error path that warrants a fix before real adapters land.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/start-new-cycle.dto.ts:3-5`
`StartNewCycleDto` has no `class-validator` decorators on either field. NestJS `ValidationPipe` cannot enforce type constraints without them. A request body like `{"fromCycleNum": "abc", "cycleDate": "not-a-date"}` passes through — `fromCycleNum` arrives as the string `"abc"` (coerced to `NaN` in numeric comparisons), and `new Date("not-a-date")` produces an `Invalid Date` object that is passed to `updateCycle` and persisted without error.
**Fix:** Add `@IsOptional()`, `@IsInt()`, `@Min(1)` on `fromCycleNum` and `@IsOptional()`, `@IsDateString()` on `cycleDate`. Confirm `ValidationPipe` is globally registered or applied to this endpoint (it is not wired in the e2e module setup).

**[security]** `apps/api/src/programs/cycle-generation.controller.ts:27,40`
The `program` path parameter is passed unvalidated to all four repository calls, and it is also embedded verbatim in the persisted `sheetName` field (constructed by `updateCycle` as `${program}_Cycle_...`). When real Sheets adapters land, a caller supplying a path-traversal sequence or injection payload in `program` will have it baked into persisted data and file-system-adjacent operations.
**Fix:** Add a `@Param('program', new ParsePatternPipe(/^[a-zA-Z0-9_-]+$/))` or equivalent allowlist pipe so non-alphanumeric identifiers are rejected at the HTTP boundary before reaching any domain or persistence code.

**[reliability]** `apps/api/src/programs/cycle-generation.service.ts:78-79`
`saveCycleDashboard` is awaited first, then `saveTrainingMaxes`. If the second write fails (transient I/O on real adapters), the cycle counter has permanently advanced but training maxes are stale. The next `startNewCycle` call will compute maxes against a now-incorrect baseline, and the error compounds with each subsequent cycle.
**Fix:** Swap the order — save maxes first (if that fails, cycle counter has not advanced and retry is safe), then save the dashboard. Add a comment explaining the ordering rationale. True atomicity can wait for real adapters; the ordering alone reduces blast radius significantly.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/cycle-generation.service.ts:46-61`
Two separate `if (dto.fromCycleNum !== undefined)` blocks — one for the empty-records guard, one for the `minDate` reduce — are 15 lines apart. The reduce's safety depends on the guard having already thrown, but a future change relaxing the guard condition without updating the reduce would crash at `liftRecords[0]`. The invariant is invisible at the call site.
**Suggestion:** Merge the two blocks so the reduce sits inside the same `if` branch as the guard, making it structurally impossible to reach the reduce when `liftRecords` is empty.

**[maintainability]** `apps/api/src/programs/programs.e2e.spec.ts:83-162`
The "write operations" `describe` block relies on cumulative singleton adapter state across tests. The third test (`fromCycleNum: 1`) expects `cycleNum: 2` only because the prior test has already advanced the cycle — the assertion looks like it validates the `fromCycleNum` path in isolation but is actually validating re-idempotence. If the second test is skipped, the third test produces a misleading result.
**Suggestion:** Add a leading assertion in each stateful test that documents its expected pre-condition (e.g., assert the current `cycleNum` before acting), so cascading failures surface their actual cause rather than failing at the wrong assertion.

---

### Questions for Author

- Should `startNewCycle` without `fromCycleNum` also return a `400` when the current cycle has no lift records, or is a silent no-op save on an empty cycle intentional (allowing cycle headers to be advanced before any lifts are logged)?
- The e2e `fromCycleNum: 1` test (third test in "write operations") asserts `cycleNum: 2` when the stored cycle is already `2`. Is this confirming that providing an older `fromCycleNum` re-pins the cycle counter, or is the assertion there to confirm the cycle was advanced and the pre-condition just happens to produce the same result?

---

### Style Notes

`@Body() dto: StartNewCycleDto = {}` — the `= {}` default is never used; NestJS injects an empty DTO itself when the body is absent. Remove it to avoid implying the default participates in injection. The controller JSDoc blocks are clear and match the implementation. The service-level atomicity comment is a good acknowledgment — consider adding a `Logger.warn` between the two saves so that if they diverge in production, the inconsistent state is visible in logs before the second write completes.

---

PR #91, opened 2026-04-29
[fix] Break circular barrel imports to resolve Jest worker OOM (#89)

brownm09:
## Summary

- All source files within \`packages/core/src/\` were importing from the root \`@src/core\` barrel, which re-exports those same files — creating circular dependency chains that caused ts-jest workers to exhaust memory.
- Replaced every \`from \"@src/core\"\` inside \`src/\` with direct sub-path imports: \`@src/core/models\`, \`@src/core/constants\`, or \`@src/core/utils/jsUtil\`.
- Also fixes a pre-existing \`TS2532: Object is possibly 'undefined'\` in \`apps/api\` that had been failing CI since 2026-04-21 (tracked in #92).
- 18 files changed.

## Root cause (OOM)

\`packages/core/src/index.ts\` is a barrel that re-exports all sub-modules (\`catalog\`, \`constants\`, \`models\`, \`services\`, \`utils\`). Files within those sub-modules were themselves importing from \`@src/core\`, completing the cycle. ts-jest's module resolution entered a pathologically deep graph traversal and OOM'd the worker.

## Acceptance criteria

- [x] \`parseCycleDashboard.test.ts\` runs without OOM
- [x] Full core test suite (\`npm test -w @lifting-logbook/core\`) passes: **92 tests, 21 suites, all green**
- [x] No \`from \"@src/core\"\` barrel imports remain in \`packages/core/src/**\`
- [x] \`apps/api\` builds without TypeScript errors (fixes #92)

## Test plan

\`\`\`bash
npm test -w @lifting-logbook/core
# Expected: 92 passed, 21 suites, 0 failures
\`\`\`

Closes #89
Closes #92

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #89, opened 2026-04-28
fix: parseCycleDashboard Jest worker OOM in packages/core test suite


brownm09:
## Problem

Running `npm test` from the repo root causes the `packages/core` Jest worker to crash with:

```
Jest worker ran out of memory and crashed
```

The failing test is `tests/core/utils/parser/parseCycleDashboard.test.ts`, introduced in #53 (lift-and-shift migration from GAS repo). The test suite runs `jest` with no `--maxWorkers` or heap-size configuration, so parallel worker memory is unbounded.

## Workaround (documented in CLAUDE.md)

```bash
npm test -w @lifting-logbook/core -- --runInBand
```

## Acceptance criteria

- [ ] `npm test` completes without OOM error from repo root
- [ ] Fix does not materially slow down the core test suite in CI
- [ ] `jest.config.base.js` or `packages/core/jest.config.js` updated as needed (e.g., `--maxWorkers`, `NODE_OPTIONS`, or test refactor to reduce peak allocation)

PR #94, opened 2026-04-29
[fix] Ban @src/core root barrel import inside packages/core

### LINKED ISSUE #93: [fix] Guard against circular barrel imports in packages/core
## Problem

PR #91 fixed a circular barrel import that caused Jest worker OOM errors. The fix resolved the issue by restructuring imports to use sub-path specifiers instead of the root barrel (`@src/core`). Without a linting rule, this pattern could regress.

## Goal

Add an ESLint `no-restricted-imports` rule to `packages/core/eslint.config.js` that bans `from "@src/core"` inside `packages/core/src/**`. The rule should direct developers to use `@src/core/models`, `@src/core/constants`, or `@src/core/utils/jsUtil` instead.

## Acceptance Criteria

- [ ] `packages/core/eslint.config.js` contains a `no-restricted-imports` pattern matching `@src/core` (exact) with a descriptive error message
- [ ] `npm run lint -w @lifting-logbook/core` passes with no violations
- [ ] The rule is scoped to `packages/core/src/**` files

## References

- Circular import incident: #89, fixed in #91
- ESLint `no-restricted-imports`: https://eslint.org/docs/latest/rules/no-restricted-imports

brownm09:
## Summary

- Adds a `no-restricted-imports` `paths` rule to `packages/core/eslint.config.js` that bans the exact specifier `@src/core` inside `packages/core/src`
- Uses `paths` (exact string match) rather than `patterns` (glob) so sub-path imports like `@src/core/models`, `@src/core/constants`, and `@src/core/utils/jsUtil` remain allowed and unaffected
- Error message directs developers to the correct sub-path alternatives

## Motivation

PR #91 (issue #89) fixed a circular barrel import in `updateCycle.ts` that caused Jest worker OOM failures. Without a lint rule, the same pattern could be reintroduced inadvertently.

## Acceptance Criteria (from #93)

- [x] `packages/core/eslint.config.js` contains a `no-restricted-imports` `paths` entry for `@src/core` with a descriptive error message
- [x] `npm run lint -w @lifting-logbook/core` passes with no violations
- [x] Sub-path imports (`@src/core/models`, etc.) are unaffected — confirmed by lint pass across all 36 existing files using those paths

## Test Plan

- [x] `npm run lint -w @lifting-logbook/core` — passes (0 errors)
- [x] Pre-commit hook ran lint across all 4 other packages — all cached green

Closes #93

### DIFF
--- a/packages/core/eslint.config.js
+++ b/packages/core/eslint.config.js
@@ -12,6 +12,19 @@ module.exports = [
       // workspaces or infrastructure packages. Violations mean the dependency
       // rule is inverted; use a port interface instead.
       'no-restricted-imports': ['error', {
+        // Exact-match ban: prevents regression of the circular barrel import
+        // that caused Jest worker OOM (issue #89, fixed in PR #91).
+        // `paths` matches the literal specifier only — sub-paths like
+        // @src/core/models are unaffected.
+        paths: [
+          {
+            name: '@src/core',
+            message:
+              "Import from the root barrel '@src/core' is not allowed inside " +
+              "packages/core/src — use '@src/core/models', '@src/core/constants', " +
+              "or '@src/core/utils/jsUtil' instead.",
+          },
+        ],
         patterns: [
           {
             group: [
--- a/packages/core/eslint.config.js
+++ b/packages/core/eslint.config.js
@@ -21,8 +21,17 @@ module.exports = [
             name: '@src/core',
             message:
               "Import from the root barrel '@src/core' is not allowed inside " +
-              "packages/core/src — use '@src/core/models', '@src/core/constants', " +
-              "or '@src/core/utils/jsUtil' instead.",
+              "packages/core/src — use a direct sub-path such as '@src/core/models', " +
+              "'@src/core/constants', or '@src/core/utils/jsUtil' instead.",
+          },
+          // '@src/core/*' in tsconfig maps @src/core/index → src/index.ts, so
+          // the explicit-index form resolves to the same barrel and must also be banned.
+          {
+            name: '@src/core/index',
+            message:
+              "Import from the root barrel '@src/core/index' is not allowed inside " +
+              "packages/core/src — use a direct sub-path such as '@src/core/models', " +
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [fix] Ban @src/core root barrel import inside packages/core

### Summary

Ready to merge. The change is minimal, correctly reasoned, and well-tested. Using `paths` (exact literal match) rather than `patterns` (glob) is the right call — the PR description explains this distinction clearly, which is exactly the kind of decision trace that prevents the same mistake recurring. One question worth answering before merge: whether `@src/core/index` can be imported to bypass the rule, depending on how the tsconfig path aliases are configured. No blocking findings.

---

### Questions for Author

- **`@src/core/index` bypass?** If the tsconfig `paths` map includes `"@src/core/index"` (or `"@src/core/*"` resolves `index` to the barrel), a developer could write `import { x } from '@src/core/index'` and the lint rule wouldn't fire. Is the alias defined in a way that makes this path resolvable? If so, worth adding a second `paths` entry for `'@src/core/index'`.

- **Error message exhaustiveness:** The error message directs developers to exactly three sub-paths (`@src/core/models`, `@src/core/constants`, `@src/core/utils/jsUtil`). As the package grows and new sub-paths are added (e.g., `@src/core/ports`), this message will silently go stale. Would it be better to link to a sub-path listing in a docs file, or is "most common imports" the intent?

---

### Non-Blocking Findings

**[maintainability]** `packages/core/eslint.config.js:21`
The error message hard-codes three specific sub-paths as the canonical alternatives.
As `packages/core` grows, new valid sub-paths (e.g., `@src/core/ports`, `@src/core/services`) will exist but won't appear in the message. A developer importing from a new sub-path may be confused if the error message doesn't list what they need.
**Suggestion:** Close the message with "or any direct sub-path" / "see packages/core/src for available sub-paths" to future-proof against package growth, rather than implying the three listed are exhaustive.

---

### Style Notes

Comment at lines 14–17 is well-placed and explains the `paths`/`patterns` distinction precisely — this is valuable institutional memory and should stay exactly as written. No style concerns.

---

PR #117, opened 2026-04-30
[feat] Configurable week grouping — remove 4-week / 2-workout hardcoding (#116)

### LINKED ISSUE #116: [feat] Make week grouping configurable — remove 2-workouts/week and 4-week cycle assumptions
## Problem

The week-grouping logic for the cycle dashboard is hardcoded to assume 2 workouts per training week and a 4-week cycle (8 workouts total). This is baked in at three levels:

- **`packages/types/src/domain.ts`** — \`WeekNumber = 1 | 2 | 3 | 4\` is a union literal type; any cycle length other than 4 weeks is a compile-time type error
- **`apps/api/src/programs/mappers.ts`** — \`MAX_WORKOUT_NUM = 8\` rejects any \`workoutNum > 8\` at the API boundary; \`Math.ceil(workoutNum / 2)\` derives week number assuming 2 workouts/week
- **`apps/web/app/cycle/[cycleNum]/page.tsx`** — same formula; week list derived from the formula result rather than any program config

The program spec CSV has an \`offset\` column (day offset from cycle start) but no field to express week structure. Week grouping is entirely implicit.

## Impact

Programs with a different cadence (e.g. Reverse Pyramid Training / Leangains — 3 days/week; a 3-week wave program; a custom 1-day/week program) will either be rejected by the API validator or grouped into the wrong weeks in the UI.

## Proposed Solution

Add a `week` column (or a program-level `workoutsPerWeek` config value) to the program spec so week grouping is declared rather than derived. Concrete changes:

1. Add `week` field to `LiftingProgramSpecResponse` (and the CSV schema) — each lift entry declares which week of the cycle it belongs to
2. Remove `MAX_WORKOUT_NUM` hard limit from the API controller (or make it derived from the spec)
3. Change `WeekNumber` to `number` in `packages/types/src/domain.ts`
4. Replace `Math.ceil(workoutNum / 2)` in both the API mapper and the web page with a lookup from the spec's `week` field
5. Update pre-made program spec CSVs (531, RPT) with the new `week` column

## References

- Discovered during review of #104 (cycle dashboard screen)
- `apps/api/src/programs/mappers.ts` already has a comment: *"Real programs vary — when that becomes relevant, source week from the program spec instead of deriving."*

brownm09:
## Summary

- Adds a `Week` column to the program spec CSV so week membership is declared in data, not derived at runtime
- Removes `WeekNumber = 1|2|3|4` type constraint — week is now `number`
- Removes `MAX_WORKOUT_NUM = 8` API limit — workoutNum is validated as any positive integer
- Replaces `Math.ceil(workoutNum / 2)` in the API mapper and web dashboard page with a lookup from the spec's `week` field

## Acceptance Criteria

- [x] `WeekNumber` is `number` (not a `1|2|3|4` union)
- [x] `MAX_WORKOUT_NUM = 8` is removed from the API validator
- [x] `LiftingProgramSpecResponse` includes `week: number`
- [x] Program spec CSVs have a `Week` column
- [x] `Math.ceil(workoutNum / 2)` is eliminated from both `mappers.ts` and `page.tsx`
- [x] `WorkoutsController` sources week from spec instead of deriving it
- [x] Week grouping in the cycle dashboard page uses `w.week` from spec

## Test plan

- All 9 API test suites pass (33 tests + 1 skipped)
- All 10 web workout-plan tests pass including new `sources week from the first lift in each offset group` test
- `tsconfig.spec.json` path fix ensures ts-jest resolves `@lifting-logbook/types` from source (consistent with `moduleNameMapper`)

Closes #116

### DIFF
--- a/apps/api/src/adapters/in-memory/fixtures.ts
+++ b/apps/api/src/adapters/in-memory/fixtures.ts
@@ -68,6 +68,7 @@ export const seedLiftRecords = (): LiftRecord[] => [
 
 export const seedProgramSpec = (): LiftingProgramSpec[] => [
   {
+    week: 1,
     offset: 0,
     lift: 'Squat',
     increment: 5,
@@ -80,6 +81,7 @@ export const seedProgramSpec = (): LiftingProgramSpec[] => [
     activation: 'compound',
   },
   {
+    week: 1,
     offset: 0,
     lift: 'Bench Press',
     increment: 5,
--- a/apps/api/src/programs/mappers.ts
+++ b/apps/api/src/programs/mappers.ts
@@ -10,7 +10,6 @@ import {
   LiftingProgramSpecResponse,
   SetResponse,
   TrainingMaxResponse,
-  WeekNumber,
   WorkoutLiftResponse,
   WorkoutResponse,
 } from '@lifting-logbook/types';
@@ -43,6 +42,7 @@ export const toLiftRecordResponse = (r: LiftRecord): LiftRecordResponse => ({
 export const toLiftingProgramSpecResponse = (
   s: LiftingProgramSpec,
 ): LiftingProgramSpecResponse => ({
+  week: s.week,
   lift: s.lift,
   order: s.order,
   offset: s.offset,
@@ -70,23 +70,19 @@ export const toCycleDashboardResponse = (
   weeks: [],
 });
 
-// Largest workoutNum whose derived week still fits in WeekNumber (1..4)
-// assuming 2 workouts per training week. Real programs vary — when that
-// becomes relevant, source week from the program spec instead of deriving.
-export const MAX_WORKOUT_NUM = 8 as const;
-
 export const isValidWorkoutNum = (n: number): boolean =>
-  Number.isInteger(n) && n >= 1 && n <= MAX_WORKOUT_NUM;
+  Number.isInteger(n) && n >= 1;
 
 /**
  * Groups a workout's lift records into the WorkoutResponse shape.
- * Assumes 2 workouts per training week; caller must validate `workoutNum`
- * with `isValidWorkoutNum` before invoking.
+ * Caller must validate `workoutNum` with `isValidWorkoutNum` and supply
+ * the `week` sourced from the program spec before invoking.
  */
 export const toWorkoutResponse = (
   program: string,
   cycleNum: number,
   workoutNum: number,
+  week: number,
   records: LiftRecord[],
 ): WorkoutResponse => {
   const liftMap = new Map<string, SetResponse[]>();
@@ -104,7 +100,6 @@ export const toWorkoutResponse = (
   const lifts: WorkoutLiftResponse[] = Array.from(liftMap.entries()).map(
     ([lift, sets]) => ({ lift, sets }),
   );
-  const week = Math.ceil(workoutNum / 2) as WeekNumber;
   const date = records[0] ? isoDate(records[0].date) : isoDate(new Date());
   return { program, cycleNum, workoutNum, week, date, lifts };
 };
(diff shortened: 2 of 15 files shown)


brownm09:
## Review: [feat] Configurable week grouping — remove 4-week / 2-workout hardcoding (#116)

### Summary

This PR is close to merge-ready. The architectural change is correct and well-scoped: all three hardcoding sites are cleanly eliminated and the data flows consistently from CSV → domain model → API response → web. One blocking correctness issue: removing `MAX_WORKOUT_NUM` without replacing it with a spec-length guard means out-of-range `workoutNum` values now return 200 with `week: 1` instead of 400. This is a behavior regression that needs a fix and a test before merge.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/workouts.controller.ts:55–62`

When `workoutNum` exceeds the number of offset groups in the spec (e.g., spec has 3 workout days, caller requests workoutNum=5), `offsetForWorkout` is `undefined` and the code silently falls back to `week = 1`, then calls `getWorkout` and returns a 200 response. Previously `MAX_WORKOUT_NUM = 8` with `isValidWorkoutNum` caught this at the front door and returned 400. Now `isValidWorkoutNum` only checks `n >= 1`, so the upper bound is gone with no replacement. A well-formed caller won't hit this, but an out-of-range request now gets a plausible-looking successful response with wrong week data instead of an error.

**Fix:** After computing `offsetForWorkout`, throw if it's undefined:
```typescript
if (offsetForWorkout === undefined) {
  throw new BadRequestException(
    `workoutNum ${workoutNum} exceeds program spec (${offsetsSorted.length} workout days)`,
  );
}
const week = spec.find((s) => s.offset === offsetForWorkout)?.week ?? 1;
```

**[correctness]** `apps/api/src/programs/workouts.controller.spec.ts`

No test covers the case where `workoutNum` is a valid positive integer but exceeds the number of offset groups. The removal of `MAX_WORKOUT_NUM` is a behavioral change to the validation contract — the spec should verify that the replacement validation (spec-length guard above) fires correctly. Without it, the fix above could regress silently.

**Fix:** Add a test where specRepo returns a 1-offset spec and workoutNum=2 is requested — assert `BadRequestException` is thrown.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/workouts.controller.ts:50–62`

The offset-grouping logic (Set dedup → sort ascending → index by workoutNum) mirrors the same algorithm in `buildWorkoutDays` on the web. If the mapping strategy changes (e.g., week conflicts between lifts at the same offset, or a different tie-breaking rule), both sites need updating independently.

**Suggestion:** Extract a `weekForWorkoutNum(spec: LiftingProgramSpec[], workoutNum: number): number` pure function in `mappers.ts`. This co-locates the lookup with `toWorkoutResponse`, keeps the controller thin, and creates a single testable unit for the logic.

---

### Questions for Author

- The `?? 1` fallback in both `buildWorkoutDays` (`lifts[0]?.week ?? 1`) and the controller (`?? 1` when `offsetForWorkout` is defined but no spec entry matches) treats week=1 as a safe default. Is this intentional policy ("unknown week = first week"), or is it a defensive pattern that should actually be an error? If the spec is always well-formed, these paths are unreachable — but the fallback hides malformed specs silently.
- `tests/fixtures/rpt_program_spec.csv` at the repo root appears to duplicate `packages/core/tests/fixtures/rpt_program_spec.csv`. Both are updated consistently here. Is there a plan to unify them, or is the duplication load-bearing for separate test harnesses?

---

### Style Notes

The comment added in `workouts.controller.ts` explaining the offset-grouping algorithm is appropriate — this is exactly the kind of non-obvious invariant that warrants a comment. The rest of the changes are clean with well-named identifiers.

---

PR #122, opened 2026-04-30
[chore] Extract weekForWorkoutNum() helper into mappers.ts (#119)

### LINKED ISSUE #119: [chore] Extract weekForWorkoutNum() helper from WorkoutsController into mappers.ts
## Problem

The offset-grouping week-lookup logic in `WorkoutsController.getWorkout` duplicates the same algorithm used by `buildWorkoutDays` on the web side: both deduplicate offsets, sort ascending, index by `workoutNum`, and read the `week` field from the first matching spec entry.

If the lookup strategy changes (e.g., tie-breaking when multiple spec entries at the same offset disagree on week), both sites need updating independently.

## Proposed change

Extract a pure function into `apps/api/src/programs/mappers.ts`:

```typescript
export const weekForWorkoutNum = (
  spec: LiftingProgramSpec[],
  workoutNum: number,
): number | undefined => {
  const offsets = [...new Set(spec.map((s) => s.offset))].sort((a, b) => a - b);
  const offset = offsets[workoutNum - 1];
  return offset !== undefined
    ? (spec.find((s) => s.offset === offset)?.week ?? 1)
    : undefined;
};
```

`WorkoutsController` calls `weekForWorkoutNum` and throws `BadRequestException` on `undefined`. Co-locates the logic with `toWorkoutResponse` and makes it independently testable.

## Acceptance Criteria

- [ ] `weekForWorkoutNum(spec, workoutNum)` exported from `mappers.ts`
- [ ] `WorkoutsController` calls the helper instead of inline logic
- [ ] Unit test for `weekForWorkoutNum` in mappers tests
- [ ] All existing tests continue to pass

brownm09:
## Summary

Replaces the hardcoded `Math.ceil(workoutNum / 2)` week derivation in `toWorkoutResponse` with a spec-driven helper that reads the `week` field from `LiftingProgramSpec[]`. Eliminates the two-workouts-per-week assumption and co-locates the derivation logic with the other mappers where it can be independently tested.

## Changes

- **`packages/core/src/models/LiftingProgramSpec.ts`** — adds `week?: number` (optional; existing callers unchanged)
- **`apps/api/src/programs/mappers.ts`** — exports `weekForWorkoutNum(spec, workoutNum)`; updates `toWorkoutResponse` to accept `week: WeekNumber` as a parameter
- **`apps/api/src/programs/workouts.controller.ts`** — injects `ILiftingProgramSpecRepository`; fetches spec in parallel with workout records; calls `weekForWorkoutNum`; throws `BadRequestException` on `undefined`
- **`apps/api/src/programs/mappers.spec.ts`** — new file with unit tests for `weekForWorkoutNum`
- **`apps/api/src/programs/workouts.controller.spec.ts`** — adds `specRepo` mock and seeds it in the existing test

## Acceptance Criteria

- [x] `weekForWorkoutNum(spec, workoutNum)` exported from `mappers.ts`
- [x] `WorkoutsController` calls the helper instead of inline logic
- [x] Unit test for `weekForWorkoutNum` in mappers tests
- [x] All existing tests continue to pass

## Testing

```
```

Result: **10 test suites passed, 40 tests passed, 1 skipped** (pre-existing skip unchanged).

Closes #119

### DIFF
--- a/apps/api/src/programs/mappers.spec.ts
+++ b/apps/api/src/programs/mappers.spec.ts
@@ -1,0 +1,67 @@
+import { LiftingProgramSpec } from '@lifting-logbook/core';
+import { weekForWorkoutNum } from './mappers';
+
+const baseFields: Omit<LiftingProgramSpec, 'offset' | 'lift' | 'week'> = {
+  increment: 5,
+  order: 1,
+  sets: 3,
+  reps: 5,
+  amrap: true,
+  warmUpPct: '0.4,0.5,0.6',
+  wtDecrementPct: 0.1,
+  activation: 'compound',
+};
+
+const spec = (offset: number, lift: string, week?: number): LiftingProgramSpec => ({
+  ...baseFields,
+  offset,
+  lift: lift as LiftingProgramSpec['lift'],
+  ...(week !== undefined ? { week } : {}),
+});
+
+describe('weekForWorkoutNum', () => {
+  it('returns 1 (default) when spec has no week field and workoutNum is 1', () => {
+    const s = [spec(0, 'Squat'), spec(0, 'Bench Press')];
+    expect(weekForWorkoutNum(s, 1)).toBe(1);
+  });
+
+  it('returns undefined for empty spec', () => {
+    expect(weekForWorkoutNum([], 1)).toBeUndefined();
+  });
+
+  it('returns undefined when workoutNum exceeds distinct offset count', () => {
+    const s = [spec(0, 'Squat'), spec(2, 'Deadlift')];
+    expect(weekForWorkoutNum(s, 3)).toBeUndefined();
+  });
+
+  it('returns explicit week value when spec entries carry week', () => {
+    const s = [spec(0, 'Squat', 1), spec(2, 'Deadlift', 1), spec(4, 'OHP', 2)];
+    expect(weekForWorkoutNum(s, 3)).toBe(2);
+  });
+
+  it('correctly maps workoutNum across multiple distinct offsets', () => {
+    const s = [
+      spec(0, 'Squat', 1),
+      spec(0, 'Bench Press', 1),
+      spec(2, 'Deadlift', 1),
+      spec(4, 'OHP', 2),
+    ];
+    expect(weekForWorkoutNum(s, 1)).toBe(1);
+    expect(weekForWorkoutNum(s, 2)).toBe(1);
+    expect(weekForWorkoutNum(s, 3)).toBe(2);
+    expect(weekForWorkoutNum(s, 4)).toBeUndefined();
+  });
+
+  it('deduplicates offsets — two lifts at the same offset count as one workout', () => {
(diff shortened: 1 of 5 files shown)


brownm09:
## Review: [chore] Extract weekForWorkoutNum() helper into mappers.ts (#119)

### Summary

This is a clean, well-scoped refactor. The pure helper is correct, the algorithm is clearly implemented, and the seven-case test suite for `weekForWorkoutNum` is thorough. One blocking finding: the `BadRequestException` path added to the controller for when `weekForWorkoutNum` returns `undefined` is a new behavior that has no test. One non-blocking finding: the `week as WeekNumber` cast in the controller bypasses the union type constraint without a runtime guard. Otherwise ready to merge after the test gap is closed.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/workouts.controller.spec.ts` — missing test for the `week === undefined` branch

The controller now has a second rejection path: when `weekForWorkoutNum` returns `undefined` (spec has fewer distinct offsets than `workoutNum`), it throws `BadRequestException`. The only rejection test in the spec (`'rejects non-numeric workoutNum'`) fires before the spec repo is even called, so this new branch is never exercised. The `mappers.spec.ts` tests verify `weekForWorkoutNum` itself returns `undefined` correctly, but the controller's response to that return value — throwing with a specific message — is untested.

**Fix:** Add a test in `workouts.controller.spec.ts` that seeds `specRepo.getProgramSpec` with a spec containing only one distinct offset (e.g., offset 0), calls `controller.getWorkout('5-3-1', '2')`, and asserts the result rejects with `BadRequestException`.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/workouts.controller.ts:65` — `week as WeekNumber` cast is unconstrained

`weekForWorkoutNum` returns `number | undefined`. After the undefined guard, `week` is narrowed to `number` — but `WeekNumber` is `1 | 2 | 3 | 4`. The cast assumes spec data is always valid. If a spec entry carries `week: 5` or `week: 0`, the cast propagates the invalid value into the `WorkoutResponse` without error. This is a data-integrity gap rather than a crash risk, but it silently violates the API contract.

**Suggestion:** Narrow the `LiftingProgramSpec.week` field from `number` to `WeekNumber` in `packages/core`, or add a runtime check after the undefined guard (`if (week < 1 || week > 4) throw new BadRequestException(...)`). The former is cleaner and propagates the constraint to parsers.

---

### Questions for Author

- The `?? 1` fallback in `weekForWorkoutNum` means that when `LiftingProgramSpec` entries have no `week` field, *every* distinct offset maps to week 1 — including the second, third, etc. workoutNum. The test at line 63 ("uses `?? 1` fallback...") verifies this: `weekForWorkoutNum(s, 2)` returns `1` even though offset 7 exists. Is this the intended behavior for programs that haven't been updated with `week` fields? (The PR description says yes — backward compatibility — just confirming this is a deliberate choice and not a gap to fill before v0.3.)

---

Issue #129, opened 2026-04-30
[feat] Initial training max discovery (estimation + test week)


brownm09:
## Problem

Training maxes are the single input that drives every working set weight calculation in the app. \`updateMaxes\` throws an error if a max for any configured lift is missing, which means a user cannot start their first cycle without maxes already in the system. The training max management screen (shipped, #108) addresses users who already know their 1RMs, but two other onboarding paths have no data model support:

**Estimation (mechanism 2):** The user does not know their exact 1RM but has recent performance data — e.g., "I did 225 lbs for 5 reps last month." A standard rep-max formula (Brzycki) can back-calculate a training max from any (weight, reps) pair. The system has no utility for this, and the training max management screen has no input for it.

**Test week (mechanism 3):** The user has no reliable data and needs to determine their maxes by actually lifting. This requires a structured "test week" — a special cycle phase with a ramp-up protocol (ascending sets culminating in a heavy set of 1–3 reps) — so the system can generate weights, accept logged results, and derive the training max. Currently, \`CycleDashboard\`, \`LiftingProgramSpec\`, and \`generateLiftPlan\` have no concept of a week type or phase; a test week is indistinguishable from a regular training week at the data and API layer.

## Proposed Solution

Three coordinated changes:

**1. \`WeekType\` — a new shared type**

Add \`WeekType\` (\`'training' | 'test' | 'deload'\`) to \`packages/types/src/domain.ts\`. Add an optional \`weekType?: WeekType\` field to \`LiftingProgramSpec\` (default: \`'training'\`). \`CycleDashboard\` gains a derived \`currentWeekType: WeekType\` field. The Sheets parser reads an optional \`Week Type\` column; blank/missing maps to \`'training'\`.

**2. \`estimateTrainingMax\` — a new core utility**

Add \`estimateTrainingMax(weight: number, reps: number): number\` to \`packages/core\` using the Brzycki formula: \`1RM = weight / (1.0278 − 0.0278 × reps)\`. Result rounds to nearest 5 lbs. Exported from the package's public API.

**3. Test-week ramp-up protocol**

When \`weekType === 'test'\`, \`generateLiftPlan\` produces a 5-set ascending ramp-up ending in a heavy set of 1–3 reps. \`updateMaxes\` is relaxed for test weeks: the \`reps >= spec.reps\` gate is replaced by "any completed final set" — that weight becomes the new training max (no increment applied).

## Acceptance Criteria

[...]

Issue #108, opened 2026-04-30
[feat] Training max management screen


brownm09:
## Problem

Training maxes (1RMs) are the single input that drives every working set weight calculation in the app. The API supports reading and updating them (`GET /training-maxes`, `PATCH /training-maxes`), but there is no web UI. A lifter who wants to review their current maxes, correct an entry, or manually set a starting max for a new cycle has no way to do so without querying the API directly. Without this screen the calculated weights shown on the cycle dashboard and workout logging screen cannot be trusted or corrected.

## Proposed Solution

Build a training max management screen at `/settings/training-maxes`. The screen lists every lift in the user's program alongside its current training max and the date it was last updated. Each row is editable — the lifter taps a weight to change it, then saves. Saving submits the changed values via `PATCH /training-maxes` and confirms the update in-place. The screen respects the user's unit preference (lbs / kg) throughout. It is accessible from the cycle dashboard via a Settings link in the nav and is usable before the first cycle is created to set opening maxes.

## Acceptance Criteria

- [ ] `GET /training-maxes` is called on load; each lift is shown with its current training max and last-updated date
- [ ] Each row is editable; tapping the weight field opens an inline editor (or number input on desktop)
- [ ] Saving changed values submits via `PATCH /training-maxes`; the row reflects the updated value and date on success
- [ ] Validation rejects non-positive weights and non-numeric input before submission
- [ ] Unit preference (lbs / kg) is respected: values are displayed and accepted in the user's preferred unit
- [ ] Lifts with no recorded max are shown with a placeholder and can be set for the first time
- [ ] Screen is usable at ≥ 375 px viewport width; touch targets ≥ 44 px

---
Proposal: [docs/proposals/2026-04-29-training-max-management.md](docs/proposals/2026-04-29-training-max-management.md)

PR #130, opened 2026-04-30
[docs] Propose: Initial training max discovery

brownm09:
## Summary

Adds a proposal doc for the two remaining initial training max discovery paths that have no data model support:
- **Estimation** — `estimateTrainingMax` utility using the Brzycki formula, surfaced in the training max management screen
- **Test week** — `WeekType` shared type, test-week ramp-up protocol in `generateLiftPlan`, relaxed `updateMaxes` gate

Also inserts the proposal row in the v0.3 Proposals table of ROADMAP.md.

Mechanism (1) — "user already knows their maxes" — is covered by #108 (shipped) and is out of scope here.

## Test plan

- [ ] Proposal file renders correctly on GitHub (links resolve, checkboxes display)
- [ ] ROADMAP row links to the proposal file
- [ ] Issue [#129](https://github.com/brownm09/lifting-logbook/issues/129) is assigned milestone `v0.2 — Core API` and Epic `Shared Types` in project #2

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -91,6 +91,7 @@ Web and mobile clients functional end-to-end. Key user-facing features implement
 | [Workout Logging Screen](docs/proposals/2026-04-29-workout-logging-screen.md) | Per-exercise logging with warm-ups, bodyweight gate, and whole-workout overview toggle | [#106](https://github.com/brownm09/lifting-logbook/issues/106) | draft |
 | [Training Max Management Screen](docs/proposals/2026-04-29-training-max-management.md) | View and edit per-lift 1RMs at `/settings/training-maxes`; drives all working set calculations | [#108](https://github.com/brownm09/lifting-logbook/issues/108) | shipped |
 | [Strength Goal Tracking](docs/proposals/2026-04-29-strength-goal-tracking.md) | Per-lift, per-tier strength standards (intermediate/advanced/elite) with target and observed dates | [#111](https://github.com/brownm09/lifting-logbook/issues/111) | shipped |
+| [Initial Training Max Discovery](docs/proposals/2026-04-30-initial-training-max-discovery.md) | Estimation utility (Brzycki) and test-week cycle phase for users with no existing training maxes | [#129](https://github.com/brownm09/lifting-logbook/issues/129) | draft |
 
 ---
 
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: [docs] Propose: Initial training max discovery

### Summary

Ready to merge with minor changes. The proposal is well-structured and the ROADMAP row is correct. Two findings worth addressing before implementation begins: `'deload'` is included in `WeekType` but has no defined behavior in this proposal (deload logic is explicitly Out of Scope), which will leave implementers guessing; and the Brzycki formula has a mathematical constraint (denominator goes to zero at reps ≥ 37) that should be documented so the implementation knows where to validate.

---

### Blocking Findings

**[correctness]** `docs/proposals/2026-04-30-initial-training-max-discovery.md:36`
`WeekType` includes `'deload'` but the Out of Scope section explicitly defers deload-week progression logic. This means implementers will be adding a type value with no defined behavior for `generateLiftPlan`, `updateMaxes`, or `CycleDashboard` — and nothing in the proposal tells them what to do when `weekType === 'deload'` is encountered at runtime (fall through to training behavior? throw? no-op?).
**Fix:** Either remove `'deload'` from the initial `WeekType` union and add it in the future deload proposal, or add a note in the Proposed Solution or Out of Scope section stating: "`'deload'` is included for type completeness; `generateLiftPlan` and `updateMaxes` treat it identically to `'training'` until a deload proposal defines its behavior."

---

### Non-Blocking Findings

**[reliability]** `docs/proposals/2026-04-30-initial-training-max-discovery.md:47`
The Brzycki formula `1RM = weight / (1.0278 − 0.0278 × reps)` produces undefined (division-by-zero) at reps = 37 and a negative (nonsensical) result for reps > 37. The proposal doesn't specify an input constraint.
**Suggestion:** Add a note to the `estimateTrainingMax` AC or Proposed Solution that the utility should enforce a valid rep range (e.g., 1–36, or the more conservative 1–20 typical in lifting applications) and return an error or throw for out-of-range inputs.

---

### Questions for Author

- The AC says `updateMaxes` for test weeks uses "any completed final set" as the trigger — but "completed" isn't defined. Does this mean any non-zero reps on the final set? What happens if the user skips the final set entirely? Worth one sentence in the AC to close this.
- `CycleDashboard.currentWeekType` is "derived from the active week's spec rows" — but what's the rule if rows within the same week have mixed `weekType` values? Does the parser enforce week-level uniqueness, or does the derivation take a majority / first-row wins / throws on conflict?
- The Sheets parser AC says "blank/missing → `'training'`" — is `Week Type` a per-row column or a per-week column? If per-row, does every row in a week need to agree, or does the parser accept mixed values and let `CycleDashboard` resolve conflicts?

---

Issue #136, opened 2026-05-01
[feat] User-tier-aware LLM provider selection for cycle planning


brownm09:
The cycle planning agent (introduced in #54) currently selects its LLM provider via the `CYCLE_AGENT_PROVIDER` env var — a server-wide toggle. Once paid tiers exist, the toggle should be per-user:

- **Free tier:** local/free OSS provider (Ollama running Llama 3.x via OpenAI-compatible API).
- **Paid tier:** Anthropic Claude Sonnet (or another premium provider).

## Scope

- User account / tier model (depends on auth landing — see ADR-003)
- Per-request provider resolution (read user tier, select adapter)
- New port `ICyclePlanningProviderResolver` to encapsulate the selection logic
- Move provider selection from module-init factory (current implementation) to a request-scoped resolver
- Update `CyclePlanController` to call the resolver instead of injecting `ICyclePlanningAgent` directly
- Cost guardrail: rate-limit free-tier calls to prevent abuse of the local inference endpoint

## Acceptance Criteria

- [ ] `ICyclePlanningProviderResolver` port interface defined in `apps/api/src/ports/`
- [ ] Resolver implementation maps `AuthUser.tier` → adapter
- [ ] `CyclePlanController` uses resolver; `CYCLE_AGENT_PROVIDER` env var is removed or relegated to a fallback default for unauthenticated/dev mode
- [ ] Integration test covers free-tier user → OSS adapter selected, paid-tier user → Anthropic adapter selected
- [ ] ADR addendum or new ADR documents the resolver pattern and rate-limit decision

## Dependencies

- Auth implementation (ADR-003)
- User account / tier data model

PR #137, opened 2026-05-01
[feat] Cycle planning agent with swappable LLM providers

### LINKED ISSUE #54: [feat] Cycle planning agent — LLM-powered training cycle recommendations
## Overview

Add a cycle planning agent to the API that reasons over a user's lift history and program spec to propose training cycle modifications. The user describes a goal (e.g., "peak my squat for a meet in 8 weeks") and the agent iteratively calls domain tools to produce a structured cycle plan with reasoning.

## Motivation

Demonstrates agentic LLM integration in a domain-rich context: tool use, multi-step reasoning, and clean hexagonal boundaries. The LLM adapter drives inward through existing port interfaces — `packages/core` is not modified.

## Acceptance Criteria

- [ ] Agent is exposed via a dedicated endpoint in `apps/api` (REST or GraphQL TBD in ADR)
- [ ] Agent has access to at least three domain tools: `get_lift_history`, `get_training_max`, `get_program_spec`
- [ ] Agent can propose `UpdateCycleOverrides` mutations and explain its reasoning
- [ ] LLM adapter lives behind a port interface — swappable without core changes
- [ ] Integration test covers at least one full agent turn with tool calls
- [ ] ADR documents model selection, tool schema design, and adapter boundary decisions (see sub-issue)

## Dependencies

- Authenticated lift history endpoints (v0.2)
- ADR for agent design (see sub-issue — tracked separately, targets v0.2)

## Notes

Uses the Anthropic SDK (`@anthropic-ai/sdk`) tool use feature. Primary reference: https://docs.anthropic.com/en/docs/tool-use

brownm09:
## Summary

Implements the LLM-powered cycle planning agent from [ADR-016](docs/adr/ADR-016-cycle-planning-agent.md) — `POST /cycle-plan` accepts `{ program, goal, cycleNum }` and returns proposed training-max changes with reasoning. The agent runs an agentic tool-use loop against five read tools (lift history, training maxes, program spec, cycle dashboard, program philosophy) and a terminal `propose_cycle_plan` tool.

**Two key changes vs. ADR-016:**

- **Default provider is open-source.** `OpenAICompatibleCyclePlanningAdapter` targets Ollama via the OpenAI Chat Completions API (default `http://localhost:11434/v1`, model `llama3.1`). `AnthropicCyclePlanningAdapter` is selected when `CYCLE_AGENT_PROVIDER=anthropic`. Tier-based selection (free → Ollama, paid → Anthropic) is tracked separately in #136 against v0.4 once auth lands.
- **Program philosophy as a new port.** `IProgramPhilosophyRepository` exposes curated guidance for 5/3/1, 5/3/1 BBB, 5/3/1 BBS, Starting Strength, StrongLifts 5×5, and RPT/Leangains via a `get_program_philosophy` tool. `CycleDashboard` gains an optional `programType` so the agent can resolve philosophy from the dashboard.

Both adapters share a 30s deadline (`Promise.race`), a 5-round tool budget, and a 4096-token cap. Exceeding any of these returns `partial: true` with empty changes rather than throwing.

## Test plan
- [x] `npm test` — full monorepo suite (10/10 workspaces pass; 64 API tests pass)
- [x] `npm run build -w @lifting-logbook/api` — clean
- [x] Controller integration tests for happy path and partial responses
- [x] Unit tests for `dispatchTool` covering all 5 read tools, error handling, and proposal parsing
- [ ] Manual smoke test against a running Ollama (out of scope for CI)

Closes #54

## References
- [ADR-016 — Cycle Planning Agent](docs/adr/ADR-016-cycle-planning-agent.md)
- [OpenAI Chat Completions API](https://platform.openai.com/docs/api-reference/chat) — used for Ollama compatibility
- [Anthropic Messages API — Tool use](https://docs.anthropic.com/en/docs/build-with-claude/tool-use)
- [Ollama OpenAI compatibility](https://github.com/ollama/ollama/blob/main/docs/openai.md)


brownm09:
## Review: [feat] Cycle planning agent with swappable LLM providers

### Summary

Needs revision before merge. The architecture (shared `agent-tools.ts` dispatcher, factory provider, partial-result envelope) is sound and the test coverage is appropriate for the dispatcher and controller, but the LLM adapters have correctness and security issues that will bite in production. The single highest-severity finding is that **the `POST /cycle-plan` endpoint has no auth, no rate limiting, and no per-user scoping** — combined with a free-form `goal` field that flows directly into LLM prompts and (when `CYCLE_AGENT_PROVIDER=anthropic`) drives paid API calls. The deadline implementation is also broken in a way that wastes tokens on every timeout. The 1092-line diff is appropriately scoped — splitting would just create lockstep PRs.

---

### Blocking Findings

**[security]** `apps/api/src/programs/cycle-plan.controller.ts`
The endpoint has no auth guard, no throttle, and `program` is trusted from the request body rather than scoped to a caller identity.
In production with `CYCLE_AGENT_PROVIDER=anthropic`, anonymous traffic drains the API key budget; the free-form `goal` field also turns the endpoint into a free LLM proxy via prompt injection. Other write endpoints in this codebase will gain auth (per the `Scope.REQUEST` comment in `programs.module.ts`); this one should not ship without it.
**Fix:** Add the same auth guard the rest of the API will use (or gate the controller behind a feature flag until #136 lands), add per-IP/per-user rate limiting, and resolve `program` from the authenticated principal rather than the body.

**[correctness/reliability]** `apps/api/src/adapters/llm/openai-compatible-cycle-planning.adapter.ts` & `anthropic-cycle-planning.adapter.ts` (the `Promise.race` block in `plan()`)
`Promise.race` against `setTimeout` resolves the caller, but `runLoop` keeps running — subsequent SDK calls and tool dispatches continue after the HTTP response is sent, the underlying `client.messages.create` / `client.chat.completions.create` request is never cancelled, and the deadline `setTimeout` handle is never cleared on the success path.
Every timeout pays for the remaining tokens on Anthropic; on Ollama the slow generation continues consuming GPU. Under any concurrency this stacks in-flight requests. The uncleared timer also keeps the event loop alive on every successful call.
**Fix:** Create an `AbortController` per `plan()` call, pass `{ signal: controller.signal }` into each SDK call, `controller.abort()` from the timeout branch, and `clearTimeout` the deadline handle when `runLoop` resolves first. Check `signal.aborted` between rounds before issuing the next call.

**[security]** `apps/api/src/adapters/llm/openai-compatible-cycle-planning.adapter.ts` (constructor) + `agent-tools.ts` (system/user prompt assembly)
The OpenAI adapter falls back to `http://localhost:11434/v1` and the literal API key string `'ollama'` when env vars are unset, and `goal` is concatenated into the user prompt with no delimiter or "treat as data" instruction.
A misconfigured deploy silently ships requests to localhost (or fails opaquely). Separately, the system prompt has no defenses against `goal` overriding it — a user can instruct the agent to ignore prior tools, propose absurd training maxes, or exfiltrate the system prompt. Since the propose result is rendered to the user, prompt-injection output becomes the product.
**Fix:** Require `CYCLE_AGENT_BASE_URL` and `CYCLE_AGENT_API_KEY` in non-development environments — throw at construction if absent rather than defaulting. In `buildUserMessage`, wrap goal text in explicit delimiters (`<user_goal>...</user_goal>`) and add a system-prompt clause that goal text is data, not instructions.

**[correctness]** `apps/api/src/adapters/llm/agent-tools.ts` — `parseProposal`
The function casts `args.proposedChanges` to the typed array after only checking `Array.isArray`. Element shape, numeric types, and `LiftName` membership are never validated. The OpenAI adapter additionally `JSON.parse`s `call.function.arguments` and on parse failure substitutes `{}`, so a malformed propose call returns an empty plan that `partial: false` claims is valid.
The result is returned to clients as a typed `CyclePlanResponse`; the LLM can emit `proposedWeight: "heavy"`, missing fields, negative numbers, or a `lift` outside the `LiftName` union and downstream callers will trust it. Future write paths that apply these as TM updates inherit the contamination.
**Fix:** Validate each element with zod / class-validator / a hand-rolled guard — reject non-finite numbers, enforce `LiftName` membership, drop or error on malformed entries. On `JSON.parse` failure for `propose_cycle_plan` specifically, return `partial: true` rather than `{}`.

**[correctness]** `openai-compatible-cycle-planning.adapter.ts` & `anthropic-cycle-planning.adapter.ts` (mixed tool-call batches)
When the model emits multiple tool calls in one turn that include `propose_cycle_plan`, both adapters return on the first propose without producing tool results for the other calls in the same batch. The earlier non-propose calls in the array were already dispatched, but their results are pushed onto a `messages` array that is then discarded; non-propose calls *after* propose in the array are silently skipped.
Behavior is non-deterministic and depends on call ordering inside a single assistant turn. Today's tools are read-only so the user-visible impact is small, but the conversation transcript is malformed (`tool_calls` without matching `tool_result` blocks) — any future retry/resume logic, write tool, or transcript logging breaks.
**Fix:** Process all non-propose tools first and append their results, then check whether any call in the batch was propose and return after. Alternatively, treat a propose call as terminal and explicitly ignore sibling calls — but document that choice in the code.

---

### Non-Blocking Findings

**[maintainability]** Tool schemas in `openai-compatible-cycle-planning.adapter.ts` and `anthropic-cycle-planning.adapter.ts`
Six tool schemas with identical names, descriptions, and parameter shapes are duplicated across both adapters.
The dispatcher in `agent-tools.ts` is already shared — schemas are the obvious next thing to share, and any drift means the two providers silently diverge in what they advertise to the model.
**Suggestion:** Define the tool catalog once in `agent-tools.ts` (a provider-neutral shape) and write thin translators to `Anthropic.Tool[]` and `ChatCompletionTool[]`. Same pattern as the dispatcher.

**[maintainability]** `runLoop` skeletons in both adapters
Round counter, terminal-tool check, dispatch loop, partial-on-exhaustion return — near-identical between the two adapters.
Bug fixes from this review (deadline leak, error handling, batch handling) all need to land twice and stay in lockstep.
**Suggestion:** Hoist the loop into `agent-tools.ts` taking provider-specific callbacks for "request next turn" and "format tool result." Leaves each adapter responsible only for SDK shape translation.

**[reliability]** No try/catch around the SDK calls in either adapter
A single transient 5xx, network blip, or rate-limit response rejects the whole controller.
The adapters already have a `partial: true` envelope designed for degraded outcomes — using it for transient SDK errors fits the existing contract better than a 500.
**Suggestion:** Wrap the SDK call, retry once or twice on 5xx / ECONNRESET / 429-with-Retry-After, and on terminal failure return `partial: true` with the error reason rather than throwing.

**[reliability]** No log lines in either adapter
The injected `Logger` is unused. For a 30-second LLM operation that can silently exhaust the tool budget or hit the deadline and return `partial: true` with empty changes, this is the gap that makes production triage impossible.
**Suggestion:** Per round, log round index, tool calls dispatched, dispatch outcome, and SDK-reported `usage`. On terminal exit log a single summary line (success / deadline / budget / error) with total elapsed ms.

---

### Questions for Author

- Was the auth gap intentional pending #136, or an oversight? If intentional, should the controller be gated behind a feature flag or excluded from the main router until auth lands?
- The 4096-token output cap with a 5-round budget is tight for a model that emits structured tool calls — was this validated against a real Ollama run, or copied from ADR-016 unchanged?
- `partial: true` with `proposedChanges: []` is returned both for "deadline exceeded" and "budget exhausted" — should the response distinguish these so the UI can give a useful retry hint?

---

### Style Notes

The shared dispatcher and factory provider patterns are clean and the test layout (unit-test the dispatcher, integration-test the controller with a mocked agent) is exactly right for this shape of code. The `programType` addition to `CycleDashboard` is well-placed. One minor nit: the `<!-- references -->` link cluster in the PR body is good but `apps/api/src/programs/programs.module.ts` has a docstring referring to `programs.e2e.spec.ts` as the forcing function for request-scoped wiring — worth verifying that file still has the `.skip` it claims to have, since this PR didn't touch it.

---

PR #138, opened 2026-05-01
[docs] Define A/B comparison exit criteria and CI event taxonomy enforcement

brownm09:
## Summary

- **ADR-008 amendment:** Adds a *Comparison Conclusion Criteria* section with an 8-week minimum comparison duration, five concrete metric thresholds (≥10% delta sustained over 4 weeks), a BigQuery-backed decision process, and an archival plan for the losing codebase — eliminating the open-ended dual-codebase maintenance burden flagged in the ARB review.
- **CI taxonomy enforcement:** Adds `scripts/validate-analytics-taxonomy.mjs` (a Node.js ESM script) and a new CI step that parses event names from `packages/types/src/analytics.ts` and validates they are present in `AnalyticsConstants.kt`. The step skips gracefully when the Kotlin file does not yet exist, and fails on first taxonomy drift once the Kotlin app is scaffolded.
- **ADR-012 amendment:** Adds a *CI Taxonomy Enforcement* section documenting the script, the per-event update workflow, and the CI configuration snippet.
- **`packages/types/src/analytics.ts`:** Creates the shared event taxonomy file that was defined only in ADR-012 prose — now an actual source file exported from the package root.

## Acceptance Criteria

- [x] ADR-008 amended with a "Comparison Conclusion Criteria" section: minimum comparison duration, metric thresholds, decision process for selecting a winner, and archival plan for the losing codebase
- [x] CI step added that extracts event names from `packages/types/src/analytics.ts` and validates they are present in `apps/mobile-kotlin/.../AnalyticsConstants.kt` (or equivalent path); CI fails on drift
- [x] CI validation documented in ADR-012

## Test plan

- `npm test` — all 10 tasks pass (49 tests + 1 skipped)
- `node scripts/validate-analytics-taxonomy.mjs` — exits 0 with INFO message (Kotlin file not yet present; expected)

Closes #41

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,3 +32,6 @@ jobs:
 
       - name: Lint & Test
         run: npx turbo run lint test
+
+      - name: Validate analytics taxonomy
+        run: node scripts/validate-analytics-taxonomy.mjs
--- a/docs/adr/ADR-008-mobile-strategy.md
+++ b/docs/adr/ADR-008-mobile-strategy.md
@@ -87,6 +87,57 @@ accumulate.
 
 ---
 
+## Comparison Conclusion Criteria
+
+The two-codebase maintenance period is explicitly time-bounded. The following criteria define
+when the comparison is complete and which client becomes the production standard.
+
+### Minimum comparison duration
+
+Both clients must have been deployed to internal testing tracks and instrumented for **at least
+8 weeks** before a conclusion is drawn. This ensures the Firebase Analytics data set is large
+enough to detect meaningful differences in session length, crash rate, and render performance
+with reasonable confidence.
+
+### Decision thresholds
+
+A winner is selected when **at least three of the five primary metrics** show a consistent
+advantage (≥ 10% delta sustained over the final 4 weeks of the comparison period):
+
+| Metric | Source | Direction |
+|---|---|---|
+| Screen render time (p95) | Firebase Performance | Lower is better |
+| Crash-free session rate | Firebase Crashlytics | Higher is better |
+| App startup time (cold, p95) | Firebase Performance | Lower is better |
+| `workout_completed` / `workout_started` ratio | Firebase Analytics | Higher is better |
+| Install size | Play Console | Lower is better |
+
+If no client meets the threshold after 8 weeks, extend the comparison in 4-week increments
+(up to a maximum of 20 weeks) until thresholds are met or the decision is made by judgment
+with documented rationale.
+
+### Decision process
+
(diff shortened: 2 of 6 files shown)


brownm09:
## Review: [docs] Define A/B comparison exit criteria and CI event taxonomy enforcement

### Summary

Ready to merge. This PR closes a real gap — the A/B comparison was previously open-ended, and the CI taxonomy enforcement had no implementation behind it. Both ADR amendments are precise and actionable. The validation script is clean and does exactly what it claims. One non-blocking concern about the `EventProperties` interface being reference-only with no enforcement mechanism, and a question about monorepo git tag semantics.

---

### Blocking Findings

None.

---

### Non-Blocking Findings

**[maintainability]** `packages/types/src/analytics.ts:22`

`EventProperties` defines a typed map from event name to expected property shape, but nothing in this PR (or the existing codebase) enforces it at call sites. Consumers can log `SET_LOGGED` with any properties object and TypeScript won't error unless they explicitly annotate with `EventProperties['set_logged']`. The interface is reference documentation, not a compiler contract.

**Suggestion:** When the logging wrapper is built (in either the RN or Kotlin scaffolding PR), add a typed `logEvent<T extends AnalyticsEventName>(name: T, props: EventProperties[T])` helper that ties the interface to actual usage. Until then, `EventProperties` will silently drift from what clients actually emit.

---

**[maintainability]** `scripts/validate-analytics-taxonomy.mjs:25`

The regex `[^}]+` captures the body of `AnalyticsEvent` by matching anything that isn't a closing brace. This works for the current flat structure but would silently produce wrong results if a nested object were ever added inside `AnalyticsEvent` (the regex would stop at the first inner `}`). The script would then report zero events found and exit 1, which surfaces the problem — but the failure message would be cryptic: "No event names found" rather than "nested structure not supported."

**Suggestion:** Add a brief comment on the regex noting the flat-object assumption, so future editors know why a nested extension would break it.

---

### Questions for Author

- The archival plan (ADR-008) says `git tag ab-comparison-final` after the comparison concludes. In a monorepo, that tag lands on the full-repo HEAD, not on the losing app directory specifically. If both apps are active right up to the conclusion date, the tag will capture both — which is arguably fine, but means "the last state of the losing client" requires a path filter (`git log ab-comparison-final -- apps/mobile/`) to extract. Is that the intended reading, or was a more app-specific ref in mind (e.g., tagging the subtree, or just noting the commit SHA in the archival `README.md`)?

---

### Style Notes

`analytics.ts` uses inline category comments (`// Workout logging`, `// Navigation`, etc.) to group the const entries. These are reasonable given this file doubles as the canonical event reference, and the groupings aren't obvious from the key names alone.

---

Issue #144, opened 2026-05-01
[feat] Per-user repository factory — IRepositoryFactory adapter and controller threading


brownm09:
## Context

ADR-003 specifies that the data store adapter is resolved per-user, per-request via `IRepositoryFactory`. The port interface exists (`apps/api/src/ports/factory.ts`) but no adapter exists. Controllers currently inject singleton in-memory repositories directly — they are not user-scoped.

This issue is the second half of auth: enforcing identity (issue #143) is not meaningful without also scoping data to that identity.

## Scope

- `apps/api/src/adapters/factory/in-memory-repository-factory.ts` — implements `IRepositoryFactory`; returns the existing singleton in-memory `RepositoryBundle` for any user (appropriate for single-user dev and e2e tests)
- `apps/api/src/adapters/factory/system-db-repository-factory.ts` — implements `IRepositoryFactory`; reads `user_data_source` (system Postgres), resolves adapter type, returns the correct `RepositoryBundle`; includes TTL cache keyed by `user_id`
- System Postgres schema: `user_data_source` table migration
- `ProgramsModule` refactor: controllers receive `IRepositoryFactory` + `AuthUser` (via `@CurrentUser()`), call `forUser(user)` instead of using directly-injected singleton repositories
- Unskip and implement the `.skip` isolation test in `programs.e2e.spec.ts` (line 174)

## Acceptance Criteria

- [ ] `IRepositoryFactory` is registered in `AppModule`; selection between in-memory and system-DB factory is environment-driven (e.g., `SYSTEM_DATABASE_URL` present → system-DB factory)
- [ ] All six program controllers use `forUser(user)` to obtain their `RepositoryBundle`
- [ ] In-memory factory: two requests for different `user_id` values return isolated `RepositoryBundle` instances (each user gets their own in-memory state)
- [ ] System-DB factory: adapter config is read from `user_data_source` and cached with ~5-minute TTL
- [ ] The previously-skipped adapter isolation e2e test passes
- [ ] No controller imports a concrete adapter class directly — all access is through the factory

## Dependencies

- Auth guard (#143) — `AuthUser` must be on the request before `forUser(user)` can be called
- ADR-003 (accepted)
- ADR-014 — `adapter_config` column encryption at rest (credential storage for Sheets adapter)

## Notes

The system-DB factory is the production path. The in-memory factory keeps e2e tests and local dev working without a running Postgres instance.

PR #145, opened 2026-05-01
[feat] Auth guard — Clerk adapter, DevAuthProvider, and NestJS global guard

### LINKED ISSUE #143: [feat] Auth guard — Clerk adapter and NestJS request guard
## Context

Port interfaces `IAuthProvider` and `AuthUser` are defined in `apps/api/src/ports/auth.ts`. No adapter or guard exists. All API endpoints are currently unguarded.

ADR-005 specifies: Clerk as the managed identity provider, `IAuthProvider` as the abstraction boundary, JWT Bearer token on every request.

## Decision

Provider selection is environment-driven:
- `CLERK_SECRET_KEY` present → `ClerkAuthProvider` (production / staging)
- `CLERK_SECRET_KEY` absent → `DevAuthProvider` (local dev, CI, e2e tests)

`DevAuthProvider` accepts any Bearer token value and returns a hardcoded `AuthUser` configurable via `DEV_USER_ID` and `DEV_USER_EMAIL` env vars. This keeps local dev and CI working without a Clerk account.

## Scope

- `apps/api/src/adapters/clerk/clerk-auth.adapter.ts` — implements `IAuthProvider` using `@clerk/backend`
- `apps/api/src/adapters/dev/dev-auth.adapter.ts` — implements `IAuthProvider` for local/CI use
- `apps/api/src/auth/auth.guard.ts` — NestJS guard: extracts Bearer token, calls `verifyToken`, attaches `AuthUser` to request; returns 401 if token is missing or invalid
- `apps/api/src/auth/current-user.decorator.ts` — `@CurrentUser()` parameter decorator
- `apps/api/src/auth/auth.module.ts` — NestJS module; selects adapter based on env
- Register guard globally in `AppModule`
- Update e2e spec to send `Authorization: Bearer dev-token` on all requests

## Acceptance Criteria

- [ ] `GET /health` (or a designated public endpoint) is exempt from the guard
- [ ] All other endpoints return 401 when no `Authorization` header is present
- [ ] With `CLERK_SECRET_KEY` unset, any Bearer value resolves to the configured dev user
- [ ] With `CLERK_SECRET_KEY` set, an invalid token returns 401; a valid Clerk JWT resolves to the correct `AuthUser`
- [ ] Existing e2e suite passes with the updated auth headers
- [ ] `AuthUser` is accessible via `@CurrentUser()` in any controller method

## Dependencies

- ADR-005 (accepted)
- `IAuthProvider` port (exists: `apps/api/src/ports/auth.ts`)

## Notes

Controllers are not changed in this issue — they still use injected singleton repositories. Per-user data scoping comes in the follow-up issue. The guard enforces identity; factory threading enforces isolation.

brownm09:
## Summary

Implements the NestJS auth guard per ADR-005. All API endpoints now require `Authorization: Bearer <token>`; `GET /health` is exempt via `@Public()`.

- **`DevAuthProvider`** — selected when `CLERK_SECRET_KEY` is absent (local dev, CI, e2e tests); accepts any Bearer value and returns a configurable `AuthUser` via `DEV_USER_ID` / `DEV_USER_EMAIL` env vars
- **`ClerkAuthProvider`** — selected when `CLERK_SECRET_KEY` is present; verifies JWT via `@clerk/backend`, fetches user profile from Clerk API
- **`AuthGuard`** — global `APP_GUARD`; extracts Bearer token, calls `verifyToken`, stores `AuthUser` on `request.user`; returns 401 on missing or invalid token
- **`@Public()`** — metadata decorator to exempt endpoints from the guard
- **`@CurrentUser()`** — param decorator to inject `AuthUser` into controller methods (ready for issue #144)
- **`AuthModule`** — env-driven adapter selection, registers guard globally via `APP_GUARD`
- e2e spec updated: all requests send `Authorization: Bearer dev-token`; new tests assert 401 without auth and that health is public

## Acceptance Criteria

- [x] `GET /health` returns 200 without auth
- [x] All other endpoints return 401 when no `Authorization` header is present
- [x] With `CLERK_SECRET_KEY` unset, any Bearer value resolves to the configured dev user
- [x] `AuthUser` is accessible via `@CurrentUser()` in any controller method
- [x] Existing e2e suite passes (70 tests, 1 intentionally skipped pending #144)

## Test plan

```
```

All 70 tests pass, 1 skipped (adapter isolation — unblocked by #144).

## Notes

Controllers are unchanged — they still use injected singleton repositories. Per-user data scoping is issue #144, which depends on this PR.

Closes #143

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.92.0",
+    "@clerk/backend": "^3.4.4",
     "@lifting-logbook/core": "*",
     "@lifting-logbook/types": "*",
     "@nestjs/common": "^10.0.0",
--- a/apps/api/src/adapters/clerk/clerk-auth.adapter.ts
+++ b/apps/api/src/adapters/clerk/clerk-auth.adapter.ts
@@ -1,0 +1,26 @@
+import { createClerkClient, verifyToken } from '@clerk/backend';
+import { UnauthorizedException } from '@nestjs/common';
+import { AuthUser, IAuthProvider } from '../../ports/auth';
+
+export class ClerkAuthProvider implements IAuthProvider {
+  private readonly client = createClerkClient({
+    secretKey: process.env.CLERK_SECRET_KEY,
+  });
+
+  async verifyToken(token: string): Promise<AuthUser> {
+    try {
+      const payload = await verifyToken(token, {
+        secretKey: process.env.CLERK_SECRET_KEY!,
+      });
+      const user = await this.client.users.getUser(payload.sub);
+      return {
+        id: user.id,
+        email: user.primaryEmailAddress?.emailAddress ?? '',
+        provider: 'clerk',
+        displayName: user.fullName ?? undefined,
+      };
+    } catch {
+      throw new UnauthorizedException('Invalid or expired token');
+    }
+  }
+}
--- a/apps/api/src/adapters/clerk/clerk-auth.adapter.ts
+++ b/apps/api/src/adapters/clerk/clerk-auth.adapter.ts
@@ -1,18 +1,35 @@
+import {
+  Injectable,
+  ServiceUnavailableException,
+  UnauthorizedException,
+} from '@nestjs/common';
 import { createClerkClient, verifyToken } from '@clerk/backend';
-import { UnauthorizedException } from '@nestjs/common';
 import { AuthUser, IAuthProvider } from '../../ports/auth';
 
+@Injectable()
 export class ClerkAuthProvider implements IAuthProvider {
   private readonly client = createClerkClient({
     secretKey: process.env.CLERK_SECRET_KEY,
   });
 
   async verifyToken(token: string): Promise<AuthUser> {
+    // Step 1: verify the JWT — throws UnauthorizedException on invalid/expired token
+    let sub: string;
     try {
       const payload = await verifyToken(token, {
         secretKey: process.env.CLERK_SECRET_KEY!,
+        ...(process.env.CLERK_AUTHORIZED_PARTY
+          ? { authorizedParties: [process.env.CLERK_AUTHORIZED_PARTY] }
+          : {}),
       });
-      const user = await this.client.users.getUser(payload.sub);
+      sub = payload.sub;
+    } catch {
+      throw new UnauthorizedException('Invalid or expired token');
+    }
+
+    // Step 2: fetch user profile — throws ServiceUnavailableException on Clerk API error
+    try {
+      const user = await this.client.users.getUser(sub);
       return {
         id: user.id,
         email: user.primaryEmailAddress?.emailAddress ?? '',
@@ -20,7 +37,7 @@ export class ClerkAuthProvider implements IAuthProvider {
         displayName: user.fullName ?? undefined,
       };
     } catch {
-      throw new UnauthorizedException('Invalid or expired token');
+      throw new ServiceUnavailableException('Auth service unavailable');
     }
   }
 }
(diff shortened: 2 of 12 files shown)


brownm09:
## Review: [feat] Auth guard — Clerk adapter, DevAuthProvider, and NestJS global guard

### Summary

Ready to merge with one fix. The guard wiring, `@Public()` bypass, adapter pattern, and e2e coverage are all correct and well-structured. One security finding needs addressing before merge: an empty `CLERK_SECRET_KEY` (a common container/Helm misconfiguration) silently selects `DevAuthProvider` rather than failing closed. Three non-blocking findings are worth noting before the Clerk production path is activated.

---

### Blocking Findings

**[security]** `apps/api/src/auth/auth.module.ts:11`

`process.env.CLERK_SECRET_KEY` is falsy for both the unset case (local/CI — intentional) and the empty-string case (`""` — a common Helm/container misconfiguration where the secret placeholder is defined but not filled). Both resolve to `DevAuthProvider`, which accepts any Bearer value with no validation. A misconfigured staging/prod deploy would silently become a fully open API with no error at boot.

**Fix:** Add a startup guard that throws if `CLERK_SECRET_KEY` is present but empty — or if `NODE_ENV === 'production'` and the key is missing:
```typescript
if (process.env.CLERK_SECRET_KEY === '') {
  throw new Error('CLERK_SECRET_KEY is set but empty — refusing to fall back to DevAuthProvider');
}
```

---

### Non-Blocking Findings

**[security]** `apps/api/src/adapters/clerk/clerk-auth.adapter.ts:12`

`verifyToken` is called without `authorizedParties` or `issuer` options. Per [Clerk's backend SDK docs](https://clerk.com/docs/references/backend/verify-token), omitting `authorizedParties` means tokens minted for any Clerk application (including an attacker's own Clerk dev instance) will pass verification. The risk is low today since the app isn't deployed against a real Clerk project, but this should be set before going live.

**Suggestion:** Pass `authorizedParties: [process.env.CLERK_AUTHORIZED_PARTY ?? '']` to `verifyToken`. Add `CLERK_AUTHORIZED_PARTY` to the documented env vars.

---

**[reliability]** `apps/api/src/adapters/clerk/clerk-auth.adapter.ts:22`

The bare `catch { }` block converts all exceptions — JWT validation failures *and* Clerk API network errors — to `UnauthorizedException`. A transient Clerk outage will surface to clients as 401 ("Invalid or expired token") rather than 503, making outages invisible and hard to diagnose.

**Suggestion:** Catch `@clerk/backend` verification errors specifically and rethrow others (or log them before rethrowing as `ServiceUnavailableException`). Worth resolving before cloud deployment but not urgent for the current dev/CI-only code path.

---

**[maintainability]** `apps/api/src/auth/current-user.decorator.ts:6`

`CurrentUser` declares its return type as `AuthUser`, but on `@Public()` routes the guard skips setting `request.user`, so the decorator returns `undefined` at runtime. Any controller method on a public route that uses `@CurrentUser()` will receive `undefined` while TypeScript thinks it's `AuthUser`. No such controller exists yet, but the type lie will surface as a silent bug when #144 starts wiring controllers.

**Suggestion:** Change the return type to `AuthUser | undefined`, or document via JSDoc that `@CurrentUser()` must not be used on `@Public()` routes.

---

### Questions for Author

- `DevAuthProvider` and `ClerkAuthProvider` have no `@Injectable()` decorator. This works today (no constructor dependencies), but NestJS DI won't properly scan them if dependencies are added later. Was this intentional, or should they be decorated for consistency?

---

### Style Notes

`catch { }` with no bound variable is valid TypeScript 4.0+ syntax. It reads cleanly here, but losing the original error means production incidents will show "Invalid or expired token" with no underlying cause in logs. Worth adding a `catch (err)` + `logger.error(err)` before the `throw` when logging is wired up.

---

PR #146, opened 2026-05-02
[feat] Per-user repository factory — IRepositoryFactory adapter and controller threading

### LINKED ISSUE #144: [feat] Per-user repository factory — IRepositoryFactory adapter and controller threading

brownm09:
## Summary

- Introduces `IRepositoryFactory` port with `forUser(user: AuthUser): Promise<RepositoryBundle>` — all 6 repo bundles resolved per authenticated user
- `InMemoryRepositoryFactory` provides per-user isolated in-memory repos via `Map<userId, RepositoryBundle>`
- `SystemDbRepositoryFactory` resolves adapter type from a `user_data_source` Postgres table with a 5-minute TTL cache
- `RepositoryFactoryModule` is `@Global()` so all feature modules get the token without circular imports
- All 8 bundle-using controllers refactored to inject `REPOSITORY_FACTORY` and call `factory.forUser(user)`
- `CycleGenerationService` made stateless — repos passed as method params instead of constructor injection
- `DevAuthProvider` updated to use the Bearer token value as user ID, enabling per-user isolation in E2E tests
- Migration `001_create_user_data_source.sql` added for the new table

## Acceptance Criteria

- [x] `IRepositoryFactory` port defined in `apps/api/src/ports/`
- [x] `InMemoryRepositoryFactory` implementation with per-user isolation
- [x] `SystemDbRepositoryFactory` with Postgres lookup and TTL cache
- [x] `RepositoryFactoryModule` wired into `AppModule`
- [x] All controllers use `factory.forUser(user)` pattern
- [x] All unit tests updated and passing (71/71)
- [x] E2E isolation test verifying Alice and Bob see separate data
- [x] Migration SQL for `user_data_source` table

## Testing

```bash
```

All 71 tests pass including the new per-user isolation E2E test.

Closes #144

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -22,12 +22,14 @@
     "fastify": "^4.0.0",
     "openai": "^6.35.0",
     "reflect-metadata": "^0.2.0",
+    "pg": "^8.13.0",
     "rxjs": "^7.8.0"
   },
   "devDependencies": {
     "@nestjs/cli": "^10.0.0",
     "@nestjs/schematics": "^10.0.0",
     "@nestjs/testing": "^10.0.0",
+    "@types/pg": "^8.11.0",
     "ts-node": "^10.9.0",
     "tsconfig-paths": "^4.2.0"
   }
--- a/apps/api/src/adapters/dev/dev-auth.adapter.ts
+++ b/apps/api/src/adapters/dev/dev-auth.adapter.ts
@@ -1,8 +1,14 @@
-import { Injectable, UnauthorizedException } from '@nestjs/common';
+import { Injectable, InternalServerErrorException, UnauthorizedException } from '@nestjs/common';
 import { AuthUser, IAuthProvider } from '../../ports/auth';
 
 @Injectable()
 export class DevAuthProvider implements IAuthProvider {
+  constructor() {
+    if (process.env.NODE_ENV === 'production') {
+      throw new InternalServerErrorException('DevAuthProvider must not run in production');
+    }
+  }
+
   async verifyToken(token: string): Promise<AuthUser> {
     if (!token) throw new UnauthorizedException();
     // Use the token value as the user ID so different tokens → different users,
--- a/apps/api/src/adapters/dev/dev-auth.adapter.ts
+++ b/apps/api/src/adapters/dev/dev-auth.adapter.ts
@@ -5,9 +5,13 @@ import { AuthUser, IAuthProvider } from '../../ports/auth';
 export class DevAuthProvider implements IAuthProvider {
   async verifyToken(token: string): Promise<AuthUser> {
     if (!token) throw new UnauthorizedException();
+    // Use the token value as the user ID so different tokens → different users,
+    // enabling per-user isolation tests without a real auth provider.
+    // DEV_USER_ID overrides this for local dev when a fixed identity is preferred.
+    const id = process.env.DEV_USER_ID ?? token;
     return {
-      id: process.env.DEV_USER_ID ?? 'dev-user',
-      email: process.env.DEV_USER_EMAIL ?? 'dev@example.com',
+      id,
+      email: process.env.DEV_USER_EMAIL ?? `${id}@dev.example.com`,
       provider: 'dev',
     };
   }
--- a/apps/api/src/adapters/factory/in-memory-repository-factory.ts
+++ b/apps/api/src/adapters/factory/in-memory-repository-factory.ts
@@ -1,0 +1,28 @@
+import { Injectable } from '@nestjs/common';
+import { AuthUser } from '../../ports/auth';
+import { IRepositoryFactory, RepositoryBundle } from '../../ports/factory';
+import { InMemoryCycleDashboardRepository } from '../in-memory/cycle-dashboard.adapter';
+import { InMemoryLiftingProgramSpecRepository } from '../in-memory/lifting-program-spec.adapter';
+import { InMemoryLiftRecordRepository } from '../in-memory/lift-record.adapter';
+import { InMemoryProgramPhilosophyRepository } from '../in-memory/program-philosophy.adapter';
+import { InMemoryTrainingMaxRepository } from '../in-memory/training-max.adapter';
+import { InMemoryWorkoutRepository } from '../in-memory/workout.adapter';
+
+@Injectable()
+export class InMemoryRepositoryFactory implements IRepositoryFactory {
+  private readonly bundles = new Map<string, RepositoryBundle>();
+
+  async forUser(user: AuthUser): Promise<RepositoryBundle> {
+    if (!this.bundles.has(user.id)) {
+      this.bundles.set(user.id, {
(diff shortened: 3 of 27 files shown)


brownm09:
## Review: [feat] Per-user repository factory — IRepositoryFactory adapter and controller threading

### Summary
The design is clean and the refactor is thorough — the factory pattern wires correctly, all 8 controllers follow the same idiom, and the E2E isolation test is a meaningful gate. Three blocking findings prevent merging as-is: the `SystemDbRepositoryFactory` TTL cache silently discards all user data every 5 minutes (the most severe), the migration SQL asserts a security property (ADR-014 encryption) that does not exist, and `DevAuthProvider`'s token-as-user-ID expansion creates impersonation risk if the adapter is ever misconfigured in production.

---

### Blocking Findings

**[correctness]** `apps/api/src/adapters/factory/system-db-repository-factory.ts:32–50`
The TTL cache evicts a user's `RepositoryBundle` after 5 minutes and calls `makeBundle` again, which returns a brand-new set of empty in-memory repos. Any data the user wrote during that session is silently discarded — the next request starts from zero. This affects every user in any deployment where `SYSTEM_DATABASE_URL` is set (the intended "production" path for v0.3).
**Fix:** Don't use TTL eviction on bundles that hold mutable in-memory state. Either cache bundles permanently keyed by `user.id` (matching `InMemoryRepositoryFactory`'s approach), or split the concerns — cache the adapter-type DB lookup separately with a TTL, but hold onto the bundle object indefinitely. If a real persistent adapter lands later, this constraint goes away naturally.

**[security]** `infra/migrations/001_create_user_data_source.sql:2`
The migration comment states `adapter_config stores adapter-specific credentials (encrypted at rest per ADR-014)`. ADR-014 does not exist in this repo (ADRs stop at ADR-012), and no encryption is implemented — `adapter_config` is plain JSONB. Any future adapter that stores OAuth tokens or database credentials in this column will write them as plaintext.
**Fix:** Remove the ADR-014 reference. Replace it with a `TODO: encryption required before non-in-memory adapters land` note (or equivalent), making the gap explicit rather than falsely resolved.

**[security]** `apps/api/src/adapters/dev/dev-auth.adapter.ts:8–13`
`DevAuthProvider` now mints `{ id: token }` for any non-empty bearer value. Previously the blast radius of accidental production use was a single shared `dev-user` identity; now it grants full impersonation of any user ID. There is no guard preventing this adapter from being wired in production — the selection appears to rely on `CLERK_SECRET_KEY` being absent, which is an implicit assumption.
**Fix:** Add a hard guard in the constructor: `if (process.env.NODE_ENV === 'production') throw new Error('DevAuthProvider must not run in production')`. This provides defense-in-depth against misconfiguration at zero runtime cost in dev.

---

### Non-Blocking Findings

**[reliability]** `apps/api/src/adapters/factory/in-memory-repository-factory.ts:6`
The `bundles` Map is unbounded. With the `DevAuthProvider` change, each distinct bearer token generates a new `user.id` and a new bundle entry — an unauthenticated caller hitting the dev server with random tokens accumulates unbounded memory.
**Suggestion:** Add an LRU cap (e.g., 1 000 entries) or document explicitly that this adapter is dev/test-only and must not run in processes exposed to the network.

**[maintainability]** `apps/api/src/adapters/factory/repository-factory.module.ts:10`
`useClass: process.env.SYSTEM_DATABASE_URL ? SystemDbRepositoryFactory : InMemoryRepositoryFactory` is evaluated once at file-import time, not at NestJS provider instantiation. Setting the env var after import (e.g., in a test `beforeEach`) will have no effect. This is a subtle gotcha distinct from standard `useClass` behavior.
**Suggestion:** Switch to `useFactory: () => new (process.env.SYSTEM_DATABASE_URL ? SystemDbRepositoryFactory : InMemoryRepositoryFactory)()` so the decision is evaluated at the expected time and both factories can be exercised in the same test process.

**[performance]** `apps/api/src/adapters/factory/system-db-repository-factory.ts:41–44`
Concurrent first-time requests for the same user both miss the cache and issue parallel `pool.query` calls. At v0.3 this is low risk (the query is cheap and the result is always `in-memory`), but once real adapter types land, N concurrent logins for the same user fan out N DB reads.
**Suggestion:** Coalesce in-flight loads by caching the `Promise<RepositoryBundle>` rather than the resolved value — a single-flight pattern that eliminates the fan-out with minimal code.

---

### Questions for Author

- **Isolation test seeding:** `aliceBaseline.json().find((m) => m.lift === 'Squat').weight` at line 1232 will throw if Alice's bundle is empty (fresh user, no training maxes). How does Alice's bundle get a Squat entry? Does an earlier test in the file seed training maxes using `Bearer user-alice` headers, or is `aliceOriginalSquat` expected to be `undefined` (in which case the `expect(bobSquat).toBe(aliceOriginalSquat)` assertion is `undefined === undefined`, and the real isolation check is solely the `not.toBe(999)` line)?

---

PR #153, opened 2026-05-02
[feat] Real database adapter — Prisma schema, repositories, and factory wiring

### LINKED ISSUE #150: [feat] Real database adapter (SystemDbRepositoryFactory + Prisma)
## Problem

`SystemDbRepositoryFactory` exists as a dispatch layer but has no real database implementation. All production traffic falls back to in-memory adapters — every server restart loses all data. There is no Prisma schema, no ORM setup, and no migration runner wired in.

## Goal

Implement the real-DB adapter so the app can persist data to Cloud SQL (Postgres) and survive restarts.

## Acceptance Criteria

- [ ] Prisma schema defined for all domain entities: programs, training maxes, lift records, body weight entries, cycle state
- [ ] `SystemDbRepositoryFactory` wired to Prisma-backed repository implementations for each port interface
- [ ] Migration runner integrated (Prisma Migrate or raw SQL via the existing `infra/migrations/` convention)
- [ ] `001_create_user_data_source.sql` migration reviewed and either superseded by Prisma or confirmed as the baseline
- [ ] `DATABASE_URL` env var documented in `.env.example`
- [ ] App boots and serves real data with `DATABASE_URL` set; falls back to in-memory when not set (dev default)

## Notes

- Port interfaces (`ILiftRecordRepository`, `ITrainingMaxRepository`, etc.) are already defined in `packages/core` — Prisma implementations should conform to those interfaces, not the other way around
- The `DevAuthProvider` / in-memory path must continue to work locally without a database configured

## Epic

CI/CD Foundation

Closes #???

brownm09:
## Summary

Closes #150

Introduces Prisma (v5) as the persistence layer for per-user lift data, replacing the all-in-memory default when a database is configured.

- **`apps/api/prisma/schema.prisma`** — three tables: `lift_record`, `training_max`, `cycle_dashboard`, each scoped by `userId` + unique constraints; initial migration SQL included
- **`apps/api/src/adapters/prisma/`** — four Prisma repository implementations conforming to the existing port interfaces (`ILiftRecordRepository`, `ITrainingMaxRepository`, `ICycleDashboardRepository`, `IWorkoutRepository`); `ILiftingProgramSpecRepository` and `IProgramPhilosophyRepository` stay in-memory (static data)
- **`PrismaRepositoryFactory`** — shared-DB factory: when `DATABASE_URL` is set, all users get Prisma-backed repos
- **`SystemDbRepositoryFactory`** — updated to wire Prisma repos for `adapter_type = 'postgres'` (per-user dispatch path); lazy `PrismaClient` init, torn down in `onModuleDestroy`
- **`repository-factory.module.ts`** — selection order: `DATABASE_URL` → Prisma; `SYSTEM_DATABASE_URL` → SystemDb dispatch; else → in-memory (dev default unchanged)
- **`apps/api/.env.example`** — documents both env vars with usage notes
- **`infra/migrations/001_create_user_data_source.sql`** confirmed as baseline for the system dispatch DB (`SYSTEM_DATABASE_URL`); orthogonal to the Prisma user-data DB (`DATABASE_URL`)

## Acceptance Criteria

- [x] Prisma schema defined for all per-user domain entities (programs, training maxes, lift records, cycle state)
- [x] `SystemDbRepositoryFactory` wired to Prisma-backed implementations for `adapter_type = 'postgres'`
- [x] Migration runner integrated (Prisma Migrate — `apps/api/prisma/migrations/`)
- [x] `001_create_user_data_source.sql` confirmed as baseline for system dispatch DB
- [x] `DATABASE_URL` documented in `apps/api/.env.example`
- [x] App boots with in-memory adapters when `DATABASE_URL` is unset; uses Prisma when set

## Test plan

- [x] `npm run build -w @lifting-logbook/api` — passes, no type errors
- [x] `npm test` — 70 passed, 1 pre-existing failure (`isolates adapter state between users`, also failing on `main` before this branch); no regressions introduced

## Notes

- Prisma 5 (not 6) required — Node 20.11.1 is below Prisma 6's Node 20.19+ floor
- `ILiftingProgramSpecRepository` and `IProgramPhilosophyRepository` are intentionally kept in-memory; they hold static/global program data, not user-specific records
- The `PrismaService.$connect()` is guarded by a `DATABASE_URL` check so the NestJS module starts cleanly without a database in local dev


brownm09:
## Review: [feat] Real database adapter — Prisma schema, repositories, and factory wiring

### Summary

This PR is well-structured and the hexagonal adapter pattern is applied consistently. The schema, migration, and factory wiring are clean. **Needs revision before merge** — three correctness issues were found: `saveTrainingMaxes` is non-atomic, `saveWorkout` uses record-supplied coordinates instead of the function arguments when inserting, and `SystemDbRepositoryFactory.getOrCreatePrisma()` silently initializes Prisma with an empty URL in the SYSTEM_DATABASE_URL-only deploy path (the exact configuration this path is designed to support).

---

### Blocking Findings

**[correctness]** `apps/api/src/adapters/prisma/training-max.repository.ts:23`
`saveTrainingMaxes` fires N independent `upsert` calls via `Promise.all`. A failure on the third of five maxes leaves the user with three updated and two stale, with no indication of partial success.
`saveWorkout` in the same PR correctly wraps its mutations in `$transaction` — the same pattern should apply here.
**Fix:** Wrap with `this.prisma.$transaction(maxes.map(m => this.prisma.trainingMax.upsert({...})))`.

---

**[correctness]** `apps/api/src/adapters/factory/system-db-repository-factory.ts:88`
`getOrCreatePrisma()` reads `process.env.DATABASE_URL ?? ''`. But `SystemDbRepositoryFactory` is only selected when `SYSTEM_DATABASE_URL` is set and `DATABASE_URL` is *not* set. Any user with `adapter_type='postgres'` will trigger Prisma initialization with an empty connection string, producing an opaque Prisma connection error on the first query instead of a clear misconfiguration message.
This is the configuration path that #150 specifically enables.
**Fix:** Introduce a separate env var (e.g. `USER_DATA_DATABASE_URL`) for the per-user postgres pool used by SystemDbRepositoryFactory, distinct from `DATABASE_URL` (which is the shared-DB path). Validate it is present at construction time and throw a descriptive `ConfigurationError` if missing. Update `.env.example` and the factory module selection comment accordingly.

---

**[correctness]** `apps/api/src/adapters/prisma/workout.repository.ts:40`
`saveWorkout(program, cycleNum, workoutNum, records)` deletes by the function arguments `(userId, program, cycleNum, workoutNum)` but creates using `r.cycleNum`/`r.workoutNum` from each record. If a caller passes records whose `cycleNum`/`workoutNum` differ from the function args, the delete removes the correct slot and the create inserts into a different slot — leaving stale data and potentially duplicating records in an adjacent workout.
**Fix:** Override `cycleNum`/`workoutNum` in the create payload with the function arguments (`cycleNum` and `workoutNum`), not `r.cycleNum`/`r.workoutNum`. This matches the intent of a replace-in-place operation.

---

### Non-Blocking Findings

**[reliability]** `apps/api/src/adapters/prisma/lift-record.repository.ts:44`
`updateLiftRecord` performs two sequential queries: `findUnique` to check existence, then `update` by PK. If the record is deleted between the two calls, the `update` throws Prisma's `P2025` (Record not found) rather than returning `null` as the caller expects, producing an unhandled exception in the API response path.
**Suggestion:** Use `update` directly on the composite unique key and catch `PrismaClientKnownRequestError` with code `P2025` to return `null`. This eliminates the TOCTOU window and one round-trip.

**[performance]** `apps/api/src/adapters/factory/repository-factory.module.ts:10`
`PrismaService` is registered as a provider unconditionally. In in-memory and SystemDb-only deploys, this instantiates a `PrismaClient` (and loads the Prisma engine binary) even though no Prisma repository is ever used.
**Suggestion:** Consider a conditional provider or lazy initialization so the Prisma engine is not loaded when `DATABASE_URL` is absent. Lower priority if startup time is acceptable.

**[maintainability]** `apps/api/src/adapters/prisma/workout.repository.ts:17`
`PrismaWorkoutRepository` re-implements the row→`LiftRecord` mapping inline rather than sharing the `rowToLiftRecord` function from `lift-record.repository.ts`. The function is currently unexported, so the duplication is forced.
**Suggestion:** Export `rowToLiftRecord` from `lift-record.repository.ts` and import it in `workout.repository.ts` to eliminate the two copies.

---

### Questions for Author

- Is `_config` (the `adapter_config` column value, passed but ignored in `makeBundle`) intentionally a no-op? If per-user connection strings are a future feature of the postgres dispatch path, the ignored parameter should either be documented as a stub or have a `// TODO` explaining when it will be wired.
- The `parseLiftRecordId` comment says lift names "may contain hyphens" — is it also intentional that program names may contain hyphens (e.g. `"5-3-1"`), relying on the full-prefix strip to disambiguate? If so, adding a test case for a hyphenated program name would guard against future ID format changes.

---

### Style Notes

The `getOrCreatePrisma()` method in `SystemDbRepositoryFactory` constructs a raw `PrismaClient` while `PrismaRepositoryFactory` receives an injected `PrismaService`. A brief inline comment explaining *why* these are separate client instances (different connection lifecycle — SystemDb manages its own Prisma teardown; PrismaService is NestJS-managed) would help the next reader.

---

*Reviewed by claude-opus-4-5 via Claude Code*

PR #155, opened 2026-05-02
[fix] Isolate InMemoryTrainingMaxRepository state between per-user factory bundles

### LINKED ISSUE #154: [fix] In-memory adapter state leaks between users (isolates adapter state between users)
## Problem

The e2e test \`isolates adapter state between users\` in \`src/programs/programs.e2e.spec.ts\` is failing. Bob reads training maxes for a program and receives a Squat entry (weight: 315) that belongs to seeded/shared state, violating per-user isolation.

\`\`\`
expect(bobSquat).toBeUndefined()
Received: {"dateUpdated": "2026-04-13", "lift": "Squat", "unit": "lbs", "weight": 315}
\`\`\`

The seeded weight (315) does not match Alice's write (999), so this is not a direct bleed-through from Alice's request — it appears to be either pre-seeded training max data in a shared \`InMemoryTrainingMaxRepository\` instance, or test setup state that persists across user bundles.

## Expected behaviour

Each user gets an independent \`InMemoryTrainingMaxRepository\` instance. A training max written for Alice must not appear when Bob reads the same program.

## Steps to reproduce

\`\`\`bash
\`\`\`

Fails at \`Programs HTTP (e2e, in-memory adapters) › isolates adapter state between users\`.

## Notes

- This failure exists on \`main\` before any changes from \`feat/issue-150-prisma-db-adapter\`
- The Prisma adapter work in #153 does not affect in-memory adapter isolation

brownm09:
## Summary

- `InMemoryTrainingMaxRepository` pre-seeded itself with the 5-3-1 training maxes in the field initializer, so every new per-user factory bundle started with the same 315 Squat — including users who never wrote any data (Bob)
- Added a `preSeed` constructor flag (`false` by default) to `InMemoryTrainingMaxRepository`
- `InMemoryRepositoryFactory.forUser()` passes `true` only for the dev seed user (`DEV_USER_ID ?? 'dev-token'`), preserving the existing e2e test fixture data while new users start with an empty repository

## Acceptance criteria

- [x] `isolates adapter state between users` passes: Alice writes weight 999, Bob reads the same program and gets an empty list (no Squat entry)
- [x] All existing tests continue to pass — the seed user ('dev-token') still starts with the 315 Squat fixture

## Test run

```
Test Suites: 13 passed, 13 total
Tests:       77 passed, 77 total
```

Closes #154

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Generate Prisma client
+        run: npx prisma generate --schema=apps/api/prisma/schema.prisma
+
       - name: Lint & Test
         run: npx turbo run lint test
 
--- a/apps/api/src/adapters/factory/in-memory-repository-factory.ts
+++ b/apps/api/src/adapters/factory/in-memory-repository-factory.ts
@@ -18,13 +18,14 @@ export class InMemoryRepositoryFactory implements IRepositoryFactory {
 
   async forUser(user: AuthUser): Promise<RepositoryBundle> {
     if (!this.bundles.has(user.id)) {
+      const preSeed = user.id === SEED_USER_ID;
       this.bundles.set(user.id, {
-        cycleDashboard: new InMemoryCycleDashboardRepository(),
-        liftingProgramSpec: new InMemoryLiftingProgramSpecRepository(),
-        liftRecord: new InMemoryLiftRecordRepository(),
+        cycleDashboard: new InMemoryCycleDashboardRepository(preSeed),
+        liftingProgramSpec: new InMemoryLiftingProgramSpecRepository(preSeed),
+        liftRecord: new InMemoryLiftRecordRepository(preSeed),
         programPhilosophy: new InMemoryProgramPhilosophyRepository(),
-        trainingMax: new InMemoryTrainingMaxRepository(user.id === SEED_USER_ID),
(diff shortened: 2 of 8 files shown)


brownm09:
## Review: [fix] Isolate InMemoryTrainingMaxRepository state between per-user factory bundles

### Summary

Ready to merge. The fix is minimal, targeted, and correct: replacing the always-seeded field initializer with a `preSeed` constructor flag eliminates the cross-user state leak while preserving existing test fixture behavior for the dev seed user. All 77 tests pass. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/adapters/in-memory/training-max.adapter.ts:7`
`@Injectable()` is retained on `InMemoryTrainingMaxRepository`, but the class is now only ever constructed manually by `InMemoryRepositoryFactory` — NestJS DI will never call it with the right `preSeed` value because that value is computed at runtime by the factory. The decorator is harmless but misleading: it implies the class can be safely injected directly, when in practice doing so always produces an empty (unseeded) repository.
**Suggestion:** Remove `@Injectable()` from this class to signal that it is a manually-constructed implementation detail of the factory, not a directly-injectable provider.

---

### Questions for Author

- Should the other in-memory adapters (`InMemoryCycleDashboardRepository`, `InMemoryLiftRecordRepository`, etc.) also receive a `preSeed` flag to support future fixture-seeded tests, or is training maxes the only adapter that carries fixture data?

---

PR #156, opened 2026-05-02
[test] Database integration tests against real Postgres

### LINKED ISSUE #151: [test] Database integration tests against real Postgres
## Problem

The existing e2e suite (`programs.e2e.spec.ts`) boots the full NestJS app but runs against in-memory adapters. Two gaps remain:

1. The real-DB path (`SystemDbRepositoryFactory` + Prisma adapters) is completely untested
2. The `PATCH /programs/:program/training-maxes` endpoint has no e2e coverage (helper is defined with an `_` prefix but no `it()` block uses it)

## Goal

Add an integration test suite that runs against a real Postgres instance, covering the full stack from HTTP request through Prisma to database and back.

## Acceptance Criteria

- [ ] A second e2e suite (`programs.db.e2e.spec.ts` or similar) runs against a real Postgres database
- [ ] Test database is spun up via Docker Compose or a test container; `DATABASE_URL` injected at test time
- [ ] Covers all read endpoints currently tested in the in-memory suite
- [ ] Covers write endpoints including `PATCH /programs/:program/training-maxes` (the currently skipped path)
- [ ] CI pipeline runs DB integration tests in a separate job with a Postgres service container
- [ ] In-memory e2e suite retained as a fast smoke test; DB suite is the authoritative integration gate

## Notes

- Prerequisite: issue #150 (real DB adapter) must land first
- The in-memory isolation test (`isolates adapter state between users`) should have a DB-backed equivalent verifying row-level isolation

## Epic

CI/CD Foundation

Closes #???

brownm09:
## Summary

- Adds `programs.db.e2e.spec.ts` — a second e2e suite that runs against a real Postgres instance via `PrismaRepositoryFactory`, mirroring the full in-memory spec plus two gaps it could not cover
- Adds explicit `PATCH /programs/:program/training-maxes` coverage (the `_patchJson` helper existed but had no `it()` block)
- Extends the user isolation test with a direct Prisma assertion verifying row-level separation in the DB
- Adds `docker-compose.test.yml` for local Postgres on port 5433
- Adds `test:db` script to `apps/api/package.json` for targeted local runs
- Adds `db-integration` CI job (separate from `lint-and-test`) with a Postgres service container that runs migrations then fires the DB suite

## Acceptance Criteria

- [x] `programs.db.e2e.spec.ts` runs against real Postgres when `DATABASE_URL` is set
- [x] Test database spun up via `docker-compose.test.yml`; `DATABASE_URL` injected at test time
- [x] Covers all read endpoints currently tested in the in-memory suite
- [x] Covers `PATCH /programs/:program/training-maxes` (previously untested write path)
- [x] CI pipeline runs DB integration tests in a separate job with a Postgres service container
- [x] In-memory e2e suite retained as a fast smoke test — suite skips cleanly when `DATABASE_URL` is unset

## Test Results

`npm test -w @lifting-logbook/api` (no DATABASE_URL):
- **77 passed, 16 skipped** (DB suite skips), 0 failures — in-memory smoke tests unaffected

DB suite (with DATABASE_URL + migrations applied):
- Runs via `npm run test:db -w @lifting-logbook/api`

Closes #151

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,6 @@ jobs:
           DATABASE_URL: postgresql://lifting:lifting@localhost:5432/lifting_test
 
       - name: DB Integration Tests
-        run: npm run test:db -w @lifting-logbook/api
+        run: npx turbo run test --filter=@lifting-logbook/api -- --testPathPattern=db\\.e2e
         env:
           DATABASE_URL: postgresql://lifting:lifting@localhost:5432/lifting_test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,3 +38,47 @@ jobs:
 
       - name: Validate analytics taxonomy
         run: node scripts/validate-analytics-taxonomy.mjs
+
+  db-integration:
+    name: DB Integration Tests
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:16-alpine
+        env:
+          POSTGRES_USER: lifting
+          POSTGRES_PASSWORD: lifting
+          POSTGRES_DB: lifting_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20.11.1'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Generate Prisma client
+        run: npx prisma generate --schema=apps/api/prisma/schema.prisma
+
+      - name: Run migrations
+        run: npx prisma migrate deploy --schema=apps/api/prisma/schema.prisma
+        env:
+          DATABASE_URL: postgresql://lifting:lifting@localhost:5432/lifting_test
+
+      - name: DB Integration Tests
+        run: npm run test:db -w @lifting-logbook/api
+        env:
+          DATABASE_URL: postgresql://lifting:lifting@localhost:5432/lifting_test
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -7,7 +7,8 @@
     "dev": "nest start --watch",
     "start": "node dist/main.js",
     "lint": "eslint src",
-    "test": "jest"
+    "test": "jest",
+    "test:db": "jest --testPathPattern=db\\.e2e"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.92.0",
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: [test] Database integration tests against real Postgres

### Summary

Ready to merge with one small fix. The test suite is well-structured: clean skip-without-DATABASE_URL mechanic, deterministic seed/cleanup lifecycle, correct PrismaClient isolation from the app's own connection, and a solid DB-layer assertion in the isolation test that goes meaningfully beyond what HTTP assertions alone can catch. One non-blocking but realistic CI reliability risk is worth addressing before merge — see the build-step finding below.

---

### Blocking Findings

None.

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/ci.yml` — `db-integration` job, `DB Integration Tests` step

The `db-integration` job calls `npm run test:db -w @lifting-logbook/api` directly, bypassing Turborepo. The `lint-and-test` job uses `npx turbo run lint test`, which respects the pipeline and ensures workspace packages like `@lifting-logbook/types` are built before the API tests run. If any of the API's transitive imports resolve to compiled `dist/` output (rather than TypeScript source mapped through ts-jest), the CI job will fail with a module-not-found error that doesn't reproduce locally after `npm ci`.

**Fix:** Add a turbo build step before the test run, or run the test through turbo with an appropriate filter: `npx turbo run test --filter=@lifting-logbook/api -- --testPathPattern=db\.e2e`.

---

**[maintainability]** `apps/api/src/programs/programs.db.e2e.spec.ts:60`

`patchJson` declares a third `headers` parameter that defaults to `AUTH`, but is never called with a custom header argument anywhere in the file — the isolation test uses `injectRaw` directly instead. The parameter is dead API surface that implies the helper is more general than it is.

**Suggestion:** Drop the third parameter; inline `headers: { 'content-type': 'application/json', ...AUTH }` like the other helpers, or remove the default and let the two call sites be explicit.

---

**[maintainability]** `apps/api/src/programs/programs.db.e2e.spec.ts:208`

The `fromCycleNum` test asserts only on `body.cycleNum` (the response) but doesn't follow up with a `GET /cycles/current` to confirm the DB-persisted state changed, unlike the `POST /cycles` test immediately above it which does verify via a GET. For an integration test against a real DB this asymmetry is noticeable — the whole point of the DB suite is confirming persistence, not just response shape.

**Suggestion:** Add a `GET /cycles/current` assertion after the POST, same pattern as the preceding test.

---

### Questions for Author

- Does `@lifting-logbook/types` publish compiled JS or TypeScript source as its package entry point? If the latter, the build-step concern above is a non-issue and the CI job is fine as written.
- The `fromCycleNum=1` test advances from cycle 2 and produces `cycleNum=2` again (same number). The comment says "re-pins — not normal advancement." Is this an intentional domain rule (re-pinning replaces the current cycle in place), or is it emergent from the test state and worth a more explicit assertion that `GET /cycles/current` still reports 2 afterward?

---

### Style Notes

`ALL_TEST_USERS` is declared at module scope but is only referenced inside `cleanTestUsers`. It's a minor coupling — if someone adds a test-scoped user and updates the array, they may not realize `cleanTestUsers` is the only consumer. Consider inlining it in `cleanTestUsers` or making it a parameter.

---

PR #157, opened 2026-05-03
[infra] Populate deployment infrastructure (GKE, Cloud Run, Terraform)

### LINKED ISSUE #149: [infra] Populate deployment infrastructure (GKE, Cloud Run, or Terraform)
## Problem

All three infra directories are empty stubs with only `.gitkeep` files:

- `infra/kubernetes/` — GKE Autopilot manifests and Helm charts: empty
- `infra/cloud-run/` — Cloud Run service YAML: empty
- `infra/terraform/` — Shared infrastructure (VPC, load balancer, DNS, IAM): empty

Only `infra/migrations/001_create_user_data_source.sql` exists. The app cannot be deployed anywhere.

## Goal

Choose a deployment target and populate the infra to produce a runnable, accessible deployment of the web app and API.

## Decision needed first

Before implementation, decide:
1. **Target platform:** Cloud Run (simpler, serverless) vs GKE Autopilot (more control, more ops overhead)
2. **Database:** Which managed database service to use (Cloud SQL, Neon, etc.) and how to wire the migration
3. **Auth:** Clerk is configured in the app — confirm the production Clerk app and env vars
4. **Domain:** Whether to set up a custom domain via the Terraform load balancer/DNS config or use a generated URL

## Acceptance Criteria

- [ ] Platform decision documented (Cloud Run vs GKE — update ADR or add a new one)
- [ ] `infra/terraform/` contains modules for the chosen shared infrastructure
- [ ] Service manifest(s) populated for the chosen runtime (Cloud Run YAML or GKE Helm chart)
- [ ] `DATABASE_URL` and Clerk env vars documented in a `.env.example` at the repo root
- [ ] CI/CD pipeline updated to deploy on merge to `main`
- [ ] App is reachable at a known URL with a working login flow

## Notes

- The Dockerfile exists in `apps/web/` and `apps/api/` — build is already containerized
- The migration SQL file is at `infra/migrations/001_create_user_data_source.sql` — needs a migration runner wired in

brownm09:
## Summary

- Populates all three previously-empty infra directories with production-ready infrastructure-as-code following ADR-009 (GKE Autopilot + Cloud Run) and ADR-014 (KMS credential encryption)
- Expands the existing build-only GitHub Actions workflow into a full 5-job staging → production pipeline with a manual approval gate
- Adds `docs/deploy.md` bootstrap guide and root `.env.example` covering all required environment variables

### Terraform (`infra/terraform/`)
- `main.tf` — VPC, Cloud SQL Postgres 15, Secret Manager secrets, Cloud KMS key ring, Artifact Registry, IAM service accounts, Workload Identity Federation (keyless GitHub Actions auth)
- `gke.tf` — GKE Autopilot cluster with private nodes and Workload Identity for the API pod
- `cloud-run.tf` — Cloud Run services for API and web (Serverless VPC Connector for Cloud SQL access)
- `variables.tf` / `outputs.tf` / `terraform.tfvars.{staging,production}` — full parameterisation with staging/production workspace support

### Helm charts (`infra/kubernetes/charts/`)
- `charts/api/` and `charts/web/` — Deployment, Service, HPA (min 2 / max 10 replicas, 70% CPU target), ConfigMap, ServiceAccount (annotated for GKE Workload Identity)
- `values/staging.yaml` and `values/production.yaml` — per-environment overrides

### Cloud Run (`infra/cloud-run/`)
- `api-service.yaml` and `web-service.yaml` — service reference definitions (deployed via `gcloud run deploy` in CI)

### CI/CD (`.github/workflows/deploy.yml`)
Five-job pipeline on push to `main`:
1. **terraform-staging** — `terraform apply` staging workspace
2. **build-images** — Docker build + push to Artifact Registry
3. **deploy-staging** — Helm upgrade (GKE) + Cloud Run deploy, secrets synced from Secret Manager
4. **smoke-test** — HTTP health checks against staging URLs
5. **deploy-production** — gated by `environment: production` (requires manual approval in GitHub), then Terraform + Helm + Cloud Run for production

## Acceptance criteria

- [x] Platform decision documented — ADR-009 (accepted); this PR implements it
- [x] `infra/terraform/` contains modules for the chosen infrastructure
- [x] Service manifests populated (Helm charts for GKE + Cloud Run YAMLs)
- [x] `DATABASE_URL` and Clerk env vars documented in `.env.example` at repo root
- [x] CI/CD pipeline updated to deploy on merge to `main`
- [ ] App is reachable at a known URL — requires bootstrap steps in `docs/deploy.md` (GCP project creation, Clerk account, GitHub secrets) — see that doc for the one-time setup sequence

## Testing

`npm test` passes (no TypeScript or application code was changed — infra-only PR).

The CI/CD pipeline itself will be exercised on the first push to `main` after the bootstrap steps in `docs/deploy.md` are completed.

Closes #149


brownm09:
## Review: [infra] Populate deployment infrastructure (GKE, Cloud Run, Terraform)

### Summary

This PR implements a well-structured, production-aware deployment pipeline that follows the ADRs correctly and handles the hard parts (keyless WIF auth, KMS, secret sync to K8s). The scope is appropriate for a single PR. However, **there are five blocking issues that will cause the pipeline to fail on first run**: a double-`https://` URL bug that breaks Cloud Run inter-service communication, a Helm values nesting mismatch that silently applies no overrides, a Cloud Run deploy-before-image-push ordering problem, and two security issues (public API IAM vs. `--no-allow-unauthenticated` conflict; open GKE control plane). All five are straightforward fixes. Four additional non-blocking observations follow.

---

### Blocking Findings

**[correctness]** `infra/terraform/cloud-run.tf:135`
The web service's `API_URL` is set to `"https://${google_cloud_run_v2_service.api.uri}"`. The `uri` attribute of a `google_cloud_run_v2_service` resource already returns the full HTTPS URL including scheme (e.g., `https://foo-abc123-uc.a.run.app`). This constructs `https://https://foo-abc123-uc.a.run.app`, an invalid URL that will cause every API call from the Cloud Run web container to fail with a connection error at runtime.
**Fix:** Remove the `https://` prefix: `value = google_cloud_run_v2_service.api.uri`

---

**[correctness]** `infra/kubernetes/values/staging.yaml:1` and `values/production.yaml:1`
Both env-specific values files wrap all keys under a top-level `api:` / `web:` namespace (e.g., `api.replicaCount: 1`, `api.image.repository: ""`). However, the CI/CD runs per-chart Helm upgrades: `helm upgrade api ./infra/kubernetes/charts/api --values infra/kubernetes/values/staging.yaml`. Helm applies the values file against the chart's own root namespace, where `replicaCount`, `image.repository`, etc. live at the top level. The `api:` wrapper key is unknown to the chart; every override silently falls back to `charts/api/values.yaml` defaults (2 replicas, `pullPolicy: Always`, `NODE_ENV: production`). Staging will deploy with production defaults.
**Fix:** Remove the `api:` / `web:` wrapper and flatten the file to match the chart's top-level key structure, or create per-chart files (`staging-api.yaml`, `staging-web.yaml`) with matching structure.

---

**[reliability]** `.github/workflows/deploy.yml:146-151` (terraform-staging job)
The `terraform-staging` job runs `terraform apply` with `-var="image_tag=SHA"`, embedding a specific git SHA into the Cloud Run service definition. This job runs before `build-images`. Cloud Run will attempt to pull the named image immediately on service creation or update; since the image with that SHA has not been pushed yet, the new revision will fail to start. Every deploy will log a revision failure until the subsequent CI steps push the image — and Cloud Run may not recover automatically if the previous revision was the first.
**Fix:** Two options: (a) remove `image_tag` from the Terraform-managed Cloud Run service config entirely and rely solely on the `gcloud run deploy --image=...` step in `deploy-staging` to set the image (the `gcloud run deploy` already handles this correctly); or (b) set `image_tag = "latest"` in Terraform and only pin the SHA in the `gcloud run deploy` step.

---

**[security]** `infra/terraform/cloud-run.tf:158-164` + `.github/workflows/deploy.yml:308-310`
Terraform creates `google_cloud_run_v2_service_iam_member.api_public` granting `allUsers` `roles/run.invoker` on the API service — making the API publicly accessible at the IAM layer. The CI/CD step then deploys with `--no-allow-unauthenticated`, which only prevents gcloud from *adding* a new allUsers binding; it does not remove the one Terraform already created. The API will be publicly reachable in both staging and production regardless of the deploy flag.
**Fix:** Decide which behavior is intended. If the API should be public (simplest for a portfolio project), remove `--no-allow-unauthenticated` from both deploy steps. If the API should require auth (web-to-API calls use a service identity token), remove `google_cloud_run_v2_service_iam_member.api_public` from `cloud-run.tf` and add token attachment to the web service.

---

**[security]** `infra/terraform/gke.tf:56-60`
`master_authorized_networks_config` allows `0.0.0.0/0` ("all") for both staging and production. The inline comment says "tighten for production" but no tightening is applied. In production, any host on the internet can reach the GKE API server; only credential-based auth remains as the barrier.
**Fix:** For production, restrict to GitHub Actions IP ranges (published at `https://api.github.com/meta` under `actions`) using the same conditional-expression pattern used elsewhere in this PR. Alternatively, accept the risk explicitly and update the comment to reflect that this is a known, intentional portfolio trade-off.

---

### Non-Blocking Findings

**[maintainability]** `infra/terraform/terraform.tfvars.staging:4` and `terraform.tfvars.production:4`
Both files commit `billing_account = "XXXXXX-XXXXXX-XXXXXX"` as a placeholder. Once filled in with a real value, the billing account ID will be tracked in version history.
**Suggestion:** Move `billing_account` out of the committed tfvars files and pass it via a `GCP_BILLING_ACCOUNT` GitHub secret using `-var="billing_account=..."` in CI/CD, or document that contributors must fill it locally and never commit the change.

---

**[reliability]** `.github/workflows/deploy.yml:331-343`
The smoke test step is named "Wait for GKE ingress IP" but actually polls the Cloud Run web URL. The GKE deployment (helm upgrade web) is not smoke-tested at all.
**Suggestion:** Rename the step to "Smoke test Cloud Run staging" and add a second check that curls the GKE ingress IP (available from `kubectl get ingress -n staging -o jsonpath=...`) to verify the GKE path is also healthy.

---

**[maintainability]** `infra/kubernetes/charts/web/values.yaml:132-134`
The ingress uses both `ingress.className: ""` and the deprecated annotation `kubernetes.io/ingress.class: "gce"`. In Kubernetes >= 1.18 the `ingressClassName` field takes precedence, but here it is empty, leaving the deprecated annotation as the sole class selector.
**Suggestion:** Set `ingress.className: "gce"` and remove the deprecated annotation, or leave both and add a comment explaining the reason (e.g., compatibility with older GKE versions).

---

**[maintainability]** `infra/kubernetes/charts/api/values.yaml:47-51`
`autoscaling.minReplicas: 2` and `replicaCount: 2` are both set at the chart default level. Once the values nesting bug is fixed, `minReplicas: 1` will be overridden for staging, but `replicaCount` (used when autoscaling is disabled) will still be 2.
**Suggestion:** Add `replicaCount: 1` to the staging values file alongside `autoscaling.minReplicas: 1` so the count is consistent regardless of whether autoscaling is enabled.

---

### Questions for Author

- **Cloud Run API access model:** Is the API intentionally public (anyone can call `/health`, workout endpoints, etc. directly without the web frontend), or should it only be reachable from the web service via service-to-service auth? The answer determines the correct fix for the IAM / `--no-allow-unauthenticated` conflict.
- **GKE smoke test gap:** The smoke test only checks Cloud Run staging. Is GKE deployment health verified anywhere, or is the intent that Cloud Run is the primary smoke signal with GKE failures surfaced only by Helm's `--wait --timeout=5m`?

---

### Style Notes

The Terraform files use a consistent, readable comment-banner style. The deploy.yml section headers match job names well. One minor inconsistency: `gke.tf` has a `# Kubernetes namespace` section with only a comment and no resource. This is reasonable for intent documentation but may surprise readers scanning for resources by section.

---

PR #163, opened 2026-05-04
[feat] Training max reduction flagging and multi-workout progression test

### LINKED ISSUE #141: [test] Integration test — multiple workouts in a cycle (training max progression)
## Context

The existing e2e suite (`programs.e2e.spec.ts`) exercises cycle reads and the cycle-advance write path but does not touch the workout logging flow at all. There is no test that verifies the progression math end-to-end: log sets → recalculate maxes → verify which lifts increased and which didn't.

## Goal

Add a stateful scenario to the e2e spec (in-memory adapters, Fastify inject) that simulates completing multiple workouts within a single cycle and verifies training max behaviour in both directions.

## Acceptance Criteria

- [ ] A test scenario logs bodyweight at session start, then logs sets for all exercises in at least two workouts within the same cycle
- [ ] At least one lift hits its target weight across the required sets (should trigger a max increase on recalculate)
- [ ] At least one lift does not hit its target (max should remain unchanged on recalculate)
- [ ] After calling `POST /programs/:program/training-maxes/recalculate`, the test asserts the correct per-lift outcome for both cases
- [ ] The workout endpoint (`GET /programs/:program/workouts/:workoutNum`) reflects correct completion status after sets are logged
- [ ] All assertions use deterministic fixture data — no floating-point tolerance hacks

## Notes

Fixture data must be designed so the pass/fail outcomes of the progression math are known in advance. The in-memory adapter is a singleton within the test suite; see the ordering warning in the existing `write operations` describe block.

Depends on: existing e2e harness (no new infrastructure needed).

brownm09:
## Summary

Addresses two review findings from PR #163:

1. **TM reduction flagging** — `updateMaxes` in `packages/core` now returns `{ maxes, flagged }`. If a completed set computes a new TM that would be *lower* than the current TM, the result is added to a `flagged: MaxReductionFlag[]` list rather than silently applied. The `recalculate` endpoint returns `RecalculateMaxesResponse { maxes, flagged }` so callers can surface the flags to the user for review.

2. **Full program-spec seed** — `seedProgramSpec()` in the in-memory adapter now covers all 3 weeks (3×5, 3×3, 5/3/1) for all 4 lifts (Squat, Bench Press, Deadlift, Overhead Press), matching the real data structure.

Also fixes a worktree-specific test resolution issue: npm workspace symlinks in the worktree resolve to the main repo's compiled `dist/` (not the worktree's modified source), causing stale behavior. `jest.config.js` `moduleNameMapper` now redirects `@lifting-logbook/core` and `@lifting-logbook/types` to the worktree's TypeScript source, consistent with the existing `@src/core` mapping.

## Acceptance Criteria

- [x] `updateMaxes` returns `{ maxes: TrainingMax[], flagged: MaxReductionFlag[] }`
- [x] TM reductions are flagged, not applied — idempotent on re-run
- [x] `RecalculateMaxesResponse` in `packages/types` includes `flagged` list
- [x] `POST /programs/:program/training-maxes/recalculate` returns both `maxes` and `flagged`
- [x] `seedProgramSpec` covers all 3 weeks × 4 lifts (12 entries total)
- [x] All existing tests pass; new tests added for flagged-reduction behavior

## Test Results

- `@lifting-logbook/core`: 124/124 passing
- `@lifting-logbook/api`: 81 passing, 2 pre-existing failures in `PrismaLiftRecordRepository` (unrelated, fail on main too)
- e2e `programs.e2e.spec.ts`: 21/21 passing

Closes #141


brownm09:
## Review: [test] Integration test — multiple workouts in a cycle (training max progression)

### Summary

Ready to merge. The shared-store refactor correctly mirrors Prisma's single-table semantics, and all six new tests use deterministic hardcoded values with a clear derivation trace in the header comment. No blocking findings. The one question worth confirming is whether the Squat TM regression (210→205) in the test is intentional — it's correct per the formula but may surprise future readers.

---

### Non-Blocking Findings

**[reliability]** `apps/api/src/programs/programs.e2e.spec.ts:301`
`find('Squat')!` uses a non-null assertion. If `recalculate` returns a response that omits a lift, the test throws `TypeError: Cannot read properties of undefined (reading 'weight')` rather than producing a readable Jest assertion failure.
**Suggestion:** Replace `find(lift)!.weight` with an explicit existence check (`expect(find('Squat')).toBeDefined()` before `.weight`) so failures report which lift was missing, not an uncaught exception.

**[maintainability]** `apps/api/src/programs/programs.e2e.spec.ts` (multi-workout scenario describe)
The new describe block runs after `write operations` and depends on the singleton adapter state it leaves behind (cycleNum=3). If any `write operations` test fails mid-sequence, the new tests will fail for an unrelated reason with no clear indication of the actual cause.
**Suggestion:** The existing pattern already accepts this tradeoff, so no change is strictly required. Adding a precondition guard (e.g., `expect(cycleNum).toBe(3)` as the first line of the first new test) would make the dependency explicit and dramatically narrow the failure surface.

---

### Questions for Author

- The Squat TM goes from 210 (after `write operations` recalculate) down to 205 after cycle-3 recalculate, because the cycle-3 set was logged at 200 lbs. `updateMaxes` applies `weight + increment` unconditionally — it doesn't floor at the current TM. Is this TM regression intentional (demonstrating the formula's behavior in a deload scenario)? The test is correct, but a future reader may find "5 reps hit → TM decreases" surprising.
- The new `seedProgramSpec` entries use `week: 1` only, matching the existing Squat/Bench entries. If future tests exercise multi-week program behavior (weeks 2 and 3), Deadlift and OHP will be absent from those weeks. Is this a conscious YAGNI call, or should the seed match the full week structure of the existing lifts?

---

### Style Notes

The comment block above the new describe is exactly the right level of documentation for a stateful, order-sensitive test — it spells out the preconditions, the expected values, and the gate logic that produces them. The inline `// → progression gate passes/fails` annotations within individual tests continue the same style. Well done.

---

brownm09:
## Review: [feat] Training max reduction flagging and multi-workout progression test

### Summary

Ready to merge. The core domain change — `updateMaxes` returning `{ maxes, flagged }` and the guard preventing silent TM reductions — is correct and well-tested. The shared-store refactor in the in-memory adapters is a clean fix. One non-blocking issue worth noting: `startNewCycle` silently discards the `flagged` list, so users advancing a cycle won't be told about blocked reductions. That's a scope call, not a bug — flagging it as a question.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/cycle-generation.service.ts:44`
`startNewCycle` destructures only `{ maxes: newMaxes }` from `updateMaxes`, silently dropping the `flagged` list. Any TM reductions blocked during cycle advancement are never surfaced — the cycle advances but the user gets no indication that some maxes weren't updated.
**Suggestion:** Either return `flagged` from `startNewCycle` (and propagate through the controller + response type), or add a comment explicitly calling this a deliberate policy. Right now it looks like an oversight.

**[maintainability]** `packages/types/src/api.ts:754`
`RecalculateMaxesResponse.flagged` uses an inline type `Array<{ lift: LiftName; currentWeight: number; proposedWeight: number }>` rather than referencing `MaxReductionFlag` from `@lifting-logbook/core`. The two shapes can drift silently — the core type uses `string` for `lift` while the API type narrows to `LiftName`.
**Suggestion:** Re-export or re-type `MaxReductionFlag` through `packages/types` and reference it here, or at minimum add a comment tying the two shapes together. The coupling should be explicit.

---

### Questions for Author

- Is the `jest.config.js` `moduleNameMapper` fix intentional to land on `main`? Once merged, all `@lifting-logbook/api` tests (not just worktree runs) will resolve `@lifting-logbook/core` and `@lifting-logbook/types` from TypeScript source rather than compiled `dist/`. That's probably desirable (tests always reflect current source), but it does mean the dist is no longer exercised by the test suite.
- Should `startNewCycle` surface flagged reductions to the caller, or is the policy "recalculate is the dedicated surface for flags, cycle advancement just doesn't apply them"? If the latter, it's worth encoding that as a code comment so future readers don't wonder whether the drop was accidental.

---

### Style Notes

The inline one-liner format in the expanded `seedProgramSpec()` works well here — 12 entries at full verbosity would be hard to scan, and the week-grouping comments (`// Week 1: 3×5`) make the intent clear.

---

PR #182, opened 2026-05-05
[feat] Training max history with PR flags and source tracking

### LINKED ISSUE #174: [feat] Training max history with PR flags and source tracking
## Problem

The current `training_max` table stores only the latest max per (user, program, lift). The mockup at `docs/mockups/user-journeys.html` (Cycle Dashboard journey → Training Maxes view) shows a timeline / list history with PR flags, test-week vs. program-cycle source labels, and "goal met" markers. None of these are queryable today.

## Goal

Persist a history row each time a training max is updated (cycle advance, recalculate, manual edit) and expose it via API so the web UI can render the timeline / list views from the mockup.

## Architectural decision

**New `training_max_history` table** rather than deriving from `lift_record` rows. Rationale: lift_record doesn't carry PR or test-vs-program annotations; computing them heuristically would be fragile. The "Programs: catalog metadata only" and "Training max history: new table" decisions are documented in the implementation plan at `~/.claude/plans/c-users-brown-downloads-lifting-logbook-flickering-volcano.md`. An ADR will be added under `docs/adr/` as part of this PR.

### Database
Add `training_max_history` model to `apps/api/prisma/schema.prisma`:
- Fields: `id`, `userId`, `program`, `lift`, `weight`, `reps`, `date`, `isPR` (bool), `source` ("test" | "program"), `goalMet` (bool), `createdAt`
- Indexes: `(userId, program)` and `(userId, program, lift)`
- Migration via `prisma migrate dev`

### API
- New types in `packages/types/src/api.ts`: `TrainingMaxHistoryEntryResponse`, `TrainingMaxHistoryResponse`, `UpdateTrainingMaxHistoryRequest`
- New endpoints under `apps/api/src/programs/`:
  - `GET /programs/:program/training-maxes/history?lift=&source=&isPR=` → `TrainingMaxHistoryResponse`
  - `PATCH /programs/:program/training-maxes/history/:id` → toggle `isPR` / `goalMet`
- New port + Prisma adapter + in-memory adapter (mirror existing pattern under `apps/api/src/adapters/`)
- Wire `cycle-generation.service.ts` to append a history row for each lift whose max changes during cycle advance / recalculate

### Web
- New component `apps/web/app/settings/training-maxes/MaxHistory.tsx`
  - Timeline / list toggle
  - Filter pills: All / PRs / Test Week / Goal Met
  - Per-lift filter
  - PR toggle button per entry → calls `PATCH /history/:id`
- Embed `<MaxHistory />` below the existing form on the training-maxes settings page

### Tests
- Unit: history controller and service `.spec.ts` files (mocked repositories)
- E2E in-memory (`programs.e2e.spec.ts`):
  - Cycle advance writes history rows with `source: 'program'`
  - `?lift=` filter
  - `?isPR=` filter
  - `PATCH /history/:id` toggles `isPR`
- E2E DB (`programs.db.e2e.spec.ts`): same scenarios against Postgres
- ADR added under `docs/adr/` capturing the new-table-vs-derive decision

## Acceptance Criteria
- [ ] `training_max_history` Prisma model + migration committed
- [ ] `GET` and `PATCH` history endpoints implemented and tested
- [ ] `cycle-generation.service.ts` writes history rows on max changes
- [ ] In-memory adapter for the new repository so existing e2e tests continue to pass
- [ ] `MaxHistory.tsx` panel renders on `/settings/training-maxes` and matches the mockup behavior
- [ ] ADR under `docs/adr/` documenting the new-table decision
- [ ] `npm test` passes

## References
- Implementation plan: `~/.claude/plans/c-users-brown-downloads-lifting-logbook-flickering-volcano.md` (Workstream B)
- Mockup: `docs/mockups/user-journeys.html` — Cycle Dashboard journey, "Training Maxes" view
- Related: closed #172 (mockup file added)

brownm09:
## Summary

- Adds a `training_max_history` table that records a row each time a training max changes (cycle advance or recalculate), capturing the weight, source (`test` | `program`), and user-toggleable `isPR` / `goalMet` flags
- Implements `ITrainingMaxHistoryRepository` port with Prisma and in-memory adapters following the existing hexagonal adapter pattern (ADR-002, ADR-004)
- `CycleGenerationService.startNewCycle` and `recalculateMaxes` diff prev vs. new maxes and append history entries; source is `'test'` when `currentWeekType === 'test'`
- New endpoints: `GET /programs/:program/training-maxes/history` (with `?lift`, `?source`, `?isPR` filters) and `PATCH /programs/:program/training-maxes/history/:id` for toggling flags
- `MaxHistory` client component added to the settings/training-maxes page with timeline/list toggle and filter pills (All / PRs / Test Week / Goal Met)
- ADR-017 documents the new-table-vs-derive decision

## Acceptance Criteria

- [x] History row written on every cycle advance (source = test or program based on weekType)
- [x] History row written on every recalculate (source = program)
- [x] `GET /programs/:program/training-maxes/history` returns entries with lift, weight, unit, date, isPR, source, goalMet
- [x] `?lift=`, `?source=`, `?isPR=` query filters work
- [x] `PATCH /programs/:program/training-maxes/history/:id` toggles isPR / goalMet; returns 404 for unknown id
- [x] MaxHistory component renders below TrainingMaxesForm on the settings page
- [x] Unit tests, in-memory e2e, and DB e2e tests added
- [x] ADR-017 written with References section

## Test Results

`npm test` — 14 suites passed, 1 skipped (DB e2e; requires DATABASE_URL), 102 tests passed, 20 skipped.

Closes #174

### DIFF
--- a/apps/api/prisma/migrations/20260505000000_add_training_max_history/migration.sql
+++ b/apps/api/prisma/migrations/20260505000000_add_training_max_history/migration.sql
@@ -1,0 +1,22 @@
+-- CreateTable
+CREATE TABLE "training_max_history" (
+    "id" TEXT NOT NULL,
+    "userId" TEXT NOT NULL,
+    "program" TEXT NOT NULL,
+    "lift" TEXT NOT NULL,
+    "weight" DOUBLE PRECISION NOT NULL,
+    "reps" INTEGER NOT NULL DEFAULT 1,
+    "date" TIMESTAMP(3) NOT NULL,
+    "isPR" BOOLEAN NOT NULL DEFAULT false,
+    "source" TEXT NOT NULL,
+    "goalMet" BOOLEAN NOT NULL DEFAULT false,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    CONSTRAINT "training_max_history_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "training_max_history_userId_program_idx" ON "training_max_history"("userId", "program");
+
+-- CreateIndex
+CREATE INDEX "training_max_history_userId_program_lift_idx" ON "training_max_history"("userId", "program", "lift");
--- a/apps/api/prisma/migrations/20260505000001_add_training_max_source_check/migration.sql
+++ b/apps/api/prisma/migrations/20260505000001_add_training_max_source_check/migration.sql
@@ -1,0 +1,4 @@
+-- AddCheckConstraint
+ALTER TABLE "training_max_history"
+ADD CONSTRAINT "training_max_history_source_check"
+CHECK (source IN ('test', 'program'));
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -55,3 +55,21 @@ model CycleDashboard {
   @@index([userId, program])
   @@map("cycle_dashboard")
 }
+
+model TrainingMaxHistory {
+  id        String   @id @default(cuid())
+  userId    String
+  program   String
+  lift      String
+  weight    Float
+  reps      Int      @default(1)
+  date      DateTime
+  isPR      Boolean  @default(false)
+  source    String
+  goalMet   Boolean  @default(false)
+  createdAt DateTime @default(now())
+
+  @@index([userId, program])
+  @@index([userId, program, lift])
+  @@map("training_max_history")
+}
--- a/apps/api/src/adapters/factory/in-memory-repository-factory.ts
+++ b/apps/api/src/adapters/factory/in-memory-repository-factory.ts
@@ -7,6 +7,7 @@ import { InMemoryLiftingProgramSpecRepository } from '../in-memory/lifting-progr
 import { InMemoryLiftRecordRepository } from '../in-memory/lift-record.adapter';
 import { InMemoryProgramPhilosophyRepository } from '../in-memory/program-philosophy.adapter';
 import { InMemoryTrainingMaxRepository } from '../in-memory/training-max.adapter';
+import { InMemoryTrainingMaxHistoryRepository } from '../in-memory/training-max-history.adapter';
 import { InMemoryWorkoutRepository } from '../in-memory/workout.adapter';
 import { SEED_PROGRAM, seedLiftRecords } from '../in-memory/fixtures';
 
@@ -32,6 +33,7 @@ export class InMemoryRepositoryFactory implements IRepositoryFactory {
         liftRecord: new InMemoryLiftRecordRepository(sharedRecords),
         programPhilosophy: new InMemoryProgramPhilosophyRepository(),
         trainingMax: new InMemoryTrainingMaxRepository(preSeed),
+        trainingMaxHistory: new InMemoryTrainingMaxHistoryRepository(),
         workout: new InMemoryWorkoutRepository(sharedRecords),
       });
     }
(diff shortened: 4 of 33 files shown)


brownm09:
## Review: [feat] Training max history with PR flags and source tracking

### Summary

Ready to merge with two changes addressed first. The feature is well-structured: the port/adapter/controller layering is clean, both adapters implement the interface correctly, the compile-time contract test and e2e tests cover the key paths, and the ADR is thorough. The two blocking findings are a missing input-validation step on the PATCH endpoint (risk: bad input reaches the DB instead of getting a 400) and a floating-point comparison in `buildHistoryEntries` that can produce spurious history entries after `recalculateMaxes`. Three non-blocking items are noted below.

---

### Blocking Findings

**[security]** `apps/api/src/programs/training-max-history.controller.ts` — `@Patch` endpoint

`@Body() body: UpdateTrainingMaxHistoryRequest` is passed directly to `updateHistoryEntry`. TypeScript types are erased at runtime; NestJS does not validate or strip fields without a `ValidationPipe`. A client can send `{ isPR: "yes" }` and the Prisma adapter will attempt to write a string to a boolean column (resulting in a Prisma error and a 500 instead of a 400), or send `{ isPR: true, goalMet: false, weight: 9999 }` and the extra fields will be silently ignored only because both adapters happen to destructure only `isPR`/`goalMet` — a fragile guarantee.

**Fix:** Add `@IsBoolean() @IsOptional()` class-validator decorators to `UpdateTrainingMaxHistoryRequest` in `packages/types/src/api.ts` and enable `ValidationPipe({ whitelist: true, forbidNonWhitelisted: true })` globally (in `main.ts`) or on this controller. This makes the 400 response explicit and the safety invariant declarative rather than accidental.

---

**[correctness]** `apps/api/src/programs/cycle-generation.service.ts` — `buildHistoryEntries`

```ts
.filter((m) => prevMap.get(m.lift) !== m.weight)
```

`weight` is `DOUBLE PRECISION` in the schema. The `updateMaxes` domain function derives new maxes by multiplying lift-record weights by percentages, which can introduce floating-point drift (e.g., `252.4999999999` vs `252.5`). Comparing with strict `!==` will treat these as a change and append a spurious history entry. In `recalculateMaxes` this fires against freshly computed weights without a normalization step, so the risk is active on every recalculate.

**Fix:** Apply the same rounding `updateMaxes` uses (or `Math.round(w * 10) / 10` for 0.1-lb precision) to both `prevMap` weights and `newMaxes` weights before diffing. Alternatively, round the result inside `updateMaxes` itself so the domain always returns normalized values.

---

### Non-Blocking Findings

**[reliability]** `apps/api/src/programs/cycle-generation.service.ts` — `recalculateMaxes`

`recalculateMaxes` passes `new Date()` as the history entry date while `startNewCycle` passes `newCycle.cycleDate`. History from a recalculate will show today's wall-clock date regardless of when the cycle began, breaking chronological ordering when cycle date ≠ today and making tests non-deterministic.

**Suggestion:** Fetch `cycleDate` from the already-loaded `dashboard` object (`dashboard.cycleDate`) and pass it to `buildHistoryEntries` in the `recalculateMaxes` path.

---

**[maintainability]** `apps/api/prisma/schema.prisma` — `source` field

`source` is declared as `String` — no Prisma enum, no `CHECK` constraint in the migration SQL. The application-layer cast `r.source as 'test' | 'program'` silently lies to TypeScript if an invalid value ever reaches the DB (direct SQL, future third source written before a schema change, etc.).

**Suggestion:** Promote to a Prisma `enum TrainingMaxSource { test program }` (generates both a DB `CREATE TYPE` and TypeScript types), eliminating the cast entirely. Requires a new migration.

---

**[maintainability]** In-memory vs Prisma adapter update semantics

In-memory uses `update.isPR ?? current.isPR` (nullish coalescing); Prisma uses `...(update.isPR !== undefined && { isPR: update.isPR })` (undefined-guard spread). Both produce the same result for booleans today, but using different idioms for the same operation makes the in-memory adapter a less reliable test double.

**Suggestion:** Standardize on the `!== undefined` spread in both adapters so they share identical update semantics.

---

### Questions for Author

- Do other controllers in this codebase apply `@UseGuards()` explicitly, or is auth enforced by a global guard in `main.ts`/`AppModule`? The `@CurrentUser() user?` optional pattern with `user!` is consistent with the existing controllers, but confirming the global guard covers this new controller would be worth a quick check — the PATCH endpoint writes user-scoped data and must not be reachable unauthenticated.

---

PR #188, opened 2026-05-07
[feat] Lift metadata — muscle groups, substitutions, foundational lift

### LINKED ISSUE #179: [feat] Lift metadata — muscle groups, substitutions, foundational lift
## Problem

The mockup at `docs/mockups/user-journeys.html` shows an **Edit Lift** screen where users customize a lift's target muscle groups, substitution exercises, and foundational lift. No DB storage, no API, no UI exists today.

### Database
New `lift_metadata` model in `apps/api/prisma/schema.prisma`:
- `id`, `userId`, `lift`, `muscleGroups` (String[]), `substitutions` (String[]), `foundational` (String)
- Unique on `(userId, lift)`

### API
Lift metadata is per-user, not per-program — new top-level controller:
- `GET /lifts/:lift/metadata`
- `PATCH /lifts/:lift/metadata` — body: `{ muscleGroups?, substitutions?, foundational? }`
- Port + Prisma adapter + in-memory adapter

### Web
New component `apps/web/components/LiftEditor.tsx`:
- Name field, muscle groups (comma-separated input), substitutions (comma-separated), foundational
- Save / Reset to Foundational buttons
- Used inline from the manage-lifts / lift-db flows

### Tests
- E2E in-memory + DB:
  - PATCH metadata → GET → verify muscleGroups / substitutions
  - User isolation (User A sets metadata, User B gets defaults)

## Acceptance Criteria
- [ ] `lift_metadata` model + migration
- [ ] GET / PATCH endpoints
- [ ] `LiftEditor.tsx` component renders and persists
- [ ] All new tests pass

## References
- Plan workstream G
- Mockup: `docs/mockups/user-journeys.html` — "Edit Lift" view
- Depends on: issue for Manage Lifts (mounts the editor)

brownm09:
## Summary

- Adds per-user, per-lift metadata storage: muscle groups, substitutions, and foundational lift designation
- New `lift_metadata` Prisma model with `GET /lifts/:lift/metadata` (returns defaults when no record) and `PATCH /lifts/:lift/metadata` (upserts partial fields)
- Full hexagonal architecture stack: `ILiftMetadataRepository` port → `InMemoryLiftMetadataRepository` adapter → `PrismaLiftMetadataRepository` adapter → registered in all three factory implementations
- New `LiftsModule` NestJS module registered in `AppModule`
- Web: `LiftEditor` client component with comma-separated inputs + Save/Reset; edit page at `manage-lifts/edit/[lift]/`; "Edit" link added per lift row in `ManageLiftsActions`
- Also fixes pre-existing `exactOptionalPropertyTypes` errors in strength-goals code and fills a missing `workoutLiftOverride` gap in `SystemDbRepositoryFactory`

## Acceptance Criteria

- [x] `GET /lifts/:lift/metadata` returns defaults (`muscleGroups: []`, `substitutions: []`, `foundational: ""`) when no record exists
- [x] `PATCH /lifts/:lift/metadata` persists and returns the updated metadata
- [x] Partial PATCH retains unpatched fields
- [x] Unauthenticated requests return 401
- [x] User isolation: User B gets empty defaults when User A has metadata for the same lift
- [x] Edit page accessible from manage-lifts via "Edit" button per lift
- [x] Save → fields persist on returning to edit the same lift
- [x] Reset clears the form fields client-side

## Test Results

All 171 API tests pass including 5 new lift metadata e2e tests:
- `GET /lifts/:lift/metadata without auth returns 401`
- `GET /lifts/:lift/metadata returns defaults when no record exists`
- `PATCH then GET persists all fields`
- `Partial PATCH retains unpatched fields`
- `User isolation: User B gets separate metadata`

Closes #179

### DIFF
--- a/apps/api/prisma/migrations/20260507000001_add_lift_metadata/migration.sql
+++ b/apps/api/prisma/migrations/20260507000001_add_lift_metadata/migration.sql
@@ -5,7 +5,7 @@ CREATE TABLE "lift_metadata" (
     "lift" TEXT NOT NULL,
     "muscleGroups" TEXT[],
     "substitutions" TEXT[],
-    "foundational" TEXT NOT NULL DEFAULT '',
+    "foundational" BOOLEAN NOT NULL DEFAULT false,
 
     CONSTRAINT "lift_metadata_pkey" PRIMARY KEY ("id")
 );
--- a/apps/api/prisma/migrations/20260507000001_add_lift_metadata/migration.sql
+++ b/apps/api/prisma/migrations/20260507000001_add_lift_metadata/migration.sql
@@ -1,0 +1,17 @@
+-- CreateTable
+CREATE TABLE "lift_metadata" (
+    "id" TEXT NOT NULL,
+    "userId" TEXT NOT NULL,
+    "lift" TEXT NOT NULL,
+    "muscleGroups" TEXT[],
+    "substitutions" TEXT[],
+    "foundational" TEXT NOT NULL DEFAULT '',
+
+    CONSTRAINT "lift_metadata_pkey" PRIMARY KEY ("id")
+);
+
+-- CreateIndex
+CREATE INDEX "lift_metadata_userId_idx" ON "lift_metadata"("userId");
+
+-- CreateUniqueIndex
+CREATE UNIQUE INDEX "lift_metadata_userId_lift_key" ON "lift_metadata"("userId", "lift");
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -119,3 +119,16 @@ model WorkoutLiftOverride {
   @@index([userId, program, cycleNum, workoutNum])
   @@map("workout_lift_override")
 }
+
+model LiftMetadata {
+  id            String   @id @default(uuid())
+  userId        String
+  lift          String
+  muscleGroups  String[]
+  substitutions String[]
+  foundational  String   @default("")
+
+  @@unique([userId, lift])
+  @@index([userId])
+  @@map("lift_metadata")
+}
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -126,7 +126,7 @@ model LiftMetadata {
   lift          String
   muscleGroups  String[]
   substitutions String[]
-  foundational  String   @default("")
+  foundational  Boolean  @default(false)
 
   @@unique([userId, lift])
   @@index([userId])
--- a/apps/api/src/adapters/factory/in-memory-repository-factory.ts
+++ b/apps/api/src/adapters/factory/in-memory-repository-factory.ts
@@ -3,6 +3,7 @@ import { LiftRecord } from '@lifting-logbook/core';
 import { AuthUser } from '../../ports/auth';
 import { IRepositoryFactory, RepositoryBundle } from '../../ports/factory';
 import { InMemoryCycleDashboardRepository } from '../in-memory/cycle-dashboard.adapter';
+import { InMemoryLiftMetadataRepository } from '../in-memory/lift-metadata.adapter';
 import { InMemoryLiftingProgramSpecRepository } from '../in-memory/lifting-program-spec.adapter';
 import { InMemoryLiftRecordRepository } from '../in-memory/lift-record.adapter';
 import { InMemoryProgramPhilosophyRepository } from '../in-memory/program-philosophy.adapter';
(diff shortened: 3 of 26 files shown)


brownm09:
## Review: [feat] Lift metadata — muscle groups, substitutions, foundational lift

### Summary
The architecture is solid — the port/adapter/controller stack is well-structured, the conditional-spread pattern for `exactOptionalPropertyTypes` is applied consistently, and the e2e test coverage is good. Three blocking issues need addressing before merge: a race condition in the Prisma upsert, a double-decode bug in the edit page, and missing input-size validation on the DTO. None require significant rework.

---

### Blocking Findings

**[correctness]** `apps/api/src/adapters/prisma/lift-metadata.repository.ts:upsertMetadata`
`upsertMetadata` calls `getMetadata` (a `SELECT`) then builds `next` from the result, then calls `upsert`. Two concurrent PATCHes for the same `(userId, lift)` — e.g., one patching `muscleGroups` and another patching `substitutions` — will both read the same `existing`, both compute a `next` that overwrites the other's field with the stale value, and the second write silently drops the first client's change. This is also an unnecessary extra round-trip per PATCH.
**Fix:** Build `update` and `create` objects conditionally from `patch` alone — only include keys present in the patch — and let `create` supply defaults inline. The `upsert` then becomes a single atomic operation with no pre-read needed:
```typescript
const update = {
  ...(patch.muscleGroups !== undefined ? { muscleGroups: patch.muscleGroups } : {}),
  ...(patch.substitutions !== undefined ? { substitutions: patch.substitutions } : {}),
  ...(patch.foundational !== undefined ? { foundational: patch.foundational } : {}),
};
const row = await this.prisma.liftMetadata.upsert({
  where: { userId_lift: { userId: this.userId, lift } },
  update,
  create: { userId: this.userId, lift, muscleGroups: [], substitutions: [], foundational: '', ...update },
});
```

**[correctness]** `apps/web/app/cycle/[cycleNum]/workout/[workoutNum]/detail/manage-lifts/edit/[lift]/page.tsx:712`
`decodeURIComponent(liftParam)` is called on a value that Next.js App Router already URL-decodes before handing it to the page component. For most lift names this is a no-op, but a lift name containing a literal `%` (e.g. `"100%"`), or any malformed percent sequence, causes `decodeURIComponent` to throw a `URIError` — crashing the render instead of reaching `notFound()`.
**Fix:** Remove the `decodeURIComponent` call. Next.js dynamic segments are pre-decoded:
```typescript
const lift = liftParam; // already decoded by Next.js
```
If a guard is still wanted, wrap in `try/catch` and call `notFound()` on `URIError`.

**[correctness]** `apps/api/src/lifts/patch-lift-metadata.dto.ts`
`muscleGroups` and `substitutions` accept an unbounded `string[]` with no per-item length cap, no array size cap, and no `foundational` length cap. A client can PATCH a 10 MB array that persists to `TEXT[]` columns; every subsequent GET pays that cost. There is also no trimming or normalization, so `"Quads"` and `" quads "` become distinct stored values.
**Fix:** Add size constraints and trim in the controller or a pipe:
```typescript
@IsArray()
@ArrayMaxSize(20)
@IsString({ each: true })
@MaxLength(100, { each: true })
@IsOptional()
muscleGroups?: string[];
```
Apply the same pattern to `substitutions`, and add `@MaxLength(200)` to `foundational`.

---

### Non-Blocking Findings

**[reliability]** `apps/api/src/programs/programs.e2e.spec.ts` (partial-PATCH test)
The test `"PATCH is partial — unpatched fields retain prior values"` asserts `muscleGroups` equals `['Quads', 'Glutes']`, but those values were written only by the preceding `"PATCH persists values"` test. If tests are ever filtered with `--testNamePattern` or the describe order shifts, this test fails for a reason unrelated to the partial-PATCH behavior.
**Suggestion:** Seed the metadata inside this test (PATCH all three fields first, then PATCH only `substitutions`, then GET and assert all three) so it can run in isolation.

**[maintainability]** `apps/web/components/LiftEditor.tsx:849–851`
Muscle groups and substitutions are stored in React state as comma-joined strings and split on save. Any entry containing a comma (e.g., `"Chest, Anterior Deltoid"` as a single group) silently splits into two entries. It also ties the component to an ad-hoc serialization format that diverges from the API contract (`string[]`).
**Suggestion:** Keep state as `string[]` with explicit add/remove handlers, or at minimum document the comma constraint in the input placeholder. For the current scope this is workable; just worth revisiting if the UI evolves.

---

### Questions for Author

- `SystemDbRepositoryFactory._buildInMemoryBundle()` — does it return a fresh `InMemoryLiftMetadataRepository` instance on every call, or is the in-memory bundle cached/shared across users? The in-memory adapter's store is keyed only by `lift` (no `userId`), which works correctly only if each user gets a distinct instance. If the bundle is shared in the in-memory fallback path, the user isolation test would pass (because both factories create per-user instances) but the actual runtime behavior for the SystemDb in-memory path could silently expose User A's data to User B.
- What is the intended semantic of `foundational`? The schema uses `String` (defaulting to `""` as falsy), but the field name implies a boolean flag. If it's "which lift this is the foundational movement for", the string type is appropriate — but if it's "is this lift foundational (yes/no)", a `Boolean` column would eliminate empty-string-as-false ambiguity and future inconsistency.

---

PR #198, opened 2026-05-07
[chore] Local dev setup: docker-compose Postgres and dev auth wiring

### LINKED ISSUE #197: chore: Local dev setup — docker-compose Postgres and dev auth wiring
## Problem

The application cannot run end-to-end locally because two things are missing:

1. **No `docker-compose.yml`** — `apps/api/.env.example` and the root `.env.example` both document `docker compose up db` as the way to start a local Postgres instance, but the file was never committed.

2. **Web app sends no auth headers in local dev** — `apps/web/lib/api.ts` and `client-api.ts` return empty auth headers when the API URL is not a Cloud Run HTTPS endpoint. The API's `AuthGuard` rejects all requests with 401 because no Bearer token is present. A `DevAuthProvider` already exists in the API (`apps/api/src/adapters/dev/clerk-auth.adapter.ts`) and is automatically selected when `CLERK_SECRET_KEY` is unset, but the web never sends a token to trigger it.

## Acceptance Criteria

- [ ] `docker-compose.yml` at repo root starts a `postgres:16` container on port 5432
- [ ] `apps/web/lib/api.ts` reads `DEV_AUTH_TOKEN` and adds `Authorization: Bearer \<token>` when `!isCloudRun`
- [ ] `apps/web/lib/client-api.ts` reads `NEXT_PUBLIC_DEV_AUTH_TOKEN` and adds the same header when not in Cloud Run context
- [ ] `apps/web/.env.example` documents both new variables
- [ ] `npm run dev` starts both servers without errors when `.env` files are configured
- [ ] All existing tests pass

brownm09:
## Summary

- Adds `docker-compose.yml` (postgres:16 on port 5432) so `docker compose up db` works as documented in the `.env.example` files
- Wires `DEV_AUTH_TOKEN` / `NEXT_PUBLIC_DEV_AUTH_TOKEN` into `apps/web/lib/api.ts` and `client-api.ts` so the web sends a Bearer token in local dev, letting the API's `DevAuthProvider` identify the caller without Clerk
- Updates `apps/api/.env.example` and `apps/web/.env.example` to document the new variables and explain the auth fallback

## Root cause

`DevAuthProvider` (selected when `CLERK_SECRET_KEY` is unset) was already wired in the API, but `getAuthHeaders()` returned `{}` for non-Cloud Run URLs so every request arrived unauthenticated → 401.

## Test plan

- [x] 124 API + core unit tests pass (`npm test -w @lifting-logbook/api -w @lifting-logbook/core`)
- [x] `jest-environment-jsdom` failure in `@lifting-logbook/web` is pre-existing on `main` (confirmed by running on `main` before changes)
- [ ] Manual: `docker compose up db`, copy `.env.example` → `.env` in both apps, `npm run dev`, navigate to `http://localhost:3000`

Closes #197

### DIFF
--- a/.claude/hook-config.json
+++ b/.claude/hook-config.json
@@ -16,5 +16,7 @@
     "v0.1 — Foundation",
     "v0.2 — Core API",
     "v0.3 — Client Applications"
-  ]
+  ],
+  "status_field_id": "PVTSSF_lAHOAjEKvM4BTuEFzhA7F7E",
+  "done_option_id": "98236657"
 }
--- a/README.md
+++ b/README.md
@@ -54,12 +54,50 @@ runs only core's tests, with no knowledge of the rest of the monorepo.
 
 - Node.js >= 20.11.1 (use `.nvmrc`: `nvm use`)
 - npm >= 10 (bundled with Node 20)
+- Docker (for local Postgres)
 
-### Install
+### Local Development
 
 ```sh
+# 1. Install dependencies
 npm install
+
+# 2. Start Postgres
+docker compose up db -d
+
+# 3. Create env files from examples
+cp apps/api/.env.example apps/api/.env
+cp apps/web/.env.example apps/web/.env
 ```
+
+Edit `apps/api/.env` and set:
+```
+DATABASE_URL=postgresql://postgres:postgres@localhost:5432/lifting_logbook
+SYSTEM_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/lifting_logbook
+# Leave CLERK_SECRET_KEY unset — DevAuthProvider is used automatically
+DEV_USER_ID=dev-user
+DEV_USER_EMAIL=dev@example.com
+```
+
+Edit `apps/web/.env` and set:
+```
+API_URL=http://localhost:3004
+NEXT_PUBLIC_API_URL=http://localhost:3004
+NEXT_PUBLIC_DEFAULT_PROGRAM=5-3-1
+DEV_AUTH_TOKEN=dev-user
+NEXT_PUBLIC_DEV_AUTH_TOKEN=dev-user
+```
+
+```sh
+# 4. Apply database migrations
+cd apps/api && npx prisma migrate dev && cd ../..
+
+# 5. Start all dev servers
+npm run dev
+```
+
+- Web: http://localhost:3000
+- API: http://localhost:3004
 
 ### Common Commands
 
--- a/README.md
+++ b/README.md
@@ -73,7 +73,6 @@ cp apps/web/.env.example apps/web/.env
 Edit `apps/api/.env` and set:
 ```
 DATABASE_URL=postgresql://postgres:postgres@localhost:5432/lifting_logbook
-SYSTEM_DATABASE_URL=postgresql://postgres:postgres@localhost:5432/lifting_logbook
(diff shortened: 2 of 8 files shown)


brownm09:
## Review: [chore] Local dev setup: docker-compose Postgres and dev auth wiring

### Summary

Nearly ready to merge — the docker-compose addition and server-side auth fix (`api.ts`) are clean. One blocking correctness gap: two client-side functions in `client-api.ts` call `fetch` directly instead of going through `clientFetch`, so they won't receive the dev auth headers and will still 401 in local dev. Fix those two functions and this is good to go.

---

### Blocking Findings

**[correctness]** `apps/web/lib/client-api.ts:64` and `:129`

`rescheduleWorkout` (line 64) and `deleteLiftOverride` (line 129) call the native `fetch` directly rather than routing through `clientFetch`. The PR updates `clientFetch` to inject `Authorization: Bearer <devToken>` in local dev, but these two functions are not covered by that change — they remain unauthenticated. A user trying to reschedule a workout or delete a lift override in local dev will get a 401 from `AuthGuard`, which is exactly the failure mode this PR is intended to fix.

**Fix:** Route both through `clientFetch`. For `rescheduleWorkout`:
```typescript
export async function rescheduleWorkout(...): Promise<void> {
  await clientFetch<void>(
    `/programs/${encodeURIComponent(program)}/cycles/${cycleNum}/workouts/${workoutNum}/reschedule`,
    { method: 'PATCH', headers: { 'Content-Type': 'application/json' }, body: JSON.stringify({ newDate }), cache: 'no-store' },
  );
}
```
Same pattern for `deleteLiftOverride` — wrap the existing `fetch` call in `clientFetch<void>`.

---

### Non-Blocking Findings

**[reliability]** `scripts/dev-setup.sh:97`

`npx prisma migrate dev --name init` with a hardcoded `--name` flag will create a new migration file named `init` (or `TIMESTAMP_init`) if any schema drift is detected. On a fresh clone of a repo that already has committed migrations this is harmless, but if the schema has pending changes the script silently creates a migration as a side effect of setup. Consider `prisma migrate dev` without `--name` (Prisma will prompt for a name only if it needs to create one) or `prisma migrate deploy` if the intent is always to apply existing migrations without creating new ones.

**[reliability]** `docker-compose.yml`

No `healthcheck` is defined on the `db` service. The setup script compensates with a manual `pg_isready` loop, so this doesn't affect script users. But anyone who runs `docker compose up db --wait` (or a future CI step using that flag) won't get the wait behaviour they'd expect. Consider adding:
```yaml
healthcheck:
  test: ["CMD-SHELL", "pg_isready -U postgres"]
  interval: 5s
  timeout: 5s
  retries: 6
```

---

### Questions for Author

- `dev-setup.sh` patches `SYSTEM_DATABASE_URL` to point at `lifting_logbook` (same as `DATABASE_URL`). Is that intentional — i.e., the system database and user database share the same Postgres DB in local dev — or should a second database be created? A short comment in `.env.example` would clarify for future contributors.

---

### Style Notes

`scripts/dev-setup.sh` uses `echo -e` throughout for colour output. `echo -e` is bash-specific (not POSIX), but the shebang is `#!/usr/bin/env bash` so this is consistent. No change needed.

The fallback default in `client-api.ts` (`?? 'http://localhost:3001'`) still references port 3001 rather than 3004. This is pre-existing and out of scope here, but worth a follow-up issue since mismatched defaults can silently route to a dead port.

---

PR #217, opened 2026-05-09
[chore] Replace head-based 100% sampling with tail-based policy

### LINKED ISSUE #210: [chore] Revisit OTel head-based 100% sampling policy
ADR-018 locked in head-based 100% sampling at start because there is no production traffic to inform a more nuanced policy. This issue tracks the revisit.

**Trigger:** revisit when **either** of the following holds:
- Sustained ingest crosses **1,000 spans/minute** averaged over a 24-hour window
- Monthly Grafana Cloud trace ingest cost exceeds **$25**

**Scope of the revisit:**
- Move from head-based to tail-based sampling, or to a head-based ratio < 100%
- Decide on tail-based policies (always-keep errors, always-keep slow, sampled fraction of clean traces)
- Document the new policy as an addendum or superseding ADR

Refs [#199](https://github.com/brownm09/lifting-logbook/issues/199), [ADR-018](../blob/main/docs/adr/ADR-018-observability-stack.md)

brownm09:
## Summary

Implements ADR-020: replaces the head-based 100% sampling policy (locked in by ADR-018 at project start) with tail-based sampling at the OTel Collector level.

**Three policies (OR logic — keep if any matches):**
- `errors` — always keep any trace containing a span with `STATUS_CODE_ERROR`
- `slow_traces` — always keep any trace with total duration ≥ 1,000 ms
- `probabilistic` — keep 20% of all remaining clean traces

**Key design choices:**
- Tail-based (Collector-side) rather than head-based ratio (SDK) so errors and slow traces are always retained regardless of sampling rate
- `decision_wait: 10s` to see the full `apps/web → apps/api → Postgres` span chain before deciding
- `OTEL_TAIL_SAMPLE_RATE` env var makes the rate tunable per-environment without code changes
- Local dev overrides to 100% in `docker-compose.yml` — no debugging friction

## Changes

| File | Change |
|---|---|
| `infra/observability/otel-collector.yaml` | Added `tail_sampling` processor, wired into traces pipeline |
| `infra/kubernetes/charts/otel-collector/templates/configmap.yaml` | Mirrored processor config; added `OTEL_TAIL_SAMPLE_RATE` env to ConfigMap |
| `infra/kubernetes/charts/otel-collector/values.yaml` | Added `tailSampleRate: "20"` default |
| `docker-compose.yml` | Added `OTEL_TAIL_SAMPLE_RATE: "100"` for local dev |
| `docs/adr/ADR-020-tail-based-sampling-policy.md` | New ADR documenting the decision |
| `docs/README.md` | Added ADR-019 and ADR-020 rows (ADR-019 was missing) |
| `docs/adr-references.md` | Added two ADR-020 references |

## Test plan

- [x] 114 unit tests pass (`npm test -w @lifting-logbook/core -w @lifting-logbook/api`)
- [x] Pre-existing e2e failures (`nestjs-pino`, `@opentelemetry/sdk-trace-node` missing packages) are unrelated to these changes
- [ ] Smoke: `docker compose up otel-collector` — confirm collector starts without errors
- [ ] Smoke: send test trace to `localhost:4318`, confirm it appears in Grafana Tempo

Closes #210

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,7 @@ services:
       OTEL_COLLECTOR_LOKI_ENDPOINT: http://loki:3100/loki/api/v1/push
       OTEL_COLLECTOR_LOKI_AUTH_HEADER: ""
       OTEL_TAIL_SAMPLE_RATE: "100"   # local dev: keep every trace
+      OTEL_DECISION_WAIT: "2s"      # local dev: short buffer so spans appear in Tempo promptly
     ports:
       - "4317:4317"   # OTLP gRPC
       - "4318:4318"   # OTLP HTTP
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,7 @@ services:
       OTEL_COLLECTOR_OTLP_AUTH_HEADER: ""
       OTEL_COLLECTOR_LOKI_ENDPOINT: http://loki:3100/loki/api/v1/push
       OTEL_COLLECTOR_LOKI_AUTH_HEADER: ""
+      OTEL_TAIL_SAMPLE_RATE: "100"   # local dev: keep every trace
     ports:
       - "4317:4317"   # OTLP gRPC
       - "4318:4318"   # OTLP HTTP
--- a/docs/README.md
+++ b/docs/README.md
@@ -150,6 +150,8 @@ monorepo/
 | [ADR-016](adr/ADR-016-cycle-planning-agent.md)                | Cycle Planning Agent                                          | Accepted |
 | [ADR-017](adr/ADR-017-training-max-history-table.md)          | Training Max History — Dedicated Table vs. Derived            | Accepted |
 | [ADR-018](adr/ADR-018-observability-stack.md)                 | Observability Stack — OpenTelemetry + Grafana Cloud           | Accepted |
+| [ADR-019](adr/ADR-019-slo-methodology.md)                    | SLO Methodology — Burn-Rate Alerting over Threshold Alerting  | Accepted |
+| [ADR-020](adr/ADR-020-tail-based-sampling-policy.md)         | Tail-Based Sampling Policy — Errors Always, Slow Always, 20% Clean | Accepted |
 
 ---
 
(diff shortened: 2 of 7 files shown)


brownm09:
## Review: [chore] Replace head-based 100% sampling with tail-based policy

### Summary

Ready to merge with one suggestion. The three-policy tail-sampling configuration is correct — processor ordering (`tail_sampling` → `batch`) is right, env var wiring through the Helm chart is confirmed (`-config` ConfigMap → `envFrom: configMapRef` in `daemonset.yaml` → runtime OTel env substitution), and OR-logic across policies is exactly what's needed. The ADR is the strongest in the project: it calls out the p50/p75 bias, metric undercounting, and memory window tradeoffs explicitly. One non-blocking reliability concern and one question worth resolving before the smoke test runs.

---

### Non-Blocking Findings

**[reliability]** `infra/observability/otel-collector.yaml:23`, `infra/kubernetes/charts/otel-collector/templates/configmap.yaml:36`

`sampling_percentage: ${env:OTEL_TAIL_SAMPLE_RATE}` has no inline fallback. If the env var is absent, the OTel Collector v0.104.0 config parser substitutes an empty string, producing a YAML parse error and a Collector startup failure. The standard paths (docker-compose and Kubernetes) both set the value, so this only bites someone running the standalone `otel-collector.yaml` directly without a `.env` file. OTel's `${env:VAR}` syntax does not support `:-default` fallback notation.

**Suggestion:** Add a comment to both YAML files directly above the `sampling_percentage` line noting the env var is required and where the defaults live. Example:
```yaml
# Required: set by docker-compose (100) or Helm tailSampleRate value (default: 20)
sampling_percentage: ${env:OTEL_TAIL_SAMPLE_RATE}
```
Alternatively, add `OTEL_TAIL_SAMPLE_RATE` to any `.env.example` file at the repo root if one exists.

---

### Questions for Author

- **`decision_wait: 10s` applies in local dev too** — with `OTEL_TAIL_SAMPLE_RATE=100`, the Collector still buffers every trace for 10 seconds before exporting, so spans appear in local Grafana Tempo with ~10 s lag. Is that latency acceptable for local debugging sessions, or would a shorter `decision_wait` (e.g., 2 s) make sense as a local dev override alongside the rate override?

---

### Style Notes

`docs/README.md` ADR table: the two new rows (ADR-019 and ADR-020) use fewer padding spaces in the `|` columns compared to the surrounding rows, so the Markdown source misaligns in editors that render raw text. The rendered table is fine. Worth normalizing if the file is opened for any other reason.

---

PR #221, opened 2026-05-10
[feat] Validate cycle dashboard before reschedule and lift-override writes

### LINKED ISSUE #193: [feat] Validate program/cycle exists before upserting reschedule and lift-override entries
## Problem

The reschedule controller (\`PATCH /programs/:program/cycles/:cycleNum/workouts/:workoutNum/reschedule\`) and the lift-override controller (\`POST /programs/:program/cycles/:cycleNum/workouts/:workoutNum/lift-overrides\`) both upsert their respective overrides using URL path parameters directly, without first verifying that the program exists in the user's cycle dashboard or that the requested \`cycleNum\` is valid.

This means:
- A client receives no feedback (no 404) when rescheduling a cycle that has never been started.
- Dangling overrides can accumulate for cycles that don't exist, and are silently ignored at read time.
- The behavior is inconsistent with write endpoints that *do* validate the dashboard (cycle-advance, training-max write, etc.).

Surfaced during the issue #181 integration-test review.

## Proposed Change

Both controllers should validate that the program exists in the user's cycle dashboard before upserting the override. If the program is not found, return 404 (same as the cycle-advance and training-max endpoints).

Optionally, validate that \`cycleNum\` is the current cycle or within a reasonable range — though this is a softer constraint and may be deferred.

## Acceptance Criteria
- [ ] \`PATCH .../reschedule\` returns 404 if the program is not in the user's cycle dashboard
- [ ] \`POST .../lift-overrides\` returns 404 if the program is not in the user's cycle dashboard
- [ ] Existing e2e isolation tests updated: user-isolation tests for both controllers must seed a cycle dashboard for the test user (or restructured to use the seeded dev-token user)
- [ ] All existing tests pass

## Notes

The user-isolation tests in \`programs.e2e.spec.ts\` (added in issue #181) currently rely on the absence of dashboard validation — fresh tokens with no dashboard can POST overrides. Those tests will need to be updated when this issue is addressed.

brownm09:
## Summary

- `PATCH .../reschedule` and `POST .../lift-overrides` both now call `getCycleDashboard(program)` before upserting. If the program is not in the user's cycle dashboard, `ProgramNotFoundError` propagates to `DomainNotFoundFilter` → 404, consistent with cycle-advance and training-max endpoints.
- Unit test mocks for both controllers updated to include `cycleDashboard` in the factory mock.
- E2e isolation tests restructured: the writing party is now the pre-seeded dev user (has a cycle dashboard) rather than a fresh user with no dashboard; the read-only party remains a fresh token.

## Acceptance Criteria

- [x] `PATCH .../reschedule` returns 404 if the program is not in the user's cycle dashboard
- [x] `POST .../lift-overrides` returns 404 if the program is not in the user's cycle dashboard
- [x] Isolation tests updated (dev-token user as the writer, fresh token for the read-only assertion)
- [x] All existing tests pass

## Testing

`npm test -w @lifting-logbook/api` — 173 passed, 31 skipped (DB suite), 0 failed.

Closes #193

### DIFF
--- a/apps/api/src/programs/manage-lifts.controller.spec.ts
+++ b/apps/api/src/programs/manage-lifts.controller.spec.ts
@@ -1,6 +1,7 @@
 import { BadRequestException, NotFoundException } from '@nestjs/common';
 import { Test, TestingModule } from '@nestjs/testing';
 import { LIFT_NAMES } from '@lifting-logbook/types';
+import { ICycleDashboardRepository } from '../ports/ICycleDashboardRepository';
 import { ILiftingProgramSpecRepository } from '../ports/ILiftingProgramSpecRepository';
 import { IWorkoutLiftOverrideRepository } from '../ports/IWorkoutLiftOverrideRepository';
 import { IRepositoryFactory } from '../ports/factory';
@@ -19,6 +20,8 @@ describe('ManageLiftsController', () => {
   let liftOverrideRepoB: jest.Mocked<IWorkoutLiftOverrideRepository>;
   let specRepoA: jest.Mocked<ILiftingProgramSpecRepository>;
   let specRepoB: jest.Mocked<ILiftingProgramSpecRepository>;
+  let dashboardRepoA: jest.Mocked<ICycleDashboardRepository>;
+  let dashboardRepoB: jest.Mocked<ICycleDashboardRepository>;
   let factory: jest.Mocked<IRepositoryFactory>;
 
   beforeEach(async () => {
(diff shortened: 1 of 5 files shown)


brownm09:
## Review: [feat] Validate cycle dashboard before reschedule and lift-override writes

### Summary

Merge with minor changes. The production code change is minimal and correct — one `await cycleDashboard.getCycleDashboard(program)` call added to each controller, consistent with the established pattern in `cycle-advance` and `training-max` endpoints. One blocking gap: neither controller's unit-test spec nor the e2e suite adds a case that actually asserts 404 when the dashboard is missing, so the central AC ("returns 404 if program is not in the user's cycle dashboard") goes unverified by any test.

---

### Blocking Findings

**[correctness]** `reschedule.controller.spec.ts` / `manage-lifts.controller.spec.ts` — no test for the new 404 path

Both specs add `dashboardRepoA`/`B` mocks that always resolve successfully, but never add a case where `getCycleDashboard` rejects. The e2e restructuring removed the fresh-user write tests (which previously confirmed no-validation behavior) without adding replacement tests that assert a 404 is returned when the dashboard is absent. The PR body's AC checkbox "returns 404 if program is not in the user's cycle dashboard" is unchecked by any test.

**Fix:** In each controller spec, add one test case that calls `dashboardRepoA.getCycleDashboard.mockRejectedValue(new ProgramNotFoundError(...))` and expects the call to `rejects.toBeInstanceOf(NotFoundException)` (or verify the raw error propagates). Mirrors the existing "throws 404 when program is not found" spec tests for the spec-check path.

---

### Non-Blocking Findings

**[reliability]** `programs.e2e.spec.ts` — reschedule e2e uses a hardcoded `RESCHEDULE_URL` (cycle 1) while the lift-override test now dynamically fetches `cycleNum`

The lift-override isolation test was updated to query `GET /programs/${PROGRAM}/cycles/current` and use the live `cycleNum`, but the reschedule isolation test still targets a fixed URL (comment says "cycle 1, workout 1"). Both tests now share `AS_ALICE = AUTH` (the dev user), whose current cycle may have advanced past cycle 1 as earlier tests run. If the dev user's cycle dashboard reflects a later cycle, a reschedule request to cycle 1 could produce a 404 or silent no-op depending on the implementation.

**Suggestion:** Apply the same dynamic `cycleNum` lookup to the reschedule isolation test that was applied to the lift-override test.

---

### Questions for Author

- The `DELETE .../lift-overrides` path is intentionally left without dashboard validation ("per issue scope" per the PR body). Is this a deliberate long-term design choice (idempotent delete = no harm to validate) or a deferral that warrants a follow-up issue? A brief comment on the `deleteOverride` method would preserve the intent.

---

PR #230, opened 2026-05-12
[docs] Fix nvm use instructions for nvm-windows compatibility

### LINKED ISSUE #229: Fix nvm use instructions for nvm-windows compatibility
nvm-windows does not support reading `.nvmrc` automatically — that is a macOS/Linux nvm feature. On Windows Git Bash, `nvm use` without a version argument fails with "A version argument is required but missing."\n\nAll documentation and scripts tell users to run `nvm use` bare, which silently fails on Windows.\n\n**Fix:** replace every bare `nvm use` with `nvm use $(cat .nvmrc)` — this works in Git Bash on Windows (the shell expands `.nvmrc` to `20.11.1` before nvm-windows sees it) and is identical in behavior on macOS/Linux.\n\n**Files affected:**\n- `docs/onboarding.md` (3 occurrences)\n- `README.md` (1 occurrence)\n- `CLAUDE.md` (1 occurrence)\n- `scripts/dev-setup.sh` (1 occurrence)

brownm09:
## Summary

- Replaces all bare `nvm use` calls with `nvm use $(cat .nvmrc)` across docs and scripts
- `nvm-windows` does not support reading `.nvmrc` automatically; bare `nvm use` fails with "A version argument is required but missing"
- `nvm use $(cat .nvmrc)` works in Git Bash on Windows and is identical in behavior on macOS/Linux

## Files changed

| File | Occurrences fixed |
|------|-------------------|
| `docs/onboarding.md` | 3 |
| `README.md` | 1 |
| `CLAUDE.md` | 1 |
| `scripts/dev-setup.sh` | 1 |

## Acceptance Criteria

- [x] No bare `nvm use` (without a version argument) remains in any docs or scripts
- [x] Instructions work on Windows (Git Bash + nvm-windows) without error
- [x] `bash -n scripts/dev-setup.sh` passes (syntax clean)

## Test Plan

```bash
bash -n scripts/dev-setup.sh
grep -rn "nvm use" README.md docs/onboarding.md CLAUDE.md scripts/dev-setup.sh
# All lines show nvm use $(cat .nvmrc) — no bare nvm use
```

Closes #229

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,7 +9,7 @@ Include in your opening brief only: the issue you are working on, current branch
 ## Platform & Environment
 
 - **OS:** Windows 11, Git Bash terminal
-- **Node:** 20.11.1 (managed by nvm for Windows; `.nvmrc` is set — run `nvm use` at session start if not already active)
+- **Node:** 20.11.1 (managed by nvm for Windows; `.nvmrc` is set — run `nvm use $(cat .nvmrc)` at session start if not already active)
 - **Package manager:** npm (workspaces)
 - **`jq` is NOT available.** Use `node -e` with a temp file in the working directory for JSON parsing:
   ```bash
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ For a guided walkthrough from clone to first PR, see [docs/onboarding.md](docs/o
 
 ### Prerequisites
 
-- Node.js >= 20.11.1 (use `.nvmrc`: `nvm use`)
+- Node.js >= 20.11.1 (use `.nvmrc`: `nvm use $(cat .nvmrc)`)
 - npm >= 10 (bundled with Node 20)
 - Docker (for local Postgres and the observability stack)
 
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: [docs] Fix nvm use instructions for nvm-windows compatibility

### Summary

Ready to merge. The fix is correct, applied consistently across all 6 occurrences in 4 files, and the PR description includes a solid test plan that was verified. No blocking findings. One non-blocking observation about shell expansion behavior in `dev-setup.sh` that is actually better UX than literal text — worth being aware of.

---

### Non-Blocking Findings

**[maintainability]** `scripts/dev-setup.sh:22`
`$(cat .nvmrc)` inside a double-quoted string in the `die` message is evaluated by bash at runtime — the error the user sees will read `Run: nvm use 20.11.1`, not `Run: nvm use $(cat .nvmrc)`. Since the script `cd`s to the repo root before this line runs, `.nvmrc` is always readable. This is actually better UX (user sees a concrete, copy-pasteable command), but it differs from the intent as written — a future author might change it to a literal string and lose the dynamic resolution.
**Suggestion:** Add a brief inline comment (`# expands to version at runtime`) so the behavior is self-documenting.

---

PR #235, opened 2026-05-12
[fix] Annotate .map() callback to fix implicit any in workout-lift-override repository

### LINKED ISSUE #224: fix: implicit `any` type error in workout-lift-override.repository.ts (Prisma types need regeneration)
## Problem

`npm test` fails with a TypeScript error in a fresh worktree:

```
apps/api/src/adapters/prisma/workout-lift-override.repository.ts(20,28): error TS7006: Parameter 'r' implicitly has an 'any' type.
```

The error is on the `.map((r) => ({` call over a raw Prisma query result. The same code exists on `main` — this is not a regression introduced by a recent PR.

## Root cause (suspected)

Prisma client types are not generated in a fresh worktree after `npm ci` because `DATABASE_URL` is not set at install time (no `.env` file). Without generated types, TypeScript cannot infer the row shape returned by Prisma queries.

## Steps to reproduce

1. Clone the repo (or open a fresh worktree)
2. Run `npm ci` without setting `DATABASE_URL`
3. Run `npm test` — TS error fires

## Fix

Run `scripts/dev-setup.sh` first (sets `DATABASE_URL` via `.env`), then `npx prisma generate`, then `npm test`. Or add an explicit return-type annotation to the repository method so the build is not blocked by missing generated types.

## Acceptance Criteria

- [ ] `npm test` passes in a fresh worktree after following the documented setup steps
- [ ] Either Prisma types are regenerated correctly via setup, or the repository method carries an explicit return-type annotation that removes the implicit-`any` dependency on generated types

brownm09:
## Summary

- Adds an explicit inline type annotation to the `.map((r) => ...)` callback parameter in `PrismaWorkoutLiftOverrideRepository.getOverrides()`
- Removes the TS7006 implicit-any error that fires in a fresh worktree when Prisma types have not been generated

## Root cause

`npm ci` succeeds without `DATABASE_URL` set, but no `postinstall`/`prepare` script runs `prisma generate`. Without generated types, TypeScript cannot infer the shape of rows returned by `findMany`, making `r` implicitly `any`. Adding `r: { lift: string; action: string; replacedBy: string | null }` mirrors the Prisma schema exactly and satisfies the compiler without requiring generated types.

## Test plan

- [x] `npm test -w @lifting-logbook/api` — 175 passed, 31 skipped, 0 failed
- [x] Lint passes (pre-commit hook, turbo lint)

Closes #224

### DIFF
--- a/apps/api/src/adapters/prisma/strength-goal.repository.ts
+++ b/apps/api/src/adapters/prisma/strength-goal.repository.ts
@@ -15,7 +15,8 @@ export class PrismaStrengthGoalRepository implements IStrengthGoalRepository {
       where: { userId: this.userId, program },
       orderBy: { lift: 'asc' },
     });
-    return rows.map((r) => ({
+    // mirrors StrengthGoal schema
+    return rows.map((r: { lift: string; goalType: string; unit: string; updatedAt: Date; target: number | null; ratio: number | null }) => ({
       lift: r.lift,
       goalType: r.goalType as 'absolute' | 'relative',
       unit: r.unit as 'lbs' | 'kg',
--- a/apps/api/src/adapters/prisma/training-max-history.repository.ts
+++ b/apps/api/src/adapters/prisma/training-max-history.repository.ts
@@ -21,7 +21,8 @@ export class PrismaTrainingMaxHistoryRepository implements ITrainingMaxHistoryRe
       },
       orderBy: { date: 'desc' },
     });
-    return rows.map((r) => ({
+    // mirrors TrainingMaxHistory schema
+    return rows.map((r: { id: string; lift: string; weight: number; reps: number; date: Date; isPR: boolean; source: string; goalMet: boolean }) => ({
       id: r.id,
       lift: r.lift,
       weight: r.weight,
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: [fix] Annotate .map() callback to fix implicit any in workout-lift-override repository

### Summary
Ready to merge. The annotation is correct — `lift: string`, `action: string`, and `replacedBy: string | null` exactly mirror the three fields the callback actually accesses, and are consistent with the Prisma schema. One non-blocking maintainability note and one question about systemic coverage.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/adapters/prisma/workout-lift-override.repository.ts:20`
The inline type is a manual mirror of the `WorkoutLiftOverride` Prisma model. If the schema evolves (e.g., `action` becomes an enum column, `replacedBy` gets a different type), this annotation must be updated by hand — unlike the generated `WorkoutLiftOverride` type, which updates with `prisma generate`.
**Suggestion:** Add a brief inline comment — `// mirrors WorkoutLiftOverride schema` — to signal to future maintainers that this type must stay in sync with `prisma/schema.prisma`. Low cost, makes the coupling explicit.

---

### Questions for Author
- Other Prisma repositories (`training-max-history.repository.ts`, `lift-record.repository.ts`, etc.) also use `.map((r) => ...)` without explicit `r` annotations. Do they also fail TS7006 in a fresh worktree, or is there something specific to `WorkoutLiftOverride` that causes only this file to trigger the error? If they fail too, a follow-up issue to annotate them consistently would close out the systemic gap.

---

PR #236, opened 2026-05-12
[feat] Historical lift data backfill via CSV upload

### LINKED ISSUE #225: [feat] Historical lift data backfill via CSV upload
## Problem

Intermediate lifters migrating from Google Sheets / Apps Script implementations arrive with months or years of historical lift records they want to carry over. Without a backfill path, the app's progression logic (RPT top-set targeting, 5/3/1 cycle planning) starts from zero on day one, breaking continuity with the program the lifter is mid-cycle on. This blocks the primary persona's adoption: the Consistent Intermediate Lifter will not switch tools if doing so erases their training history.

## Proposed Solution

Add a CSV upload path that ingests historical `LiftRecord` rows for a given program in a single transactional batch. A new `POST /programs/:program/lift-records/import` endpoint accepts a CSV file, parses it via the existing `parseLiftRecords()` utility, resolves lift abbreviations through `DEFAULT_SLOT_MAP`, and persists rows via the existing `appendLiftRecords()` repository method. A complementary file-upload component in `apps/web` calls the endpoint and surfaces validation errors. Validation is all-or-nothing: if any row fails parsing, lift resolution, or schema checks, the request returns the complete list of errors and writes nothing.

## Acceptance Criteria

- [ ] `POST /programs/:program/lift-records/import` accepts a `multipart/form-data` CSV upload and returns `201` with a count of records written and a list of skipped duplicate rows (by row number and natural key) on success.
- [ ] On any validation failure (unparseable row, unknown lift abbreviation, invalid date, non-numeric weight/reps), the endpoint returns `400` with a structured list of all errors including row numbers, and writes zero records to the database.
- [ ] The endpoint resolves lift abbreviations via `DEFAULT_SLOT_MAP` and rejects rows whose abbreviation is not mapped.
- [ ] Successful imports use `appendLiftRecords()` so duplicate rows (matching on natural key) are silently skipped via Prisma `skipDuplicates`.
- [ ] `apps/web` exposes a file-upload component on the program detail page that calls the import endpoint and renders the validation error list when the upload is rejected, and the skipped-duplicates list when the upload succeeds.
- [ ] Integration test covers a happy-path import of the ~900-row fixture already used by `parseLiftRecords()` tests.
- [ ] Integration test covers a rejected file with at least two distinct error types and asserts the database state is unchanged.

---
Proposal: [docs/proposals/2026-05-11-historical-lift-data-backfill.md](docs/proposals/2026-05-11-historical-lift-data-backfill.md)

brownm09:
## Summary

- Adds `POST /programs/:program/lift-records/import` that accepts a `multipart/form-data` CSV upload, validates every row, and bulk-inserts via the existing `appendLiftRecords()` with `skipDuplicates`
- All-or-nothing validation: any bad row returns `400` with a complete list of errors (row number, field, message); nothing is written
- Duplicate detection via a new `findExistingRecords()` port method — skipped rows are reported in the `201` response body by row number and natural key
- `LiftRecordsImportForm` client component added to the program detail page — shows success counts + skipped rows list, or a formatted error table

## Changes

| Area | What changed |
|---|---|
| `packages/types` | `ImportError`, `SkippedRecord`, `ImportLiftRecordsResponse` |
| `packages/core` | `parseCsvText` (csv-parse promoted to runtime dep), `validateLiftImport` |
| `apps/api` | `@fastify/multipart` plugin, `findExistingRecords` on port + both adapters, `importLiftRecords` controller action |
| `apps/api` (tests) | DB e2e: happy-path, re-import-as-duplicates, all-or-nothing rejection |
| `apps/web` | `importLiftRecords` client-api fn, `LiftRecordsImportForm` component on program page |

## Acceptance criteria checklist

- [x] `POST /programs/:program/lift-records/import` accepts multipart/form-data and returns `201` with `written` + `skipped` on success
- [x] Any validation failure returns `400` with a structured error list and writes zero records
- [x] Lift abbreviations resolved via `DEFAULT_SLOT_MAP`; unknown abbreviations → validation error
- [x] Successful imports use `appendLiftRecords()` with `skipDuplicates`
- [x] `apps/web` exposes a file-upload component on the program detail page
- [x] Integration test: happy-path import of `lift_records.csv` fixture (~1398 rows)
- [x] Integration test: rejected file with two distinct error types, DB unchanged

## Test plan

- `npm test` — all unit + integration tests pass (DB e2e skipped without `DATABASE_URL`)
- DB e2e (with Postgres): `DATABASE_URL=... npm test -w @lifting-logbook/api -- --testPathPattern=db.e2e`
- Manual: start API + web, navigate to program detail page, upload `packages/core/tests/fixtures/lift_records.csv`, verify `201` + record count; re-upload to verify skipped list; upload a malformed CSV and verify error list

Closes #225

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -13,6 +13,7 @@
   "dependencies": {
     "@anthropic-ai/sdk": "^0.92.0",
     "@clerk/backend": "^3.4.4",
+    "@fastify/multipart": "^10.0.0",
     "@lifting-logbook/core": "*",
     "@lifting-logbook/types": "*",
     "@nestjs/common": "^10.0.0",
--- a/apps/api/src/adapters/in-memory/lift-record.adapter.ts
+++ b/apps/api/src/adapters/in-memory/lift-record.adapter.ts
@@ -13,6 +13,16 @@ export class InMemoryLiftRecordRepository implements ILiftRecordRepository {
     this.store.set(program, [...existing, ...records]);
   }
 
+  async findExistingRecords(program: string, candidates: LiftRecord[]): Promise<LiftRecord[]> {
+    const stored = this.store.get(program) ?? [];
+    const existingKeys = new Set(
+      stored.map((r) => `${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}`),
+    );
+    return candidates.filter((r) =>
+      existingKeys.has(`${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}`),
+    );
+  }
+
   async updateLiftRecord(
     program: string,
     id: string,
--- a/apps/api/src/adapters/in-memory/lift-record.adapter.ts
+++ b/apps/api/src/adapters/in-memory/lift-record.adapter.ts
@@ -1,4 +1,4 @@
-import { LiftRecord } from '@lifting-logbook/core';
+import { LiftRecord, liftRecordNaturalKey } from '@lifting-logbook/core';
 import { ILiftRecordRepository } from '../../ports/ILiftRecordRepository';
 
 export class InMemoryLiftRecordRepository implements ILiftRecordRepository {
@@ -8,19 +8,16 @@ export class InMemoryLiftRecordRepository implements ILiftRecordRepository {
     return (this.store.get(program) ?? []).filter((r) => r.cycleNum === cycleNum);
   }
 
-  async appendLiftRecords(program: string, records: LiftRecord[]): Promise<void> {
+  async appendLiftRecords(program: string, records: LiftRecord[]): Promise<number> {
     const existing = this.store.get(program) ?? [];
     this.store.set(program, [...existing, ...records]);
+    return records.length;
   }
 
   async findExistingRecords(program: string, candidates: LiftRecord[]): Promise<LiftRecord[]> {
     const stored = this.store.get(program) ?? [];
-    const existingKeys = new Set(
-      stored.map((r) => `${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}`),
-    );
-    return candidates.filter((r) =>
-      existingKeys.has(`${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}`),
-    );
+    const existingKeys = new Set(stored.map(liftRecordNaturalKey));
+    return candidates.filter((r) => existingKeys.has(liftRecordNaturalKey(r)));
   }
 
   async updateLiftRecord(
--- a/apps/api/src/adapters/prisma/lift-record.repository.ts
+++ b/apps/api/src/adapters/prisma/lift-record.repository.ts
@@ -35,6 +35,31 @@ export class PrismaLiftRecordRepository implements ILiftRecordRepository {
     });
   }
 
+  async findExistingRecords(program: string, candidates: LiftRecord[]): Promise<LiftRecord[]> {
+    if (candidates.length === 0) return [];
+    const rows = await this.prisma.liftRecord.findMany({
+      where: {
+        userId: this.userId,
+        program,
+        OR: candidates.map((r) => ({
+          cycleNum: r.cycleNum,
+          workoutNum: r.workoutNum,
+          lift: r.lift,
+          setNum: r.setNum,
+        })),
+      },
+    });
+    const existingKeys = new Set(
+      rows.map(
+        (r: { cycleNum: number; workoutNum: number; lift: string; setNum: number }) =>
+          `${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}`,
+      ),
+    );
+    return candidates.filter((r) =>
+      existingKeys.has(`${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}`),
+    );
+  }
+
   async updateLiftRecord(
     program: string,
     id: string,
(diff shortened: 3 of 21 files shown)


brownm09:
## Review: [feat] Historical lift data backfill via CSV upload

### Summary

This is a solid implementation — the hexagonal architecture is respected, the all-or-nothing validation design is correct, and the test structure covers the right cases. However, **there is a data contract bug that causes the `skipped` list to always be empty**: the controller builds `skipped` by iterating `parsed` (which contains CSV abbreviations like `"Bench P."`) but matches against `duplicates` (which contain canonical IDs like `"bench-press"`) — the natural-key strings never match. The second e2e test (`re-import returns written=0 and skipped=all`) would catch this if run against a live DB. Fix that, the missing 413 error handling, and the `written` overcount before merging.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/lift-records.controller.ts:189–197`
The `skipped` list is built by iterating `parsed` — the pre-validation array where `lift` is still the raw CSV abbreviation (e.g., `"Bench P."`) — and filtering against `dupKeys`, which contains canonical lift IDs (e.g., `"bench-press"`) from `duplicates`. The natural-key strings (`cycleNum:workoutNum:lift:setNum`) never match because one side uses abbreviations and the other uses canonical IDs. `skipped` will always be `[]`, even when rows were skipped as duplicates.
**Fix:** Replace `parsed` with `records` (the post-canonicalization array) when building `skipped`. Row index is preserved: since validation is all-or-nothing, `records[i]` corresponds to the same logical row as `parsed[i]` when this code is reached.
```typescript
const skipped: SkippedRecord[] = records
  .map((r, i) => ({ r, row: i + 1 }))
  .filter(({ r }) => dupKeys.has(`${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}`))
  .map(({ r, row }) => ({ row, naturalKey: `${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}` }));
```

---

**[reliability]** `apps/api/src/programs/lift-records.controller.ts:183–184`
`findExistingRecords` and `appendLiftRecords` are two separate non-transactional calls. Between them, a concurrent import (e.g., double-click) could insert rows, causing `skipDuplicates` to drop records silently — and `written = records.length - duplicates.length` will overcount. More broadly, the "how many were actually written" source of truth is the Prisma `createMany` count, not the pre-query arithmetic.
**Fix:** The cleanest path without changing the port interface is to change `appendLiftRecords` to return `Promise<number>` (the Prisma `createMany` response already provides `count`) and use that as the `written` value. Alternatively, wrap both calls in `prisma.$transaction`. At minimum, add a comment documenting the race so the next maintainer doesn't silently widen the window.

---

**[reliability]** `apps/api/src/programs/lift-records.controller.ts:167–168`
When the uploaded file exceeds the `fileSize` limit (5 MB), `@fastify/multipart` throws `RequestFileTooLargeError` inside `file.toBuffer()`. This is not caught, so it surfaces as an unhandled exception and the NestJS exception filter returns a `500`. The client sees a server error instead of a `413 Payload Too Large`.
**Fix:** Wrap `file.toBuffer()` in a `try/catch` (or check `file.file.truncated` after the call) and throw `new PayloadTooLargeException('File exceeds the 5 MB limit')`. The `@fastify/multipart` error class is `FST_FILES_LIMIT` / `RequestFileTooLargeError` — import it or catch by `err.code === 'FST_FILES_LIMIT'`.

---

### Non-Blocking Findings

**[performance]** `apps/api/src/adapters/prisma/lift-record.repository.ts:42–55`
`findExistingRecords` issues a single `findMany` with an `OR` array containing one tuple per candidate row. For the fixture CSV (1,398 rows × 4 fields = 5,592 bound parameters), this is within Postgres's ~32k parameter limit, but query planning with a 1,398-clause OR is slow. Any future import approaching ~8,000 rows will exceed the parameter limit and fail outright.
**Suggestion:** Batch the OR query in chunks of ~500 rows, or narrow the query to `cycleNum BETWEEN min AND max AND workoutNum BETWEEN min AND max` and do the final Set-membership check in memory. Add a max-row guard (e.g., 5,000 rows → 400 before validation) to make the upper bound explicit rather than implicit at the DB layer.

---

**[maintainability]** `apps/api/src/programs/lift-records.controller.ts:167` and adapters
The natural-key template string `${r.cycleNum}:${r.workoutNum}:${r.lift}:${r.setNum}` appears in at least four places: the controller (twice), `PrismaLiftRecordRepository`, and `InMemoryLiftRecordRepository`. If the natural key ever gains a field (e.g., `date` were included), all four sites must be updated in sync.
**Suggestion:** Extract a `liftRecordNaturalKey(r: Pick<LiftRecord, 'cycleNum' | 'workoutNum' | 'lift' | 'setNum'>): string` helper in `packages/core` and use it everywhere. This also eliminates the inline Fastify type cast risk — separately, consider a `@fastify/multipart` module augmentation file (`multipart.d.ts`) rather than the inline intersection type cast on the request.

---

### Questions for Author

- **Lift-program compatibility**: `validateLiftImport` accepts any abbreviation in `DEFAULT_SLOT_MAP` regardless of which program the route targets. A "Squat" row can be imported into a bench-only program. Is this intentionally deferred (e.g., programs don't currently have a canonical lift list), or should the controller filter valid records against the program's expected lifts post-canonicalization?

- **`skipped` row numbers**: The `row` value in `SkippedRecord` is the 1-based index within the validated record set, not the original CSV row number (which includes the header). For a 1-row CSV with header, the skipped record reports `row: 1`. Is this the intended numbering, and is it documented enough for clients to present useful feedback to users?

---

### Style Notes

The `naturalKey` field in `SkippedRecord` surfaces `"cycleNum:workoutNum:lift:setNum"` as a raw colon-delimited string. This is an internal implementation detail leaking into the API contract — a future consumer who needs to parse it back has no schema to rely on. Consider either documenting the format explicitly in the `SkippedRecord` JSDoc or replacing it with structured fields (`cycleNum`, `workoutNum`, `lift`, `setNum`) on the response type.

---

PR #246, opened 2026-05-13
[fix] "Choose This Program" fails for first-time users — add cycle initialization endpoint

### LINKED ISSUE #245: fix: "Choose This Program" fails for first-time users — no initial cycle creation path
## Problem

The "Choose This Program" button in the onboarding flow always fails for first-time users. Clicking it shows:

> "Failed to start your program. Please try again."

## Root Cause

The button calls `POST /programs/{programId}/cycles` (the *advance-cycle* endpoint). That endpoint calls `getCycleDashboard()` first, which throws `ProgramNotFoundError` when no `CycleDashboard` row exists yet. The API returns 404, and the onboarding server action swallows the error into a generic message.

There is no code path to create cycle 1 from scratch — `updateCycle()` in the core library also requires an existing `CycleDashboard` as input.

## Fix

Add a dedicated `POST /programs/{programId}/cycles/initialize` endpoint that creates the first cycle from scratch (cycleNum=1, today's date, zero training maxes), and wire the onboarding flow to call it instead of the advance endpoint.

## Acceptance Criteria

- [ ] New endpoint `POST /programs/{programId}/cycles/initialize` creates a cycle-1 dashboard for the authenticated user
- [ ] Returns 409 Conflict if a cycle for that program already exists
- [ ] Returns 400 Bad Request for an unrecognized program ID
- [ ] Onboarding "Choose This Program" button redirects to `/cycle/1` on success
- [ ] All existing tests continue to pass; new tests cover the new endpoint and service method

brownm09:
## Summary

- Adds `POST /programs/:program/cycles/initialize` — a dedicated endpoint for creating the first cycle dashboard from scratch (cycle 1) for a brand-new user
- Wires the onboarding \"Choose This Program\" button to call this new endpoint instead of the advance-cycle endpoint
- Adds `PROGRAM_DEFAULTS` registry in the service mapping program IDs to their `cycleUnit`/`programType` metadata

## Root Cause

The advance-cycle endpoint (`POST /programs/:program/cycles`) calls `getCycleDashboard()` first, which throws `ProgramNotFoundError` when no row exists yet. The core library's `updateCycle()` also requires an existing `CycleDashboard`. There was no path to bootstrap cycle 1 from scratch, so every new user who clicked \"Choose This Program\" hit a 404 that the server action swallowed into a generic error message.

## Changes

| File | Change |
|---|---|
| `apps/api/src/programs/initialize-cycle.dto.ts` | New DTO — optional `cycleDate` |
| `apps/api/src/programs/cycle-generation.service.ts` | `initializeFirstCycle()` method + `PROGRAM_DEFAULTS` registry |
| `apps/api/src/programs/cycle-generation.controller.ts` | `POST cycles/initialize` route (declared before `POST cycles` to avoid route conflict) |
| `apps/web/lib/api.ts` | `initializeCycle()` client function |
| `apps/web/app/onboarding/actions.ts` | Call `initializeCycle` instead of `createCycle` |
| `apps/api/src/programs/cycle-generation.service.spec.ts` | 4 new tests for `initializeFirstCycle` |
| `apps/api/src/programs/cycle-generation.controller.spec.ts` | 3 new tests for the new route |

## Acceptance Criteria

- [x] `POST /programs/:program/cycles/initialize` creates a cycle-1 dashboard for the authenticated user
- [x] Returns 409 Conflict if a cycle already exists
- [x] Returns 400 Bad Request for an unrecognized program ID
- [x] Onboarding \"Choose This Program\" calls the initialize endpoint and redirects to `/cycle/1` on success
- [x] All existing tests continue to pass; 7 new tests cover the new endpoint and service method

## Test Results

```
PASS src/programs/cycle-generation.service.spec.ts
PASS src/programs/cycle-generation.controller.spec.ts
Tests: 20 passed, 20 total (was 14 — 6 new)
```

Pre-existing failures in `programs.e2e.spec.ts` (lift metadata 401 errors) are unchanged — 86 failing before and after this PR.

Closes #245

### DIFF
--- a/apps/api/src/programs/cycle-generation.controller.spec.ts
+++ b/apps/api/src/programs/cycle-generation.controller.spec.ts
@@ -22,13 +22,14 @@ const stubCycleDashboard = () => ({
 
 describe('CycleGenerationController', () => {
   let controller: CycleGenerationController;
-  let service: jest.Mocked<Pick<CycleGenerationService, 'startNewCycle' | 'recalculateMaxes'>>;
+  let service: jest.Mocked<Pick<CycleGenerationService, 'startNewCycle' | 'recalculateMaxes' | 'initializeFirstCycle'>>;
   let factory: jest.Mocked<IRepositoryFactory>;
 
   beforeEach(async () => {
     service = {
       startNewCycle: jest.fn(),
       recalculateMaxes: jest.fn(),
+      initializeFirstCycle: jest.fn(),
     };
     factory = {
       forUser: jest.fn().mockResolvedValue(MOCK_BUNDLE),
@@ -82,6 +83,55 @@ describe('CycleGenerationController', () => {
       await expect(controller.startNewCycle('unknown', {}, MOCK_USER)).rejects.toThrow(
         'Program not found',
       );
+    });
+  });
+
+  describe('initializeFirstCycle', () => {
+    const stubInitDashboard = () => ({
+      program: PROGRAM,
+      cycleUnit: 'week' as const,
+      cycleNum: 1,
+      cycleDate: new Date('2026-05-12T00:00:00.000Z'),
+      sheetName: '5-3-1_Cycle_1_20260512',
+      cycleStartWeekday: Weekday.Monday,
+      currentWeekType: 'training' as const,
+      programType: '5-3-1',
+    });
+
+    it('calls service with repos, program, and dto, returns mapped response', async () => {
+      service.initializeFirstCycle.mockResolvedValue(stubInitDashboard());
+
+      const result = await controller.initializeFirstCycle(PROGRAM, {}, MOCK_USER);
+
+      expect(factory.forUser).toHaveBeenCalledWith(MOCK_USER);
+      expect(service.initializeFirstCycle).toHaveBeenCalledWith(MOCK_BUNDLE, PROGRAM, {});
+      expect(result).toEqual({
+        program: PROGRAM,
+        cycleNum: 1,
+        cycleStartDate: '2026-05-12',
+        weeks: [],
+        currentWeekType: 'training',
+      });
+    });
+
+    it('passes optional cycleDate through to service', async () => {
+      service.initializeFirstCycle.mockResolvedValue(stubInitDashboard());
+
+      await controller.initializeFirstCycle(PROGRAM, { cycleDate: '2026-05-12' }, MOCK_USER);
+
+      expect(service.initializeFirstCycle).toHaveBeenCalledWith(
+        MOCK_BUNDLE,
+        PROGRAM,
+        { cycleDate: '2026-05-12' },
+      );
+    });
+
+    it('propagates service errors (e.g. ConflictException)', async () => {
+      service.initializeFirstCycle.mockRejectedValue(new Error('Already exists'));
+
+      await expect(
+        controller.initializeFirstCycle(PROGRAM, {}, MOCK_USER),
+      ).rejects.toThrow('Already exists');
     });
   });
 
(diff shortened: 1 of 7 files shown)


brownm09:
## Review: [fix] "Choose This Program" fails for first-time users — add cycle initialization endpoint

### Summary

**Needs revision before merge** — one correctness bug: the weekday lookup in `initializeFirstCycle` returns a lowercase key (`'monday'`) from `WEEKDAY_MAP` but assigns it to `cycleStartWeekday`, which has type `Weekday` (PascalCase enum values: `'Monday'`). The `as Weekday` cast suppresses the TypeScript error, so the mismatch ships silently. No test asserts `cycleStartWeekday` on the result, leaving the bug undetected. The rest of the implementation is solid — the guard pattern, the `PROGRAM_DEFAULTS` registry, route ordering, and test coverage are all correct.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/cycle-generation.service.ts` lines 124–127

`WEEKDAY_MAP` (from `@lifting-logbook/core`) uses lowercase string keys: `{ sunday: 0, monday: 1, ..., saturday: 6 }`. `Object.keys(WEEKDAY_MAP).find(...)` therefore returns a lowercase string like `'monday'`. The `Weekday` enum, however, uses PascalCase values: `"Sunday"`, `"Monday"`, …. The `as Weekday` cast silences the type error and stores `'monday'` in a field that the rest of the codebase treats as `Weekday.Monday` (`"Monday"`). Any downstream `=== Weekday.Monday` comparison or Prisma enum constraint check will fail silently, and no test currently asserts the `cycleStartWeekday` field on the saved dashboard.

**Fix:** Search by enum values instead of map keys, so PascalCase is preserved:

```typescript
const weekdayName = Object.values(Weekday).find(
  (v) => WEEKDAY_MAP[v.toLowerCase()] === cycleDate.getUTCDay(),
);
if (!weekdayName) {
  throw new Error(`No Weekday mapping for UTC day index ${cycleDate.getUTCDay()}`);
}
const dashboard: CycleDashboard = {
  // ...
  cycleStartWeekday: weekdayName,
};
```

Also add `expect(result.cycleStartWeekday).toBe(Weekday.Tuesday)` (for the `2026-05-12` test date, which is a Tuesday) to the "creates and saves a cycle-1 dashboard" service spec.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/cycle-generation.controller.ts` line 23 (comment)

The route-ordering comment says "NestJS does not treat 'initialize' as a dynamic segment match against an existing `:id` param." There is no `POST cycles/:id` route in this controller — the only other `POST` route is the bare `POST cycles`. The comment's stated reason is therefore misleading; the actual concern is plain route-prefix ordering. The ordering itself is correct and harmless to keep; the comment just needs updating.

**Suggestion:** Change the comment to: `// Declare before @Post('cycles') so 'initialize' is matched as a literal path segment, not swallowed by the advance-cycle handler.`

---

**[maintainability]** `apps/api/src/programs/cycle-generation.service.ts` lines 22–24 (`PROGRAM_DEFAULTS`)

The `PROGRAM_DEFAULTS` registry is the sole source of `cycleUnit`/`programType` for cycle initialization. Omitting an entry for a new program causes `400 Bad Request` for every first-time user of that program, but there is no in-code reminder that this step is required when onboarding a new program type.

**Suggestion:** Add a JSDoc comment or inline note:
```typescript
/**
 * Metadata required to bootstrap cycle 1 for each supported program.
 * ADD A NEW ENTRY HERE when a new program is made available in onboarding —
 * omitting it causes 400 Bad Request for all first-time users of that program.
 */
const PROGRAM_DEFAULTS: Record<string, { cycleUnit: string; programType: string }> = {
```

---

### Questions for Author

- `initializeCycle()` in `apps/web/lib/api.ts` always sends an empty body. If the onboarding UI later lets users pick a program start date, the client will need to pass `cycleDate`. Is it worth accepting an optional argument now, or intentionally deferred?

---

PR #253, opened 2026-05-14
[feat] Catalog-based tag pickers in LiftEditor (substitutions + muscle groups)

### LINKED ISSUE #249: feat: catalog-based tag pickers in LiftEditor (substitutions + muscle groups)
## Problem
The LiftEditor uses comma-separated free-text inputs for Substitutions and Muscle Groups. Substitutions should be constrained to the lift catalog to ensure consistency; both fields have poor discoverability.

## Solution
- **Substitutions**: replace comma-text input with a search-as-you-type dropdown sourced from `LIFT_CATALOG`, selections rendered as removable chip tags (`CatalogTagPicker` component).
- **Muscle Groups**: replace comma-text input with a free-text chip-tag picker — type, press Enter or comma, value becomes a removable chip (`FreeTagPicker` component).

## Acceptance criteria
- [ ] Substitutions field shows a filtered dropdown from LIFT_CATALOG as user types
- [ ] Selecting a catalog lift adds it as a chip tag; clicking × removes it
- [ ] Already-selected lifts are excluded from the dropdown
- [ ] Muscle Groups field accepts free text; pressing Enter or comma commits a chip
- [ ] Muscle Groups chips are removable with ×
- [ ] Both fields produce `string[]` values passed unchanged to `patchLiftMetadata`
- [ ] Clicking outside the dropdown closes it

## New components
- `apps/web/components/CatalogTagPicker.tsx`
- `apps/web/components/FreeTagPicker.tsx`

brownm09:
## Summary

- Replaces the Substitutions free-text input with `CatalogTagPicker`: a search-as-you-type dropdown constrained to `LIFT_CATALOG`; selections render as removable chip tags; already-selected lifts are excluded; clicking outside closes the dropdown.
- Replaces the Muscle Groups free-text input with `FreeTagPicker`: type a value, press Enter or comma to commit it as a chip; Backspace removes the last chip; blur commits any pending draft.
- Both components produce `string[]` values passed unchanged to `patchLiftMetadata` — no API changes required.

## Acceptance Criteria

- [x] Substitutions field shows a filtered dropdown from LIFT_CATALOG as user types
- [x] Selecting a catalog lift adds it as a chip tag; clicking × removes it
- [x] Already-selected lifts are excluded from the dropdown
- [x] Muscle Groups field accepts free text; pressing Enter or comma commits a chip
- [x] Muscle Groups chips are removable with ×
- [x] Both fields produce `string[]` values passed unchanged to `patchLiftMetadata`
- [x] Clicking outside the dropdown closes it

## New components

- `apps/web/components/CatalogTagPicker.tsx` + `CatalogTagPicker.module.css`
- `apps/web/components/FreeTagPicker.tsx` + `FreeTagPicker.module.css`

## Testing

- `npm test` — 182 tests pass, 0 failures (pre-existing TypeScript error in `app/cycle/[cycleNum]/page.tsx` is unrelated and exists on `main`).
- Manual: navigate to a lift editor, verify both pickers render existing values as chips, accept new values, and submit correctly via Save.

Closes #249

### DIFF
--- a/apps/web/components/CatalogTagPicker.module.css
+++ b/apps/web/components/CatalogTagPicker.module.css
@@ -3,64 +3,19 @@
 }
 
 .chipList {
-  display: flex;
-  flex-wrap: wrap;
-  gap: 0.375rem;
-  align-items: center;
-  padding: 0.375rem 0.5rem;
-  border: 1px solid var(--color-border, #e5e7eb);
-  border-radius: 6px;
-  background: var(--color-surface, #f8f9fa);
-  min-height: 2.5rem;
-  cursor: text;
-}
-
-.chipList:focus-within {
-  border-color: var(--color-primary, #0070f3);
+  composes: chipList from './chip.module.css';
 }
 
 .chip {
-  display: inline-flex;
-  align-items: center;
-  gap: 0.25rem;
-  background: var(--color-surface-alt, #eef0f2);
-  border: 1px solid var(--color-border, #e5e7eb);
-  border-radius: 4px;
-  padding: 0.125rem 0.375rem;
-  font-size: 0.875rem;
+  composes: chip from './chip.module.css';
 }
 
 .chipRemove {
-  background: none;
-  border: none;
-  cursor: pointer;
-  padding: 0;
-  color: var(--color-text-secondary, #666);
-  line-height: 1;
-  font-size: 1rem;
-}
-
-.chipRemove:hover {
-  color: var(--color-text, #111);
-}
-
-.chipRemove:disabled {
-  cursor: not-allowed;
-  opacity: 0.5;
+  composes: chipRemove from './chip.module.css';
 }
 
 .textInput {
-  border: none;
-  outline: none;
-  background: transparent;
-  font-size: 0.9375rem;
-  flex: 1;
-  min-width: 120px;
-  color: var(--color-text, #111);
-}
-
-.textInput::placeholder {
-  color: var(--color-text-secondary, #666);
+  composes: textInput from './chip.module.css';
 }
 
 .dropdown {
--- a/apps/web/components/CatalogTagPicker.module.css
+++ b/apps/web/components/CatalogTagPicker.module.css
@@ -1,0 +1,96 @@
+.wrapper {
+  position: relative;
+}
+
+.chipList {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 0.375rem;
+  align-items: center;
+  padding: 0.375rem 0.5rem;
+  border: 1px solid var(--color-border, #e5e7eb);
+  border-radius: 6px;
+  background: var(--color-surface, #f8f9fa);
+  min-height: 2.5rem;
+  cursor: text;
+}
+
+.chipList:focus-within {
+  border-color: var(--color-primary, #0070f3);
+}
+
+.chip {
+  display: inline-flex;
+  align-items: center;
+  gap: 0.25rem;
+  background: var(--color-surface-alt, #eef0f2);
+  border: 1px solid var(--color-border, #e5e7eb);
+  border-radius: 4px;
+  padding: 0.125rem 0.375rem;
+  font-size: 0.875rem;
+}
+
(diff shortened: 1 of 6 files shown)


brownm09:
## Review: [feat] Catalog-based tag pickers in LiftEditor (substitutions + muscle groups)

### Summary
Good implementation of the chip-tag pattern — the `onMouseDown`/`e.preventDefault()` fix for the blur-before-select race condition is the right call, and the catalog filtering logic is clean. Two blocking correctness issues need addressing before merge: the `<label>` associations are now broken (orphaned `htmlFor` IDs), and `FreeTagPicker` allows duplicate chips. Neither is complex to fix.

---

### Blocking Findings

**[correctness]** `apps/web/components/LiftEditor.tsx:60,72`
`<label htmlFor="muscleGroups">` and `<label htmlFor="substitutions">` reference IDs that no longer exist in the DOM. The old `<input id="muscleGroups">` and `<input id="substitutions">` were replaced by `FreeTagPicker` and `CatalogTagPicker`, neither of which accepts or forwards an `id` prop to their internal `<input>`. Clicking either label does nothing — focus is not moved to the picker.
**Fix:** Add an optional `id` prop to both picker components and wire it to the `<input>` element. Then pass `id="muscleGroups"` and `id="substitutions"` from `LiftEditor`.

---

**[correctness]** `apps/web/components/FreeTagPicker.tsx:16-22`
`commit()` appends `trimmed` to `value` without a duplicate check. Typing "Quads" → Enter → "Quads" → Enter produces two chips with the same string. React emits a key-collision warning (`key={tag}` on the duplicate), and clicking × on either chip calls `value.filter((v) => v !== tag)` — removing **all** instances at once rather than just the one clicked.
**Fix:** Guard the `onChange` call: `if (trimmed && !value.includes(trimmed)) { onChange([...value, trimmed]); }`

---

### Non-Blocking Findings

**[accessibility]** `apps/web/components/CatalogTagPicker.tsx:70`
`aria-expanded` on a plain `input[type="text"]` is not a valid ARIA attribute per WAI-ARIA 1.2 — it belongs on an element with `role="combobox"`. `aria-autocomplete="list"` is also only meaningful in the combobox context; without `role="combobox"` and a paired `aria-controls` pointing to the `role="listbox"`, screen readers may ignore or misinterpret these attributes.
**Suggestion:** Either commit to the full combobox pattern (`role="combobox"` on the input, `id` on the `<ul>`, `aria-controls` linking them), or remove `aria-autocomplete` and `aria-expanded` if the full pattern is out of scope for this iteration.

---

**[accessibility]** `apps/web/components/CatalogTagPicker.tsx:79-88`
The dropdown has no keyboard navigation — Arrow Down, Escape, and Tab are unhandled. Keyboard-only users can type to filter but must use a pointer device to select an option from the list.
**Suggestion:** Add `onKeyDown` to the input: Arrow Down selects the first filtered option (or moves focus into the list); Escape clears `query`. This does not need to be a full roving-tabindex implementation — even a simple Arrow Down → `select(filtered[0])` shortcut dramatically improves keyboard accessibility.

---

**[maintainability]** `CatalogTagPicker.module.css` / `FreeTagPicker.module.css`
~50 lines of CSS are duplicated verbatim across the two files (`.chipList`, `.chip`, `.chipRemove`, `.textInput`, `:focus-within`, `::placeholder`, `:hover`, `:disabled`). A single design-token change (e.g., chip border-radius) requires two edits.
**Suggestion:** Extract the shared rules into a `chip.module.css` and reference them via CSS Modules' `composes:` directive. Low urgency — fine to defer to a follow-up.

---

### Questions for Author
- Was the `htmlFor` label association intentionally dropped (in which case the `<label>` elements could be changed to `<span>` for semantic correctness), or is the broken association an oversight?

---

### Style Notes
The `×` multiply sign (U+00D7) works fine as a chip-remove label, but it is visually lighter and less immediately recognizable as a close/remove affordance than `✕` (U+2715) or a small SVG icon. If a future design pass standardizes remove buttons across the app, this would be the place to revisit. Not urgent.

---

PR #255, opened 2026-05-15
[feat] Programs management page — browse catalog, switch programs, custom editor (#251)

### LINKED ISSUE #251: feat: programs management page (/programs) with template browse and custom editor
## Problem
The current program page (`/cycle/[cycleNum]/program`) is read-only and has a disabled 'Edit Program' button. There is no way to browse template programs or create/manage custom programs.

## Solution
New route `/programs` implementing the full Programs v3 design:
- **Category select**: Beginner / Intermediate / Advanced / Custom cards
- **Browse**: program cards with metadata (days/week, split, goal tags, status badge)
- **Program detail**: workout list, metadata, action buttons
- **Actions on templates**: Start, Edit as Custom (creates a copy), Copy to Custom
- **Custom program editor**: name/cycles/blocks fields + per-workout editor with inline catalog-based lift picker
- **Create custom**: blank editor

## Acceptance criteria
- [ ] `/programs` is accessible from the cycle layout nav
- [ ] Category grid shows program counts
- [ ] Browsing shows program cards with description and metadata
- [ ] Template detail shows workouts and all action buttons
- [ ] Custom detail shows Edit / Start / Delete buttons
- [ ] Custom editor allows renaming, setting cycles/blocks, adding/removing workouts
- [ ] Each workout editor lists current lifts, allows adding via catalog search, allows removing
- [ ] Custom programs persist (localStorage MVP if no backend CRUD endpoint exists; comment marks wiring point)
- [ ] 'Start This Template' triggers the existing cycle-start flow

## Backend dependency
Check `GET /programs/philosophies` existence before wiring; fall back to client-side template data from `onboarding/programs.ts` if not available.

brownm09:
## Summary

- **Browse tab**: 13-program catalog with experience/goal filters, inline detail expansion, active-program highlight ("Current" badge), and a confirmation dialog to switch programs
- **My Programs tab**: Full custom program editor in three modes — create from scratch, clone a template (seeds spec from 5/3/1 fixture), and edit existing custom programs (with delete)
- **Dynamic active program**: All cycle/history/settings pages now resolve the active program from `UserSettings.activeProgram` (database) with fallback to `NEXT_PUBLIC_DEFAULT_PROGRAM`
- **Navigation**: "Programs" link added to cycle nav header and dashboard quicknav

## Backend changes

- Prisma migration adds `user_settings`, `custom_program`, `custom_program_spec` tables
- `UserSettingsModule` — `GET/PATCH /users/me/settings`
- `CustomProgramsModule` — full CRUD at `GET/POST /programs/custom` and `GET/PUT/DELETE /programs/custom/:id`
- `POST /programs/:program/switch` — updates active program in UserSettings, initializes cycle if first time
- `HybridLiftingProgramSpecRepository` — routes UUID program IDs to Prisma, catalog IDs to in-memory fixtures
- `PROGRAM_DEFAULTS` expanded to all 13 catalog programs + UUID custom programs

## Architecture note

Client components in `/programs` call server-only API functions via Next.js Server Actions (`apps/web/app/programs/actions.ts`), keeping `lib/api.ts` safely server-only.

## Test plan

- [x] `npm run build` — passes, `/programs` route appears in build output
- [x] `npm test -w @lifting-logbook/api` — all 123 unit tests pass; 86 e2e failures are pre-existing DB connectivity issues (baseline 88 before this PR)
- [x] `npm test -w @lifting-logbook/core` — all tests pass
- [ ] Manual: navigate to `/programs`, browse catalog, switch to an available program
- [ ] Manual: create a custom program, verify it appears in "Edit Custom" sub-tab
- [ ] Manual: clone 5/3/1 template, verify spec table pre-populates
- [ ] Manual: switch to a custom program, verify cycle dashboard reflects new spec

Closes #251


brownm09:
## Review: [feat] Programs management page — browse catalog, switch programs, custom editor (#251)

### Summary

This is a well-structured feature with clean layering — Server Actions correctly bridge the `server-only` API client to client components, the hybrid repo pattern is a good fit, and the `getActiveProgram()` helper cleanly centralizes dynamic program resolution across all pages. Ready to merge with revisions. The highest-severity finding is a TOCTOU authorization pattern in `update()` and `delete()`: ownership is checked in a separate query from the mutation, meaning the userId constraint is not enforced atomically. Two other correctness issues — a non-atomic switch-program flow and a missing week-range constraint in the DTO — round out the blocking set. 6 additional non-blocking findings were identified; the 4 most impactful are listed below.

---

### Blocking Findings

**[security]** `apps/api/src/custom-programs/custom-programs.repository.ts` — `update()` and `delete()`

Both `update()` and `delete()` do ownership verification with `findFirst({ where: { id, userId: this.userId } })` then mutate with `prisma.customProgram.update({ where: { id } })` / `prisma.customProgram.delete({ where: { id } })` — the `userId` is dropped from the write clause.

This is a TOCTOU pattern: the ownership check and the mutation are separate queries. Any future refactor that removes or reorders the preceding `findFirst` silently produces a cross-tenant write/delete. It also means the write itself does not enforce the tenant boundary — defense-in-depth fails here.

**Fix:** Use a single tenant-scoped write. For update: switch to `updateMany({ where: { id, userId: this.userId }, data: updatePayload })` and throw `NotFoundException` when `count === 0`. For delete: use `deleteMany({ where: { id, userId: this.userId } })` with the same check. This is atomic, cheaper (one query instead of two), and the tenant boundary cannot be accidentally removed.

---

**[correctness]** `apps/api/src/programs/switch-program.controller.ts:31`

`switchProgram` runs three separate awaits with no transaction: `getCycleDashboard` → conditional `initializeFirstCycle` → `upsertSettings`. If `initializeFirstCycle` succeeds but `upsertSettings` subsequently fails, the user ends up with a cycle dashboard for a program that is not their `activeProgram`. A retry will then skip cycle initialization (the dashboard already exists), write the setting, and recover — but the window of inconsistency means a page load between failure and retry will route the user to their old program's cycle.

**Fix:** Wrap the cycle-ensure and settings-upsert in a `prisma.$transaction(async (tx) => { ... })`. The cycle initialization operations can be made transactional by threading `tx` as the Prisma client — or at minimum document that eventual consistency here is an accepted trade-off and add a comment explaining the recovery path.

---

**[correctness]** `apps/api/src/custom-programs/create-custom-program.dto.ts`

`CustomProgramSpecRowDto` validates `week` with `@IsNumber()` but no range constraint (`@Min(1) @Max(3)` or `@IsIn([1, 2, 3])`). A payload with `week: 99` passes the global `ValidationPipe` and is stored. `HybridLiftingProgramSpecRepository.getCustomSpec` then casts `s.week as 1 | 2 | 3` — a TypeScript-only lie at runtime. Downstream cycle generation logic that uses `week` to index into a 3-slot structure will produce unexpected behavior for records with out-of-range values.

**Fix:** Add `@IsIn([1, 2, 3])` (or `@Min(1) @Max(3)`) to the `week` field. Also consider `@Min(1) @Max(10)` on `sets` and `reps` to prevent degenerate specs from reaching the workout-plan logic.

---

### Non-Blocking Findings

**[performance]** `apps/web/lib/active-program.ts`

`getActiveProgram()` issues a serial `fetchUserSettings()` network+DB roundtrip on every server page render, before the page's own `Promise.all` can start. On the cycle dashboard this adds a full round-trip in front of `fetchCycleDashboard` + `fetchProgramSpec` + `fetchTrainingMaxes`. `user_settings` changes rarely (only on program switch).

**Suggestion:** Wrap `getActiveProgram` with [`React.cache`](https://react.dev/reference/react/cache) so it deduplicates within a single render pass. For the cycle-heavy pages, consider adding `activeProgram` to the cycle dashboard response so the setting is fetched in parallel with existing calls rather than serially before them.

---

**[reliability]** `apps/web/lib/active-program.ts`

The `catch` block silently swallows all errors from `fetchUserSettings()` — auth failures, 5xx, and network timeouts are all treated identically to "user has no preference." A user who has switched to a custom program will appear to be on the default program any time the API hiccups, with no indication in logs that anything failed.

**Suggestion:** At minimum, add `console.error('getActiveProgram: falling back to default', e)` in the catch. For auth/5xx errors, consider rethrowing so the Next.js error boundary handles them consistently with other API failures on the same page.

---

**[maintainability]** `apps/api/src/custom-programs/custom-programs.repository.ts` — `getSpec()`

`getSpec(id)` queries `customProgramSpec` by `programId` without a `userId` filter and is unused — `HybridLiftingProgramSpecRepository` implements its own version directly. As dead code with no tenant scoping, the first caller to wire it up would silently leak another user's spec data.

**Suggestion:** Delete `getSpec()`. If it's ever needed, it should be added with a `userId` join from the start.

---

**[maintainability]** `apps/web/app/programs/ProgramEditor.tsx` and `apps/web/lib/programs.ts`

The editor hardcodes `const weeks = [1, 2, 3]` with no UI for `weekType` or `offset`. Cloning a program with a deload or test week (which exist in the domain model) and saving will silently drop those rows. Additionally, `seedProgramSpec()` in `apps/web/lib/programs.ts` duplicates the same 12 rows already in `apps/api/src/adapters/in-memory/fixtures.ts`, creating two sources of truth that will drift.

**Suggestion:** Either drive `weeks` from the loaded spec (preserving whatever weeks exist) or add a server-side constraint that rejects specs with `week > 3` to make the truncation explicit rather than silent. Extract `seedProgramSpec` to a shared package or expose it via the catalog endpoint.

---

### Questions for Author

- In `EditorTab.tsx`, the Delete button has no confirmation dialog — one click permanently cascades all spec rows. Is this intentional given the low volume of custom programs per user, or should it match the Switch flow's confirmation dialog pattern?
- `ParseProgramPipe` allows `[a-zA-Z0-9_-]+`. UUIDs (hex + hyphens) match this pattern — was that verified explicitly, or was the pipe pre-existing and untested with UUID-shaped inputs?

---

### Style Notes

`EditorTab.tsx` uses inline `style={{ marginTop: 0 }}` and `style={{ flex: 1 }}` in a few spots where CSS module classes would keep styling centralized. Minor inconsistency with the rest of the file.

---

brownm09:
## Re-review: [feat] Programs management page — browse catalog, switch programs, custom editor (#251)

### Summary

The fixes from the first review round are correctly applied — the TOCTOU `update()`/`delete()` auth bypass, DTO validation tightening, delete confirmation UX, `React.cache` deduplication, and `console.error` logging are all present and correct. Two new **blocking** security/correctness issues surface in this pass that were not in scope of the original review, plus one correctness bug in client state management. The PR should not merge until the two security findings are addressed; the correctness bug is straightforward.

No doc-reconciliation rules defined for this repo.

---

### Blocking Findings

**[security]** `apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts:24`
`getCustomSpec(id)` queries `customProgramSpec.findMany({ where: { programId: id } })` with no `userId` filter. Any authenticated user can call `POST /programs/<uuid>/switch` with a UUID they don't own — `ParseProgramPipe` accepts any UUID-shaped string, and `SwitchProgramController` passes it straight through to `initializeFirstCycle`, which invokes `getProgramSpec`. The spec rows from another user's custom program are returned and used to generate the caller's cycles.
**Fix:** In `SwitchProgramController.switchProgram`, when the program value passes the `UUID_PATTERN` test, verify ownership before calling `initializeFirstCycle`:
```typescript
if (UUID_PATTERN.test(program)) {
  const owned = await this.prisma.customProgram.findFirst({
    where: { id: program, userId: user.id },
  });
  if (!owned) throw new ForbiddenException('Program not found');
}
```
Throw `ForbiddenException` (not `NotFoundException`) to avoid confirming UUID existence to the caller.

---

**[security]** `apps/api/src/user-settings/update-settings.dto.ts:4` + `apps/api/src/user-settings/user-settings.repository.ts:14`
`PATCH /users/me/settings` accepts `{ activeProgram?: string }` with only `@IsString()` validation and writes it directly to the DB. A user who knows another user's custom program UUID can bypass the switch endpoint entirely — `HybridLiftingProgramSpecRepository.getCustomSpec` will then serve the foreign program's spec rows on the next cycle load. The switch endpoint's ownership check (once added per the finding above) is circumvented.
**Fix:** Remove `activeProgram` from `UpdateSettingsDto` and from the `upsertSettings` path. The only supported way to change the active program should be `POST /programs/:program/switch`, which is where ownership can be enforced. If future settings (e.g. notification preferences) are added, the field names will make the distinction clear.

---

**[correctness]** `apps/web/app/programs/ProgramsTabs.tsx:19`
`handleProgramSaved(id)` calls `setCurrentActive(id)`, causing `BrowseTab` to show the just-saved program with a "Current" badge. But `ProgramEditor.handleSave` only persists the program to the DB — it does not call `switchProgram`. The `UserSettings.activeProgram` in the DB is unchanged. On the next hard refresh the old program shows as current again; the "Current" badge is a false positive.
**Fix:** Remove `setCurrentActive(id)` from `handleProgramSaved`. `currentActive` should only update when `handleSaveAndSwitch` completes (which already navigates away) or when the Browse tab switch dialog confirms. Alternatively, pass two callbacks: `onSaved` (no active-program update) and `onSavedAndSwitched` (updates active program).

---

### Non-Blocking Findings

**[reliability]** `apps/web/app/programs/SwitchProgramDialog.tsx:17`
The dialog sets `role="dialog"` and `aria-modal="true"` but has no Escape key handler. [ARIA dialog authoring practices](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) require dialogs to close on Escape. Keyboard users have no way to dismiss without clicking Cancel.
**Suggestion:** Add `onKeyDown={(e) => { if (e.key === 'Escape' && !isPending) onClose(); }}` to the overlay `<div>`.

---

**[reliability]** `apps/web/app/programs/EditorTab.tsx:26`
`isFetching` is a single boolean shared across all programs in the list. Clicking Edit on program A disables and relabels **all** Edit buttons as "Loading…". Only the clicked item should show the loading state.
**Suggestion:** Replace `const [isFetching, startFetch] = useTransition()` with `const [fetchingId, setFetchingId] = useState<string | null>(null)` (mirroring the `confirmingDeleteId` pattern already in this file). Set `fetchingId = p.id` before the fetch and `fetchingId = null` after.

---

**[maintainability]** `apps/api/src/custom-programs/create-custom-program.dto.ts:21` + `update-custom-program.dto.ts:7`
Neither DTO enforces a minimum spec count. An empty `specs: []` array passes validation; the resulting program would generate no workouts and silently produce an empty cycle if switched to.
**Suggestion:** Add `@ArrayMinSize(1)` to `specs` in both DTOs.

---

**[performance]** `apps/api/src/programs/program.pipe.ts:4`
`ParseProgramPipe` validates character class but not string length. A 50,000-character string matching `[a-zA-Z0-9_-]+` would pass the pipe and reach Prisma.
**Suggestion:** Add a `value.length > 200` check before the regex test and throw `BadRequestException`.

---

### Questions for Author

- `ProgramsTabs.handleProgramSaved` is called from both save-only and save-and-switch. Was it intentional for plain save (without switching) to show the new program as "Current" in the Browse tab? If yes, the fix in the blocking finding above needs a different approach (e.g., a separate "justCreatedId" highlight that's distinct from the real active program).

---

PR #256, opened 2026-05-15
[test] Add e2e coverage for POST cycles/initialize

### LINKED ISSUE #247: Add e2e test coverage for POST cycles/initialize endpoint
## Problem

The \`POST /programs/:program/cycles/initialize\` endpoint introduced in #246 has no e2e test coverage. The unit tests (service + controller) cover the logic in isolation, but nothing exercises the full stack through the Prisma adapter against a real database.

## Acceptance Criteria

- [ ] \`programs.db.e2e.spec.ts\` covers \`POST cycles/initialize\` with at least:
  - Happy path: creates a \`CycleDashboard\` row and returns the expected response shape
  - 409 Conflict: a second call for the same user+program returns 409
  - 400 Bad Request: an unrecognized program ID returns 400
- [ ] All existing e2e tests continue to pass

brownm09:
## Summary

- Adds a `describe('POST /programs/:program/cycles/initialize (DB)')` block to `programs.db.e2e.spec.ts` with three cases matching the issue acceptance criteria
- Happy path: creates a `CycleDashboard` row for a fresh user and returns 201 with the expected response shape; verified at the DB layer via Prisma
- 409 Conflict: a second call for the same user+program returns 409
- 400 Bad Request: an unrecognized program ID (`not-a-real-program`) returns 400
- Adds `USER_INIT = 'db-e2e-init'` to `cleanTestUsers()` so teardown covers the new rows

## Test plan

- [ ] Unit + in-memory e2e: `npm test -w @lifting-logbook/api` — 182 tests pass, db.e2e suite skipped (no `DATABASE_URL`)
- [ ] DB e2e (local Docker):
  ```bash
  docker-compose -f apps/api/docker-compose.test.yml up -d
  DATABASE_URL=postgresql://lifting:lifting@localhost:5433/lifting_test     npx prisma migrate deploy --schema=apps/api/prisma/schema.prisma
  DATABASE_URL=postgresql://lifting:lifting@localhost:5433/lifting_test     npm test -w @lifting-logbook/api -- --testPathPattern=db.e2e
  ```

Closes #247

### DIFF
--- a/apps/api/src/programs/programs.db.e2e.spec.ts
+++ b/apps/api/src/programs/programs.db.e2e.spec.ts
@@ -413,8 +413,8 @@ describeOrSkip('Programs HTTP (e2e, PrismaRepositoryFactory)', () => {
       expect(body.program).toBe('5-3-1');
       expect(body.cycleNum).toBe(1);
       expect(body.cycleStartDate).toBe('2026-06-02');
-      expect(body.currentWeekType).toBeDefined();
-      expect(Array.isArray(body.weeks)).toBe(true);
+      expect(body.currentWeekType).toBe('training');
+      expect(body.weeks).toEqual([]);
 
       // DB-level assertion
       const row = await prisma.cycleDashboard.findFirst({
--- a/apps/api/src/programs/programs.db.e2e.spec.ts
+++ b/apps/api/src/programs/programs.db.e2e.spec.ts
@@ -28,8 +28,10 @@ const TEST_USER = 'db-e2e-primary';
 const USER_ALICE = 'db-e2e-alice';
 const USER_BOB = 'db-e2e-bob';
 
+const USER_INIT = 'db-e2e-init';
+
 async function cleanTestUsers(prisma: PrismaClient): Promise<void> {
-  const users = [TEST_USER, USER_ALICE, USER_BOB];
+  const users = [TEST_USER, USER_ALICE, USER_BOB, USER_INIT];
   await prisma.liftRecord.deleteMany({ where: { userId: { in: users } } });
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [test] Add e2e coverage for POST cycles/initialize

### Summary

Merge with minor changes. The three-case structure matches the AC exactly, the `USER_INIT` isolation is sound, and teardown correctly covers all new rows. Two non-blocking assertions could be tightened to catch future regressions: the `weeks` check is vacuously true and the `currentWeekType` check is weaker than the service's own guarantee.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/programs.db.e2e.spec.ts:414`
`expect(Array.isArray(body.weeks)).toBe(true)` — `toCycleDashboardResponse` hardcodes `weeks: []` unconditionally. `Array.isArray([])` is always true, so this assertion passes regardless of what the mapper returns. It would not catch a regression where `weeks` is dropped, renamed, or populated with unexpected data.
**Suggestion:** Replace with `expect(body.weeks).toEqual([])`. This documents the actual contract at initialization (empty array) and fails if the mapper ever stops emitting `weeks` or emits a non-empty value.

**[maintainability]** `apps/api/src/programs/programs.db.e2e.spec.ts:415`
`expect(body.currentWeekType).toBeDefined()` — `initializeFirstCycle` explicitly sets `currentWeekType: 'training'`. The `toBeDefined()` check would pass for any non-undefined value, including `null`, `0`, or an incorrect string like `'test'`.
**Suggestion:** Replace with `expect(body.currentWeekType).toBe('training')`. The service guarantees this value at initialization; the test should hold it to that contract.

---

Issue #269, opened 2026-05-17
[feat] Implement Workout Scheduling Override (Phases 1–5)


brownm09:
Implements the full workout scheduling override feature per the design doc and proposal landed in #267.

**Design doc:** [`docs/design/workout-scheduling-integration.md`](docs/design/workout-scheduling-integration.md)
**Proposal:** [`docs/proposals/2026-05-17-workout-scheduling-override.md`](docs/proposals/2026-05-17-workout-scheduling-override.md)

---

## Phase 1: Settings UI & Data Model

- [ ] Add `UserWorkoutSchedule` interface to `packages/types/src/api.ts`
- [ ] Extend `UserSettingsResponse` with `workoutSchedule` field
- [ ] Create/update user settings database schema to store schedule
- [ ] Implement `GET /users/settings` endpoint in `apps/api`
- [ ] Implement `PATCH /users/settings` endpoint in `apps/api`
- [ ] Build Settings UI in `apps/web` with day selector (leverage prototype)
- [ ] Wire Settings UI to API, add error handling
- [ ] Write unit tests for API endpoints
- [ ] Write E2E test: set schedule, refresh page, verify persistence

## Phase 2: Distribution Core Logic

- [ ] Implement `distributeWorkouts()` in `packages/core/src/services/workout/`
- [ ] Implement `getScheduleWorkoutsPerWeek()` display helper (duration estimate)
- [ ] Write unit tests for distribution algorithm (fixed, rotating, edge cases)
- [ ] Add schedule-mode flag to settings or program-activation response (for UI to trigger prompt)
- [ ] Write integration test: set schedule + activate program, verify schedule-mode flag returned

## Phase 3: Cycle Dashboard Integration

- [ ] Extend `CycleDashboardResponse` to include workout dates
- [ ] Modify cycle creation to call `distributeWorkouts()` when schedule mode is active
- [ ] Store workout dates in database
- [ ] Modify `GET /programs/:program/cycles/:cycleNum` to return dates
- [ ] Write integration test: create cycle with schedule active, verify dates
- [ ] Write E2E test: create cycle in UI, view calendar, verify dates

## Phase 4: Lift Record Scheduling

- [ ] Modify lift record creation to fetch scheduled date from distribution
- [ ] Update `CreateLiftRecordRequest` to accept optional `scheduledDate`
- [ ] Modify UI to show scheduled date when logging a set
- [ ] Allow user to override scheduled date in UI
[...]

Issue #259, opened 2026-05-15
[test] Add Playwright E2E coverage for critical frontend flows


brownm09:
## Summary

No browser-level E2E tests exist. Every user-facing page is untested for rendering, navigation, and form submission. This issue tracks adding Playwright and covering all critical user flows.

## Motivation

The API layer has good coverage via Jest+NestJS in-memory and DB adapter tests, but the frontend has zero automated test coverage. Regressions in page rendering, navigation, and form submission are only caught by manual testing.

## Acceptance Criteria

- [ ] Playwright installed and configured in `apps/web/`
- [ ] CI step (GitHub Actions) runs Playwright tests against the dev server
- Smoke tests covering:
  - [ ] Unauthenticated user → redirected to login/onboarding
  - [ ] Onboarding: enter training maxes → select program → confirm → lands on cycle dashboard
  - [ ] Cycle dashboard: workouts render for current cycle
  - [ ] Workout logger: sets/reps display, submit lift records, confirm success state
  - [ ] Training maxes settings: values load, update, TM history table updates
  - [ ] History page: both Lift History and TM Timeline tabs render with data
  - [ ] Programs page: catalog loads, switching program navigates correctly
  - [ ] Strength goals: create a goal, verify it appears in list, delete it

## Out of Scope

- Full visual regression testing (screenshots)
- Mobile (React Native) — separate effort

## References

See `docs/testing/e2e-coverage.md` for the full flow-coverage matrix.

brownm09:
**Follow-up task added by #265 / PR #266:**

When this issue ships, update the **Coverage Requirements** table in [CLAUDE.md](../blob/main/CLAUDE.md) — remove the "until then, a written test plan in the PR body" clause from the frontend row so the row reads simply: "Playwright test in the same PR."

There is an HTML comment above the table flagging this dependency.

PR #270, opened 2026-05-17
[feat] Workout schedule data model + Settings UI (Phase 1 of 5)

brownm09:
First of two foundation PRs for [#269](https://github.com/brownm09/lifting-logbook/issues/269). Adds the user-level `workoutSchedule` field and a Settings UI to manage it. Pure data-model + UI — no behavior changes in cycle dashboard, lift records, or program editor yet (those land in Phases 3–5). Implements per [docs/design/workout-scheduling-integration.md](docs/design/workout-scheduling-integration.md).

## Changes

- **`packages/types`** — `UserWorkoutSchedule` interface (fixed days or rotating week patterns), extends `UserSettingsResponse`, new `UpdateUserSettingsRequest`.
- **`apps/api`** — nullable `workoutSchedule` JSONB column on `user_settings` (migration `20260517000000_add_workout_schedule`); `class-validator` DTO rejects out-of-range day indices (0-6), duplicates, and empty rotating weeks; repository round-trips both schedule shapes via `Prisma.JsonNull` for explicit clears.
- **`apps/web`** — new `/settings/schedule` page (Server Component → `ScheduleForm` client component) with day-of-week checkbox picker for fixed mode and add/remove repeater for rotating mode; references mockup [`docs/mockups/workout-scheduling.html`](docs/mockups/workout-scheduling.html).
- **Tests** — 10 controller + DTO cases (happy paths for fixed/rotating + validation rejections).

## Phase 1 acceptance criteria (from #269)

- [x] Add `UserWorkoutSchedule` interface to `packages/types/src/api.ts`
- [x] Extend `UserSettingsResponse` with `workoutSchedule` field
- [x] Create/update user settings database schema to store schedule
- [x] Implement `GET /users/settings` endpoint (extended, already existed)
- [x] Implement `PATCH /users/settings` endpoint (extended, already existed)
- [x] Build Settings UI with day selector (leverages prototype)
- [x] Wire Settings UI to API, add error handling
- [x] Write unit tests for API endpoints
- [ ] Write E2E test: set schedule, refresh page, verify persistence — Playwright deferred until [#259](https://github.com/brownm09/lifting-logbook/issues/259); manual test plan below

## Test plan

Per project CLAUDE.md `## Testing`:

- `npm test -w @lifting-logbook/api` — full suite: 192 passed / 51 skipped / 0 failed. New file `apps/api/src/user-settings/user-settings.controller.spec.ts` adds 10 cases.
- `npm run build -w @lifting-logbook/api` — clean.
- `npm run build -w @lifting-logbook/web` — clean; `/settings/schedule` registered as a dynamic route.
- `npm run lint -w @lifting-logbook/api` — clean.

**Manual UI smoke (recommended before merge):**
1. `npm run dev -w @lifting-logbook/web`
2. Visit `/settings/schedule`, choose **Fixed days** → check Mon/Wed/Fri → Save → reload → verify selection persists.
3. Switch to **Rotating weeks**, click **Add week**, set patterns → Save → reload → verify.
4. Try saving an empty fixed schedule → expect inline error "Pick at least one day."

**Coverage gate:** new API endpoint shape covered by the unit/validation spec. Frontend feature has no automated E2E (gate exception per CLAUDE.md `## Testing` until #259).

## Out of scope (follow-up PRs under #269)

- Phase 2: `distributeWorkouts()` core function (next PR — same branch / same issue).
- Phase 3-5: cycle dashboard date distribution, lift-record scheduled-date plumbing, program editor UI.

Refs #269 (Phase 1 of 5)

### DIFF
--- a/apps/api/prisma/migrations/20260517000000_add_workout_schedule/migration.sql
+++ b/apps/api/prisma/migrations/20260517000000_add_workout_schedule/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "user_settings" ADD COLUMN "workoutSchedule" JSONB;
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -135,8 +135,9 @@ model LiftMetadata {
 }
 
 model UserSettings {
-  userId        String  @id
-  activeProgram String?
+  userId          String  @id
+  activeProgram   String?
+  workoutSchedule Json?
 
   @@map("user_settings")
 }
--- a/apps/api/src/user-settings/update-settings.dto.ts
+++ b/apps/api/src/user-settings/update-settings.dto.ts
@@ -17,25 +17,18 @@ import {
   ValidatorConstraintInterface,
   ValidationArguments,
 } from 'class-validator';
+import { SCHEDULE_LIMITS, isValidSchedule } from '@lifting-logbook/types';
 import type { UserWorkoutSchedule } from '@lifting-logbook/types';
 
+// Delegates to the shared `isValidSchedule` predicate so the write-side and read-side
+// bounds (range, uniqueness, max weeks, max days/week) cannot drift.
 @ValidatorConstraint({ name: 'IsWeekPatternArray', async: false })
 class IsWeekPatternArrayConstraint implements ValidatorConstraintInterface {
   validate(value: unknown): boolean {
-    if (!Array.isArray(value) || value.length < 1 || value.length > 8) return false;
-    for (const week of value) {
-      if (!Array.isArray(week) || week.length < 1 || week.length > 7) return false;
-      const seen = new Set<number>();
-      for (const day of week) {
-        if (!Number.isInteger(day) || day < 0 || day > 6) return false;
-        if (seen.has(day)) return false;
-        seen.add(day);
-      }
-    }
-    return true;
+    return isValidSchedule({ type: 'rotating', weeks: value });
   }
   defaultMessage(): string {
-    return 'weeks must be 1-8 arrays of unique day indices (0-6)';
+    return `weeks must be 1-${SCHEDULE_LIMITS.MAX_ROTATING_WEEKS} arrays of unique day indices (0-6, max ${SCHEDULE_LIMITS.MAX_DAYS_PER_WEEK} per week)`;
   }
 }
 
@@ -62,7 +55,7 @@ export class WorkoutScheduleDto implements UserWorkoutSchedule {
   @ValidateIf((o: WorkoutScheduleDto) => o.type === 'fixed')
   @IsArray()
   @ArrayMinSize(1)
-  @ArrayMaxSize(7)
+  @ArrayMaxSize(SCHEDULE_LIMITS.MAX_DAYS_PER_WEEK)
   @ArrayUnique()
   @IsInt({ each: true })
   @Min(0, { each: true })
--- a/apps/api/src/user-settings/update-settings.dto.ts
+++ b/apps/api/src/user-settings/update-settings.dto.ts
@@ -1,3 +1,67 @@
+import { Type } from 'class-transformer';
+import {
+  ArrayMaxSize,
+  ArrayMinSize,
+  ArrayUnique,
+  IsArray,
+  IsIn,
+  IsInt,
+  IsOptional,
+  Max,
+  Min,
+  Validate,
+  ValidateIf,
+  ValidateNested,
+  ValidatorConstraint,
+  ValidatorConstraintInterface,
+} from 'class-validator';
+
 // Active program changes must go through POST /programs/:program/switch,
 // which enforces ownership of custom program UUIDs.
-export class UpdateSettingsDto {}
+
+@ValidatorConstraint({ name: 'IsWeekPatternArray', async: false })
+class IsWeekPatternArrayConstraint implements ValidatorConstraintInterface {
+  validate(value: unknown): boolean {
+    if (!Array.isArray(value) || value.length < 1 || value.length > 8) return false;
+    for (const week of value) {
+      if (!Array.isArray(week) || week.length < 1 || week.length > 7) return false;
+      const seen = new Set<number>();
+      for (const day of week) {
+        if (!Number.isInteger(day) || day < 0 || day > 6) return false;
+        if (seen.has(day)) return false;
+        seen.add(day);
+      }
+    }
+    return true;
+  }
+  defaultMessage(): string {
+    return 'weeks must be 1-8 arrays of unique day indices (0-6)';
+  }
+}
+
+export class WorkoutScheduleDto {
+  @IsIn(['fixed', 'rotating'])
+  type!: 'fixed' | 'rotating';
+
+  @ValidateIf((o: WorkoutScheduleDto) => o.type === 'fixed')
+  @IsArray()
+  @ArrayMinSize(1)
+  @ArrayMaxSize(7)
+  @ArrayUnique()
+  @IsInt({ each: true })
+  @Min(0, { each: true })
+  @Max(6, { each: true })
+  days?: number[];
+
+  @ValidateIf((o: WorkoutScheduleDto) => o.type === 'rotating')
+  @Validate(IsWeekPatternArrayConstraint)
+  weeks?: number[][];
+}
+
+export class UpdateSettingsDto {
+  // null clears the schedule; undefined leaves it unchanged.
+  @IsOptional()
+  @ValidateNested()
+  @Type(() => WorkoutScheduleDto)
+  workoutSchedule?: WorkoutScheduleDto | null;
+}
--- a/apps/api/src/user-settings/update-settings.dto.ts
+++ b/apps/api/src/user-settings/update-settings.dto.ts
@@ -6,6 +6,7 @@ import {
   IsArray,
   IsIn,
   IsInt,
+  IsObject,
   IsOptional,
   Max,
   Min,
@@ -14,10 +15,9 @@ import {
   ValidateNested,
   ValidatorConstraint,
   ValidatorConstraintInterface,
+  ValidationArguments,
 } from 'class-validator';
-
-// Active program changes must go through POST /programs/:program/switch,
-// which enforces ownership of custom program UUIDs.
+import type { UserWorkoutSchedule } from '@lifting-logbook/types';
 
 @ValidatorConstraint({ name: 'IsWeekPatternArray', async: false })
 class IsWeekPatternArrayConstraint implements ValidatorConstraintInterface {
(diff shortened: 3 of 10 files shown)


brownm09:
## Review: [feat] Workout schedule data model + Settings UI (Phase 1 of 5)

### Summary

Needs revision before merge. The PR is well-scoped foundation work — schema + DTO + UI all hang together cleanly — but the DTO has two validation gaps that let malformed payloads reach the JSONB column, and the repository read-path has no runtime validation, so any malformed row crashes the web form. Highest-severity issue: `WorkoutScheduleDto` accepts `{ type: 'fixed', days: [...], weeks: [...] }` and `{ workoutSchedule: "garbage" }` cleanly, both of which then persist to the database via the unchecked `as unknown as` cast. Diff is appropriately scoped at ~500 lines for one logical feature. Test coverage gate satisfied (10 controller/DTO cases; Playwright deferred per #259 — documented in PR body).

---

### Blocking Findings

**[correctness]** `apps/api/src/user-settings/update-settings.dto.ts:72-89`
`WorkoutScheduleDto` uses `@ValidateIf((o) => o.type === 'fixed')` on `days` and `@ValidateIf((o) => o.type === 'rotating')` on `weeks`. `@ValidateIf` only *skips* validation when its predicate is false — it does not *reject the presence* of the inactive field. A payload like `{ type: 'fixed', days: [0,2,4], weeks: [[9,9,9]] }` validates cleanly and persists both fields. Worse, `type: 'fixed'` with no `days` is accepted (the predicate is true but the field is undefined and there is no `@IsDefined` on the gated arms), producing a schedule that Phase 2's `distributeWorkouts()` will not be able to interpret.
**Fix:** Add a cross-field check — either a custom class-level validator that enforces `(type === 'fixed') ⇔ (days defined && weeks undefined)` and the converse for rotating, or model the DTO as a discriminated union via `@Type` with `discriminator`. Add tests for the two illegal-combination cases.

**[correctness]** `apps/api/src/user-settings/update-settings.dto.ts:93-96`
`@IsOptional() @ValidateNested() @Type(() => WorkoutScheduleDto) workoutSchedule?: WorkoutScheduleDto | null` accepts non-object primitives. `class-transformer` will not coerce a string/number/array into a class instance, and `@ValidateNested` silently treats non-objects as having nothing to check. `{ "workoutSchedule": "hacker" }` or `{ "workoutSchedule": 42 }` passes validation, then the repository casts it via `as unknown as Prisma.InputJsonValue` and writes garbage to JSONB.
**Fix:** Add `@ValidateIf((o) => o.workoutSchedule !== null && o.workoutSchedule !== undefined) @IsObject()` so non-objects are rejected while `null` (the explicit-clear sentinel) is still allowed. Add a test for `{ workoutSchedule: "x" }`.

**[correctness]** `apps/api/src/user-settings/user-settings.repository.ts:22, 50`
`getSettings` and `upsertSettings` both cast `row.workoutSchedule` to `UserWorkoutSchedule | null` via `as unknown as` with no runtime validation. Combined with the two DTO gaps above, malformed JSONB can reach the client. Even without those gaps, Prisma returns whatever JSON is in the column — a manual DB edit, a future schema migration, or a row written by an older code path will flow unchecked into `UserSettingsResponse`. The web client's `ScheduleForm.tsx` `initialState` accesses `schedule.type`, `.days`, `.weeks` and will crash or render incorrectly for a non-conforming shape.
**Fix:** Add a small `parseSchedule(value: unknown): UserWorkoutSchedule | null` type-guard at the read boundary that returns `null` on shape mismatch (and ideally logs a warning). Use it in both read paths so the asserted contract is actually enforced.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/user-settings/update-settings.dto.ts:72-89` vs `packages/types/src/api.ts:203-209`
The DTO and `UserWorkoutSchedule` interface define the same shape independently — three sources of truth (interface, DTO class, the custom validator's hand-rolled bounds). Adding a field (e.g., `startDate`) requires editing all three, and the `as unknown as` casts in the repository will swallow drift silently.
**Suggestion:** Have `WorkoutScheduleDto implements UserWorkoutSchedule` so a compile error fires when shapes diverge. Optionally add a type-level assertion to make the constraint explicit.

**[reliability]** `apps/web/app/settings/schedule/ScheduleForm.tsx:83-84`
`handleSave` surfaces server errors as `err instanceof Error ? err.message : 'Failed to save schedule.'`. The shared `apiFetch` in `lib/api.ts` typically throws errors whose message is `"Request failed: 400"` with the validation detail in the response body discarded. When the API rejects for one of the validator's many reasons (duplicate days, out-of-range index, empty rotating week), the user sees a generic 400 with no actionable detail — and the client-side `validate()` only catches two of those cases.
**Suggestion:** Extend `apiFetch` to parse the NestJS `{ message: string | string[] }` error body and surface it, or mirror every server constraint in client `validate()` so 400s never reach the user.

**[maintainability]** `apps/web/app/settings/schedule/actions.ts` + `page.tsx`
The page is a server component that calls `fetchUserSettings()` directly (matching the rest of `apps/web`), but writes go through a one-line `'use server'` action that just forwards to `updateUserSettings`. The action adds no value — no validation, no `revalidatePath`, not passed to a `<form action={...}>`, just invoked imperatively from a client handler that could call `updateUserSettings` directly via the same `apiFetch` path the rest of the app uses. This introduces a new architectural pattern for one PR's worth of code.
**Suggestion:** Either drop `actions.ts` and call `updateUserSettings` directly from the client, or commit to the server-action pattern by adding `revalidatePath('/settings/schedule')` after the write.

**[maintainability]** `apps/web/app/settings/schedule/ScheduleForm.tsx:43-45`
Switching mode (`fixed` → `rotating` → `fixed`) preserves both `fixedDays` and `rotatingWeeks` in state. `buildPayload` only emits the active mode's array, so the other rides along invisibly. Fine today, but a bug magnet the moment a future change auto-saves, sends both, or persists draft state. Users switching `fixed` → `none` → `fixed` will also be surprised that their days reappear.
**Suggestion:** Either clear the inactive mode's state on mode change, or surface a "you have unsaved selections in the other mode" hint — make the choice explicit rather than implicit.

---

### Questions for Author

- The controller test enables `whitelist: true` via the `validate()` call options, but is the global `ValidationPipe` in `main.ts` also configured with `whitelist: true`? Without it, extraneous sibling fields on the request body would reach the repository even after the DTO findings above are fixed.
- Is concurrent PATCH from the same user a real concern at this stage? The `upsert` is last-write-wins with no optimistic concurrency — fine for one tab, surprising across multiple tabs. Worth a one-line doc in the repository or a deferred ticket?

---

*Reviewed by claude-opus-4-7 via Claude Code*

brownm09:
## Review: [feat] Workout schedule data model + Settings UI (Phase 1 of 5) — Re-Review

### Summary
**Merge-ready.** This is a re-review after commit `86f8d7e` addressed three blocking findings (DTO `@ValidateIf` cross-field gap, non-object `workoutSchedule` bypass, missing repository read-path guard). Verified all three are properly fixed with corresponding test coverage (controller spec grew from 10 → 18 cases). Diff is 651 lines but appropriately scoped — Phase 1 of a designed 5-phase rollout with clear seams to phases 2–5. No blocking findings remain. Four non-blocking observations below worth addressing before the matching Phase 2 PR (they intersect with `distributeWorkouts()`).

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/user-settings/user-settings.repository.ts:354-377`
`parseSchedule` duplicates the validation logic in `WorkoutScheduleDto` (range 0-6, non-empty arrays, type discriminator) but is missing the DTO's upper bounds — `@ArrayMaxSize(7)` on days and `1-8` on weeks. A row manually edited to contain 50 weeks would pass the read-side guard while being rejected by any write through the API.
**Suggestion:** Extract a single `isValidSchedule(value: unknown): value is UserWorkoutSchedule` predicate (in `packages/core` or `packages/types`) that the DTO custom validators and the repository both call. Aligning the bounds in one place prevents drift between write-validation and read-validation.

**[reliability]** `apps/web/app/settings/schedule/ScheduleForm.tsx:42-45`
`initialState(initialSchedule)` runs only on mount; `useState` ignores subsequent initializer changes. The action now calls `revalidatePath('/settings/schedule')`, which will re-fetch on next navigation but does not propagate into this client component's state on the current render. The user's just-saved selections happen to match what the server stored, so the form *appears* fine — but if the server ever normalizes the payload (e.g., sorts/dedupes days), the displayed state silently diverges until a hard reload.
**Suggestion:** After a successful save, set local state from the returned `result.workoutSchedule` rather than trusting the form's optimistic state. Alternatively, accept `initialSchedule` in a `useEffect` that resets state when its identity changes.

**[maintainability]** `apps/web/app/settings/schedule/ScheduleForm.tsx:7` and `packages/types/src/api.ts` (the day-index convention)
`DAY_LABELS = ['Mon'...'Sun']` pins `0 = Monday`, but this convention lives implicitly in three places (DTO validator bounds, `parseSchedule`, and the UI) with no shared constant. JavaScript's `Date.getDay()` uses `0 = Sunday` — a Phase 2 contributor implementing `distributeWorkouts()` is one obvious off-by-one bug away from a wrong-week schedule. The fact that Phase 2 is the very next PR makes this worth fixing now rather than later.
**Suggestion:** Export a `DAY_INDEX` const (e.g., `{ MON: 0, TUE: 1, ..., SUN: 6 }`) or named tuple from `packages/types`, reference it in DTO, `parseSchedule`, and `DAY_LABELS`. At minimum, add a `// 0=Monday convention; do NOT use Date.getDay()` comment on each of the three sites.

**[reliability]** `apps/api/src/user-settings/update-settings.dto.ts:115-121` (the nested-whitelist coverage)
The DTO defends against unknown sibling keys via the global `ValidationPipe` configured `{ whitelist: true, forbidNonWhitelisted: true }` in `apps/api/src/main.ts`. The new controller spec at line 132 uses `forbidNonWhitelisted: false` for its DTO-level checks, which is correct for testing the validator decorators in isolation — but it means the spec never exercises the actual nested whitelist behavior used in production. A regression that loosens the global pipe config would not be caught by any test in this PR.
**Suggestion:** Add one test that constructs the controller via `Test.createTestingModule` with the production `APP_PIPE` provider, then issues a request with `{ workoutSchedule: { type: 'fixed', days: [0], extraField: 'x' } }` and asserts the extra field is stripped or rejected. Locks down the assumption that production wiring carries `forbidNonWhitelisted: true`.

---

### Questions for Author
- Is the day-index convention (`0 = Monday`) something you want documented in `packages/types/src/api.ts` as part of the `UserWorkoutSchedule` JSDoc, or is the plan to address it as part of Phase 2's `distributeWorkouts()` PR where it becomes load-bearing?

---

### Style Notes
None — comments added in this round (e.g., the `switchMode` rationale and the `actions.ts` server-boundary explanation) are the right kind of "why" comments for non-obvious decisions.

---

PR #271, opened 2026-05-17
[feat] Workout distribution algorithm (Phase 2 of 5)

brownm09:
## Summary

Phase 2 of the Workout Scheduling Override feature ([design](docs/design/workout-scheduling-integration.md) §Core Service Logic, lines 149–238).

Adds two pure functions to `packages/core/src/services/workout/`:

- **`distributeWorkouts(cycleWorkouts, schedule, cycleStartDate)`** — places N workouts across weeks according to a `UserWorkoutSchedule`, aligning the first week to the Monday of the start date's week, cycling through rotating week patterns, and emitting trimmed final weeks when the count isn't a multiple of the weekly pattern.
- **`getScheduleWorkoutsPerWeek(schedule)`** — display-only helper returning the (rounded average) workouts/week count for cycle-duration estimates.

No infrastructure dependencies; no date library. Small local `alignToMonday` helper uses `(getDay() + 6) % 7` to convert from JS `Date.getDay()` (0=Sun) to the project's `DAY_INDEX` convention (0=Mon), which is single-sourced in `@lifting-logbook/types`.

Empty-schedule and zero-count inputs short-circuit (`return []`) — guards against the infinite loop a naive `while (placed < target)` would hit when the pattern has no placeable days.

## Test plan

- [x] `npm test -w @lifting-logbook/core -- --testPathPatterns=distributeWorkouts` → 12 passed.
- [x] `npm test -w @lifting-logbook/core` (full suite) → 144 passed across 26 suites.
- [x] `npm run lint -w @lifting-logbook/core` → clean.
- [x] `npm run build -w @lifting-logbook/types && npm run build -w @lifting-logbook/core` → clean.

### Coverage

Pure core logic → unit tests required (covered): 12 tests for `distributeWorkouts` (fixed 12/MWF→4 weeks, partial last week, rotating 4/3-week alternation, mid-week Wednesday start aligned to Monday, single workout, empty fixed days, empty rotating weeks, zero count) and 4 tests for `getScheduleWorkoutsPerWeek` (fixed length, empty fixed, rotating rounded average, empty rotating). No frontend or API surface changes.

Refs #269 (Phase 2 of 5)

### DIFF
--- a/packages/core/src/services/workout/distributeWorkouts.ts
+++ b/packages/core/src/services/workout/distributeWorkouts.ts
@@ -1,0 +1,96 @@
+import { DAY_INDEX, UserWorkoutSchedule } from "@lifting-logbook/types";
+import { addDaysLocal } from "@src/core/utils/jsUtil";
+
+export interface DistributedWeek {
+  week: number;
+  workouts: Date[];
+}
+
+// Backs a date up to the Monday of its week. Uses local time to match addDaysLocal.
+// JS Date.getDay() returns 0=Sun … 6=Sat; the project encodes 0=Mon … 6=Sun via DAY_INDEX.
+// (jsDay + 6) % 7 converts: Sun(0)→6, Mon(1)→0, Tue(2)→1, …
+function alignToMonday(date: Date): Date {
+  const base = new Date(date.getFullYear(), date.getMonth(), date.getDate());
+  const offsetFromMonday = (base.getDay() + 6) % 7;
+  return addDaysLocal(base, -offsetFromMonday);
+}
+
+function getWeekPattern(
+  schedule: UserWorkoutSchedule,
+  weekIndex: number,
+): number[] {
+  if (schedule.type === "fixed") return schedule.days ?? [];
+  const weeks = schedule.weeks ?? [];
+  if (weeks.length === 0) return [];
+  return weeks[weekIndex % weeks.length] ?? [];
+}
+
+/**
+ * Distributes a fixed number of workouts across weeks according to a user schedule.
+ *
+ * The first week is aligned to the Monday of `cycleStartDate`'s week; days earlier in
+ * that week than the start date are still emitted, matching the design-doc contract
+ * (callers wanting "future days only" should filter the first week themselves).
+ *
+ * Returns an empty array when the schedule has no usable days (e.g. an empty fixed
+ * `days` array or rotating with all-empty weeks) — guards against infinite loops.
+ */
+export function distributeWorkouts(
+  cycleWorkouts: number,
+  schedule: UserWorkoutSchedule,
+  cycleStartDate: Date,
+): DistributedWeek[] {
+  if (cycleWorkouts <= 0) return [];
+
+  // Guard against schedules with no placeable days — without this, the while loop
+  // below would spin forever advancing weekCounter without placing anything.
+  const hasAnyDays =
+    schedule.type === "fixed"
+      ? (schedule.days?.length ?? 0) > 0
+      : (schedule.weeks ?? []).some((w) => w.length > 0);
+  if (!hasAnyDays) return [];
+
+  const distribution: DistributedWeek[] = [];
+  let workoutsPlaced = 0;
+  let weekCounter = 0;
+  let currentWeekStart = alignToMonday(cycleStartDate);
+
+  while (workoutsPlaced < cycleWorkouts) {
+    const weekPattern = getWeekPattern(schedule, weekCounter);
+    const weekWorkouts: Date[] = [];
+
+    for (const dayIndex of weekPattern) {
+      if (workoutsPlaced >= cycleWorkouts) break;
+      weekWorkouts.push(addDaysLocal(currentWeekStart, dayIndex));
+      workoutsPlaced++;
+    }
+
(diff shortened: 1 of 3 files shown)


brownm09:
## Review: [feat] Workout distribution algorithm (Phase 2 of 5)

### Summary
Merge with minor changes. Small, well-scoped pure-function PR (264 LOC, 12 targeted tests, no infrastructure). Implementation matches the design-doc pseudo-code, the infinite-loop guard for empty schedules is correct, and the `(getDay() + 6) % 7` day-of-week conversion is documented inline. No blocking findings. The most notable issue is a latent silent-skip behavior for rotating schedules with intermixed empty weeks — non-blocking because no caller currently constructs that shape, but worth knowing about before Phase 3 wires this in.

---

### Non-Blocking Findings

**[correctness]** `packages/core/src/services/workout/distributeWorkouts.ts:62-73`
When a rotating schedule has an empty pattern in the middle of the rotation (e.g., `weeks: [[MON,WED], [], [FRI]]`), the inner loop emits nothing for that iteration and `distribution.push` is skipped, but `weekCounter` still advances. The output's `week` field then jumps (1 → 3 with no entry for 2), so a downstream consumer iterating by array index sees calendar-week numbers that don't match their position. The current DTO doesn't validate against empty interior weeks, so this is reachable.
**Suggestion:** Either filter empty interior weeks at the validation layer (`isValidSchedule` already exists in `@lifting-logbook/types`) or document explicitly in the function's JSDoc that the `week` field may have gaps. Phase 3 should not assume `result[i].week === i + 1`.

**[maintainability]** `packages/core/src/services/workout/distributeWorkouts.ts:96`
`export { DAY_INDEX }` re-exports a constant that already lives in `@lifting-logbook/types`. Callers now have two valid import paths for the same value, which guarantees inconsistent imports across the codebase over time.
**Suggestion:** Drop the re-export. Callers that need `DAY_INDEX` import it from `@lifting-logbook/types` (one line); the convenience saved isn't worth the path ambiguity.

**[maintainability]** `packages/core/src/services/workout/distributeWorkouts.ts:22-26, 47-50, 88-89`
Defensive `schedule.days ?? []` / `schedule.weeks ?? []` fallbacks across three sites, despite `UserWorkoutSchedule` being a discriminated union where the relevant arm guarantees the field exists. The redundancy will eventually mask a real DTO/type bug (silently returning `[]` instead of crashing on malformed input that bypassed the validator).
**Suggestion:** Trust the type — drop `?? []`. The empty-array case is already covered by `hasAnyDays` for the only path that matters.

**[maintainability]** `packages/core/tests/core/services/workout/distributeWorkouts.test.ts:9-10`
`new Date(2026, 4, 18)` requires the reader to remember JS `Date` uses 0-indexed months. A reader checking "is May 18 a Monday?" has to mentally translate `4` → May first.
**Suggestion:** Inline-comment the month or use a named constant (`const MAY = 4`) so the assertions are self-evident on first read.

**[correctness]** `packages/core/src/services/workout/distributeWorkouts.ts:38-43`
`cycleWorkouts <= 0` silently returns `[]` for negative inputs. The type is `number` so `-1` is reachable from a misbehaving caller (e.g., a UI input bug).
**Suggestion:** Either narrow the type to a branded non-negative integer at the call site, or throw on negative input — silent acceptance of nonsense input pushes the failure downstream where the cause is harder to find.

---

### Questions for Author

- Phase 3 design doc lines 202–220 mention `scheduleLiftDates()` calling this — is the `week` field's possible non-monotonic behavior (per the first finding) acceptable for that consumer, or should we tighten the contract first?

---

brownm09:
## Re-Review (Round 2): [feat] Workout distribution algorithm (Phase 2 of 5)

### Summary
**Ready to merge.** All round-1 non-blocking findings either addressed in commit `e7c88cb` or explicitly justified. Verified against `origin/feat/issue-269-distribute-workouts` (not local worktree). No new findings introduced by the fixes.

### Status of Round 1 findings

| # | Finding | Status |
|---|---|---|
| 1 | `week` field gaps in rotating empty-week case | ✅ JSDoc documents the contract; Phase 3 advised to compute its own positional index |
| 2 | `DAY_INDEX` re-export ambiguity | ✅ Removed re-export and the now-unused import |
| 3 | Defensive `?? []` fallbacks | 🟡 **Not applied — review premise was incorrect.** `UserWorkoutSchedule` keeps `days?` and `weeks?` as optional on a single interface (not a discriminated union), so the fallbacks are required for tsc soundness. Author added an inline comment recording the constraint. Accepted; tightening the type to a true discriminated union is a Phase 1 follow-up that would touch DTO + repository + `isValidSchedule`. |
| 4 | JS Date 0-indexed month readability | ✅ `const MAY = 4` introduced; comment updated |
| 5 | Silent acceptance of negative `cycleWorkouts` | ✅ Now throws with clear message; new test asserts the throw |

### Verification (re-pulled from remote)
- 13 tests in `distributeWorkouts.test.ts` (was 12)
- Full core suite 145/145 (was 144)
- Lint + build clean

### No new findings
The throw guard, JSDoc additions, and test naming changes do not introduce new correctness, security, reliability, performance, or maintainability concerns. The retained `?? []` fallbacks are bounded by the type-system constraint and now documented.

**Merge recommended.**

---

PR #282, opened 2026-05-18
[feat] Workout Scheduling Phase 5 — Skip UI + schedule-mode prompt

### LINKED ISSUE #281: [feat] Workout Scheduling Override (Phase 5 of 5)
Implements the final phase of the workout scheduling override feature.

**Parent:** #269
**Design doc:** [docs/design/workout-scheduling-integration.md](docs/design/workout-scheduling-integration.md)

## What this closes

Phase 5 wires the remaining UI that phases 1–4.5 left headless:

### Skip/Unskip UI
- [ ] Add `skipped: boolean` to `WorkoutResponse` type and GET workout API endpoint
- [ ] Add `skipWorkout` / `unskipWorkout` client functions to `lib/client-api.ts`
- [ ] Add `'skipped'` status to `WorkoutCell` and map from cycle dashboard response
- [ ] Show "Skipped" badge in cycle dashboard grid
- [ ] Add `SkipForm` component to workout detail page (skip / undo skip with optional reason)

### Schedule-mode confirmation
- [ ] Thread `workoutSchedule` from programs page → `SwitchProgramDialog`
- [ ] Show schedule-info step after confirming program switch (when schedule is set): schedule summary + est. workouts/week

### Tests
- [ ] Unit: GET workout with skip override returns `skipped: true`
- [ ] E2E: skip a workout → Skipped badge shown; undo skip → reverts
- [ ] E2E: select program with schedule set → schedule-info step appears

## Out of scope (deferred)
- Calendar preview widget (7-column grid)
- Schedule redistribution when schedule changes while program is active

brownm09:
## Summary

- **Skip/Unskip UI**: New `SkipForm` component on the workout detail page lets users mark a workout as skipped (with optional reason) or undo a skip. The cycle dashboard now shows a Skipped badge for skipped workouts.
- **`WorkoutResponse.skipped`**: The GET workout endpoint now fetches the skip set from `workoutSkipOverride.getSkipsForCycle()` in parallel with other data and includes `skipped: boolean` in the response.
- **Schedule-mode confirmation prompt**: When a user with a `workoutSchedule` configured switches to a new program, `SwitchProgramDialog` now shows an informational second step after the switch confirming that workout dates have been distributed, displaying the schedule summary (e.g., "Mon / Wed / Fri") and estimated workouts per week.
- **Fix: "completed" status now requires actual logged lifts** — `WorkoutResponse.lifts` always includes planned entries for upcoming workouts. The cycle dashboard's `logged` check was changed from `lifts.length > 0` (which counted planned-only lifts, making every workout appear completed) to `lifts.some(l => !l.planned)` (actual logged sets only).

## Acceptance Criteria

- [x] Skip a workout from detail page — Skipped badge shown, Start Logging hidden, Undo Skip button visible
- [x] Undo skip — workout reverts to Upcoming/Missed status
- [x] Cycle dashboard card shows Skipped status for skipped workouts
- [x] Selecting a program with schedule set shows schedule-info step with schedule summary and workouts/week
- [x] Selecting a program without schedule skips the schedule-info step (navigates directly)

## Changes

| Area | Change |
|---|---|
| `packages/types/src/api.ts` | Add `skipped: boolean` to `WorkoutResponse` |
| `apps/api/src/programs/mappers.ts` | Accept `skipped` param in `toWorkoutResponse` |
| `apps/api/src/programs/workouts.controller.ts` | Fetch skip set in parallel, pass to mapper; log errors before skip-set fallback |
| `apps/web/lib/client-api.ts` | Add `skipWorkout`, `unskipWorkout`; handle 204 No Content in `clientFetch` |
| `apps/web/lib/workoutPlan.ts` | Add `'skipped'` to `WorkoutCell.status` |
| `apps/web/app/cycle/[cycleNum]/page.tsx` | Derive skipped status from dashboard; fix `logged` predicate |
| `apps/web/app/cycle/[cycleNum]/CycleDashboardGrid.tsx` | Skipped badge + card style |
| `apps/web/app/cycle/[cycleNum]/workout/[workoutNum]/detail/SkipForm.tsx` | **New** — skip/unskip toggle component |
| `apps/web/app/cycle/[cycleNum]/workout/[workoutNum]/detail/page.tsx` | Render SkipForm, hide Start Logging when skipped |
| `apps/web/app/programs/page.tsx` + `ProgramsTabs.tsx` + `BrowseTab.tsx` | Thread `workoutSchedule` to dialog |
| `apps/web/app/programs/SwitchProgramDialog.tsx` | Add schedule-info step with Close button and state reset |
| `apps/web/e2e/mock-api.mjs` | Skip state tracking + `withSchedule` reset param |
| `apps/web/e2e/scheduling.spec.ts` | **New** — 5 Playwright E2E tests |

## Design notes

- Skip is a label orthogonal to log data: a partially-logged workout can be marked skipped. No server-side guard prevents both states coexisting. The UI prevents reaching that state through normal paths (`completed` status hides SkipForm; `skipped` status hides Start Logging); this is documented with a comment in `detail/page.tsx`.

## Tests

**Unit tests:** `npm test -w @lifting-logbook/api` — 243 passed, 3 new tests covering `skipped: false`, `skipped: true`, and correct cycle scoping in the workouts controller.

**Web tests:** `npm test -w @lifting-logbook/web` — 28 passed (no regressions).

**E2E tests:** 5 new Playwright scenarios in `apps/web/e2e/scheduling.spec.ts` — skip flow, undo skip, dashboard badge, and schedule-mode prompt with and without a configured schedule.

Note: `npm run build` fails on the API package with pre-existing Prisma type errors unrelated to this PR (present on `main` before this branch).

Closes #281

### DIFF
--- a/apps/api/src/programs/mappers.ts
+++ b/apps/api/src/programs/mappers.ts
@@ -234,6 +234,7 @@ export const toWorkoutResponse = (
   overrideDate?: Date,
   plannedLifts?: string[],
   scheduledDate?: Date,
+  skipped = false,
 ): WorkoutResponse => {
   const liftMap = new Map<string, SetResponse[]>();
   for (const r of records) {
@@ -282,6 +283,7 @@ export const toWorkoutResponse = (
     week,
     date,
     ...(overrideDate !== undefined && { overrideDate: isoDate(overrideDate) }),
+    skipped,
     lifts,
   };
 };
--- a/apps/api/src/programs/workout-skip.controller.ts
+++ b/apps/api/src/programs/workout-skip.controller.ts
@@ -9,7 +9,7 @@ import {
   Param,
   Post,
 } from '@nestjs/common';
-import { IsOptional, IsString } from 'class-validator';
+import { IsOptional, IsString, MaxLength } from 'class-validator';
 import { CurrentUser } from '../auth/current-user.decorator';
 import { AuthUser } from '../ports/auth';
 import { IRepositoryFactory } from '../ports/factory';
@@ -19,6 +19,7 @@ import { isValidWorkoutNum } from './mappers';
 class SkipWorkoutDto {
   @IsOptional()
   @IsString()
+  @MaxLength(500)
   reason?: string;
 }
 
--- a/apps/api/src/programs/workouts.controller.spec.ts
+++ b/apps/api/src/programs/workouts.controller.spec.ts
@@ -7,6 +7,7 @@ import { ILiftingProgramSpecRepository } from '../ports/ILiftingProgramSpecRepos
 import { IWorkoutDateOverrideRepository } from '../ports/IWorkoutDateOverrideRepository';
 import { IWorkoutLiftOverrideRepository } from '../ports/IWorkoutLiftOverrideRepository';
 import { IWorkoutRepository } from '../ports/IWorkoutRepository';
+import { IWorkoutSkipOverrideRepository } from '../ports/IWorkoutSkipOverrideRepository';
 import { ProgramNotFoundError, WorkoutNotFoundError } from '../ports/errors';
 import { IRepositoryFactory } from '../ports/factory';
 import { REPOSITORY_FACTORY } from '../ports/tokens';
@@ -22,6 +23,7 @@ describe('WorkoutsController', () => {
   let overrideRepo: jest.Mocked<IWorkoutDateOverrideRepository>;
   let liftOverrideRepo: jest.Mocked<IWorkoutLiftOverrideRepository>;
   let scheduledWorkoutRepo: jest.Mocked<ICycleScheduledWorkoutRepository>;
+  let skipOverrideRepo: jest.Mocked<IWorkoutSkipOverrideRepository>;
   let factory: jest.Mocked<IRepositoryFactory>;
 
   beforeEach(async () => {
@@ -44,6 +46,11 @@ describe('WorkoutsController', () => {
       getScheduledWorkouts: jest.fn().mockResolvedValue([]),
       saveScheduledWorkouts: jest.fn().mockResolvedValue(undefined),
     };
+    skipOverrideRepo = {
+      getSkipsForCycle: jest.fn().mockResolvedValue(new Set<number>()),
+      skipWorkout: jest.fn().mockResolvedValue(undefined),
+      unskipWorkout: jest.fn().mockResolvedValue(undefined),
+    };
     factory = {
       forUser: jest.fn().mockResolvedValue({
         workout: workoutRepo,
@@ -52,6 +59,7 @@ describe('WorkoutsController', () => {
         liftingProgramSpec: specRepo,
         workoutDateOverride: overrideRepo,
         workoutLiftOverride: liftOverrideRepo,
+        workoutSkipOverride: skipOverrideRepo,
       }),
     };
     const module: TestingModule = await Test.createTestingModule({
@@ -319,6 +327,54 @@ describe('WorkoutsController', () => {
 
       expect(result.lifts).toHaveLength(2);
       expect(result.lifts[0]).toMatchObject({ lift: 'Squat', planned: true });
+    });
+  });
+
+  describe('skipped field', () => {
+    const dashboard = {
+      program: '5-3-1',
+      cycleUnit: 'week',
+      cycleNum: 3,
+      cycleDate: new Date('2026-04-20T00:00:00.000Z'),
+      sheetName: '',
+      cycleStartWeekday: Weekday.Monday,
+      currentWeekType: 'training' as const,
+    };
+    const spec = [
+      { week: 1, offset: 0, lift: 'Squat', increment: 5, order: 1, sets: 3, reps: 5, amrap: false, warmUpPct: '', wtDecrementPct: 0, activation: 'compound' },
+    ];
+
+    it('returns skipped: false when the workout is not in the skip set', async () => {
+      dashboardRepo.getCycleDashboard.mockResolvedValue(dashboard);
+      specRepo.getProgramSpec.mockResolvedValue(spec);
+      workoutRepo.getWorkout.mockRejectedValue(new WorkoutNotFoundError('5-3-1', 3, 1));
+      skipOverrideRepo.getSkipsForCycle.mockResolvedValue(new Set<number>());
+
+      const result = await controller.getWorkout('5-3-1', '1', MOCK_USER);
+
+      expect(result.skipped).toBe(false);
+    });
+
+    it('returns skipped: true when the workout is in the skip set', async () => {
+      dashboardRepo.getCycleDashboard.mockResolvedValue(dashboard);
+      specRepo.getProgramSpec.mockResolvedValue(spec);
+      workoutRepo.getWorkout.mockRejectedValue(new WorkoutNotFoundError('5-3-1', 3, 1));
+      skipOverrideRepo.getSkipsForCycle.mockResolvedValue(new Set([1]));
+
+      const result = await controller.getWorkout('5-3-1', '1', MOCK_USER);
+
+      expect(result.skipped).toBe(true);
+    });
+
+    it('fetches skip set for the current cycle', async () => {
+      dashboardRepo.getCycleDashboard.mockResolvedValue(dashboard);
+      specRepo.getProgramSpec.mockResolvedValue(spec);
+      workoutRepo.getWorkout.mockRejectedValue(new WorkoutNotFoundError('5-3-1', 3, 1));
+      skipOverrideRepo.getSkipsForCycle.mockResolvedValue(new Set<number>());
+
+      await controller.getWorkout('5-3-1', '1', MOCK_USER);
+
+      expect(skipOverrideRepo.getSkipsForCycle).toHaveBeenCalledWith('5-3-1', 3);
     });
   });
 });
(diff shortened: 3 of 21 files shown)


brownm09:
## Review: [feat] Workout Scheduling Phase 5 — Skip UI + schedule-mode prompt

### Summary
This is solid, well-scoped work: the skip/unskip flow, `WorkoutResponse.skipped`, and the schedule-info dialog step are all coherently implemented with matching unit tests and E2E coverage. One blocking issue — a missing error handler on `getSkipsForCycle` — will cause a 500 for all workout detail loads if the skip repository hits a transient failure. Everything else is non-blocking. Ready to merge after that catch is added.

---

### Blocking Findings

**[reliability]** `apps/api/src/programs/workouts.controller.ts:68`
`workoutSkipOverride.getSkipsForCycle(program, dashboard.cycleNum)` is added to the `Promise.all` without a `.catch` handler. Every other entry in that `Promise.all` either has an explicit catch (e.g., `workout.getWorkout` → maps to 404) or is expected to throw and bubble up as an HTTP error. If `getSkipsForCycle` rejects due to a transient DB error, the entire `Promise.all` rejects and the user gets a 500 on every workout detail page load — not just for skipped workouts.
**Fix:** Add `.catch(() => new Set<number>())` to fail open (workout displays as not-skipped, which is the safer degradation), or throw a mapped HTTP error consistent with the endpoint's error model.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/mappers.ts:237`
`skipped = false` is the 4th optional positional parameter after `overrideDate`, `plannedLifts`, and `scheduledDate`. Any future call site that needs to reach `skipped` must pass `undefined` placeholders for the preceding optional args, with no TypeScript error to warn about a wrong position. The one existing call site in this PR was updated correctly, but positional optional params with defaults become a maintenance trap as the function accretes more.
**Suggestion:** When next touching this function, consider consolidating the tail optional params into an options bag (`{ overrideDate?, plannedLifts?, scheduledDate?, skipped? }`).

**[maintainability]** `apps/web/app/cycle/[cycleNum]/workout/[workoutNum]/detail/SkipForm.tsx:31`
The catch block is bare (`catch { ... }`), discarding the original error. Skip/unskip failures (DB errors, 4xx responses, network failures) are now indistinguishable — all surface the same generic message. This matters for triaging "Failed to skip workout" reports on a new mutation path with no prior incident history.
**Suggestion:** Capture the error (`catch (e)`) and `console.error` it so it appears in browser dev tools and any telemetry. For 4xx responses with a server-provided message, surface that message instead of the generic fallback.

**[performance]** `apps/api/src/programs/workouts.controller.ts:68`
`getSkipsForCycle` returns a `Set<number>` of all skipped workout numbers for the cycle, then immediately calls `.has(workoutNum)` to answer a single boolean. On a single-workout GET request, loading the full set is over-fetching relative to the question being answered — it scales with the number of skips in the cycle, adds an extra DB round-trip for all workout detail page loads, and is called from the same endpoint that already does 4+ parallel queries.
**Suggestion:** Add an `isSkipped(program, cycleNum, workoutNum): Promise<boolean>` method to `IWorkoutSkipOverrideRepository` for the single-workout case. Reserve `getSkipsForCycle` for the dashboard where the full set is needed.

**[reliability]** `apps/web/app/programs/SwitchProgramDialog.tsx` (handleConfirm)
When `workoutSchedule` is set and the switch succeeds, the dialog transitions to the `schedule-info` step but `router.refresh()` is not called until the user clicks "Go to Cycle". If the user uses browser back-navigation after that push, the programs page may display stale active-program state.
**Suggestion:** Call `router.refresh()` when transitioning to the schedule-info step (not just in `handleGoToCycle`) so the programs page is already invalidated regardless of how the user exits.

**[maintainability]** `apps/web/app/programs/page.tsx:7`
The `.catch` fallback `{ activeProgram: null, workoutSchedule: null as UserWorkoutSchedule | null }` uses an inline cast to satisfy the `Promise.all` tuple type. If `UserSettingsResponse` gains a third field, this fallback silently becomes structurally incomplete and TypeScript will not warn at the call site.
**Suggestion:** Extract `const DEFAULT_SETTINGS: UserSettingsResponse = { activeProgram: null, workoutSchedule: null }` as a typed constant, or have `fetchUserSettings` own its own typed fallback internally.

---

### Questions for Author

- **Reason field backend validation:** `SkipForm` sends `reason` as a free-text string with no client-side `maxLength`. Does the skip endpoint (`POST .../skip`) validate and bound the `reason` field server-side? If not, an arbitrarily large string reaches the DB. A `maxLength={500}` attribute on the input would provide consistent UX even if backend validation exists.

- **`formatSchedule` fallback:** The function returns `'Custom schedule'` if `schedule.type` is neither `'fixed'` nor `'rotating'`, but `UserWorkoutSchedule['type']` is a `'fixed' | 'rotating'` union. Is this fallback intentionally unreachable dead code (serving as a runtime safety net) or is there a future schedule type planned?

---

### Style Notes

`SwitchProgramDialog.tsx` renders the schedule-info step as a separate early-return `if` block while the confirm step is a second `return` at the bottom. This works, but the component now has two `role="dialog"` roots with no shared wrapper — consider extracting both step bodies into a single `<dialog>` shell to avoid duplicating the overlay/container markup.

---

brownm09:
## Review: [feat] Workout Scheduling Phase 5 — Skip UI + schedule-mode prompt

### Summary
Needs revision before merge — two blocking issues. The skip/unskip mechanics, API changes, and E2E test coverage are solid; the highest-priority finding is the schedule-info dialog trapping users with no dismiss path once they've switched programs. A second issue is silent error swallowing in the skip-store fetch with no log trail. Four non-blocking findings and two questions follow.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/workouts.controller.ts:68`
`workoutSkipOverride.getSkipsForCycle(...).catch(() => new Set<number>())` swallows all errors silently and returns an empty set. Every other `.catch` in this controller (e.g., the `getWorkout` catch) either logs the error or re-throws. When the skip store fails, every workout silently appears as "not skipped" — users see their skips gone, re-skip, and no operator signal exists to indicate the backend issue.
**Fix:** Log the error before returning the fallback, matching the pattern of the sibling catches: `.catch((err: unknown) => { /* log err */ return new Set<number>(); })`.

**[correctness]** `apps/web/app/programs/SwitchProgramDialog.tsx:68–98`
The `schedule-info` step has no dismiss path. Escape is explicitly blocked (`step === 'confirm'` guard), there is no Cancel button, and clicking outside the overlay does nothing (no `onClick` on the outer `div`). Once the program switch succeeds server-side and the dialog advances to `schedule-info`, the user is forced to navigate to the new cycle — they cannot stay on the programs page.
**Fix:** Add a close affordance for the `schedule-info` step: either remove the `step === 'confirm'` guard so Escape works in both steps and calls `onClose()`, or add a secondary "Close" / "Stay here" button that calls `onClose()` without navigating.

---

### Non-Blocking Findings

**[reliability]** `apps/web/app/programs/SwitchProgramDialog.tsx:handleConfirm`
`router.refresh()` is called immediately on the `schedule-info` branch, before the user clicks "Go to Cycle". This refreshes the current `/programs` route while the user is still looking at the dialog — wasted work that can cause a visual flicker on the programs list behind the modal. The meaningful refresh is already called in `handleGoToCycle`.
**Suggestion:** Drop `router.refresh()` from the `schedule-info` branch of `handleConfirm`.

**[maintainability]** `apps/web/app/programs/SwitchProgramDialog.tsx:35–38`
`step` and `cycleNum` state are never reset when the dialog is closed or reopened. If the dialog is dismissed mid-flow (e.g., an error occurs, the user closes and reopens it), it may re-enter on `schedule-info` with a stale `cycleNum`.
**Suggestion:** Reset both to their initial values (`'confirm'` / `null`) in the `onClose` handler, or when the dialog mounts/unmounts.

**[maintainability]** `apps/web/app/cycle/[cycleNum]/workout/[workoutNum]/detail/SkipForm.tsx:35`
`handleConfirm` branches on `skipped` to choose the API call, the error message, and (at render time) the button label/class. The two flows already differ significantly — one has a reason input, a different info copy, and a different button style. As the flows further diverge, boolean-flag branching through a single confirm path will obscure intent.
**Suggestion:** Extract the skip and unskip paths into separate internal handlers (e.g., `handleSkip` / `handleUnskip`) called from `handleConfirm`, so each path reads linearly.

**[maintainability]** `apps/web/app/cycle/[cycleNum]/workout/[workoutNum]/detail/page.tsx:141`
`<SkipForm>` is rendered only when `status !== 'completed'`. If a workout ever has both logs and an active skip override (reachable via direct API calls, not through the UI), the page renders `status = 'completed'` and hides `SkipForm` entirely — the user has no way to undo the skip. The UI gates prevent this state through normal flows, but the server doesn't validate it.
**Suggestion:** Consider rendering `SkipForm` when `workout.skipped` is true regardless of status, or add a server-side guard that prevents logging a skipped workout. At minimum, add a comment documenting the intentional precedence decision.

---

### Questions for Author

- **`logged` predicate change** (`apps/web/app/cycle/[cycleNum]/page.tsx:65`): the check was changed from `response.lifts.length > 0` to `response.lifts.some((l) => !l.planned)`. This changes the dashboard's definition of "completed" — a workout with only planned lifts no longer counts as logged. Is this an intentional semantic fix or a side effect of the skip work? The PR body doesn't call it out separately, and no unit test pins the new semantics on the dashboard (the E2E test covers it, but a unit test would make the intent explicit).

- **Skipping a workout with logs**: the server has no guard preventing a user from calling `POST .../skip` on a workout that already has lift records. Is that intentional (skip is purely a UI label, orthogonal to log data), or should the endpoint return a 409 when logs exist?

---

### Style Notes

The `STATUS_LABELS` record object extracted from the `StatusBadge` conditional chain is a clean improvement — it makes adding future statuses a one-line change. The `formatSchedule` helper in `SwitchProgramDialog` handles the `rotating` schedule case gracefully. No style concerns.

---

PR #286, opened 2026-05-20
[fix] Fix terraform config for initial production bootstrap

### LINKED ISSUE #285: Fix terraform config for initial production bootstrap
Several issues discovered during the first `terraform apply` against a fresh GCP project:

- `replication { auto {} }` single-line syntax requires Terraform 1.9+ but config only requires 1.7
- Missing APIs: `vpcaccess.googleapis.com` not in `required_apis`
- `PORT` is a reserved Cloud Run env var and cannot be set manually
- No placeholder image strategy for initial bootstrap before CI/CD has pushed images
- `web_workload` SA missing `secretmanager.secretAccessor` IAM binding
- No `time_sleep` after API enablement causes propagation-lag 403s on fresh projects

Also adds `scripts/deploy-prod-infra.sh` to automate the terraform init/workspace/apply sequence.

brownm09:
Fixes several issues discovered during the first \`terraform apply\` against a fresh GCP project. All fixes were validated by a successful apply that completed infrastructure provisioning.

## Changes

- **Syntax compat**: Expand `replication { auto {} }` to multi-line form (single-line requires Terraform 1.9+; config only requires 1.7)
- **Missing API**: Add `vpcaccess.googleapis.com` to `required_apis` (needed by Serverless VPC connector for Cloud Run)
- **Propagation lag**: Add `hashicorp/time` provider and `time_sleep.api_propagation` (60s) between API enablement and VPC creation — prevents 403s on fresh projects
- **Reserved env var**: Remove `PORT` from both Cloud Run service definitions (Cloud Run sets it automatically)
- **Bootstrap image**: Add `image_tag = "bootstrap"` sentinel in `terraform.tfvars.production` that resolves to Google's public hello image; `lifecycle { ignore_changes = [template] }` already prevents Terraform from fighting CI/CD image updates
- **Missing IAM**: Add `secretmanager.secretAccessor` for `web_workload` SA (needed to read `clerk-publishable-key` at service startup)
- **New script**: `scripts/deploy-prod-infra.sh` automates the init → workspace → apply sequence with a `--plan-only` flag for safe previews

## Testing

Full `terraform apply` completed successfully against `lifting-logbook-prod`. All 16 resources created. Outputs confirmed:
- Cloud Run API and Web services running with placeholder image
- Workload Identity Provider and CI/CD SA email available for GitHub Actions secrets

Closes #285

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ next-env.d.ts
 
 # Claude Code local overrides (machine-specific permissions, not for team sharing)
 .claude/settings.local.json
+scripts/cloud-sql-proxy
+scripts/cloud-sql-proxy.exe
--- a/docs/deploy-single-user.md
+++ b/docs/deploy-single-user.md
@@ -110,7 +110,7 @@ single user.
    - Publishable key (`pk_live_…`)
    - Secret key (`sk_live_…`)
 
-You'll load these into Secret Manager in Step 6.
+You'll load these into Secret Manager in Step 7.
 
 ---
 
@@ -182,7 +182,78 @@ Cloud Run reuses it as the API service identity.
 
 ---
 
-## Step 6 — Populate Clerk secrets
+## Step 6 — Map a custom domain (optional)
+
+Skip this step if you're happy using the Cloud Run URL directly.
+
+### 6a — Verify domain ownership
+
+Google requires one-time domain ownership verification before a Cloud Run domain mapping will be accepted.
+
+```bash
+gcloud domains verify your-domain.com
+```
+
+This opens Google Search Console in your browser. Add the TXT record it shows at your registrar:
+
+| Field | Value |
+|---|---|
+| Type | TXT |
+| Name / Host | `@` |
+| Value | the string Google provides |
+
+> **Important:** the Name field must be exactly `@`. Other values (blank, the full domain name, etc.) fail silently.
+
+Click **Verify** in Search Console once the record is live. DNS propagation can take a few minutes.
+
+### 6b — Create the domain mappings
+
+```bash
+gcloud beta run domain-mappings create \
+  --service lifting-logbook-prod-web \
+  --domain your-domain.com \
+  --region us-central1 \
+  --project lifting-logbook-prod
+
+gcloud beta run domain-mappings create \
+  --service lifting-logbook-prod-web \
+  --domain www.your-domain.com \
+  --region us-central1 \
+  --project lifting-logbook-prod
+```
+
+### 6c — Add DNS records
+
+```bash
+gcloud beta run domain-mappings describe \
+  --domain your-domain.com \
+  --region us-central1 \
+  --project lifting-logbook-prod \
+  --format="table(status.resourceRecords[].type, status.resourceRecords[].name, status.resourceRecords[].rrdata)"
+```
+
+At your registrar, add:
+- Four `A` records pointing `@` to the IPs shown
+- Four `AAAA` records pointing `@` to the IPv6 addresses shown
+- One `CNAME` record pointing `www` to `ghs.googlehosted.com.`
+
+Replace any existing `A`/`AAAA` records for the apex; keep existing `MX` and `TXT` records.
+
+SSL is provisioned automatically once DNS propagates. Check status with:
+
+```bash
+gcloud beta run domain-mappings describe \
+  --domain your-domain.com \
+  --region us-central1 \
+  --project lifting-logbook-prod \
+  --format="value(status.conditions[0].reason)"
+```
+
+`CertificateProvisioned` means the cert is active. Provisioning typically completes within 15–60 minutes.
+
+---
+
+## Step 7 — Populate Clerk secrets
 
 Terraform created empty secret containers with
 `lifecycle.ignore_changes = [secret_data]`. Add the actual values now:
@@ -197,7 +268,7 @@ echo -n "pk_live_..." | gcloud secrets versions add \
 
 ---
 
-## Step 7 — Apply database migrations (one-time)
+## Step 9 — Apply database migrations (one-time)
 
 The API container does **not** run `prisma migrate deploy` on startup. Use the
 Cloud SQL Auth Proxy:
@@ -215,7 +286,7 @@ cd apps/api && DATABASE_URL="$DATABASE_URL" npx prisma migrate deploy
 
 ---
 
-## Step 8 — Trigger the first deploy
+## Step 10 — Trigger the first deploy
 
 Merge any branch to `main` (or push directly if you have the workflow set up
 that way). The [Deploy workflow](../.github/workflows/deploy.yml):
@@ -233,7 +304,7 @@ so the production job will pause until you approve it.
 
 ---
 
-## Step 9 — Create your Clerk user, log in, verify
+## Step 11 — Create your Clerk user, log in, verify
 
 Self-serve signup isn't wired by default. Provision yourself:
 
@@ -294,7 +365,6 @@ These come from [`docs/deploy.md`](deploy.md). Skip them for single-user; add
 them later if you ever invite a second person:
 
 - A separate `lifting-logbook-staging` project and the full A/B `90/10` split.
-- A custom domain via Cloud Load Balancer.
 - Self-serve signup / open registration.
 - Rate limiting beyond what Cloud Run gives you by default.
 - On-call runbook / alert routing.
--- a/docs/deploy-single-user.md
+++ b/docs/deploy-single-user.md
@@ -98,19 +98,29 @@ The script's `--help` flag prints the full header documentation.
 
 ---
 
-## Step 2 — Create a Clerk production app (~5 min)
+## Step 2 — Create and configure a Clerk production instance (~10 min)
 
-[Clerk](https://clerk.com) handles authentication. The free tier covers a
-single user.
+[Clerk](https://clerk.com) handles authentication. The free tier covers a single user.
 
-1. Sign up at https://clerk.com.
-2. Create an application named `lifting-logbook-production`.
-3. Pick auth methods (email + password is simplest).
-4. From the dashboard, **API Keys** → copy:
+1. Sign up / log in at https://clerk.com and create an application named `lifting-logbook-production`.
+2. Choose auth methods (email + password is simplest for a single user).
+3. **Switch to Production mode:** click the **Development** badge at the top of the dashboard → **Switch to Production**.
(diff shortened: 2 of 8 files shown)


brownm09:
## Review: [fix] Fix terraform config for initial production bootstrap

### Summary

This PR is nearly merge-ready — all seven bootstrap issues are correctly diagnosed and fixed, and a successful `terraform apply` validates the approach. Three issues should be resolved before merging: (1) the `web_workload` Secret Manager IAM binding is project-scoped when it should be secret-scoped, granting the web SA access to database credentials it shouldn't have; (2) the real billing account ID was committed despite the file header explicitly prohibiting it; and (3) `scripts/README.md` (which has a Production deploy table and explicit instructions to add entries) wasn't updated.

---

### Blocking Findings

**[security]** `cloud-run.tf:198-202`
`google_project_iam_member.web_workload_secret_accessor` grants `roles/secretmanager.secretAccessor` at the **project level**. The web service only needs to read `clerk-publishable-key`, but this binding also gives the Next.js SA read access to `database_url`, `system_database_url`, and `clerk_secret_key`. If the web container is compromised, an attacker can extract production database credentials via the metadata server — they shouldn't be in scope for the web workload at all. `clerk_secret_key` in particular is the Clerk admin key and should be API-only.
**Fix:** Replace with `google_secret_manager_secret_iam_member` scoped to the single secret:
```hcl
resource "google_secret_manager_secret_iam_member" "web_workload_publishable_key" {
  secret_id = google_secret_manager_secret.clerk_publishable_key.secret_id
  role      = "roles/secretmanager.secretAccessor"
  member    = "serviceAccount:${google_service_account.web_workload.email}"
}
```

**[correctness]** `terraform.tfvars.production:7`
The real billing account ID `01F30C-20F0EA-D44922` is committed. The file header in the *same file* reads: `# billing_account: do NOT commit a real value here — pass via CI/CD secret instead`. The original placeholder `XXXXXX-XXXXXX-XXXXXX` was correct per that convention; this diff replaces it with the live value, embedding it permanently in git history.
**Fix:** Revert to a placeholder (e.g., `REPLACE_ME`) and pass the real value at apply time via `-var="billing_account=$GCP_BILLING_ACCOUNT"`, consistent with the deploy script's existing design.

**[documentation]** `scripts/README.md` not updated
`scripts/README.md` has a "Production deploy" table that lists every production script, and its own footer says: *"When adding a new script, add a row to the appropriate table above."* `deploy-prod-infra.sh` was added without an entry.
**Fix:** Add a row to the "Production deploy" table:
```
| [`deploy-prod-infra.sh`](deploy-prod-infra.sh) | Automates terraform init → workspace select → apply for the production environment. Use `--plan-only` to preview. Run after `bootstrap-gcp-prod.sh`. |
```

---

### Non-Blocking Findings

**[maintainability]** `main.tf:65-68`
`time_sleep.api_propagation` has `create_duration = "60s"` but no comment noting that `time_sleep` only fires on resource creation, not on subsequent applies. If a new API is added to `required_apis` in a future PR, no propagation delay will occur — the resource is already in state. The PR body documents this correctly ("only fires on first apply of a fresh project") but the code comment doesn't.
**Suggestion:** Add a one-liner: `# Only fires on initial resource creation — does not re-trigger when new APIs are added incrementally.`

**[reliability]** `scripts/deploy-prod-infra.sh:55`
`--project-id` overrides the GCS state bucket name but not the vars file (`TFVARS` stays hardcoded to `terraform.tfvars.production`). If someone passes `--project-id lifting-logbook-staging` thinking they're targeting staging, they'll apply production variable values to a different project's state.
**Suggestion:** Either drop `--project-id` entirely and document this as a production-only script, or derive the vars file from the project ID: `TFVARS="terraform.tfvars.${WORKSPACE}"`.

---

brownm09:
## Review: [fix] Fix terraform config for initial production bootstrap

### Summary

This PR successfully addresses the three blockers from the first review (IAM scope, billing placeholder, README table) and the infrastructure is validated against production. However, the two new scripts (`deploy-prod-infra.sh`, `migrate-prod-db.sh`) introduce five new blocking issues that should be fixed before merge: the deploy script silently omits the required `billing_account` var from `terraform apply`; the migration script unconditionally drops a production table on every run with no guard; the public IP exposure during migration has no `authorized_networks` scope; the cleanup trap has a race window; and the production tfvars uses a quoted string boolean for `enable_gke`. Docs and Terraform changes are clean.

---

### Blocking Findings

**[correctness]** `scripts/deploy-prod-infra.sh:70`
`terraform apply -var-file="$TFVARS"` is called without `-var="billing_account=..."`. The committed `terraform.tfvars.production` now contains `billing_account = "REPLACE_ME"` (correctly reverted from the real ID). On a fresh-project apply this either fails with an invalid billing account ID or silently attempts to bind `"REPLACE_ME"` — breaking the primary use case of this script. The tfvars file header even shows the expected pattern: `-var="billing_account=$GCP_BILLING_ACCOUNT"`.
**Fix:** Add a pre-flight check that `$GCP_BILLING_ACCOUNT` is set and non-placeholder, then pass `-var="billing_account=$GCP_BILLING_ACCOUNT"` to both `terraform plan` and `terraform apply`.

---

**[correctness]** `scripts/migrate-prod-db.sh:172`
`DROP TABLE IF EXISTS user_data_source` runs unconditionally on every invocation against the hardcoded `lifting-logbook-prod` project, with no confirmation prompt and no check for whether migrations have already been applied. A re-run after the initial bootstrap (e.g., to apply a new Prisma migration) silently destroys the `user_data_source` table and all its data.
**Fix:** Gate the drop behind an explicit flag (e.g., `--bootstrap`) or detect whether `_prisma_migrations` already exists and skip the drop if it does. Add an interactive confirmation when running against a production project ID.

---

**[security]** `scripts/migrate-prod-db.sh:120-125`
`gcloud sql instances patch --assign-ip` enables a public IP on the production Cloud SQL instance with no `authorized_networks` restriction. During the migration window the instance accepts TCP connections from the entire internet; IAM authentication on the proxy is the only guard. This is avoidable: Cloud SQL Auth Proxy supports connecting over Private IP via the VPC connector, or the public IP exposure can be scoped to the operator's current address.
**Fix:** Either pass `--authorized-networks=$(curl -s ifconfig.me)/32` alongside `--assign-ip` to scope exposure to the operator's IP only, or document that `--private-ip` can be used when running from within the VPC (e.g., Cloud Shell).

---

**[reliability]** `scripts/migrate-prod-db.sh:136,161`
The `trap` that calls `remove_public_ip` is installed on line ~161, but `PUBLIC_IP_ENABLED=true` and `gcloud sql instances patch --assign-ip` run on lines ~120-125. If the script is interrupted (Ctrl-C, `set -e` triggering in the `until` polling loop, or a failed sleep) between enabling the public IP and the `trap` line, the public IP is left permanently attached to the production database. Additionally, `$PROXY_PID` is referenced in the trap but is only set after the proxy is started on line ~160; if the script exits before that assignment, `set -u` aborts the trap handler before `remove_public_ip` runs.
**Fix:** Install the trap (on `EXIT INT TERM`) immediately after `gcloud sql instances patch --assign-ip` succeeds. Guard the proxy kill: `[[ -n "${PROXY_PID:-}" ]] && kill "$PROXY_PID" 2>/dev/null || true; remove_public_ip`.

---

**[correctness]** `infra/terraform/terraform.tfvars.production:8`
`enable_gke = "false"` uses a quoted string. Terraform performs implicit string-to-bool coercion in current versions, so `"false"` → `false` works in practice, but this is fragile: it silently fails validation if the variable is declared with `type = bool` and `validation` blocks in some Terraform versions, and it communicates intent poorly to anyone reading the file.
**Fix:** Use a bare boolean: `enable_gke = false` (no quotes).

---

### Non-Blocking Findings

**[reliability]** `scripts/migrate-prod-db.sh:98`
On Windows the proxy binary is downloaded from the URL as `cloud-sql-proxy.x64.exe` but saved as `scripts/cloud-sql-proxy` (no `.exe` extension). Windows typically requires the `.exe` extension to execute a binary directly. If the proxy silently fails to start, the script proceeds to run migrations against a non-existent connection and the public IP remains enabled when the trap fires.
**Suggestion:** Detect Windows in the download path and save the file with the correct name and extension (e.g., `cloud-sql-proxy.exe`), and invoke it accordingly.

---

**[reliability]** `scripts/migrate-prod-db.sh:163`
`sleep 3` after starting the proxy assumes the proxy will be ready to accept connections within 3 seconds. On cold starts with fresh IAM credential propagation, this can be insufficient, causing the first `pg` client connection to fail.
**Suggestion:** Replace the fixed sleep with an active readiness poll — e.g., loop on `node -e "require('net').createConnection($PROXY_PORT,'127.0.0.1').on('connect',()=>process.exit(0)).on('error',()=>process.exit(1))"` with a 30-second timeout.

---

**[reliability]** `infra/terraform/main.tf:68`
`create_duration = "60s"` in `time_sleep.api_propagation` is a magic number. 60 seconds was observed to be sufficient in one run but GCP API propagation is eventually consistent and has been reported to exceed 90s on cold projects. More importantly, this sleep fires on every subsequent `terraform apply` even when APIs are already enabled.
**Suggestion:** Add a comment documenting why 60s was chosen. Consider a longer value (90s) for safety, or gate the sleep with a `count` that evaluates to 0 when the APIs are already active (this is non-trivial in Terraform, so the comment is the minimum).

---

**[maintainability]** `scripts/deploy-prod-infra.sh:22`, `scripts/migrate-prod-db.sh:32`
Both scripts hardcode `PROJECT_ID="lifting-logbook-prod"`, `REGION="us-central1"`, and `deploy-prod-infra.sh` hardcodes `liftinglogbook.com`/`www.liftinglogbook.com` and `${PROJECT_ID}-web`. These duplicate values already in `terraform.tfvars.production` and Terraform outputs. A region change or staging environment reuse will require editing the scripts separately from the tfvars.
**Suggestion:** Accept `--region` as an overridable flag (already done for `--project-id`), and derive the domain list and service name from `terraform output` rather than hardcoding them.

---

### Questions for Author

- The PR description states "lifecycle.ignore_changes = [template] already prevents Terraform from fighting CI/CD image updates" — but this block doesn't appear in the diff. Can you confirm it exists in the unchanged portion of `cloud-run.tf` for both `api` and `web` services? If it covers only `template` (not `template.containers.image` specifically), a `terraform apply` run by someone with `image_tag = "bootstrap"` in their tfvars after CI has deployed would revert the running image to the hello container.

---

brownm09:
**Self-answered (author question from review):** Confirmed `lifecycle { ignore_changes = [template] }` exists on both `google_cloud_run_v2_service.api` and `.web` in the unchanged portion of `cloud-run.tf`. This covers `template.containers.image`, so subsequent `terraform apply` runs will never revert CI-deployed images to the hello container. Not a blocking issue.

PR #290, opened 2026-05-22
[fix] Make staging environment optional in deploy workflow

### LINKED ISSUE #289: fix: deploy workflow fails when staging environment is not configured
In single-user mode (`enable_gke = false`, no staging project), the Deploy workflow fails immediately because the first job (`terraform-staging`) requires `GCP_STAGING_WORKLOAD_IDENTITY_PROVIDER` which doesn't exist. This blocks the entire pipeline — production is never reached.

## Fix

Add a `preflight` job that checks whether staging credentials are present and outputs `staging_enabled`. Guard all staging jobs with `if: needs.preflight.outputs.staging_enabled == 'true'`. Make `deploy-production` succeed when staging is skipped (not just when it passes). In production-only mode, authenticate `build-images` with production credentials and push to the production Artifact Registry.

Closes this issue.

brownm09:
## Summary

- Adds a `preflight` job that checks whether `GCP_STAGING_WORKLOAD_IDENTITY_PROVIDER` is set
- When staging is **not** configured (single-user / production-only mode):
  - `terraform-staging`, `deploy-staging`, `smoke-test` are skipped
  - `build-images` authenticates with production credentials and pushes to the production Artifact Registry
  - `deploy-production` proceeds directly once `build-images` succeeds
- When staging **is** configured the full pipeline runs unchanged

## Root cause

The original workflow required staging GCP credentials as the very first step. With no staging project, the pipeline failed immediately and production was never reached — the Cloud Run placeholder image was never replaced.

## Test plan

- [ ] Docs-only CI checks pass on this PR
- [ ] After merge: confirm the Deploy workflow runs, staging jobs are skipped (shown as skipped in Actions UI), build-images succeeds using prod credentials, and deploy-production completes — liftinglogbook.com serves the app

Closes #289

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -159,6 +159,20 @@ jobs:
           PREFLIGHT_REPO="${{ needs.preflight.outputs.ar_repo }}"
           echo "repo=${STAGING_REPO:-$PREFLIGHT_REPO}" >> "$GITHUB_OUTPUT"
 
+      - name: Resolve API URL for web image build
+        id: api-url
+        run: |
+          if [[ "${{ needs.preflight.outputs.staging_enabled }}" == "true" ]]; then
+            echo "url=${{ needs.terraform-staging.outputs.cloud_run_api_url }}" >> "$GITHUB_OUTPUT"
+          else
+            # Production-only mode: gcloud auth is already done above.
+            URL=$(gcloud run services describe lifting-logbook-prod-api \
+              --region="${{ env.REGION }}" \
+              --project="${{ secrets.GCP_PROD_PROJECT_ID }}" \
+              --format="value(status.url)")
+            echo "url=${URL}" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Configure Docker for Artifact Registry
         run: gcloud auth configure-docker ${{ env.REGISTRY }} --quiet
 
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [fix] Make staging environment optional in deploy workflow

### Summary

Ready to merge. The conditional job logic is correct and well-structured: `preflight` correctly detects staging config, the `build-images` `always()` condition handles the skipped-`terraform-staging` case, and `deploy-production` fires directly in production-only mode. One question about `NEXT_PUBLIC_API_URL` should be confirmed before or just after merge — it won't block the placeholder-page fix but could affect app functionality once deployed.

---

### Questions for Author

- **`NEXT_PUBLIC_API_URL` in production-only mode** (`build-images`, web image `build-args`): `NEXT_PUBLIC_API_URL` is set from `needs.terraform-staging.outputs.cloud_run_api_url`, which is an empty string when `terraform-staging` is skipped. If the Next.js frontend uses this variable to construct API calls (and it's baked in at build time, as all `NEXT_PUBLIC_*` vars are), the web image will be built with no API URL in production-only mode. `deploy-production` does capture `cloud_run_api_url` from the production Terraform outputs, but that's too late — the image is already built. Is `NEXT_PUBLIC_API_URL` actually used? Does the frontend fall back to a relative URL or derive it from the window origin?

---

### Style Notes

Minor: the `echo "ar_repo=" >> "$GITHUB_OUTPUT"` line in the `preflight` job explicitly writes an empty value with a trailing comment `# filled by terraform-staging`. Since the bash `${STAGING_REPO:-$PREFLIGHT_REPO}` fallback in `build-images` treats an empty string and an unset variable identically, the explicit empty write is harmless but slightly noisy. Could just omit it.

---

PR #321, opened 2026-05-23
[feat] Implement Clerk authentication in web app

### LINKED ISSUE #319: [feat] Implement Clerk authentication in web app
## Problem

The web app is live but has two critical issues:

1. **No login prompt** — `@clerk/nextjs` is not installed. The Cloud Run YAML provides `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` but there is no `<ClerkProvider>`, no `middleware.ts`, and no sign-in flow.

2. **Navigation fails** — Clicking any nav link (e.g. "Current Cycle") causes the server-side render to fail. The API's `AuthGuard` requires a Bearer token; without one it returns 401. `getActiveProgram()` catches the 401 and falls back to `'5-3-1'`, but `fetchCycleDashboard` calls `apiFetchNullable` which throws on non-404 errors — the uncaught 401 error prevents the page from rendering.

## Root Cause

Clerk is fully configured in the API (`CLERK_SECRET_KEY`, `ClerkAuthProvider`, `AuthGuard`) but the web app has never had `@clerk/nextjs` integrated.

## Acceptance Criteria

- [ ] Unauthenticated users are redirected to Clerk's sign-in page
- [ ] After sign-in, navigation to `/cycle`, `/settings/*`, `/history`, and `/programs` works correctly
- [ ] Server-side API calls include the Clerk session JWT as the Authorization header
- [ ] Client-side API calls include the Clerk session JWT as the Authorization header
- [ ] Local dev workflow continues to work using `DEV_AUTH_TOKEN`

brownm09:
## Summary

- Installs `@clerk/nextjs@6.x` in the web app
- Adds `middleware.ts` with `clerkMiddleware()` — unauthenticated users are redirected to Clerk's hosted sign-in page
- Wraps the root layout with `<ClerkProvider>`
- Replaces `google-auth-library` (GCP service identity tokens) in `lib/api.ts` with `auth().getToken()` from `@clerk/nextjs/server` so server-side API calls carry the authenticated user's Clerk JWT
- Adds `setAuthTokenGetter()` + `ClerkApiInitializer` client component so browser-side `clientFetch` calls also include the Clerk JWT in Cloud Run environments
- Local dev workflow is unchanged: `DEV_AUTH_TOKEN` is still used when `API_URL` is `http://` (non-HTTPS)

## Root cause fixed

The API's `AuthGuard` requires a `Bearer` token on every request. Without Clerk integrated in the web app, server-side rendering called the API with no token (401 → uncaught throw in `apiFetchNullable`) and client-side calls had no auth at all. This is why navigation away from the home page failed silently.

## Acceptance criteria

- [x] Unauthenticated users are redirected to Clerk's sign-in page
- [x] After sign-in, server-side API calls include the Clerk session JWT
- [x] Client-side API calls include the Clerk session JWT via `ClerkApiInitializer`
- [x] Local dev workflow unchanged (DEV_AUTH_TOKEN path)
- [x] `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` is already wired in `web-service.yaml`

## Testing

Pre-existing TypeScript errors in the web app (missing `UserWorkoutSchedule`, `workoutSchedule` etc.) are unrelated to this change. No new type errors were introduced — confirmed with `tsc --noEmit` filtered to the changed files.

Full test suite (`npm test`) has a pre-existing `ts-jest` configuration issue affecting all workspaces; unrelated to this PR.

Manual verification required:
- Deploy to staging with a valid `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` secret
- Visit the app URL unauthenticated → should redirect to Clerk sign-in
- Sign in → should navigate to `/cycle` successfully

Closes #319

### DIFF
--- a/apps/web/app/layout.tsx
+++ b/apps/web/app/layout.tsx
@@ -1,4 +1,6 @@
+import { ClerkProvider } from '@clerk/nextjs';
 import type { Metadata } from "next";
+import { ClerkApiInitializer } from '@/components/ClerkApiInitializer';
 import "./globals.css";
 
 export const metadata: Metadata = {
@@ -18,13 +20,16 @@ export default function RootLayout({
   children: React.ReactNode;
 }>) {
   return (
-    <html lang="en" suppressHydrationWarning>
-      <head>
-        <script dangerouslySetInnerHTML={{ __html: themeInitScript }} />
-      </head>
-      <body>
-        {children}
-      </body>
-    </html>
+    <ClerkProvider>
+      <html lang="en" suppressHydrationWarning>
+        <head>
+          <script dangerouslySetInnerHTML={{ __html: themeInitScript }} />
+        </head>
+        <body>
+          <ClerkApiInitializer />
+          {children}
+        </body>
+      </html>
+    </ClerkProvider>
   );
 }
--- a/apps/web/components/ClerkApiInitializer.tsx
+++ b/apps/web/components/ClerkApiInitializer.tsx
@@ -1,0 +1,15 @@
+'use client';
+
+import { useAuth } from '@clerk/nextjs';
+import { useEffect } from 'react';
+import { setAuthTokenGetter } from '@/lib/client-api';
+
+export function ClerkApiInitializer() {
+  const { getToken } = useAuth();
+
+  useEffect(() => {
+    setAuthTokenGetter(getToken);
+  }, [getToken]);
+
+  return null;
+}
--- a/apps/web/components/ClerkApiInitializer.tsx
+++ b/apps/web/components/ClerkApiInitializer.tsx
@@ -4,6 +4,15 @@ import { useAuth } from '@clerk/nextjs';
 import { useEffect } from 'react';
 import { setAuthTokenGetter } from '@/lib/client-api';
 
+/**
+ * Wires the Clerk session token into client-side API calls via setAuthTokenGetter.
+ *
+ * Ordering note: React fires useEffect bottom-up (children before parents). Any page
+ * component that calls clientFetch in its own useEffect on mount will execute before
+ * this initializer runs — meaning _getToken is still null for that request (→ 401).
+ * For this reason, mount-time client fetches in child components should be avoided;
+ * prefer interaction-triggered calls or server-side fetches via lib/api.ts instead.
+ */
 export function ClerkApiInitializer() {
   const { getToken } = useAuth();
 
--- a/apps/web/lib/api.ts
+++ b/apps/web/lib/api.ts
@@ -36,7 +36,11 @@ async function getAuthHeaders(): Promise<Record<string, string>> {
   try {
     const { getToken } = await auth();
     const token = await getToken();
-    return token ? { Authorization: `Bearer ${token}` } : {};
+    if (!token) {
+      console.warn('[getAuthHeaders] No Clerk session token in Cloud Run — request will be unauthenticated');
+      return {};
+    }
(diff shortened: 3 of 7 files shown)


brownm09:
## Review: [feat] Implement Clerk authentication in web app

### Summary

Ready to merge with minor changes. The implementation is correct and well-reasoned: `middleware.ts` protects all non-public routes via `auth.protect()`, `<ClerkProvider>` wraps the root layout, server-side `api.ts` now uses `auth().getToken()` (removing the mismatched GCP identity token flow), and `ClerkApiInitializer` wires the Clerk JWT into client-side calls via a module-level setter — cleanly avoiding 7 call-site changes. No blocking findings. Two non-blocking findings (auth header duplication, effect ordering note) and one question worth confirming before deploy.

---

### Non-Blocking Findings

**[maintainability]** `apps/web/lib/client-api.ts:190–207`
The auth header assembly block (`if (devToken) { ... } else if (_getToken) { ... }`) is duplicated verbatim in both `clientFetch` and `importLiftRecords`. `importLiftRecords` uses multipart form data and can't go through `clientFetch` (different response handling), so the duplication is structurally necessary — but it means any future change to auth logic (token audience, header name, fallback behavior) needs to be applied in two places.
**Suggestion:** Extract to a private `async function getClientAuthHeaders(): Promise<Record<string, string>>` and call it from both sites. Three-line change; eliminates the drift risk.

---

**[reliability]** `apps/web/components/ClerkApiInitializer.tsx:10–12`
React fires `useEffect` callbacks bottom-up: children before parents. `ClerkApiInitializer` sits in the root layout alongside `{children}`, so any page component that fires a client API call in its own `useEffect` on mount will execute *before* `setAuthTokenGetter` is called. This means `_getToken` is still null for those calls, and the request goes out unauthenticated (→ 401 from the API).

For this codebase this is low risk — most write operations are triggered by user interactions, not mount effects — but the assumption isn't enforced and will be silently broken if a future component adds a mount-time fetch.
**Suggestion:** Add a comment to `ClerkApiInitializer` noting the effect-ordering limitation ("initialized after children mount — mount-time `clientFetch` calls in child components will run without auth"), so future authors know the constraint before adding eager fetches.

---

### Questions for Author

**Question:** In `lib/api.ts`, when `getToken()` returns `null` (no active session) in a Cloud Run context, `getAuthHeaders` silently returns `{}`, resulting in an unauthenticated request and a 401 from the API. Is this silent path intentional, or should it log a warning?
**Context:** [`apps/web/lib/api.ts:36–42`](apps/web/lib/api.ts) — the `try/catch` only catches errors thrown by `auth()` itself (e.g., no request context in ISR or static generation). A successfully-resolved `auth()` that simply has no session (signed-out mid-session, expired Clerk token) falls through to `token ? ... : {}` silently — no log entry.
**Tradeoffs:** (a) Keep silent — the middleware already prevents unauthenticated users from reaching server components in normal SSR, so a null token in Cloud Run is only reachable in edge cases (ISR revalidation, background cron); logging there is noise. (b) Add a `console.warn('[getAuthHeaders] No Clerk session token in Cloud Run')` before the empty return — helps diagnose auth failures in staging without leaking user info.

---

### Style Notes

The `middleware.ts` matcher pattern is the standard Clerk-recommended one and is correct. Minor: the `config` export could carry a brief comment (`// Standard Clerk matcher — excludes static assets`) so authors who encounter it unfamiliar with Clerk's conventions understand why the regex looks complex.

---

*Reviewed by claude-opus-4-5 via Claude Code*

PR #327, opened 2026-05-23
[fix] Fix pre-commit hook TURBO_BINARY_PATH propagation on Windows worktrees

### LINKED ISSUE #326: [fix] pre-commit hook: TURBO_BINARY_PATH export doesn't reach cmd.exe on Windows
## Problem

The `.husky/pre-commit` hook sets `export TURBO_BINARY_PATH` to point `npx turbo` at the main repo's native binary (avoiding EUNKNOWN from paths containing `.claude`). But the exported variable does NOT propagate to the node process when running from a git worktree on Windows.

**Root cause:** MSYS2/Git Bash's `export` updates the POSIX environment block. When the hook runs `npx turbo`, sh invokes `npx.cmd` which goes through `cmd.exe`. `cmd.exe` inherits from the *Windows NT* process environment block — not the MSYS2 POSIX block — so `TURBO_BINARY_PATH` is invisible to node's `process.env`.

**Symptom:** `git commit` fails with `spawnSync ... EUNKNOWN` pointing at the worktree's turbo.exe, even though the hook correctly computes the right binary path.

**Workaround (current):** Pre-set `TURBO_BINARY_PATH` in the calling shell before running `git commit`.

## Fix

Replace `npx turbo` with `TURBO_BINARY_PATH=... node node_modules/turbo/bin/turbo` when on a Windows worktree. This invokes `node.exe` directly from MSYS2 sh (no cmd.exe hop), and MSYS2 bridges the inline env assignment into the Windows NT env block for that subprocess. The turbo JS wrapper then reads `process.env.TURBO_BINARY_PATH` correctly.

Falls back to `npx turbo` on macOS/Linux (where the `.exe` file won't exist).

brownm09:
## Summary

Fixes `git commit` failing with `EUNKNOWN` from git worktrees whose path contains `.claude`.

**Root cause (three-layer):**

1. The worktree-specific `config.worktree` sets `hooksPath` to the **main repo's** `.husky/`, so all worktree commits run `C:/Users/brown/Git/lifting-logbook/.husky/pre-commit` — not the worktree's copy. This is the file changed in this PR.

2. That hook called `npx turbo`, which goes through `npx.cmd` → `cmd.exe`. `cmd.exe` inherits the Windows NT env block, not MSYS2's POSIX env block. So `export TURBO_BINARY_PATH` in the old worktree hook was invisible.

3. The previous workaround (`export TURBO_BINARY_PATH` + `npx turbo`) only ran from worktree-specific `.husky/` files that git was never actually executing.

**Fix:** Replace `npx turbo` with `node -e "process.env.TURBO_BINARY_PATH=process.argv[1]; require('./node_modules/turbo/bin/turbo')"` when the Windows-x64 binary is found in the main repo root. Node.js is a Windows native process — its `process.env` mutations propagate to child processes through the Windows NT env block. Setting the path inside Node before requiring the turbo wrapper means the wrapper's `process.env.TURBO_BINARY_PATH` check sees the correct binary.

Falls back to `npx turbo` on macOS/Linux (no `.exe` found → condition false).

## After merging

After `git pull` in the main repo, `C:/Users/brown/Git/lifting-logbook/.husky/pre-commit` will have the fix and all worktrees will commit without needing `TURBO_BINARY_PATH` pre-set in the shell.

## Test plan

- [x] `sh .husky/pre-commit` passes without `TURBO_BINARY_PATH` in env
- [x] `git commit` succeeds in a worktree with no `TURBO_BINARY_PATH` pre-set in shell (this commit was made using the fixed hook)
- [x] All 4 lint targets pass (cache hits confirmed)

Closes #326

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,21 +2,31 @@
 
 # When running in a git worktree whose path contains directories (e.g. .claude)
 # that prevent Windows from spawning executables, turbo.exe fails with EUNKNOWN.
-# Resolve the binary from the main repo root (git common dir) to sidestep this.
+# Fix: run node directly (no cmd.exe hop) and set TURBO_BINARY_PATH inside the node
+# process. Node is a Windows native process so process.env mutations propagate to its
+# children via the Windows NT env block — unlike MSYS2 sh `export`, which only updates
+# the POSIX env and is invisible to cmd.exe subprocesses in git's hook context.
 _COMMON=$(git rev-parse --git-common-dir 2>/dev/null)
 _MAIN_ROOT=${_COMMON%/.git}
-if [ -n "$_MAIN_ROOT" ] && [ -f "$_MAIN_ROOT/node_modules/@turbo/windows-64/bin/turbo.exe" ]; then
-  export TURBO_BINARY_PATH="$_MAIN_ROOT/node_modules/@turbo/windows-64/bin/turbo.exe"
-fi
-unset _COMMON _MAIN_ROOT
+_WIN_TURBO="$_MAIN_ROOT/node_modules/@turbo/windows-64/bin/turbo.exe"
+unset _COMMON
+
+_TURBO_INLINE="process.env.TURBO_BINARY_PATH=process.argv[1]; require('./node_modules/turbo/bin/turbo')"
 
 # Excluded until scaffolded/clean:
 #   @lifting-logbook/web  — no src/ yet (issue #9)
 #   @lifting-logbook/core — 51 pre-existing GAS-migration lint errors (issue #51)
-npx turbo run lint \
-  --filter=!@lifting-logbook/web \
-  --filter=!@lifting-logbook/core
+if [ -n "$_MAIN_ROOT" ] && [ -f "$_WIN_TURBO" ]; then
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [fix] Fix pre-commit hook TURBO_BINARY_PATH propagation on Windows worktrees

### Summary

Merge as-is. The fix is technically correct and the PR body explains all three layers of the root cause clearly. The `node -e` approach is the right mechanism: Node is a Windows native process, so `process.env.TURBO_BINARY_PATH = process.argv[1]` before `require(turbo)` puts the variable in the Windows NT env block where all child processes can see it — unlike `export` in MSYS2 sh, which only updates the POSIX env. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `.husky/pre-commit:14`
`_TURBO_INLINE` holds a multi-statement Node snippet expanded into `node -e "$_TURBO_INLINE"`. If the snippet ever needs a literal double-quote, the interpolation would break. Currently safe because the snippet uses no double-quotes, but the pattern is fragile.
**Suggestion:** Document in a comment that `$_TURBO_INLINE` must remain free of double-quotes, or switch to a heredoc approach when the snippet grows. No change required now.

---

Issue #324, opened 2026-05-23
[ops] Manual staging deploy: terraform apply + migrate + seed post-#323


brownm09:
Apply the CIDR fix from #323 to the staging GCP environment, run Prisma migrations, seed the DB, and verify login→/cycle.

## Acceptance criteria
- [ ] `terraform apply` completes on the staging workspace with no errors
- [ ] Prisma migrations applied to staging Cloud SQL instance
- [ ] `infra/migrations/001_create_user_data_source.sql` applied
- [ ] Seed script populates ~268 rows across 7 tables
- [ ] Login → /cycle navigation flow verified in browser
- [ ] `migrate-staging-db.sh` script committed for future use

PR #323, opened 2026-05-23
[fix] Fix staging subnet CIDR overlap and add Prisma seed script

brownm09:
## Summary

- Fixes the staging services secondary IP range (`10.53.0.0/20` → `10.56.0.0/20`) which was contained within the pod range `10.52.0.0/14`, causing `terraform apply` to fail with a GCP validation error
- Adds `apps/api/prisma/seed.ts` with ~3 months of synthetic 5/3/1 lift data (432 lift records, 6 lifts, 12 cycles) for one seed user — all writes are idempotent upserts
- Wires `"prisma": { "seed": "..." }` in `apps/api/package.json` so `npx prisma db seed` works
- Adds Step 3.5 (seed staging data) to `docs/deploy.md` between Terraform bootstrap and Clerk setup
- Fixes `.husky/pre-commit` to resolve `turbo.exe` from the main repo root so commits from git worktrees under `.claude/` paths work on Windows

## Acceptance criteria

- [x] Seed script is idempotent (all upserts on unique constraints)
- [x] Seed data uses a synthetic Clerk user ID (`seed_user_clerkid_staging_001`) — no real PII
- [x] `docs/deploy.md` documents how to run seeding after migrations
- [x] Lint passes (all 4 packages clean)

## Test plan

- `npm test -w @lifting-logbook/api` — existing tests pass
- To verify seed script locally: `DATABASE_URL=<staging-db-url> npx --prefix apps/api prisma db seed`
- Re-run seed to verify idempotency (no errors, row counts unchanged)

Closes #322

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #330, opened 2026-05-23
[fix] Fix migrate-staging-db.sh: state bucket, secret names, and proxy SSL

### LINKED ISSUE #328: [fix] migrate-staging-db.sh: wrong state bucket and secret name pattern
## Problem

`scripts/migrate-staging-db.sh` has two bugs that prevent it from running:

1. **Wrong state bucket:** `terraform init -backend-config="bucket=${PROJECT_ID}-tfstate"` expands to `lifting-logbook-staging-tfstate`, but the actual staging state lives in `lifting-logbook-tfstate` with prefix `terraform/state`.

2. **Wrong secret name:** `gcloud secrets versions access latest --secret="${PROJECT_ID}-database-url"` expands to `lifting-logbook-staging-database-url`, but secrets are named with the `-stg-` suffix pattern from terraform locals (e.g. `lifting-logbook-stg-database-url`).

## Fix

- Change `terraform init` to use `bucket=lifting-logbook-tfstate` with `prefix=terraform/state`
- Derive secret names from the `name_prefix` (`${app_name}-stg`) pattern: `lifting-logbook-stg-database-url`

Discovered during the post-#323 staging deploy (issue #324 execution).

brownm09:
## Summary

Fixes three bugs in \`scripts/migrate-staging-db.sh\` discovered during the post-#323 staging deploy execution. All three caused the script to fail silently or partially.

**1. Wrong state bucket**
The script used \`bucket=${PROJECT_ID}-tfstate\` → \`lifting-logbook-staging-tfstate\`, but staging state lives in the shared \`lifting-logbook-tfstate\` bucket at \`prefix=terraform/state\` (as documented in \`docs/deploy.md\` line 139). This caused \`terraform output -raw database_instance_name\` to fail with "Output not found."

**2. Wrong secret name pattern**
The script used \`--secret="${PROJECT_ID}-database-url"\` → \`lifting-logbook-staging-database-url\`, but terraform's \`name_prefix\` uses the \`-stg-\` env_suffix (from \`locals { env_suffix = ... ? "prod" : "stg" }\`), making the actual secret \`lifting-logbook-stg-database-url\`.

**3. Prisma SSL failure via Cloud SQL Auth Proxy**
Prisma failed with P1001 ("Can't reach database server") even though the proxy was running and the \`pg\` node client connected successfully. Root cause: the DATABASE_URL from Secret Manager includes SSL parameters tuned for direct Cloud SQL connections. When routed through the Auth Proxy, these must be stripped and replaced with \`sslmode=disable\` — the proxy handles TLS; Prisma must not negotiate SSL independently.

## Documentation updated

- \`scripts/README.md\`: description updated to list all staging-specific differences from \`migrate-prod-db.sh\`
- \`docs/deploy.md\` Step 3.5: seed command updated to use port 5434, cloud-sql-proxy v2 syntax, and \`sslmode=disable\`

## Test plan

- [x] All three bugs reproduced and root-caused during staging deploy execution (2026-05-23)
- [ ] \`./scripts/migrate-staging-db.sh\` runs to completion against the staging Cloud SQL instance (part of issue #324 deploy verification)

Closes #328

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -178,7 +178,7 @@ the environment realistic data for testing cycle generation, reporting, and API 
 any real user PII.
 
 **When to run:** after `prisma migrate deploy` completes on the staging database (the deploy
-workflow runs migrations automatically; you can also run them manually via Cloud SQL Auth Proxy).
+workflow runs migrations automatically; for manual runs use `./scripts/migrate-staging-db.sh`).
 
 **What it creates:**
 
@@ -200,8 +200,10 @@ schema validation, API smoke tests, and query testing.
 
 ```bash
 # Via Cloud SQL Auth Proxy (connect to staging DB locally)
-cloud_sql_proxy lifting-logbook-staging:us-central1:lifting-logbook-stg-db-<suffix> &
-DATABASE_URL="postgresql://lifting-logbook-app:<password>@127.0.0.1:5432/lifting_logbook" \
+# Use port 5434 (avoids conflict with prod proxy on 5433)
+# sslmode=disable required: proxy handles TLS; Prisma must not negotiate SSL independently
+cloud-sql-proxy "lifting-logbook-staging:us-central1:lifting-logbook-stg-db-<suffix>?port=5434" &
+DATABASE_URL="postgresql://lifting-logbook-app:<password>@127.0.0.1:5434/lifting_logbook?sslmode=disable" \
   npx --prefix apps/api prisma db seed
 ```
 
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -15,7 +15,7 @@ Repository automation scripts. Grouped by lifecycle.
 | [`bootstrap-gcp-prod.sh`](bootstrap-gcp-prod.sh) | One-time GCP bootstrap for a single-user production deploy: creates the project, links billing, enables the APIs Terraform needs, and provisions the Terraform state bucket. Idempotent. See [`docs/deploy-single-user.md`](../docs/deploy-single-user.md). |
 | [`deploy-prod-infra.sh`](deploy-prod-infra.sh) | Automates `terraform init` → workspace select → `apply` for the production environment. Maps custom domains and prints GitHub Actions secrets. Use `--plan-only` to preview. Run after `bootstrap-gcp-prod.sh`. |
 | [`migrate-prod-db.sh`](migrate-prod-db.sh) | Applies all database migrations to the production Cloud SQL instance. Temporarily enables a public IP, runs Prisma migrations and the `user_data_source` infra migration via the Cloud SQL Auth Proxy, then removes the public IP. Downloads the proxy automatically. |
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: [fix] Fix migrate-staging-db.sh: state bucket, secret names, and proxy SSL

### Summary

Ready to merge with one non-blocking note. All three stated bug fixes are correct and well-reasoned: the terraform backend config now points at the right shared state bucket with the right prefix, the secret name is hardcoded to match terraform's `-stg-` env_suffix convention, and `sslmode=disable` is correctly forced on the proxy-rewritten URL. Documentation (`scripts/README.md` and `docs/deploy.md`) is updated in sync. The one reliability note is an edge case in the sslmode-stripping regex that does not affect the current Cloud SQL secret format but would silently produce a malformed URL in certain multi-param configurations.

---

### Non-Blocking Findings

**[reliability]** `scripts/migrate-staging-db.sh:201`

The `node -e` sslmode-stripping regex replaces the match including the leading `?` when `sslmode` is the first query parameter. If other parameters follow (e.g., `?sslmode=require&sslrootcert=…`), the `?` is consumed and the remainder is left as `&sslrootcert=…`, producing a malformed URL: `postgresql://…/db&sslrootcert=…?sslmode=disable`. The subsequent `u.includes('?')` check then appends another `?` in the wrong position. Tested locally — the `"sslmode first+cert"` case fails. For the current secret format (`?sslmode=require` only) this never triggers, but Cloud SQL SSL-cert-enabled configurations would hit it.

**Suggestion:** Replace the `?` capture with a lookahead or restructure to handle the `?sslmode=first` case:

```js
const u = url
  .replace(/\?sslmode=[^&]*&/, '?')   // sslmode=... is first param; promote next param to ?
  .replace(/[?&]sslmode=[^&]*/g, ''); // sslmode=... is only or non-first param
```

---

**[reliability]** `scripts/migrate-staging-db.sh:164`

`api4.ipify.org` is correctly made primary (IPv4-only), but the fallback to `ifconfig.me` can still return an IPv6 address on dual-stack networks. If `api4.ipify.org` is unreachable and `ifconfig.me` returns IPv6, the `gcloud sql instances patch --authorized-networks` call will fail with a Cloud SQL API error (IPv6 CIDR not accepted). The script exits non-zero via `set -e` and the trap removes the public IP — no resource leak — but the error message does not clearly indicate the root cause.

**Suggestion:** Either drop the `ifconfig.me` fallback entirely (fail explicitly if `api4.ipify.org` is unavailable), or validate `$OPERATOR_IP` as IPv4 before using it:

```bash
if [[ -n "$OPERATOR_IP" ]] && [[ "$OPERATOR_IP" =~ ^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```

---

### Style Notes

The `node -e` one-liner on line 201 packs a regex replace, a ternary, and a `console.log` into ~90 characters of argument string. It works, but a future author who needs to modify the sslmode logic would need to parse it without IDE support. Consider adding a comment above the line explaining the two-step intent (strip existing sslmode → re-add as disable).

---

PR #334, opened 2026-05-23
[fix] migrate-staging-db.sh: IPv4 validation, 30s proxy wait, sslmode regex

### LINKED ISSUE #333: [fix] migrate-staging-db.sh: sslmode regex edge case, IPv4 validation, and proxy wait time
Three non-blocking reliability findings from PR #330 review, plus one new issue found during execution:

**1. sslmode regex edge case (line ~201)**
The node -e regex removes `[?&]sslmode=[^&]*`, which consumes the leading `?` when sslmode is the first query param. If other params follow (e.g. `?sslmode=require&sslrootcert=...`), the remaining params lose their `?` anchor and produce a malformed URL: `db&sslrootcert=...?sslmode=disable`.

**2. ifconfig.me IPv6 fallback (line ~164)**
If `api4.ipify.org` fails and `ifconfig.me` returns an IPv6 address, the `gcloud sql instances patch --authorized-networks` call fails with a non-obvious error. Should validate IPv4 format before using the address.

**3. Style: dense node -e one-liner (line ~201)**
The sslmode-stripping one-liner packs regex, ternary, and console.log into one line of argument string. Add a comment explaining the two-step intent.

**4. Proxy tunnel wait time too short (new)**
After enabling public IP, the script waits 15s for RUNNABLE state then starts the proxy. But the proxy's connection to Cloud SQL on port 3307 timed out during a migration run — port 3307 was not yet reachable even though the instance was RUNNABLE. Need a longer wait or retry mechanism after enabling the public IP.

brownm09:
## Summary

Addresses three non-blocking findings from PR #330 review and one new issue discovered during migration execution.

**1. IPv4 validation (finding: #330 review reliability)**
After detecting `OPERATOR_IP`, validate it matches `^[0-9]+.[0-9]+.[0-9]+.[0-9]+$` before using as an authorized-network. If IPv6 is detected (e.g. from the `ifconfig.me` fallback), skip the restriction with a clear warning rather than passing an IPv6 CIDR that Cloud SQL rejects with an unintuitive gcloud error.

**2. Extended post-RUNNABLE wait: 15s → 30s (new, from execution)**
After enabling public IP and waiting for RUNNABLE state, port 3307 (used by the Cloud SQL Auth Proxy) was not yet reachable during a migration run, causing a 22-second timeout. The instance enters RUNNABLE state before port 3307 is fully accessible externally.

**3. sslmode regex fix (finding: #330 review reliability)**
The original single-pass regex `[?&]sslmode=[^&]*` consumed the leading `?` when sslmode was the first query param, leaving subsequent params without a `?` anchor and producing a malformed URL (e.g. `db&sslrootcert=x?sslmode=disable`). Fixed with two-pass approach:
- Pass 1: `?sslmode=X&` → `?` (promote next param when sslmode leads)
- Pass 2: `[?&]sslmode=X` → `` (strip any remaining sslmode)
Verified against six URL patterns.

**4. Clarifying comment (finding: #330 review style)**
Added four-line comment above the sslmode node-e one-liner explaining the two-step intent.

## Test plan

- [x] sslmode regex verified against 6 URL patterns (all pass)
- [ ] `./scripts/migrate-staging-db.sh` runs to completion against staging Cloud SQL

Closes #333

### DIFF
--- a/scripts/migrate-staging-db.sh
+++ b/scripts/migrate-staging-db.sh
@@ -159,6 +159,11 @@ echo "    Instance: $INSTANCE_CONNECTION_NAME"
 
 echo "==> Enabling temporary public IP on Cloud SQL instance ..."
 OPERATOR_IP=$(curl -fsSL api4.ipify.org 2>/dev/null || curl -fsSL ifconfig.me 2>/dev/null || echo "")
+# Validate IPv4 -- Cloud SQL authorized-networks rejects IPv6; the gcloud error is not obvious
+if [[ -n "$OPERATOR_IP" ]] && ! [[ "$OPERATOR_IP" =~ ^[0-9]+[.][0-9]+[.][0-9]+[.][0-9]+$ ]]; then
+  echo "WARNING: Detected non-IPv4 address ($OPERATOR_IP) -- skipping authorized_networks restriction." >&2
+  OPERATOR_IP=""
+fi
 if [[ -n "$OPERATOR_IP" ]]; then
   echo "    Scoping authorized_networks to operator IP: $OPERATOR_IP/32"
   gcloud sql instances patch "$INSTANCE_NAME" \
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [fix] migrate-staging-db.sh: IPv4 validation, 30s proxy wait, sslmode regex

### Summary

Ready to merge — no findings. All four changes are targeted, correct, and well-documented. The IPv4 guard uses `[.]` for literal dot matching and correctly distinguishes IPv4 from IPv6 for the purpose of authorized-network scoping. The 30s post-RUNNABLE wait addresses a real observed timeout (port 3307 not yet routable after 15s). The two-pass sslmode regex fix has been verified against six URL patterns and handles the previously failing `?sslmode=first&other=param` case correctly.

---

PR #336, opened 2026-05-23
[fix] Move ts-node compilerOptions to config block to fix Windows seed quoting

### LINKED ISSUE #335: [fix] Prisma seed fails on Windows due to ts-node --compiler-options quoting
## Problem

`npx prisma db seed` fails on Windows because the seed command in `apps/api/package.json` passes `--compiler-options '{"module":"CommonJS"}'` to ts-node. On Windows, shell single-quote stripping does not occur — ts-node receives the literal single-quote characters, producing invalid JSON:

```
SyntaxError: Unexpected token ''', "'{"module""... is not valid JSON
```

## Root cause

The `prisma.seed` field in `apps/api/package.json`:
```json
"seed": "ts-node --compiler-options '{\module\:\CommonJS\}' prisma/seed.ts"
```

On Windows, npm runs this command via `cmd.exe`, which does not strip single quotes — they are passed verbatim to ts-node.

## Fix

Move compiler options to a `ts-node` config block in `package.json`, which ts-node reads automatically without any CLI flag:

```json
"ts-node": {
  "compilerOptions": {
    "module": "CommonJS"
  }
},
"prisma": {
  "seed": "ts-node prisma/seed.ts"
}
```

## Acceptance criteria

- [ ] `npx prisma db seed` executes without error on Windows
- [ ] Seed command is cross-platform (no OS-specific quoting)
- [ ] ts-node compiler options are applied correctly via config block

Closes this issue.

brownm09:
## Summary

Fixes `npx prisma db seed` failing on Windows due to shell quoting of the `--compiler-options` flag.

## Root cause

On Windows, `cmd.exe` does not strip single quotes — ts-node received literal `'` characters as part of the JSON option value, producing:
```
SyntaxError: Unexpected token ''', "'{"module""... is not valid JSON
```

## Fix

Moved the compiler options from a `--compiler-options` CLI flag to a `"ts-node"` config block in `package.json`. ts-node v10+ reads this block automatically — no flag needed, no shell quoting involved, fully cross-platform.

## Changes

- `apps/api/package.json`: Added `"ts-node": { "compilerOptions": { "module": "CommonJS" } }` block; simplified `prisma.seed` to `"ts-node prisma/seed.ts"`

## Testing

- Linting passes (confirmed by pre-commit hook)
- Seed command verified on Windows (this was the active blocker for staging seeding)

Closes #335

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -37,8 +37,13 @@
     "reflect-metadata": "^0.2.0",
     "rxjs": "^7.8.0"
   },
+  "ts-node": {
+    "compilerOptions": {
+      "module": "CommonJS"
+    }
+  },
   "prisma": {
-    "seed": "ts-node --compiler-options '{\"module\":\"CommonJS\"}' prisma/seed.ts"
+    "seed": "ts-node prisma/seed.ts"
   },
   "devDependencies": {
     "@nestjs/cli": "^10.0.0",


brownm09:
## Review: [fix] Move ts-node compilerOptions to config block to fix Windows seed quoting

### Summary

Ready to merge. The fix correctly eliminates the Windows shell-quoting issue by moving `--compiler-options` out of the CLI invocation entirely. The diff is minimal (6 lines added, 1 removed), targeted, and correct. One non-blocking observation worth noting: the new `"ts-node"` config block is technically redundant given `tsconfig.base.json` already sets `"module": "commonjs"` — the simplified seed command alone would have fixed the issue — but the block makes the intent explicit and costs nothing.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/package.json:38–42`
The `"ts-node": { "compilerOptions": { "module": "CommonJS" } }` block duplicates what `tsconfig.base.json` already provides via `"module": "commonjs"`. The original `--compiler-options` flag was likely added as a precaution, but it was always redundant from a TypeScript perspective; it only caused harm on Windows.
**Suggestion:** The seed command works correctly as-is. If you want to simplify further, the config block can be omitted — `"seed": "ts-node prisma/seed.ts"` alone is sufficient. Keep it if you prefer the explicitness.

---

### Style Notes

The `"ts-node"` key is placed between `"dependencies"` and `"prisma"`. Both `"ts-node"` and `"prisma"` are tool-config keys (not npm standard fields), so grouping them together or moving both after `"devDependencies"` would make the layout marginally more readable. Not worth a round-trip — flag for a future housekeeping pass if desired.

---

PR #346, opened 2026-05-25
[fix] Fail staging integration tests when deploy prereqs unmet, not skip

### LINKED ISSUE #345: Activate staging integration tests: Cloud SQL startup, env secrets, branch protection
## Problem

PR #342 shipped the staging deploy pipeline. The \`Staging Integration Tests\` job exists and skips cleanly today, but several blockers prevent it from actually running and passing.

### 1 — Cloud Run API does not start (~32 s crash)

\`Deploy API to Cloud Run (staging)\` fails with *"container failed to start and listen on PORT=3000 within the allocated timeout."* The container takes ~32 seconds to crash, which matches a \`Prisma.\$connect()\` timeout rather than an instant throw.

The next PR's CI run will emit the actual crash log via the "Fetch Cloud Run API logs on failure" step (\`gcloud logging read\` against \`lifting-logbook-stg-api\`). Read that output to identify the root cause.

Leading hypothesis: Cloud SQL private-IP access from Cloud Run via the VPC connector is not working (networking or IAM). DATABASE_URL is a valid PostgreSQL URL (computed by Terraform from the Cloud SQL private IP), so the URL format is not the issue.

**Investigation steps:**
1. Open the next PR on this repo and let CI run.
2. In the \`Deploy (staging)\` job, find the "Fetch Cloud Run API logs on failure" step — it runs \`gcloud logging read\` and prints the last 50 error lines from the \`lifting-logbook-stg-api\` Cloud Run revision.
3. The crash log will show the actual NestJS/Prisma error. Fix accordingly (likely a VPC connector egress rule, Cloud SQL IAM binding, or missing private-IP route).

### 2 — GitHub environment secrets not set

The \`staging-integration-tests\` job signs in through the real Clerk UI using a dedicated test account. Two secrets must be added under **Settings → Environments → staging → Environment secrets**:

| Secret name | Value |
|---|---|
| \`STAGING_CLERK_TEST_EMAIL\` | Email of the Clerk staging test account |
| \`STAGING_CLERK_TEST_PASSWORD\` | Password of the Clerk staging test account |

**Steps:**
1. clerk.com → lifting-logbook staging app → Users → Create user; note the credentials.
2. This repo: Settings → Environments → staging → add the two secrets above.

### 3 — Branch protection rule not yet added

Add this only **after** blockers 1 and 2 are resolved and at least one CI run shows the job passing.

Settings → Branches → Branch protection rules → \`main\` → Require status checks → add **\`Staging Integration Tests\`**.

### 4 — Skip-instead-of-fail design gap (workflow bug)

The \`staging-integration-tests\` job's \`if:\` condition evaluates all three of these at once:

\`\`\`yaml
if: |
  needs.deploy-staging.result == 'success' &&
  needs.deploy-staging.outputs.clerk_configured == 'true' &&
  needs.deploy-staging.outputs.cloud_run_api_deployed == 'true'
\`\`\`

When the Cloud Run API deploy fails, \`cloud_run_api_deployed\` becomes \`false\` — but because \`deploy-staging\` uses \`continue-on-error: true\`, the job itself still reports \`success\`. The entire \`if:\` evaluates to false, so GitHub marks the job **Skipped** instead of **Failed**.

**GitHub branch protection treats Skipped as passing.** This means even after branch protection is added for \`Staging Integration Tests\`, a Cloud Run API crash silently allows the merge to proceed — the gate is not actually enforced.

**Fix:** when staging is configured (\`preflight.should_run == 'true'\`), the integration test job should always run and fail explicitly when deploy prerequisites are not met, rather than being skipped by the \`if:\` condition. Only a genuine "staging not configured" (\`should_run == 'false'\`) is a valid reason to skip.

## Acceptance criteria

- [ ] \`staging-integration-tests\` fails (not skips) when Cloud Run API deploy failed but staging is otherwise configured
- [ ] Cloud Run API starts cleanly; \`Deploy API to Cloud Run (staging)\` turns green
- [ ] \`STAGING_CLERK_TEST_EMAIL\` and \`STAGING_CLERK_TEST_PASSWORD\` are set as environment secrets
- [ ] At least one CI run shows \`Staging Integration Tests\` passing (all 4 Playwright tests green)
- [ ] Branch protection rule for \`Staging Integration Tests\` is active on \`main\`

brownm09:
## Summary

- Fixes the skip-instead-of-fail design gap in `staging.yml` identified in #345 (Blocker 4)
- `Staging Integration Tests` previously evaluated all conditions — staging configured, Clerk key set, Cloud Run API deployed — inside a single `if:` guard. When Cloud Run API crashed, the job was **Skipped**, not **Failed**
- GitHub branch protection treats Skipped as passing, so the merge gate was silently bypassed on every Cloud Run crash
- Fix: `if:` now only skips when `preflight.should_run == 'false'` (no GCP credentials). When staging is configured, the job always runs and a new **Verify deploy prerequisites** step fails explicitly with a clear `::error::` message if Clerk or Cloud Run API is not ready

## What changes

**Before:** three-condition `if:` that silently skipped when any prerequisite was absent
```yaml
if: |
  needs.deploy-staging.result == 'success' &&
  needs.deploy-staging.outputs.clerk_configured == 'true' &&
  needs.deploy-staging.outputs.cloud_run_api_deployed == 'true'
```

**After:** single condition (only skip when staging is unconfigured); explicit fail step otherwise
```yaml
if: needs.preflight.outputs.should_run == 'true'
```
Plus a "Verify deploy prerequisites" step as the first step of the job that `exit 1`s with a descriptive error message.

## Acceptance criteria (from #345)

- [x] `staging-integration-tests` fails (not skips) when Cloud Run API deploy failed but staging is otherwise configured

Remaining checklist items (Cloud Run fix, Clerk secrets, branch protection) are operational tasks tracked in #345 and can only be completed once CI is green.

## Testing

`npm test` passes (10/10 tasks, 243 unit tests + integration tests). This change is YAML-only; the Playwright staging suite itself is unchanged.

No suppressions added.

Closes #345

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -352,13 +352,14 @@ jobs:
       - name: Fetch Cloud Run API logs on failure
         if: steps.deploy-api.outcome == 'failure'
         run: |
-          echo "=== Cloud Run API crash logs (last 50 lines) ==="
+          echo "=== Cloud Run API crash logs (last 50 lines, all severities) ==="
           gcloud logging read \
-            'resource.type="cloud_run_revision" AND resource.labels.service_name="lifting-logbook-stg-api" AND severity>=ERROR' \
+            'resource.type="cloud_run_revision" AND resource.labels.service_name="lifting-logbook-stg-api"' \
             --project="${{ vars.GCP_STAGING_PROJECT_ID }}" \
             --limit=50 \
-            --format="table(timestamp,textPayload,jsonPayload.message)" \
+            --format="table(timestamp,severity,textPayload,jsonPayload.message)" \
             --order=desc \
+            --freshness=5m \
             2>&1 || true
 
       - name: Deploy web to Cloud Run (staging)
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -380,10 +380,11 @@ jobs:
     runs-on: ubuntu-latest
     needs: [preflight, terraform-staging, deploy-staging]
     # Only skip when staging is not configured at all (no GCP credentials).
-    # When staging IS configured, run always so that deploy failures produce a
-    # FAILED status rather than SKIPPED — GitHub branch protection treats
-    # SKIPPED as passing, so a silent skip bypasses the merge gate. See #345.
-    if: needs.preflight.outputs.should_run == 'true'
+    # When staging IS configured, always() ensures the job runs even when a
+    # needs job hard-fails (e.g. GCP auth failure in deploy-staging) — without
+    # it GitHub Actions auto-skips dependents on failure, which branch protection
+    # treats as passing and silently bypasses the merge gate. See #345.
+    if: always() && needs.preflight.outputs.should_run == 'true'
     permissions:
       contents: read
       id-token: write
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -453,9 +453,20 @@ jobs:
         run: npx playwright install --with-deps chromium
         working-directory: apps/web
 
+      - name: Authenticate to GCP (staging) for Secret Manager access
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ secrets.GCP_STAGING_WORKLOAD_IDENTITY_PROVIDER }}
+          service_account: ${{ secrets.GCP_STAGING_SERVICE_ACCOUNT }}
+
       - name: Wait for staging to be ready
+        id: wait-ready
         run: |
           WEB_URL="${{ needs.deploy-staging.outputs.web_url }}"
+          # Show the first response body to diagnose unexpected status codes
+          echo "=== Diagnostic curl (attempt 0) ==="
+          curl -s -L --max-redirs 0 -D - "${WEB_URL}/" 2>&1 | head -30 || true
+          echo "==="
           for i in $(seq 1 18); do
             STATUS=$(curl -s -o /dev/null -w "%{http_code}" "${WEB_URL}/" || true)
             if [ "$STATUS" = "200" ] || [ "$STATUS" = "307" ]; then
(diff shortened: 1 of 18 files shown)


brownm09:
## Review: [fix] Fail staging integration tests when deploy prereqs unmet, not skip

### Summary

The fix correctly addresses the specific Cloud Run crash path — because `deploy-staging` uses `continue-on-error: true` on the Cloud Run step, the job always reports success with `cloud_run_api_deployed = 'false'`, and the new "Verify deploy prerequisites" step now catches that with `exit 1`. One blocking finding: the fix does not cover the case where `deploy-staging` itself fails hard (e.g., GCP auth failure), which still produces a SKIPPED on `staging-integration-tests` due to GitHub Actions' dependency-failure behavior. The fix for this is a one-line change.

---

### Blocking Findings

**[correctness]** `.github/workflows/staging.yml:381`

The job's `if: needs.preflight.outputs.should_run == 'true'` condition guards against running when staging is unconfigured, but GitHub Actions skips a dependent job when any job in its `needs` list **fails** (as distinct from **skips**) — the `if:` condition is never evaluated. `deploy-staging` contains steps without `continue-on-error: true` (specifically `Authenticate to GCP (staging)` at ~line 226 and `Check Clerk secret key configured` at ~line 353); if either fails, `deploy-staging` reports FAILED, and `staging-integration-tests` is SKIPPED, not FAILED. This is the same bypass the PR is explicitly trying to close.

**Fix:** Add `always()` to the `if:` condition so GitHub Actions does not auto-skip the job on a dependency failure:

```yaml
if: always() && needs.preflight.outputs.should_run == 'true'
```

`always()` returns `true` and signals to GitHub Actions that the job should run regardless of dependency results. The `&& needs.preflight.outputs.should_run == 'true'` guard still suppresses the job when staging is not configured at all. The "Verify deploy prerequisites" step already handles the empty-output case (empty string != "true"), so it will emit the right `::error::` message and `exit 1` whether `deploy-staging` succeeded with bad outputs or was entirely skipped due to an upstream failure.

---

PR #351, opened 2026-05-27
[fix] Add GCP identity token for Cloud Run service-to-service auth

### LINKED ISSUE #350: API calls from web server components fail with 403 — need GCP identity token for service-to-service auth
## Problem

The staging API Cloud Run service requires GCP Cloud Run IAM authentication (`roles/run.invoker` on the web workload SA — see `infra/terraform/cloud-run.tf: web_invoker_on_api`). Unauthenticated HTTP requests are rejected with 403 by Cloud Run infrastructure before reaching NestJS.

All server-side API calls from `apps/web` (server components, route handlers) currently make plain `fetch()` calls without a GCP identity token. This means every API call in the staging and production web app silently fails with 403, and server components swallow the error via `.catch()` wrappers.

## Discovered via

Staging integration test #345 — test 5 (`/api/health` route handler) returned 503 with `{"error":"api health returned 403"}` when calling the backend public `GET /health` endpoint.

## Root cause

The web app needs to obtain a GCP identity token from the GCP metadata service (`http://metadata.google.internal/...`) and pass it as `Authorization: Bearer <identity-token>` on every request to `API_URL`. This is standard Cloud Run service-to-service authentication (see [GCP docs](https://cloud.google.com/run/docs/authenticating/service-to-service)).

## Scope

Every place in `apps/web` that calls `fetch(`${API_URL}/...`)` needs to be updated:
- All server components (`fetchLiftRecords`, `fetchTrainingMaxHistory`, `fetchCycleDashboard`, `fetchUserSettings`, `fetchCustomPrograms`)
- Route handlers

A shared utility (e.g., `lib/api-client.ts`) should wrap `fetch()` with identity-token injection so the pattern isn't duplicated everywhere.

## Workaround

Currently, server components have `.catch()` wrappers that render empty/default state on API failure. The app appears to work but data never loads from the staging API.

## Acceptance criteria
- [ ] `lib/api-client.ts` (or equivalent) wraps `fetch()` and adds the GCP identity token when running on Cloud Run
- [ ] All `apps/web` server components and route handlers use the shared utility
- [ ] Staging integration test 5 (`/api/health`) tests the full path including identity-token-authenticated API calls
- [ ] Local dev still works (no metadata service → no identity token → no auth header needed for `localhost:3004`)

brownm09:
## Summary

- Extracts `getGcpIdentityToken()` from the health route into a shared `apps/web/lib/gcp-identity-token.ts` utility
- Updates `getAuthHeaders()` in `api.ts` to inject the GCP identity token as `Authorization: Bearer` (Cloud Run IAM) and the Clerk JWT as `X-Clerk-Authorization: Bearer` (NestJS AuthGuard) when running on Cloud Run
- Updates the NestJS `AuthGuard` to prefer `X-Clerk-Authorization` over `Authorization`, falling back to `Authorization` for browser clients and local dev (backward-compatible)

## Root cause

Cloud Run IAM rejects requests where `Authorization: Bearer` is not a valid GCP identity token. Server-side `apiFetch()` calls were sending a Clerk JWT there, getting 403 before NestJS was ever reached.

## Acceptance criteria

- [x] `lib/gcp-identity-token.ts` wraps the metadata service fetch and is imported by both `api.ts` and the health route
- [x] All server components and route handlers use the shared utility via `getAuthHeaders()`
- [x] Local dev unchanged (no metadata service → falls through to DEV_AUTH_TOKEN)
- [x] Staging integration test 5 (`/api/health`) passes once deployed — this test was failing with `{"error":"api health returned 403"}` before this fix

## Test results

```
@lifting-logbook/api  — Test Suites: 20 passed  |  Tests: 243 passed (all passing)
@lifting-logbook/web  — Test Suites: 3 passed   |  Tests: 28 passed
@lifting-logbook/core — all passing
```

Note: `@lifting-logbook/api` TypeScript build fails with Prisma type errors — pre-existing issue on `main`, not introduced by this PR.

## Compatibility matrix

| Caller | Authorization | X-Clerk-Authorization | Cloud Run IAM | NestJS Guard |
|---|---|---|---|---|
| Browser (Clerk JWT) | Bearer \<clerk-jwt\> | — | n/a | ✅ via Authorization |
| Local dev | Bearer dev-token | — | n/a | ✅ via Authorization |
| Server component (Cloud Run) | Bearer \<identity-token\> | Bearer \<clerk-jwt\> | ✅ | ✅ via X-Clerk-Authorization |
| No Clerk session | Bearer \<identity-token\> | — | ✅ | 401 (correct) |
| `GET /health` (public) | Bearer \<identity-token\> | — | ✅ | @Public(), skipped |

Closes #350

### DIFF
--- a/apps/api/src/auth/auth.guard.spec.ts
+++ b/apps/api/src/auth/auth.guard.spec.ts
@@ -1,0 +1,87 @@
+import { ExecutionContext, UnauthorizedException } from '@nestjs/common';
+import { Reflector } from '@nestjs/core';
+import { Test } from '@nestjs/testing';
+import { AuthGuard } from './auth.guard';
+import { AUTH_PROVIDER } from '../ports/tokens';
+import type { AuthUser } from '../ports/auth';
+
+const mockUser: AuthUser = { id: 'user_1', email: 'test@example.com', provider: 'clerk' };
+const mockAuthProvider = { verifyToken: jest.fn().mockResolvedValue(mockUser) };
+
+function makeContext(headers: Record<string, string>, isPublic = false): ExecutionContext {
+  return {
+    getHandler: () => ({}),
+    getClass: () => ({}),
+    switchToHttp: () => ({
+      getRequest: () => ({ headers, user: undefined }),
+    }),
+    // Reflector.getAllAndOverride is stubbed via the module; this lets the guard
+    // read IS_PUBLIC_KEY without a real Reflector call.
+    _isPublic: isPublic,
+  } as unknown as ExecutionContext;
+}
+
+describe('AuthGuard', () => {
+  let guard: AuthGuard;
+  let reflector: Reflector;
+
+  beforeEach(async () => {
+    mockAuthProvider.verifyToken.mockClear();
+    const module = await Test.createTestingModule({
+      providers: [
+        AuthGuard,
+        {
+          provide: Reflector,
+          useValue: { getAllAndOverride: jest.fn().mockReturnValue(false) },
+        },
+        { provide: AUTH_PROVIDER, useValue: mockAuthProvider },
+      ],
+    }).compile();
+
+    guard = module.get(AuthGuard);
+    reflector = module.get(Reflector);
+  });
+
+  it('allows @Public() routes without a token', async () => {
+    (reflector.getAllAndOverride as jest.Mock).mockReturnValue(true);
+    const ctx = makeContext({});
+    await expect(guard.canActivate(ctx)).resolves.toBe(true);
+    expect(mockAuthProvider.verifyToken).not.toHaveBeenCalled();
+  });
+
+  it('throws UnauthorizedException when no token header is present', async () => {
+    const ctx = makeContext({});
+    await expect(guard.canActivate(ctx)).rejects.toThrow(UnauthorizedException);
+  });
+
+  it('prefers x-clerk-authorization over authorization when both are present', async () => {
+    const clerkJwt = 'clerk-jwt-token';
+    const gcpToken = 'gcp-identity-token';
+    const request = { headers: { 'x-clerk-authorization': `Bearer ${clerkJwt}`, authorization: `Bearer ${gcpToken}` }, user: undefined };
+    const ctx = {
+      getHandler: () => ({}),
+      getClass: () => ({}),
+      switchToHttp: () => ({ getRequest: () => request }),
+    } as unknown as ExecutionContext;
+
+    await guard.canActivate(ctx);
+
+    expect(mockAuthProvider.verifyToken).toHaveBeenCalledWith(clerkJwt);
+    expect(request.user).toEqual(mockUser);
+  });
+
+  it('falls back to authorization when x-clerk-authorization is absent', async () => {
+    const clerkJwt = 'clerk-jwt-token';
+    const request = { headers: { authorization: `Bearer ${clerkJwt}` }, user: undefined };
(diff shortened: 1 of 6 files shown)


brownm09:
## Review: [fix] Add GCP identity token for Cloud Run service-to-service auth

### Summary

Needs revision before merging. The implementation is correct for the happy path (authenticated user on Cloud Run), but there is one blocking correctness bug: `getAuthHeaders()` returns an empty headers object when `identityToken` is available but `clerkToken` is null — the opposite of what the PR body's compatibility matrix documents. The fix is a 3-line addition. Everything else is clean: the extraction of `gcp-identity-token.ts`, the dual-header routing, and the `auth.guard.ts` fallback are all correct and well-structured.

---

### Blocking Findings

**[correctness]** `apps/web/lib/api.ts:54–64`

When `isCloudRun` is true and `getGcpIdentityToken()` succeeds but `clerkToken` is null (session expired, Clerk error, or unprotected route), both conditionals (`if (identityToken && clerkToken)` and `if (clerkToken)`) evaluate false and the function returns `{}` — no `Authorization` header. Cloud Run IAM then rejects the request with 403 before it reaches NestJS.

The PR body's compatibility matrix explicitly documents this case as `Authorization: Bearer <identity-token>` | Cloud Run IAM ✅ | NestJS 401 (correct). The code contradicts that documented intent: Cloud Run IAM never passes the request, so NestJS never returns 401.

**Fix:** Add an `identityToken`-only branch before the final `return {}`:

```typescript
if (identityToken && clerkToken) {
  return {
    Authorization: `Bearer ${identityToken}`,
    'X-Clerk-Authorization': `Bearer ${clerkToken}`,
  };
}
if (identityToken) {
  // No Clerk session — passes Cloud Run IAM; NestJS returns 401 for protected endpoints
  return { Authorization: `Bearer ${identityToken}` };
}
if (clerkToken) {
  console.warn('...');
  return { Authorization: `Bearer ${clerkToken}` };
}
return {};
```

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/auth/auth.guard.ts`

`auth.guard.spec.ts` does not exist (confirmed on the PR branch), so the new header-preference logic has no unit coverage. A regression that swaps the header priority would silently break all server-to-server API calls in staging and production; the only signal would be test 5 failing on the next staging deploy.

**Suggestion:** Add an `auth.guard.spec.ts` with two cases: (a) `x-clerk-authorization` present and `authorization` present → uses `x-clerk-authorization`; (b) only `authorization` present → uses `authorization`. The existing 243 API tests don't cover this guard behavior.

---

**[performance]** `apps/web/lib/api.ts:40–52`

`getToken()` and `getGcpIdentityToken()` are called sequentially. They're independent I/O operations; the metadata fetch has a 2-second timeout. On slow GCP metadata responses, every server-component render blocks for up to 2 seconds past the Clerk token acquisition time.

**Suggestion:** Use `Promise.all` if this becomes a perceptible latency contributor — low priority for current traffic but worth tracking as the app scales.

---

### Style Notes

The comment block in `gcp-identity-token.ts` (7 lines) explains the infrastructure constraint, the IAM binding location, and the null-return contract. This is exactly the "why" documentation warranted by a non-obvious constraint. The length is proportionate.

The `as string | undefined` casts in `auth.guard.ts` could silently drop multi-value header arrays (Fastify types headers as `string | string[]`), but this is an existing pattern in the codebase and the callers are fully under your control — not worth changing.

---

PR #352, opened 2026-05-27
[fix] Restore PrismaInstrumentation via postinstall OTel cleanup

### LINKED ISSUE #348: Fix PrismaInstrumentation OTel SDK version conflict
## Problem

\`@prisma/instrumentation@5.22.0\` bundles its own copy of \`@opentelemetry/sdk-trace-base\` inside its \`node_modules/\`. When the main app's OTel SDK registers \`PrismaInstrumentation\` and Prisma's client later calls \`createEngineSpan()\` during \`$connect()\`, the bundled SDK receives a tracer object from the main app's SDK version. The two SDK versions have incompatible interfaces, producing:

\`\`\`
TypeError: parentTracer.getActiveSpanProcessor is not a function
    at new Span (.../prisma/instrumentation/node_modules/@opentelemetry/sdk-trace-base/...)
    at ActiveTracingHelper.createEngineSpan
    at mo.createEngineSpan (/app/node_modules/@prisma/client/runtime/library.js)
\`\`\`

This crash happens at runtime during \`PrismaService.onModuleInit()\` → \`$connect()\`, after NestJS has started successfully, and kills the process — preventing the Cloud Run startup probe from passing.

## Workaround (applied in PR #346)

\`PrismaInstrumentation\` was removed from the OTel instrumentations array in \`otel.ts\`. Prisma spans are not captured, but the app starts successfully.

## Fix

Resolve the OTel SDK version conflict so \`PrismaInstrumentation\` can be safely included. Options:
1. Pin the main app's \`@opentelemetry/sdk-trace-base\` to match the version bundled by \`@prisma/instrumentation@5.22.0\`
2. Upgrade \`@prisma/client\` and \`@prisma/instrumentation\` to a version that uses the same OTel SDK major as the main app
3. Use \`npm overrides\` to force a single \`@opentelemetry/sdk-trace-base\` version across all packages

brownm09:
## Summary

- `@prisma/instrumentation@5.22.0` bundles `@opentelemetry/sdk-trace-base@1.30.1` in its own `node_modules/`. When registered with the main app's NodeSDK (which uses `sdk-trace-base@2.7.1`), Prisma's v1.x `Span` constructor calls `parentTracer.getActiveSpanProcessor()` — a method removed in the v2.x tracer API — crashing `PrismaService.$connect()` and blocking Cloud Run startup.
- PR #346 removed `PrismaInstrumentation` as a workaround. This PR fixes the root cause and restores it.
- **Fix:** A `postinstall` script in `apps/api/package.json` deletes `@prisma/instrumentation/node_modules/@opentelemetry/` after every `npm install` and `npm ci`, forcing Node.js module resolution to traverse up to the hoisted `sdk-trace-base@2.7.1`.

## Changes

- **`apps/api/package.json`** — adds `postinstall` script that `rmSync`s the nested `@opentelemetry/` directory (idempotent)
- **`apps/api/src/otel.ts`** — re-enables `PrismaInstrumentation` import and registration; comment updated to reference the postinstall mechanism
- **`docs/adr/ADR-024-prisma-otel-sdk-override.md`** — new ADR documenting the decision; includes the npm `overrides` exploration and why postinstall was chosen
- **`docs/adr-references.md`** — three new references for ADR-024
- **`docs/README.md`** — adds ADR-023 and ADR-024 rows to the ADR index table
- **`apps/api/src/observability/otel.e2e.spec.ts`** — adds `PrismaInstrumentation` instantiation regression test for #348
- **`package-lock.json`** — adds `"hasInstallScript": true` to the `apps/api` workspace entry (tells npm to run the postinstall script during `npm ci`); all other lockfile content is identical to main

## Why postinstall instead of npm overrides

`npm overrides` is the idiomatic fix, but with npm 11.x and an existing lockfile, scoped overrides do not update already-pinned lockfile entries for workspace-level transitive packages. Deleting the lockfile to force fresh resolution triggered peer-dep conflicts (OTel v1/v2 coexistence); resolving those with comprehensive overrides changed package hoisting in a way that broke Jest module resolution for the NestJS test suite (17 suites failed). The postinstall approach avoids all of this: it preserves the original lockfile and hoisting, applies on every `npm install` and `npm ci`, and is fully idempotent. Full exploration documented in ADR-024.

## Test plan

- [x] **Structural:** `apps/api/node_modules/@prisma/instrumentation/node_modules/@opentelemetry/` does not exist after `npm install`
- [x] **Unit/smoke (api):** `npm test -w @lifting-logbook/api` — 20 suites passed, 1 skipped, 244 tests passed, 0 failures (includes `otel.e2e.spec.ts` and new PrismaInstrumentation regression test)
- [x] **Unit (web):** `npm test -w @lifting-logbook/web` — 3 suites passed, 28 tests passed, 0 failures (verified lockfile change does not affect apps/web)
- [ ] **Staging:** Cloud Run startup probe passes; staging traces show at least one `prisma:engine` or `prisma:client` span confirming `PrismaInstrumentation` is wired in

## Suppressions

None added.

Closes #348

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -8,7 +8,8 @@
     "start": "node dist/main.js",
     "lint": "eslint src",
     "test": "jest",
-    "test:db": "jest --testPathPattern=db\\.e2e"
+    "test:db": "jest --testPathPattern=db\\.e2e",
+    "postinstall": "node -e \"const fs=require('fs');try{fs.rmSync('node_modules/@prisma/instrumentation/node_modules/@opentelemetry',{recursive:true,force:true})}catch(e){}\""
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.92.0",
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -8,8 +8,7 @@
     "start": "node dist/main.js",
     "lint": "eslint src",
     "test": "jest",
-    "test:db": "jest --testPathPattern=db\\.e2e",
-    "postinstall": "node -e \"const fs=require('fs');try{fs.rmSync('node_modules/@prisma/instrumentation/node_modules/@opentelemetry',{recursive:true,force:true})}catch(e){}\""
+    "test:db": "jest --testPathPattern=db\\.e2e"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.92.0",
--- a/apps/api/src/auth/auth.guard.spec.ts
+++ b/apps/api/src/auth/auth.guard.spec.ts
@@ -1,0 +1,87 @@
+import { ExecutionContext, UnauthorizedException } from '@nestjs/common';
+import { Reflector } from '@nestjs/core';
+import { Test } from '@nestjs/testing';
+import { AuthGuard } from './auth.guard';
+import { AUTH_PROVIDER } from '../ports/tokens';
+import type { AuthUser } from '../ports/auth';
+
+const mockUser: AuthUser = { id: 'user_1', email: 'test@example.com', provider: 'clerk' };
+const mockAuthProvider = { verifyToken: jest.fn().mockResolvedValue(mockUser) };
+
+function makeContext(headers: Record<string, string>, isPublic = false): ExecutionContext {
+  return {
+    getHandler: () => ({}),
+    getClass: () => ({}),
+    switchToHttp: () => ({
+      getRequest: () => ({ headers, user: undefined }),
+    }),
+    // Reflector.getAllAndOverride is stubbed via the module; this lets the guard
+    // read IS_PUBLIC_KEY without a real Reflector call.
+    _isPublic: isPublic,
+  } as unknown as ExecutionContext;
+}
+
+describe('AuthGuard', () => {
+  let guard: AuthGuard;
+  let reflector: Reflector;
+
+  beforeEach(async () => {
+    mockAuthProvider.verifyToken.mockClear();
+    const module = await Test.createTestingModule({
+      providers: [
+        AuthGuard,
+        {
+          provide: Reflector,
+          useValue: { getAllAndOverride: jest.fn().mockReturnValue(false) },
+        },
+        { provide: AUTH_PROVIDER, useValue: mockAuthProvider },
+      ],
+    }).compile();
+
+    guard = module.get(AuthGuard);
+    reflector = module.get(Reflector);
+  });
+
+  it('allows @Public() routes without a token', async () => {
+    (reflector.getAllAndOverride as jest.Mock).mockReturnValue(true);
+    const ctx = makeContext({});
+    await expect(guard.canActivate(ctx)).resolves.toBe(true);
+    expect(mockAuthProvider.verifyToken).not.toHaveBeenCalled();
+  });
+
+  it('throws UnauthorizedException when no token header is present', async () => {
+    const ctx = makeContext({});
+    await expect(guard.canActivate(ctx)).rejects.toThrow(UnauthorizedException);
+  });
+
+  it('prefers x-clerk-authorization over authorization when both are present', async () => {
+    const clerkJwt = 'clerk-jwt-token';
+    const gcpToken = 'gcp-identity-token';
+    const request = { headers: { 'x-clerk-authorization': `Bearer ${clerkJwt}`, authorization: `Bearer ${gcpToken}` }, user: undefined };
+    const ctx = {
+      getHandler: () => ({}),
+      getClass: () => ({}),
+      switchToHttp: () => ({ getRequest: () => request }),
+    } as unknown as ExecutionContext;
+
+    await guard.canActivate(ctx);
+
+    expect(mockAuthProvider.verifyToken).toHaveBeenCalledWith(clerkJwt);
+    expect(request.user).toEqual(mockUser);
+  });
+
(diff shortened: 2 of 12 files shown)


brownm09:
## Review: [fix] Restore PrismaInstrumentation via postinstall OTel cleanup

### Summary

Three fixes needed before merge. The core approach — postinstall cleanup of the nested `@prisma/instrumentation` OTel v1.x directory — is technically sound (confirmed: the package is workspace-local in `apps/api/node_modules/`, so the relative path in the script targets the right directory). Two issues require changes before merging: the `otel.ts` comment names the wrong mechanism, and `docs/README.md`'s ADR index was not updated. One additional reliability concern about an unintended lockfile side-effect in `apps/web` needs verification or explanation before merge.

---

### Blocking Findings

**[correctness]** `apps/api/src/otel.ts:7–8`

The comment says "npm overrides in root `package.json` force `@prisma/instrumentation` to share our `sdk-trace-base` version" — but no `overrides` field was added to root `package.json`. The actual fix mechanism is the `postinstall` script in `apps/api/package.json`. A developer following this comment will open root `package.json`, find nothing, and have no idea how the crash was fixed or how to maintain the fix.

**Fix:** Replace the comment with one that accurately describes the postinstall mechanism, e.g.:
```typescript
// A postinstall script in apps/api/package.json deletes @prisma/instrumentation's
// nested node_modules/@opentelemetry/ after every npm install, forcing resolution
// to the hoisted sdk-trace-base@2.7.1 and eliminating the v1/v2 tracer API mismatch.
// See ADR-024 and issue #348.
import { PrismaInstrumentation } from '@prisma/instrumentation';
```

---

**[documentation]** `docs/adr/ADR-024-prisma-otel-sdk-override.md` (new file)

`docs/README.md` contains an ADR index table that is the authoritative listing of accepted ADRs. It currently ends at ADR-022; ADR-024 was added in this PR but the table was not updated. Note: ADR-023 also appears absent — this PR should address ADR-024 at minimum.

**Fix:** Add a row to the `docs/README.md` ADR table:
```markdown
| [ADR-024](adr/ADR-024-prisma-otel-sdk-override.md) | Prisma OTel SDK Version Conflict — postinstall Cleanup | Accepted |
```

---

**[reliability]** `package-lock.json` — `apps/web/node_modules/@opentelemetry/` entries removed

Running `npm install` in the worktree (required by CLAUDE.md worktree setup) removed seven `apps/web/node_modules/@opentelemetry/` entries from the lockfile as an unintended side effect: `sdk-trace-base@1.30.1`, `core@1.30.1`, `resources@1.30.1`, `sdk-metrics@1.30.1`, `sdk-logs@0.57.2`, `api-logs@0.57.2`, and `instrumentation@0.57.2`. These were previously installed at the workspace level to satisfy `@vercel/otel`'s `peerDependencies: { "@opentelemetry/sdk-trace-base": ">=1.19.0 <2.0.0" }`. The hoisted root version (`sdk-trace-base@2.7.1`) does not satisfy the `< 2.0.0` upper bound. The PR's own ADR states "apps/web `@vercel/otel` nested sdk-trace-base is unaffected by the scoped override — this is intentional," but the lockfile diff shows it was affected. `apps/web` tests were not run in this PR.

**Fix:** Run `npm test -w @lifting-logbook/web` against a clean install state to confirm apps/web behaves correctly with the updated lockfile. If tests fail with an OTel-related error, investigate why npm removed the `apps/web/node_modules/@opentelemetry/` entries and whether the `@vercel/otel` peer dep constraint can still be satisfied. If tests pass, document in the PR body that apps/web was verified and explain why the `< 2.0.0` peer dep is satisfied despite v2.7.1 being resolved.

---

### Non-Blocking Findings

**[maintainability]** `package-lock.json` — PR body characterization

The PR body states "package-lock.json — reset to HEAD (original lockfile preserved; postinstall handles runtime cleanup)". The actual diff shows ~150 lines of lockfile changes beyond `"hasInstallScript": true`: removal of the `apps/web/node_modules/@opentelemetry/` v1.x entries and addition of `"peer": true` to dozens of packages. The characterization obscures a meaningful lockfile change.

**Suggestion:** Update the PR body to describe the actual lockfile state: "`hasInstallScript: true` intentionally added for `apps/api`. Running `npm install` in the worktree also reorganized peer-dep metadata and removed `apps/web/node_modules/@opentelemetry/` v1.x nested entries as an unintended side-effect (see the reliability finding above)."

---

**[maintainability]** Missing regression test for PrismaInstrumentation registration

CLAUDE.md's coverage gate for bug fixes requires "a regression test that fails before the fix and passes after." The existing `otel.e2e.spec.ts` smoke test doesn't call `$connect()`, so it would pass on the base branch (where PrismaInstrumentation was absent) as well as after this fix. There is no test that specifically asserts PrismaInstrumentation appears in the SDK's `instrumentations` array. Such a test would fail on the base branch and pass after this fix, without requiring `DATABASE_URL`.

**Suggestion:** Add a test in `otel.e2e.spec.ts` or a new unit file that calls `startOtel()` with a mocked or no-op exporter and asserts the returned SDK config includes a `PrismaInstrumentation` instance. This closes the coverage gap and creates an explicit signal if PrismaInstrumentation is accidentally removed again.

---

brownm09:
Known staging prerequisite gap — `lifting-logbook-stg-clerk-secret-key` is still the placeholder value, unrelated to this fix. Staging tests will run once the secret is configured per docs/deploy.md Step 4.

Issue #356, opened 2026-05-27
Align parseCycleDashboard validation with parseTrainingMaxes: throw on missing required keys


brownm09:
## Background

Surfaced by the audit in #354.

`packages/core/src/utils/parser/parseCycleDashboard.ts` silently defaults missing required keys to sentinel values:

| Key | Missing-value result |
|---|---|
| `Program` | `""` |
| `Cycle Unit` | `""` |
| `Cycle #` | `Number(undefined)` → `NaN` |
| `Cycle Date` | `new Date("")` → `Invalid Date` |
| `Sheet Name` | `""` |
| `Start Weekday` | `""` (cast to `Weekday`) |

The sibling parser `parseCycleDashboard`'s peer, [\`parseTrainingMaxes.ts\`](https://github.com/brownm09/lifting-logbook/blob/main/packages/core/src/utils/parser/parseTrainingMaxes.ts), takes the opposite approach: it throws on missing or invalid fields. Both parsers consume the same shape of CSV input from the same dashboard, so the inconsistency is most likely accidental rather than a deliberate graceful-degradation strategy. The sentinels (\`NaN\`, \`Invalid Date\`) are not useful values — they push failures downstream to whatever renders the dashboard.

## Scope

- Audit callers of \`parseCycleDashboard\` to confirm none depend on the silent-default behavior.
- Change \`parseCycleDashboard\` to throw on missing required keys, matching \`parseTrainingMaxes\`.
- Update the pinned-current-behavior test added in #354 to assert on the throw instead of the sentinel values.
- Decide whether \`currentWeekType\` (hard-coded to \`'training'\` and labeled as derived elsewhere) should be moved out of the parser entirely.

## Acceptance criteria

- [ ] \`parseCycleDashboard\` throws a descriptive error when any required key is missing or unparseable.
- [ ] Test added in #354 (\`returns sentinel defaults when required keys are missing\`) is replaced with a throw assertion.
- [ ] All callers handle or propagate the new throw appropriately; no caller silently catches and continues with bad data.

## Related

- #354 — audit that surfaced this
- #349 — broader skewed-test audit / standard

Issue #349, opened 2026-05-26
Audit and remediate skewed tests across the codebase


brownm09:
## Problem

A skewed test passes under both the success condition it claims to verify *and* a failure condition it should detect. The test gives a false green signal.

This pattern appeared in PR #346 when server components were given graceful-degradation wrappers (correct product behavior) without corresponding updates to the test assertions:

- `ProgramsPage` — `fetchUserSettings().catch(() => DEFAULT_SETTINGS)` / `fetchCustomPrograms().catch(() => [])`. The staging test asserts the Programs heading and a "Choose This Program" button are visible. Both render from static component data regardless of whether the API returned real settings or the fallback defaults.
- `HistoryPage` — `fetchLiftRecords().catch(() => [])` / `fetchTrainingMaxHistory().catch(() => { entries: [] })`. The staging test asserts the two tab elements are visible. Tabs always render; their presence gives no signal that the API responded successfully.
- `CyclePage` — `try { … } catch { redirect('/onboarding') }`. The staging test accepts either `/cycle/N` or `/onboarding` as valid. An API failure redirects to `/onboarding`, which is indistinguishable from a fresh account with no active cycle.

PR #346 added an explicit auth propagation test (test 5 in `staging.spec.ts`) to address the staging case. The broader pattern has not been audited.

## Scope

Skewed tests can occur at any layer where:
- Error handling returns a neutral/empty value (`.catch(() => [])`, `?? default`, try/catch redirect)
- The test assertion checks structure or presence only — not that real data was loaded
- The production behavior (don't crash on transient failure) is correct, but the test cannot distinguish success from graceful failure

## Acceptance criteria

- [ ] Audit all existing tests (unit, integration, component, e2e) for the skewed pattern: error-swallowing code paths where the test assertion passes under both success and failure
- [ ] For each instance found, determine the right fix:
  - Add an assertion on actual data when a predictable non-empty response is guaranteed (e.g., a catalog endpoint that always returns rows)
  - Add a separate test that explicitly verifies the success path (analogous to the auth propagation test in `staging.spec.ts`)
  - Document intentional "structure only" assertions with a comment explaining why data-level assertion is not appropriate (e.g., staging test user has no data)
- [ ] Update the `## Testing` section in `CLAUDE.md` or `docs/standards/` with a rule that prevents new skewed tests from landing: when adding a `.catch()` or try/catch to a server component or API boundary, the reviewer must verify that any existing tests for that code path still detect the error condition

[...]

PR #357, opened 2026-05-27
[test] Add fallback-branch tests for packages/core neutral-return paths

### LINKED ISSUE #354: Audit packages/core and packages/types for skewed test patterns
## Problem

Issue #349 audits skewed tests in `apps/web` and `apps/api` where I/O boundaries swallow errors with fallback values. `packages/core` and `packages/types` were declared out of scope under the assumption that pure-logic packages have no I/O fallbacks — and therefore no analogous skewed pattern. That assumption has not been verified.

A pure-logic equivalent of the skewed pattern is plausible: e.g., a domain function that returns a neutral value (`null`, `[]`, `0`) when given malformed input, paired with a test that only asserts the function does not throw. Such a test passes whether the function correctly computes the result or silently returns the neutral value on a real bug.

## Scope

- Audit all `.catch()`, `?? default`, try/catch returning neutral values, and "return null on invalid input" branches in `packages/core` and `packages/types`.
- For each, check the corresponding test: does the assertion distinguish the success path from the neutral-value fallback path?
- Document any findings under the same remediation rubric used in #349 (data assertion / explicit success-path test / documented intentional structure-only).

## Acceptance criteria

- [ ] Inventory of every error-swallowing or neutral-return code path in `packages/core` and `packages/types`.
- [ ] For each, a verdict: skewed / not skewed / no test exists.
- [ ] Remediation PRs (or a single PR) addressing any confirmed skewed instances per the rubric in `docs/standards/error-fallback-test-coverage.md`.
- [ ] If no skewed instances are found, close with a one-line note in `docs/standards/error-fallback-test-coverage.md` recording that the audit was performed and found no issues in pure-logic packages.

## Related

- #349 — original audit and standard

brownm09:
## Summary

Closes #354.

Audit of `packages/core` and `packages/types` for skewed test patterns (the assumption #349 set aside: pure-logic packages have no analogous swallow-and-default fallbacks) found four error-swallowing branches with weak or absent test coverage. This PR adds tests for each, locks in current behavior, and creates the standard doc the audit was asked to record itself in.

## Audit results

| File | Pattern | Previous coverage | Remediation |
|---|---|---|---|
| `packages/core/src/utils/parser/tableToObjects.ts` | `if (!data \|\| data.length < 2) return []` | Success path tested with data assertions; fallback path never exercised | Added empty-array and header-only tests; added headerless test |
| `packages/core/src/utils/parser/parseCycleDashboard.ts` | Silent defaults (`NaN`, `Invalid Date`, `""`) for missing required keys | Only happy-path tested | Added test pinning current sentinel defaults |
| `packages/core/src/utils/jsUtil.ts:62` (`formatDateYYYYMMDD`) | `return String(date)` fallback + silent `NaNNaNNaN` for 3-part garbage | Only formatted-string and Date paths tested | Added tests for both fallback paths |
| `packages/core/src/utils/jsUtil.ts:90` (`weekTypeForDate`) | `?? 'training'` fallback when no spec match | No tests at all | Added 4 tests covering matching week, clamp, missing weekType, empty spec |

Eight other files in scope (`parseLiftingProgramSpec`, `parseTrainingMaxes`, `parseLiftRecords`, `generateLiftSpec`, `extractLiftRecords`, `distributeWorkouts`, `updateMaxes`, `updateCycle`) already have data-asserting tests for their neutral-return branches — not skewed. The three type-only files (`StrengthGoalEntry`, `TrainingMaxHistoryEntry`, `packages/types/src/domain.ts`) have no executable paths.

## Follow-up filed

While auditing `parseCycleDashboard`, found that its silent-default behavior is inconsistent with sibling `parseTrainingMaxes`, which throws on missing required fields. The sentinels (`NaN`, `Invalid Date`) are not useful values — they push failure downstream. Tracked separately as [#356](https://github.com/brownm09/lifting-logbook/issues/356) since changing it is a production behavior change that deserves its own review. This PR pins current behavior so #356's change is deliberate and visible.

## No production code changes

All edits are in `packages/core/tests/`, plus `docs/standards/error-fallback-test-coverage.md` (new draft — full rubric owned by #349) and one row added to the CLAUDE.md Coding Standards table.

## Testing

```
npm test -w @lifting-logbook/core
```

```
Tests: 155 passed, 0 skipped, 0 failed (7.7 s)
Test Suites: 26 passed, 26 total
```

Mutation sensitivity confirmed: the `not-a-date` test surfaced a latent code behavior (`NaNNaNNaN` output) that would have been masked by a structural-only assertion — proving the new tests actually exercise the production behavior.

Full-suite `npm test` does not run in the Claude worktree (turbo binary `EFTYPE` — known issue, see `.husky/pre-commit`). Verified per-workspace instead:

| Workspace | Result | Pre-existing on `main`? |
|---|---|---|
| `@lifting-logbook/core` | 155 / 155 passed | n/a (changed) |
| `@lifting-logbook/types` | passed (no tests) | n/a |
| `@lifting-logbook/mobile` | passed (no tests) | n/a |
| `@lifting-logbook/api` | 19 suites fail (Resolver module-not-found) | Yes — same failures on `origin/main` baseline |
| `@lifting-logbook/api-legacy` | 1 suite fail | Yes — same failure on `origin/main` baseline |
| `@lifting-logbook/web` | 1 suite fail | Yes — same failure on `origin/main` baseline |

No regressions introduced by this PR.

## Suppression / test-integrity check

- Suppression grep: one match — `spec[0]!` in `jsUtil.test.ts`. **Justification:** `spec` is a 2-element const literal declared in the same `describe` block; the assertion is on a statically known non-empty array, not an unchecked external value. Per `docs/standards` suppression policy this is a legitimate narrowing, not an error-silencer.
- Test-integrity grep: clean. No skips, deletes, threshold changes, or bypass flags.

## Test plan

- [x] `npm test -w @lifting-logbook/core` passes (155 / 155)
- [x] New tests confirmed to exercise real fallback branches (one surfaced a latent NaN behavior)
- [x] No production code touched

### DIFF
--- a/docs/standards/error-fallback-test-coverage.md
+++ b/docs/standards/error-fallback-test-coverage.md
@@ -77,6 +77,6 @@ When the staging test user genuinely has no records, this is acceptable **only i
 
 ## Audit history
 
-| Date | Scope | Issue | PR |
-|---|---|---|---|
-| 2026-05-27 | `packages/core/**`, `packages/types/**` — 4 untested or under-tested neutral-return branches in `tableToObjects`, `parseCycleDashboard`, `formatDateYYYYMMDD`, `weekTypeForDate`. All four remediated with explicit fallback-branch tests. One production-code inconsistency in `parseCycleDashboard` (silent defaults vs. sibling `parseTrainingMaxes` which throws) deferred to a follow-up issue. | [#354](https://github.com/brownm09/lifting-logbook/issues/354) | [#357](https://github.com/brownm09/lifting-logbook/pull/357) |
+| Date | Scope | Findings | Issue | PR |
+|---|---|---|---|---|
+| 2026-05-27 | `packages/core/**`, `packages/types/**` | 4 untested or under-tested neutral-return branches in `tableToObjects`, `parseCycleDashboard`, `formatDateYYYYMMDD`, `weekTypeForDate`. All four remediated with explicit fallback-branch tests; one production-code inconsistency in `parseCycleDashboard` (silent defaults vs. sibling `parseTrainingMaxes` which throws) deferred to a follow-up issue. | [#354](https://github.com/brownm09/lifting-logbook/issues/354) | [#357](https://github.com/brownm09/lifting-logbook/pull/357) |
--- a/docs/standards/error-fallback-test-coverage.md
+++ b/docs/standards/error-fallback-test-coverage.md
@@ -72,3 +72,11 @@ When the staging test user genuinely has no records, this is acceptable **only i
 - [Issue #349 — Audit and remediate skewed tests](https://github.com/brownm09/lifting-logbook/issues/349) — the audit that produced this standard
 - [PR #346 — staging deploy with Playwright integration test gate](https://github.com/brownm09/lifting-logbook/pull/346) — the incident that surfaced the pattern, and the first explicit success-path test (test 5)
 - [ADR-023 — Staging Integration Test Design](../adr/ADR-023-staging-integration-test-design.md) — Consequences section names this limitation
+
+---
+
+## Audit history
+
+| Date | Scope | Issue | PR |
+|---|---|---|---|
+| 2026-05-27 | `packages/core/**`, `packages/types/**` — 4 untested or under-tested neutral-return branches in `tableToObjects`, `parseCycleDashboard`, `formatDateYYYYMMDD`, `weekTypeForDate`. All four remediated with explicit fallback-branch tests. One production-code inconsistency in `parseCycleDashboard` (silent defaults vs. sibling `parseTrainingMaxes` which throws) deferred to a follow-up issue. | [#354](https://github.com/brownm09/lifting-logbook/issues/354) | [#357](https://github.com/brownm09/lifting-logbook/pull/357) |
--- a/packages/core/tests/core/utils/jsUtil.test.ts
+++ b/packages/core/tests/core/utils/jsUtil.test.ts
@@ -2,7 +2,7 @@ import {
   addDaysUTC,
   formatDateYYYYMMDD,
   getNextDate,
-  LiftingProgramSpec,
+  type LiftingProgramSpec,
   weekTypeForDate,
 } from "@src/core";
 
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: [test] Add fallback-branch tests for packages/core neutral-return paths

### Summary
Merge with one minor fix. The audit is well-scoped, tests pin behavior with honest comments distinguishing "current behavior" from "contract," and the parser-inconsistency deferral to #356 is the right call. One real finding: the new standard doc's audit-history table mis-labels an issue link as the PR. Everything else is style-grade.

---

### Non-Blocking Findings

**[documentation]** `docs/standards/error-fallback-test-coverage.md:27`
The "Audit history" table's `PR` column links `#354`, which is the tracking *issue*, not the PR. The remediation PR is `#357`. As more audit rows accrue under #349's broader effort, future readers using this column to jump to the actual diff that landed each remediation will hit issue pages instead.
**Suggestion:** Replace the `[#354]` cell with `[#357](https://github.com/brownm09/lifting-logbook/pull/357)`, or split into `Issue | PR` columns if you want both.

**[maintainability]** `packages/core/tests/core/utils/jsUtil.test.ts:5`
`LiftingProgramSpec` is imported as a value-side import alongside the runtime functions. If the repo later turns on `verbatimModuleSyntax` (or stricter `isolatedModules` enforcement), this will error.
**Suggestion:** Promote to `import type { LiftingProgramSpec }` in a separate import group, or use the inline `import { type LiftingProgramSpec, ... }` form.

**[maintainability]** `packages/core/tests/core/utils/jsUtil.test.ts:80`
The "clamps to the max week" test name asserts a specific design (clamping past-range to maxWeek). The test passes today, but if the implementation actually returns `'test'` because both the spec's max and the fallback happen to differ from `'training'` in opposite ways, a future refactor could quietly invalidate the *name* without failing the assertion. Worth a quick second assertion or a comment confirming the clamp is the source of `'test'` here.
**Suggestion:** Add `expect(weekTypeForDate(cycleStart, spec, new Date(2026, 0, 19))).toBe("test")` for the in-range week-2 case in the same describe — together with the past-range case, that pins the clamp specifically.

---

### Style Notes

The `const { weekType: _omit, ...rest } = spec[0]!; void _omit;` pattern is unusual but justified — `exactOptionalPropertyTypes` rules out the more obvious `{ ...spec[0]!, weekType: undefined }`. Worth a one-line comment ("// destructure to omit weekType — TS2375 under exactOptionalPropertyTypes") so the next person doesn't try to "simplify" it.

Comment blocks tying tests to audit issue #354 are well-done and the right pattern — keep doing this.

---

PR #355, opened 2026-05-27
[test] Remediate skewed tests + add fallback-coverage standard

brownm09:
## Summary

Audits and remediates the skewed-test pattern surfaced by PR #346 — tests that pass under both the success path of an upstream call and the silent-fallback path of its `.catch(() => default)` wrapper, giving a false-green signal.

- `apps/web/e2e/staging.spec.ts` — Programs / History / Cycle tests now carry explanatory comments naming the swallowed-fallback source lines and explain why structure-only assertion is intentional for the staging test user. Auth/API success-path detection remains delegated to test 5 (`/api/health`).
- `apps/api/src/programs/workouts.controller.spec.ts` — adds two paired tests covering the previously untested fallback branches: `getCycleDashboard` rethrows non-`ProgramNotFoundError` errors (asserts the error is not narrowed into the fallback), and `getSkipsForCycle` treats failure as an empty skip set (and logs).
- `docs/standards/error-fallback-test-coverage.md` — new standard codifying the rule (data-level assertion, separate success-path test, or documented structure-only intent) with good/bad examples and enforcement notes. Added as a row to the Coding Standards table in `CLAUDE.md`.
- `CLAUDE.md` `## Testing` — new "Skewed-test rule" subsection pointing at the standard, plus a "Failure-tracking rule" subsection requiring every reported test failure to be fixed in the PR or tracked by an open GitHub issue cited in the body (prose alone no longer acceptable). Global equivalent tracked in [dev-env#281](https://github.com/brownm09/dev-env/issues/281).
- `apps/api/package.json` — adds `"postinstall": "prisma generate"` to fix a worktree-only ambient-failure class (4 API suite-load failures from an ungenerated `@prisma/client`). CI was already clean because the workflow generates the client explicitly; this closes the local/CI parity gap.

Closes #349

## Acceptance criteria

- [x] Audit all existing tests for the skewed pattern — see audit summary in the issue
- [x] For each instance, apply the right fix (data assertion / separate success-path test / documented structure-only)
- [x] Update `CLAUDE.md` `## Testing` with a rule that prevents new skewed tests from landing, backed by `docs/standards/error-fallback-test-coverage.md`

## Out of scope (filed as follow-ups)

- [#353](https://github.com/brownm09/lifting-logbook/issues/353) — Add static-analysis check to detect skewed tests
- [#354](https://github.com/brownm09/lifting-logbook/issues/354) — Audit `packages/core` / `packages/types` for analogous neutral-return patterns
[...]

PR #359, opened 2026-05-27
[feat] Add no-uncovered-error-fallback ESLint rule (closes #353)

### LINKED ISSUE #353: Add static-analysis check to detect skewed tests
## Problem

Issue #349 audits and remediates existing skewed tests (assertions that pass under both the success path and the silent-fallback path of error-swallowing code) and adds a written standard plus a CLAUDE.md rule. Enforcement is currently reviewer-only — a follow-up author can still introduce a `.catch(() => default)` next to a structure-only assertion and the test suite will pass.

## Scope

Build a static check that flags candidate skewed tests at lint or CI time. Approaches to evaluate:

- A custom ESLint rule that detects, in a test file, an assertion against a route/component whose source file contains an error-swallowing fallback, without a paired assertion that the fallback was *not* taken (e.g., a real-data assertion or a sibling test calling a health/data endpoint and asserting HTTP 200).
- A repo-level grep gate in CI that lists every `.catch(() =>` / `?? default` / `try { … } catch { redirect(` in `apps/web/app/**/page.tsx` and `apps/api/src/**` and requires each location to be referenced (by source path or comment) from at least one test file.
- A lighter heuristic: any new `.catch(() =>` in a server component must appear in the same PR diff as a change to a `.spec.ts` file under `apps/web/e2e/` or `apps/api/src/`.

## Acceptance criteria

- [ ] Decide on the approach (custom lint rule vs. CI grep vs. PR-diff heuristic) with a one-paragraph rationale referencing this issue.
- [ ] Implement the check and wire it into `npm run lint` and/or CI.
- [ ] Verify it would have flagged the three pre-#349 instances (`ProgramsPage`, `HistoryPage`, `CyclePage`).
- [ ] Verify it does not flag the post-#349 remediated state.
- [ ] Update `docs/standards/error-fallback-test-coverage.md` Enforcement section to point at the new check.

## Related

- #349 — the audit and standard this follow-up enforces automatically
- PR #346 — original incident that introduced the skewed pattern
- ADR-023 — staging test design and stated limitation

brownm09:
## Summary

Adds a custom ESLint rule that automates the [`docs/standards/error-fallback-test-coverage.md`](docs/standards/error-fallback-test-coverage.md) enforcement, closing the reviewer-only gap left after #349.

For every `.catch(arrow returning a neutral default)` or `try { … } catch { redirect(…) }` in `apps/web/app/**/*.{ts,tsx}` or `apps/api/src/**/*.ts`, the rule requires one of:

1. A test-file comment naming the fallback as `<path>:<line>` or `<path>:<start>-<end>` — matches the pattern #355 adopted in `apps/web/e2e/staging.spec.ts`.
2. `// allow-skewed: <reason>` — documented exception path.
3. `// fallback-covered-by: <path-to-spec-file>` for specs that exercise both branches directly. The rule verifies the file exists on disk.

## Approach choice (acceptance criterion 1)

Custom ESLint rule was chosen over CI grep gate or PR-diff heuristic because it is the only approach that:
- runs locally in `npm run lint` and in editors (precise file:line diagnostics)
- can verify a *named* test file actually references the fallback's `file:line`, not just that *some* test file was touched
- mirrors the existing fetch-cache-semantics enforcement at [eslint.config.js](eslint.config.js)

CI grep gates can't validate that the test reference points at a real source line; PR-diff heuristics catch additions but allow drift on later commits.

## Acceptance criteria

- [x] Decide on the approach with a one-paragraph rationale referencing this issue → above.
- [x] Implement the check and wire it into `npm run lint` and CI → registered in `eslint.config.js` as `lifting-logbook/no-uncovered-error-fallback`; runs via existing `turbo run lint`.
- [x] Verify it would have flagged the three pre-#349 instances → checked out `f5c4ab3` (the commit before #355) and ran `npx eslint` on the three pages. Result: exactly 5 violations — programs:10, programs:11, history:37, history:38, cycle:10.
- [x] Verify it does not flag the post-#349 remediated state → `npx eslint 'apps/web/app/**/*.{ts,tsx}' 'apps/api/src/**/*.ts'` returns 0 violations on this branch.
- [x] Update `docs/standards/error-fallback-test-coverage.md` Enforcement section to point at the new check → replaces the "Future static check" bullet with documentation of the rule and its three annotation forms.

## Annotation additions in this PR

- `apps/api/src/otel.ts:36` — `// allow-skewed: SIGTERM/SIGINT shutdown handler — no business path to assert against.`
- `apps/api/src/programs/workouts.controller.ts` — `// fallback-covered-by: apps/api/src/programs/workouts.controller.spec.ts` on each of the three `.catch()` sites; the spec already exercises both branches per #355.

`WorkoutLogger.tsx:616` is not matched (its `.catch` body has no neutral-return statement, only `console.error` + state restoration — recovery UX, not silent business swallow).

## Testing

`npm run lint` from repo root → **7 successful** tasks (all workspaces lint clean).

Workspace tests excluding `@lifting-logbook/api`: **Tests: 183 passed, 0 skipped, 0 failed (~16s)** — core (145), web (28), eslint-rules RuleTester (10).

`@lifting-logbook/api` build and tests fail in this environment due to a `@types/node`/`@nestjs` module resolution error from running under Node v24 instead of the project-required Node 20.11.1 (`.nvmrc`). Reproduced on a clean stash of this branch (same failure on `origin/main`), so this is pre-existing environmental drift — not caused by this PR. CI runs the correct Node version and will execute the api workspace cleanly.

Suppression grep: clean. Test-integrity grep: clean. No suppressions, skips, deleted tests, or coverage threshold changes added.

## Test plan

- [x] Custom rule RuleTester suite (10 cases) in `tools/eslint-rules/no-uncovered-error-fallback.test.js` covers: bare `.catch()`, test-comment reference, `allow-skewed`, valid `fallback-covered-by`, missing-target `fallback-covered-by`, try/catch/redirect bare and referenced, re-throwing `.catch` not matched, conditional re-throw with neutral default matched, range references.
- [x] AC verification (pre/post #349) documented above.
- [ ] CI runs `npm run lint` via the new rule and confirms zero violations on the merged tree.


brownm09:
## Review: [feat] Add no-uncovered-error-fallback ESLint rule (closes #353)

### Summary

Merge with minor changes. The rule is well-scoped, matches the test-side annotation pattern #355 actually shipped, and the RuleTester suite covers the important branches. The acceptance-criteria verification (5 violations at `f5c4ab3`, 0 on this branch) is concrete and reproducible. Highest-severity finding is non-blocking: the module-level reference cache never invalidates within a single ESLint process, which is surprising in editor / `--fix --watch` workflows. No correctness or security blockers.

---

### Non-Blocking Findings

**[reliability]** `tools/eslint-rules/no-uncovered-error-fallback.js:14-15,53-60`
`cachedReferences` / `cachedRepoRoot` are module-level state that is only rebuilt when the repo root changes. Inside a single ESLint process (IDE integration, `eslint --fix --watch`, an editor plugin that reuses workers), adding a new `<path>:<line>` comment in a staging spec will not be picked up until the process restarts. `npm run lint` exits per invocation so this is invisible there, but the eventual IDE/lint-on-save story will be confusing — a comment that *should* satisfy the rule appears not to, until reload.
**Suggestion:** Either (a) add a brief comment in the rule explaining that the cache is one-shot-per-process and document the IDE-restart workaround near the rule's docstring, or (b) when reporting an `uncovered` violation, do one last `fs.statSync` + diff-against-cache check for the most recently modified test files before reporting. (a) is the cheap fix and matches the next-session-context already in the journal stub.

**[maintainability]** `tools/eslint-rules/` (no README)
A new top-level `tools/` workspace was introduced with one custom rule and three annotation forms. The only place these are documented is `docs/standards/error-fallback-test-coverage.md`. A future contributor adding a second rule under `tools/eslint-rules/` (e.g., when #354 extends scope to `packages/core`) will not find a quick orientation file.
**Suggestion:** Add a short `tools/eslint-rules/README.md` covering: (1) what lives here and why it is a workspace rather than inline in `eslint.config.js`, (2) the three annotation forms with one example each, (3) how to add a new rule (file + register in `index.js` + add to `eslint.config.js`), (4) how to run the local test suite (`npm test -w @lifting-logbook/eslint-rules`). The diff added 273 lines of rule code with no in-repo entry point for it.

**[performance]** `tools/eslint-rules/no-uncovered-error-fallback.js:74-91`
`buildReferenceSet` walks five top-level directories synchronously on first lint invocation and reads every `.spec.ts/.spec.tsx/.test.ts/.test.tsx` in full. Today's scale (~10s of spec files) makes this a few hundred ms at worst. Once the audit (#354) extends to `packages/core` and the test surface grows, the first-file-linted invocation pays the whole scan cost.
**Suggestion:** Not actionable today, but worth a one-line `TODO` near `TEST_GLOB_DIRS` noting the eager-scan tradeoff so the future extender to `packages/*` doesn't accidentally widen it to a full-repo walk. The walker already skips `node_modules`/`dist`/`.turbo`/`.next` — also consider skipping `coverage` and `.next-test`.

**[correctness]** `tools/eslint-rules/no-uncovered-error-fallback.js:102-105`
`isNeutralReturnValue` returns `true` for *any* `Identifier`. That means `.catch(() => buildErrorResponse)` (returning a function reference) or `.catch(() => someCachedFallback)` (returning a precomputed object via identifier) are treated identically to `.catch(() => [])`. This is consistent with the rule's stated intent ("identifier (neutral default)") and is what the conditional-rethrow test case in `workouts.controller.ts` relies on (`return [] as LiftRecord[]` collapsing through `TSAsExpression`), but it is broader than the four documented neutral shapes (literal, array literal, object literal, `undefined`).
**Suggestion:** Either restrict `Identifier` to a named allowlist (`undefined`, `null`, identifiers ending in `Default`/`Fallback`/`Empty`), or update `docs/standards/error-fallback-test-coverage.md` to note that "neutral default" includes any identifier-returning arrow. The current behavior is reasonable but undocumented.

**[correctness]** `tools/eslint-rules/no-uncovered-error-fallback.js:178-194`
`gatherCandidateComments` walks up `node.parent` and unions all `getCommentsBefore` results. For a `.catch()` inside `Promise.all([...])`, that walk reaches the enclosing `VariableDeclaration` / `ExportNamedDeclaration` — so an `// allow-skewed:` comment placed above an *entire* export statement opts out **every** `.catch()` in that statement, not just the one a reader expects. That is technically what the standard says ("immediately before the fallback **or before an ancestor expression / statement**"), but it makes single-statement blast radius easy to under-estimate.
**Suggestion:** Add an example in the standard doc showing this multiplier — specifically, that the three `fallback-covered-by` comments in `workouts.controller.ts` *had* to be placed individually inside the `Promise.all` arrays for granularity, and that a single comment above `const [dashboard, spec] = await Promise.all([...])` would have covered both `.catch()`s in that array. This is the kind of thing a future audit could miss.

**[correctness]** `tools/eslint-rules/no-uncovered-error-fallback.js:62-72`
The reference regex matches anything path-shaped followed by `:digits`. In practice it will pick up strings like `node:test` references, npm package names with extensions in comments, or `error.code:42` in a test fixture. None of these will accidentally match a real `posixRel:line` lookup (path mismatch), so the set just bloats. Memory is bounded by total spec-file size, which is small.
**Suggestion:** Either tighten the regex to require a leading path segment with at least one `/` (current pattern accepts bare `file.ts:1`), or document that the set is intentionally over-broad and the safety is in the lookup side. No code change needed if you document it — a one-line comment near the regex is sufficient.

**[reliability]** `tools/eslint-rules/no-uncovered-error-fallback.js:135-148`
`tryHandlerCallsRedirect` only matches when the catch body calls a bare identifier `redirect()`. If a future page does `import { redirect as nextRedirect }` or `nav.redirect('/x')`, the rule silently skips it. The current single use site in `apps/web/app/cycle/page.tsx` uses bare `redirect`, so this is fine today; the pattern is fragile though.
**Suggestion:** Add an `// allow-skewed:` test fixture exercising an aliased import to confirm the rule behaviour, and decide explicitly: matches by-name regardless of alias, or skips. If matches, broaden to inspect `MemberExpression` callees whose property name is `redirect`. If skips, note the limitation in the standard doc alongside the `?? null / [] / {}` carve-out already documented.

**[maintainability]** `tools/eslint-rules/package.json`
The workspace declares `eslint` as a `devDependency` but the test file also imports `@typescript-eslint/parser` (`const tsParser = require('@typescript-eslint/parser');`). This resolves only via npm-workspace hoisting from the monorepo root. If someone ever runs the tests from inside `tools/eslint-rules/` after a non-hoisted install, they will get `Cannot find module '@typescript-eslint/parser'`.
**Suggestion:** Add `"@typescript-eslint/parser": "^8.53.0"` to the workspace's `devDependencies` to make the dependency explicit, mirroring the root.

---

### Questions for Author

**Question:** Should the existing `// fallback-covered-by:` references in `workouts.controller.ts` validate that the named spec actually exercises both branches, or is filesystem-existence the contract?
**Context:** `tools/eslint-rules/no-uncovered-error-fallback.js:236-243` only does `fs.existsSync(targetAbs)`. The PR body and the standard both say "the spec exercises both branches", but the rule cannot verify that.
**Tradeoffs:** (a) Keep filesystem-only — current state; reviewer is the spirit-of-rule guardian (already documented in the standard). (b) Require the spec to contain a backreference comment (`// covers: apps/api/src/programs/workouts.controller.ts:42`) and assert it exists — turns this into the same mechanism as `<path>:<line>`, eliminating the special case. (c) Parse the spec file for matching `describe`/`it` names containing "fallback" or "rejects" — fragile and brittle. (a) is the explicit current choice and the standard doc backs it up; flagging in case (b) was considered and rejected, since it would have removed the third annotation form entirely.

---

### Style Notes

The rule file has clear function names and short helpers. Two small nits, both optional: (1) `findRepoRoot` carries a multi-line comment block explaining the two-tier heuristic — that comment is exactly the kind of thing that belongs in a `tools/eslint-rules/README.md` "Implementation notes" section (see the maintainability finding above). (2) The `messageId: 'uncovered'` message string repeats the file path twice (once as `{{file}}:{{line}}`, once in the implicit ESLint header). Consider trimming to just the rule explanation since ESLint already prefixes the location.

---

*Reviewed by `claude-opus-4.7` via Claude Code*

PR #360, opened 2026-05-27
[fix] Throw on missing required keys in parseCycleDashboard

### LINKED ISSUE #356: Align parseCycleDashboard validation with parseTrainingMaxes: throw on missing required keys

brownm09:
## Summary

Aligns `parseCycleDashboard` validation with sibling parser `parseTrainingMaxes`: missing or unparseable required keys now throw a descriptive `Error` instead of yielding sentinel values (`""`, `NaN`, `Invalid Date`) that pushed failures downstream into renderers.

Replaces the pinned-current-behavior test added in #354 with throw assertions covering empty input, a missing `Cycle #` key, and an unparseable `Cycle Date`.

## Caller audit

`parseCycleDashboard` has one production caller in the active codebase (`src/api/repositories/CycleDashboardRepository.ts`, the legacy Apps Script repo). It returns the parsed object directly with no try/catch — a throw propagates naturally. `apps/api` uses the Prisma/in-memory repositories and does not call this parser directly. No caller depends on sentinel values.

## `currentWeekType` decision

Left in the parser as the hardcoded `'training'` placeholder. The api controller already overwrites it via `weekTypeForDate(...)` at request time. Removing it from the parser would require making it optional in `CycleDashboard` (`packages/core/src/models/CycleDashboard.ts`) and `packages/types/src/api.ts`, with downstream churn across fixtures, the Prisma schema, and seed scripts — broader than this issue and worth its own ticket.

## Acceptance criteria

- [x] `parseCycleDashboard` throws a descriptive error when any required key is missing or unparseable.
- [x] Test added in #354 replaced with throw assertions (3 tests: empty input, missing `Cycle #`, unparseable `Cycle Date`).
- [x] All callers handle or propagate the new throw appropriately.

## Testing

```
npm test -w @lifting-logbook/core
```

Results:
- `@lifting-logbook/core`: Tests: 157 passed, 0 skipped, 0 failed (8.5s)
- `@lifting-logbook/api`: Tests: 250 passed, 51 skipped, 0 failed (11.1s)
- `@lifting-logbook/web`: Tests: 28 passed, 0 skipped, 0 failed (9.0s)

The 51 skipped tests in the api workspace are pre-existing (confirmed via `git stash` against `main`) and unrelated to this change.

Suppression check: clean. Test-integrity check: clean. No deleted test files, no lowered thresholds.

### Coverage

This is a bug fix with regression coverage: the existing happy-path test still asserts the parsed shape; three new tests assert the throw paths (empty input, missing required key, unparseable date) — satisfying the coverage rule for bug fixes (regression test that fails before the fix and passes after).

## Test plan

- [x] `npm test -w @lifting-logbook/core` passes
- [x] `npm test -w @lifting-logbook/api` passes
- [x] `npm test -w @lifting-logbook/web` passes
- [x] Pre-PR suppression grep clean
- [x] Pre-PR test-integrity grep clean

Closes #356

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2590,11 +2590,12 @@
       }
     },
     "node_modules/@clerk/backend": {
-      "version": "3.4.13",
-      "resolved": "https://registry.npmjs.org/@clerk/backend/-/backend-3.4.13.tgz",
-      "integrity": "sha512-3BABnKE1YZpQqJ/S8QD5FpzE7jq+mgaMrO7rLiWTI8Bfs/xk11XYSp1lMEf3BTo/rzEtaUsXaVDGvccYogKapg==",
+      "version": "3.4.14",
+      "resolved": "https://registry.npmjs.org/@clerk/backend/-/backend-3.4.14.tgz",
+      "integrity": "sha512-0iaMT7k4wDk31QVC3HMaoeVFttblwsCECTHKNQpbRzIyD8j2gHdKEw/FNjffoyqyBqPw869IQlk1YokUlwVAqQ==",
+      "license": "MIT",
       "dependencies": {
-        "@clerk/shared": "^4.13.1",
+        "@clerk/shared": "^4.14.0",
         "standardwebhooks": "^1.0.0",
         "tslib": "2.8.1"
       },
@@ -2603,10 +2604,11 @@
       }
     },
     "node_modules/@clerk/backend/node_modules/@clerk/shared": {
-      "version": "4.13.1",
-      "resolved": "https://registry.npmjs.org/@clerk/shared/-/shared-4.13.1.tgz",
-      "integrity": "sha512-DyUtvNHgMmqjtTM0q285jKaAXUmCDSyItiGQTt1dNL0M6DZ3bxqsJz7wXPjh9zezmU4BAnLpwhj5gsM3OuNPzA==",
+      "version": "4.14.0",
+      "resolved": "https://registry.npmjs.org/@clerk/shared/-/shared-4.14.0.tgz",
+      "integrity": "sha512-StZCFJ2rg0ITE3fYjIN9CKuKq8ZACW6l2+5qGCFPPYd4hZphA+VDselmh/lHPsF1ex/WRVUSHvquykAHR8cQbQ==",
       "hasInstallScript": true,
+      "license": "MIT",
       "dependencies": {
         "@tanstack/query-core": "^5.100.6",
         "dequal": "2.0.3",
--- a/packages/core/src/utils/parser/parseCycleDashboard.ts
+++ b/packages/core/src/utils/parser/parseCycleDashboard.ts
@@ -23,13 +23,44 @@ export function parseCycleDashboard(data: SpreadsheetCell[][]): CycleDashboard {
   data.forEach(([key, value]) => {
     map[String(key)] = value;
   });
+
+  const program = String(map[PROGRAM_KEY] ?? "");
+  const cycleUnit = String(map[CYCLE_UNIT_KEY] ?? "");
+  const cycleNum = Number(map[CYCLE_NUM_KEY]);
+  const cycleDate = new Date(String(map[CYCLE_DATE_KEY] ?? ""));
+  const sheetName = String(map[SHEET_NAME_KEY] ?? "");
+  const cycleStartWeekday = toTitleCase(
+    String(map[CYCLE_START_WEEKDAY_KEY] ?? ""),
+  ) as Weekday;
+
+  if (program.length === 0) {
+    throw new Error(`Invalid ${PROGRAM_KEY} value: ${String(map[PROGRAM_KEY])}`);
+  }
+  if (cycleUnit.length === 0) {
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: [fix] Throw on missing required keys in parseCycleDashboard

### Summary
Merge with minor changes. Tight, well-scoped bug fix that mirrors the sibling `parseTrainingMaxes` validation pattern. Tests, PR body, and caller audit are all in order. One non-blocking edge case in `cycleNum` validation and one non-blocking gap in `cycleStartWeekday` validation are worth a follow-up but do not block merge.

---

### Non-Blocking Findings

**[correctness]** `packages/core/src/utils/parser/parseCycleDashboard.ts:29`
`cycleNum` uses `Number(map[CYCLE_NUM_KEY])` then `isNaN(cycleNum)`. `Number(undefined)` is `NaN` (caught), but `Number("")` is `0` and `Number(" ")` is `0` — so an empty-string or whitespace-only `Cycle #` cell passes validation and yields `cycleNum: 0`, which is not a valid cycle number. The sibling `parseTrainingMaxes` has the same shape, so this is consistent — but downstream code now sees `0` instead of a clear error, which is exactly the class of sentinel-pushed-downstream failure this PR set out to eliminate.
**Suggestion:** Tighten to `if (!Number.isFinite(cycleNum) || cycleNum <= 0)` (or add a separate empty-string check before `Number(...)`). Worth filing as a follow-up issue tied to both parsers, since the same hardening applies to `parseTrainingMaxes`.

---

**[correctness]** `packages/core/src/utils/parser/parseCycleDashboard.ts:51`
`cycleStartWeekday` is cast to `Weekday` after only a length check on the title-cased string. A value like `"Funday"` passes the `length === 0` gate and is cast to `Weekday`, propagating an invalid enum into the model. This mirrors prior behavior and is out of scope for #356 strictly, but the PR's stated goal — fail fast on unparseable values — would naturally cover it.
**Suggestion:** Validate against the `Weekday` enum values (`Object.values(Weekday).includes(cycleStartWeekday)`) and throw the same `Invalid <key> value` shape. Defer to a follow-up if you want to keep this PR scoped to the audit finding.

---

**[maintainability]** `packages/core/src/utils/parser/parseCycleDashboard.ts:36-55`
Six near-identical `if (...) throw new Error(\`Invalid ${KEY} value: ${String(map[KEY])}\`)` blocks. Readable as-is at six keys, but if either parser grows new required fields, the duplication will compound. The sibling `parseTrainingMaxes` has the same pattern, so a small shared helper (e.g., `assertRequired(map, key, validator)`) could be extracted once and reused.
**Suggestion:** Not for this PR. If `parseTrainingMaxes` and this file diverge again in the future, consider extracting a shared validator.

---

### Questions for Author

**Question:** Should the `currentWeekType` placeholder be tracked in an issue before this PR merges, or is filing it later acceptable?
**Context:** The PR body and the journal stub both note that `currentWeekType: 'training' as WeekType` at [parseCycleDashboard.ts:65](packages/core/src/utils/parser/parseCycleDashboard.ts:65) is a placeholder overwritten by `weekTypeForDate(...)` in the controller, with removal deferred. No issue number is referenced for the follow-up.
**Tradeoffs:** (a) File the follow-up issue now and reference it inline with the comment at line 64 — keeps the deferral discoverable from the code. (b) Leave as-is — the PR body documents the decision but the placeholder will outlive memory of the rationale. (c) Inline the controller logic into the parser now — out of scope for #356 and broader than the issue warrants.

---

PR #362, opened 2026-05-27
[refactor] Remove currentWeekType placeholder from parser / CycleDashboard model

### LINKED ISSUE #361: Remove currentWeekType placeholder from parseCycleDashboard / CycleDashboard model
## Context

Surfaced during the `/review` of PR #360 (issue #356).

`parseCycleDashboard` currently emits `currentWeekType: 'training' as WeekType` as a hardcoded placeholder. The api controller (`apps/api/src/programs/cycle-dashboard.controller.ts:28`) immediately overwrites it via `weekTypeForDate(dashboard.cycleDate, programSpec)` at request time, so the value the parser emits is never read in the api path. It exists only because `CycleDashboard` (the core model) requires the field.

The placeholder is misleading: a future reader of the parser will reasonably believe the dashboard sheet stores a week type, when in fact the week type is derived from the program spec and date at request time.

## Proposed change

Make `currentWeekType` optional on the core `CycleDashboard` model (or move it off the model entirely, since it isn't a property of the dashboard sheet) and stop emitting it from the parser.

## Scope (why this is its own issue)

Touches at least:

- `packages/core/src/models/CycleDashboard.ts`
- `packages/core/src/utils/parser/parseCycleDashboard.ts`
- `packages/types/src/api.ts` (`CycleDashboardResponse` keeps the field on the response, but the internal model may diverge)
- `apps/api/prisma/schema.prisma` (column on the dashboard table)
- `apps/api/prisma/seed.ts`
- `apps/api/src/adapters/prisma/cycle-dashboard.repository.ts`
- `apps/api/src/adapters/in-memory/fixtures.ts`
- `apps/api/src/programs/cycle-dashboard.controller.ts` (still assigns the derived value before building the response)
- `apps/api/src/programs/cycle-generation.service.ts` and its spec (writes the field)
- `apps/api/src/programs/mappers.ts` and spec
- `apps/api/src/programs/cycle-dashboard.controller.spec.ts`, `workouts.controller.spec.ts`, `lift-records.controller.spec.ts`, `cycle-generation.controller.spec.ts`, `programs.db.e2e.spec.ts`
- `apps/web/e2e/mock-api.mjs`
- `packages/core/tests/core/services/dashboard/updateCycle.test.ts` and the mapper test
- `docs/proposals/2026-04-30-initial-training-max-discovery.md` and `docs/adr/ADR-017-training-max-history-table.md` (text references)

A Prisma migration is also required if the field is removed from the dashboard table.

## Acceptance criteria

- [ ] `parseCycleDashboard` no longer emits `currentWeekType`.
- [ ] Either the `CycleDashboard` core model drops the field, or it is made optional and clearly documented as "derived at request time, not stored in the sheet".
- [ ] `cycle-dashboard.controller.ts` still produces a correct `CycleDashboardResponse` with `currentWeekType` set from `weekTypeForDate(...)`.
- [ ] Prisma migration generated and the schema updated if the field is removed at the storage layer (alternative: keep storing it for legacy reasons, document why).
- [ ] All tests and fixtures updated.

## Related

- Predecessor: #356 (this followup was deferred from PR #360 to keep that PR scoped).
- Audit chain: #354 → #356 → this issue.

brownm09:
## Summary

- `parseCycleDashboard` no longer emits the hardcoded `currentWeekType: 'training'` placeholder; the field is now optional on `CycleDashboard` with a comment that it is derived at request time via `weekTypeForDate(date, programSpec)`, not stored on the sheet.
- Drop the `currentWeekType` column from `cycle_dashboard` via Prisma migration. Nothing in production ever wrote `'test'` to it, so the only read that branched on its value (`CycleGenerationService.startNewCycle`) was effectively dead; that read is replaced with `weekTypeForDate(...)` against the program spec.
- Refactor mappers (`toCycleDashboardResponse`, `buildCycleDashboardResponse`) to take an explicit `currentWeekType: WeekType` parameter rather than relying on callers mutating the dashboard before mapping. `CycleDashboardResponse.currentWeekType` stays required on the API contract; the web UI is unaffected.

Audit chain: #354 → #356 → this PR. Surfaced during `/review` of #360.

Closes #361

## Acceptance Criteria

- [x] `parseCycleDashboard` no longer emits `currentWeekType`.
- [x] `CycleDashboard.currentWeekType` is optional with a doc comment that it is derived at request time.
- [x] `cycle-dashboard.controller.ts` still produces a correct `CycleDashboardResponse` with `currentWeekType` from `weekTypeForDate(...)`.
- [x] Prisma migration drops the column; schema and adapter updated.
- [x] Fixtures, tests, ADR-017, and the proposal updated.

## Test plan

- [x] `npm test` (via `turbo run test --force`): all workspaces pass — `Tests: 250 passed, 51 skipped, 0 failed (11.094 s)` for `@lifting-logbook/api`; `155 passed, 0 skipped, 0 failed` for `@lifting-logbook/core`; `28 passed` for `@lifting-logbook/web`; `0 tests` for `@lifting-logbook/types` and `@lifting-logbook/mobile`. The skipped count in api is pre-existing (DB-backed e2e suite gated on `DATABASE_URL`).
- [x] `nest build` passes for `apps/api`.
- [x] Pre-PR suppression grep: no new suppressions introduced. Matches are context lines around unchanged `?? null` and a removed `!` in `seed.ts`.
- [x] Pre-PR test-integrity grep: clean. No deleted tests, no lowered thresholds, no new skips.
- [ ] DB-backed e2e (`programs.db.e2e.spec.ts`) requires `DATABASE_URL`; pre-existing skipped in this environment — covered by CI.
- [ ] Manual smoke test against a running api: `GET /programs/:program/cycles/current` returns a correct `currentWeekType` derived from `weekTypeForDate(...)`. (Deferred — staging deploy on merge.)

### DIFF
--- a/apps/api/prisma/migrations/20260527000000_remove_cycle_dashboard_current_week_type/migration.sql
+++ b/apps/api/prisma/migrations/20260527000000_remove_cycle_dashboard_current_week_type/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "cycle_dashboard" DROP COLUMN "currentWeekType";
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -49,7 +49,6 @@ model CycleDashboard {
   cycleDate         DateTime
   sheetName         String
   cycleStartWeekday String
-  currentWeekType   String
   programType       String?
 
   @@unique([userId, program])
--- a/apps/api/prisma/seed.ts
+++ b/apps/api/prisma/seed.ts
@@ -170,13 +170,11 @@ async function main() {
 
   // CycleDashboard — current cycle state
   const currentCycle = NUM_CYCLES;
-  const currentWeek = WEEK_TYPES[(currentCycle - 1) % 3]!;
   await prisma.cycleDashboard.upsert({
     where: { userId_program: { userId: SEED_USER_ID, program: PROGRAM } },
     update: {
       cycleNum: currentCycle,
       cycleDate: now,
-      currentWeekType: currentWeek.weekType,
     },
     create: {
       userId: SEED_USER_ID,
@@ -186,7 +184,6 @@ async function main() {
       cycleDate: now,
       sheetName: '531',
       cycleStartWeekday: 'Monday',
-      currentWeekType: currentWeek.weekType,
       programType: '531',
     },
   });
--- a/apps/api/src/adapters/in-memory/fixtures.ts
+++ b/apps/api/src/adapters/in-memory/fixtures.ts
@@ -21,7 +21,6 @@ export const seedCycleDashboard = (): CycleDashboard => ({
   cycleDate: new Date('2026-04-20T00:00:00.000Z'),
   sheetName: '',
   cycleStartWeekday: Weekday.Monday,
-  currentWeekType: 'training',
   programType: '5-3-1',
 });
 
--- a/apps/api/src/adapters/prisma/cycle-dashboard.repository.ts
+++ b/apps/api/src/adapters/prisma/cycle-dashboard.repository.ts
@@ -21,7 +21,6 @@ export class PrismaCycleDashboardRepository implements ICycleDashboardRepository
       cycleDate: row.cycleDate,
       sheetName: row.sheetName,
       cycleStartWeekday: row.cycleStartWeekday as Weekday,
-      currentWeekType: row.currentWeekType as CycleDashboard['currentWeekType'],
       ...(row.programType !== null && { programType: row.programType }),
     };
   }
@@ -37,7 +36,6 @@ export class PrismaCycleDashboardRepository implements ICycleDashboardRepository
         cycleDate: dashboard.cycleDate,
         sheetName: dashboard.sheetName,
         cycleStartWeekday: dashboard.cycleStartWeekday,
-        currentWeekType: dashboard.currentWeekType,
         programType: dashboard.programType ?? null,
       },
       update: {
@@ -46,7 +44,6 @@ export class PrismaCycleDashboardRepository implements ICycleDashboardRepository
         cycleDate: dashboard.cycleDate,
         sheetName: dashboard.sheetName,
         cycleStartWeekday: dashboard.cycleStartWeekday,
-        currentWeekType: dashboard.currentWeekType,
         programType: dashboard.programType ?? null,
       },
     });
--- a/apps/api/src/programs/cycle-dashboard.controller.spec.ts
+++ b/apps/api/src/programs/cycle-dashboard.controller.spec.ts
@@ -12,15 +12,13 @@ import { CycleDashboardController } from './cycle-dashboard.controller';
 
 const MOCK_USER = { id: 'test-user', email: 'test@example.com', provider: 'dev' };
 
-const stubDashboard = (overrides: Partial<{ currentWeekType: 'training' | 'test' | 'deload' }> = {}) => ({
+const stubDashboard = () => ({
   program: '5-3-1',
   cycleUnit: 'week' as const,
   cycleNum: 2,
   cycleDate: new Date('2026-04-20T00:00:00.000Z'),
   sheetName: '',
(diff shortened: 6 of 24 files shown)


brownm09:
## Review: [refactor] Remove currentWeekType placeholder from parser / CycleDashboard model

### Summary
Clean, tightly scoped refactor. Removes a dead-but-misleading placeholder field, drops the Prisma column that mirrored it, and routes the (already-derived) value through an explicit mapper parameter instead of caller-side mutation. Tests, fixtures, ADR-017, and the originating proposal are all updated in lockstep. No blocking findings. Two non-blocking notes around deploy ordering of the destructive migration and a duplicate `getProgramSpec` read introduced in `cycle-generation.controller.ts`.

---

### Non-Blocking Findings

**[reliability]** `apps/api/prisma/migrations/20260527000000_remove_cycle_dashboard_current_week_type/migration.sql:2`
`ALTER TABLE "cycle_dashboard" DROP COLUMN "currentWeekType"` is destructive and runs in a single step with no app-version coordination. Once the migration runs, any pod still running the previous image will fail on every write to `cycleDashboard.upsert` — the old client passes `currentWeekType: ...` to Prisma, which will now reject it as an unknown column. If the deploy pipeline runs `prisma migrate deploy` before all old API replicas are drained, request errors will surface during the rolling deploy window.
**Suggestion:** Confirm the deploy strategy (single-replica or maintenance-window) makes this safe, or note explicitly in the PR body that brief 5xx during rollout is acceptable. The standard two-step alternative (deploy new code that ignores the column first; drop the column in a follow-up PR) is the safer pattern when it matters, but probably overkill for this app's traffic profile — worth a deliberate decision either way.

---

**[performance]** `apps/api/src/programs/cycle-generation.controller.ts:34-35,47-48`
Both `initializeFirstCycle` and `startNewCycle` now call `repos.liftingProgramSpec.getProgramSpec(program)` *after* the service has already returned. The service implementations load the same program spec internally to do their work (and `startNewCycle` definitely does — it's how it computes the next cycle). The controller now re-fetches it to derive `currentWeekType`. In the Prisma adapter this is an extra DB round trip per write request on a hot path.
**Suggestion:** Either have the service return `{ dashboard, programSpec }` (or `{ dashboard, currentWeekType }`) so the controller doesn't re-query, or add a small cache on the repo. The cleanest variant matches what `cycle-dashboard.controller.ts:25-27` already does — load both in parallel — but here the service already has the spec in hand, so returning it is the cheapest fix.

---

**[maintainability]** `apps/api/src/programs/cycle-generation.service.ts:163`
`const source = weekTypeForDate(dashboard.cycleDate, programSpec) === 'test' ? 'test' : 'program';` — `dashboard` here is the *previous* cycle's dashboard (not `newCycle`). The semantic ("source is 'test' when the cycle we just finished was a test week") is the same as before, but the previous code read it from a field on the dashboard, so the relationship was implicit. With the derived call, a future reader might reasonably wonder whether `dashboard.cycleDate` or `newCycle.cycleDate` is correct here.
**Suggestion:** Add a one-line comment naming the invariant: `// Source reflects the week type of the cycle being closed, not the new one — hence dashboard, not newCycle.`

---

### Questions for Author

**Question:** Was the `currentWeekType` column ever observed to hold a value other than `'training'` (or the placeholder default) in production?
**Context:** The PR description states "nothing in production ever wrote `'test'` to it," which the diff confirms for current code, but `prisma/seed.ts` was previously writing the rotating `WEEK_TYPES[(currentCycle - 1) % 3]` value, so a dev DB could have `'test'` or `'deload'` rows that the new code never reads. The migration drops the column outright, so any informational value those rows held is gone.
**Tradeoffs:** (a) Drop now (current PR) — simplest; fine if the column truly carried no signal. (b) Backfill into `training_max_history` first — preserves audit history if any prior cycle was recorded with `source = 'test'` by virtue of the now-removed read in `startNewCycle`. (c) Leave the column for one release as a soft-deprecation — buys time to migrate consumers you may not have inventoried. Worth a quick sanity check on staging/prod data before merging.

---

Issue #363, opened 2026-05-28
[fix] StepProgram.test.tsx fails: jest-dom matchers not registered


brownm09:
## Problem

\`apps/web/app/onboarding/steps/StepProgram.test.tsx\` fails to compile with errors like:

\`\`\`
Property 'toBeInTheDocument' does not exist on type 'JestMatchers<HTMLElement>'.
Property 'toBeDisabled' does not exist on type 'JestMatchers<HTMLElement>'.
Property 'toHaveTextContent' does not exist on type 'JestMatchers<HTMLElement>'.
\`\`\`

The test uses \`@testing-library/jest-dom\` matchers but the type augmentation is not picked up by the test's TypeScript compilation. Other web tests (e.g., \`programPlan.test.ts\`, \`workoutPlan.test.ts\`) pass — this test is the only one in \`apps/web\` that uses jest-dom matchers in this form.

## Repro

\`\`\`bash
npm test -w @lifting-logbook/web
\`\`\`

## Likely fix

Either add a \`jest.setup.ts\` that imports \`@testing-library/jest-dom\` and register it in jest config, or import \`@testing-library/jest-dom/jest-globals\` at the top of the test file. Verify \`@testing-library/jest-dom\` is in devDependencies.

## Scope

Pre-existing — not introduced by recent work. Surfaced during PR for #343.

Issue #343, opened 2026-05-24
[fix] Replace seed.ts TypeScript suppressions with proper fixes


brownm09:
## Problem

PR #338 added 6 TypeScript suppressions to \`apps/api/prisma/seed.ts\` to satisfy strict-mode compiler errors rather than fixing the underlying code. Under the new suppression policy (introduced in dev-env), all suppressions must either be replaced with a proper fix or have a documented justification for why a fix is not appropriate.

### \`noUncheckedIndexedAccess\` violations — non-null assertions

These five assertions tell the compiler "trust me, this index is in bounds" without the compiler being able to verify it:

| Line | Code | Rule violated |
|------|------|---------------|
| 96 | \`WEEK_TYPES[cycle % 3]!\` | \`noUncheckedIndexedAccess\` |
| 127 | \`WORKOUT_LIFTS[workoutNum]!\` | \`noUncheckedIndexedAccess\` |
| 130 | \`LIFTS.find((l) => l.name === liftName)!\` | \`noUncheckedIndexedAccess\` / \`.find()\` returns \`T | undefined\` |
| 134 | \`weekType.pct[setNum - 1]!\` | \`noUncheckedIndexedAccess\` |
| 173 | \`WEEK_TYPES[(currentCycle - 1) % 3]!\` | \`noUncheckedIndexedAccess\` |

### \`exactOptionalPropertyTypes\` / index signature — nullish coalescing

| Line | Code | Rule violated |
|------|------|---------------|
| 213 | \`GOAL_RATIOS[lift.name] ?? null\` | Index signature returns \`T | undefined\`; column expects \`number | null\` |

## Acceptance criteria

For each suppression, choose one of:

- **Proper fix**: restructure the code so TypeScript can prove the access is safe (e.g., explicit bounds check that throws or skips, typed tuple/record instead of plain array/object, \`.find()\` result checked before use)
- **Justified retention**: if the suppression is genuinely the right call (e.g., the bounds are provably safe by construction and a runtime check would add noise), replace the bare \`!\` with an inline comment stating the invariant, e.g.:
  \`\`\`ts
  // cycle % 3 is always 0, 1, or 2 — same length as WEEK_TYPES tuple
  const weekType = WEEK_TYPES[cycle % 3] as (typeof WEEK_TYPES)[number];
  \`\`\`

No suppression should be silently retained. Every case must have a decision.

PR #364, opened 2026-05-28
[fix] Replace seed.ts TypeScript suppressions with proper fixes

brownm09:
## Summary

Replaces the 5 remaining TypeScript suppressions in `apps/api/prisma/seed.ts` with guarded lookups and a `Map`-based goal ratio table.

- **Lines 96, 127, 130, 134** — bounded `!` assertions replaced with `if (!x) throw new Error(...)` guards. Throwing gives clearer diagnostics than `Cannot read property of undefined` if a future edit ever violates the invariant.
- **Line 210** — `GOAL_RATIOS` converted from `Record<string, number>` to `Map<string, number>`. `Map.get()` natively returns `T | undefined`, so the trailing `?? null` is a legitimate semantic coercion to the nullable DB column, not a suppression of an index-signature warning.

Note: the issue listed 6 suppressions, but the line-173 `WEEK_TYPES[(currentCycle - 1) % 3]!` was already removed in commit `0c0aaba` (PR #362) when the `currentWeekType` placeholder was dropped.

## Acceptance criteria

- [x] Every suppression in `seed.ts` either fixed or justified — all 5 replaced with proper fixes
- [x] No new suppressions introduced
- [x] `seed.ts` typechecks under strict + `noUncheckedIndexedAccess` + `exactOptionalPropertyTypes`

## Testing

Standalone strict typecheck of `seed.ts` (clean):

```
npx tsc --noEmit --strict --noUncheckedIndexedAccess --exactOptionalPropertyTypes \
  --esModuleInterop --target ES2020 --module commonjs --moduleResolution node \
  --skipLibCheck apps/api/prisma/seed.ts
```

Full suite (`npm test`):

- `@lifting-logbook/core`: Tests: 160 passed, 0 skipped, 0 failed (~52s)
- `@lifting-logbook/api`: passed
- `@lifting-logbook/web`: Tests: 23 passed, 0 skipped, 1 suite failed to compile — `StepProgram.test.tsx` jest-dom matcher registration issue, **pre-existing** (not touched by this PR). Tracked in #363.

Tests: 183+ passed, 0 skipped, 1 suite-load failure (pre-existing, tracked #363).

## Suppression policy check

```bash
git diff origin/main -- apps/api/prisma/seed.ts | grep -E '(ts-ignore|ts-expect-error|eslint-disable|![.[]|!\s*[;,)]|\?\? null)'
```

Yields a single `?? null` match on the `GOAL_RATIOS.get(...)` line. Justification: `Map.get()` returns `T | undefined` natively, so this is type coercion at a DB nullable-column boundary, not silencing a strict-mode error. Net change: 5 suppressions removed, 0 added.

## Test integrity check

No skip markers added, no tests deleted, no coverage thresholds changed.

### DIFF
--- a/apps/api/prisma/seed.ts
+++ b/apps/api/prisma/seed.ts
@@ -135,6 +135,8 @@ async function main() {
 
         for (let setNum = 1; setNum <= 3; setNum++) {
           const pct = weekType.pct[setNum - 1];
+          // Explicit undefined check — 0 would be a valid (if degenerate) percentage,
+          // so !pct would be incorrect here even though the other guards in this file use it.
           if (pct === undefined) throw new Error(`weekType.pct missing index ${setNum - 1}`);
           const weight = weightForSet(max, pct);
           const isAmrap = setNum === 3;
--- a/apps/api/prisma/seed.ts
+++ b/apps/api/prisma/seed.ts
@@ -93,7 +93,8 @@ async function main() {
   const workoutDate = new Date(startDate);
 
   for (let cycle = 0; cycle < NUM_CYCLES; cycle++) {
-    const weekType = WEEK_TYPES[cycle % 3]!;
+    const weekType = WEEK_TYPES[cycle % 3];
+    if (!weekType) throw new Error(`WEEK_TYPES missing index ${cycle % 3}`);
     const cycleNum = cycle + 1;
 
     // TrainingMaxHistory — one entry per lift at cycle start (representing a PR test)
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [fix] Replace seed.ts TypeScript suppressions with proper fixes

### Summary

Merge. Tight, well-scoped 17/13 diff replacing five `!` suppressions with explicit guards and converting `GOAL_RATIOS` from `Record` to `Map` so the `?? null` becomes a legitimate `T | undefined → T | null` coercion at the DB-nullable boundary. PR body documents the suppression-policy check, the test-integrity check, the pre-existing #363 link, and the scope discrepancy with issue #343 (line-173 already removed in PR #362). No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/prisma/seed.ts:138`
`if (pct === undefined)` correctly uses strict undefined comparison rather than `if (!pct)`, since `0` would be a valid (if degenerate) percentage. The other three guards use `if (!x)` because they narrow object/array references where falsy-non-undefined values aren't reachable from the data definitions. The asymmetry is correct but easy to misread as inconsistent on a quick scan.
**Suggestion:** Optional one-line comment on the `pct` guard noting "explicit undefined check — 0 is a valid percentage" so future readers don't normalize the four guards toward a single style.

**[maintainability]** `apps/api/prisma/seed.ts:96,129,132,138`
All four guards encode invariants that hold by construction of the surrounding data tables (`WEEK_TYPES` length 3, `WORKOUT_LIFTS` keys 1–3, `LIFTS` covers every name referenced by `WORKOUT_LIFTS`, `pct` length 3). The throws are defensive belt-and-braces, which is the right call for a one-shot staging seed — but they will never fire under the current data, so they're untested.
**Suggestion:** None required. Worth a sentence in the PR body (already mostly there) noting these guards are diagnostic-only and the invariants live in the data definitions above, so reviewers don't waste cycles looking for a test that would exercise them.

---

Issue #366, opened 2026-05-28
[fix] apps/api build fails on canonical repo: Prisma InputJsonValue not exported (stale generated client)


brownm09:
## Problem

Running \`npm test\` from the canonical repo (\`C:/Users/brown/Git/lifting-logbook\`) fails at \`@lifting-logbook/api:build\`:

\`\`\`
src/user-settings/user-settings.repository.ts:57:75 - error TS2694: Namespace ... '.Prisma' has no exported member 'InputJsonValue'.
\`\`\`

11 errors total, all centered on missing exports from the generated Prisma client.

## Likely cause

Generated Prisma client is stale or was generated against a different prisma version than the consuming code expects. A \`npx prisma generate\` in \`apps/api\` likely fixes it, but should be wired so it runs automatically (e.g. as a \`postinstall\` step) — similar to the lesson from #355 / #349.

## Acceptance

- \`npm test\` runs cleanly on a fresh \`npm install\` of the canonical repo without manual \`prisma generate\` invocation.

## Discovered while working on

Issue #358 (staging deploy timing). Workflow-only change; not blocking the PR but cited per the failure-tracking rule.

Issue #365, opened 2026-05-28
[fix] Worktree npm install leaves jest unable to resolve ts-jest preset / jest-circus runner


brownm09:
## Problem

Running \`npm test\` from a fresh worktree fails for \`@lifting-logbook/web\`, \`@lifting-logbook/types\`, and \`@lifting-logbook/mobile\`:

\`\`\`
@lifting-logbook/types:test:   Preset ts-jest not found relative to rootDir
@lifting-logbook/web:test:     Preset ts-jest not found relative to rootDir
@lifting-logbook/mobile:test:  Module .../node_modules/jest-circus/build/runner.js in the testRunner option was not found.
\`\`\`

The canonical repo at \`C:/Users/brown/Git/lifting-logbook\` does not exhibit these failures (it fails differently — see #companion issue for the Prisma build failure). The worktree's \`node_modules\` install appears to omit or misplace these jest dependencies on Windows.

## Repro

1. Create a new worktree under \`.claude/worktrees/<name>\`.
2. \`npm install\` (completes successfully).
3. \`npm test\` (with TURBO_BINARY_PATH workaround for #similar) — fails as above.

## Acceptance

- Identify whether \`ts-jest\` and \`jest-circus\` should be promoted to root \`devDependencies\`, or whether each affected workspace needs its own direct dep.
- Validate \`npm test\` runs cleanly in a fresh worktree.

## Discovered while working on

Issue #358 (staging deploy timing). Workflow-only change; not blocking the PR but cited per the failure-tracking rule.

brownm09:
Verified resolved by #374 (commit `fc3e052`, bumped `ts-jest` to `^29.4.11`).

Fresh worktree under `.claude/worktrees/`, clean `rm -rf node_modules && npm install`, Node 24.13.0 on Windows 11:

- `npm test -w @lifting-logbook/types` → exit 0, `No tests found` (`--passWithNoTests`)
- `npm test -w @lifting-logbook/mobile` → exit 0, `No tests found` (`--passWithNoTests`)
- `npm test -w @lifting-logbook/web` → **Tests: 28 passed, 0 skipped, 0 failed (8.855 s)** across 3 suites including `StepProgram.test.tsx`

[...]

PR #367, opened 2026-05-28
[infra] Skip image rebuild on non-image diffs; parallelize api/web deploys

### LINKED ISSUE #358: [infra] Staging deploy takes ~21 min end-to-end; cut to under 10 on test/docs-only PRs
## Problem

Staging deploys on every PR push take ~21 minutes end-to-end (observed on PR #355 staging run: 17:50:51Z start → 18:12:20Z Deploy job complete; PR #357 currently in flight on the same path).

This sits at the upper end of the normal band for a Cloud Run + GKE Autopilot + Helm + Playwright stack (industry reference: 15–25 min), but with multiple PR pushes per session the wait time becomes the dominant cost on small/test-only changes.

## Job breakdown (from PR #357 timings)

| Job | Duration | Notes |
|---|---|---|
| Preflight | 3s | fine |
| Lint & Test | 1m10s | fine |
| DB Integration Tests | 1m20s | fine |
| Observability Smoke | 5s | fine |
| Playwright E2E (preview) | 1m49s | fine |
| Terraform (staging) | 30s | fine — nothing to do most runs |
| Build & Push Images | 4m44s | rebuilt unconditionally even on test-only diffs |
| Deploy (staging) | ~15m | bulk of the wait |

### 1. Image rebuild on every push regardless of diff scope

`Build & Push Images` runs on every PR push, including test-only changes that touch no Dockerized code. PR #357 added 4 test files and edited 2 markdown files — zero changes under `apps/api/src`, `apps/web/src`, or any other build input — yet still paid 4m44s for image rebuild.

**Fix:** Path-filter the Build & Push Images job (or its individual matrix entries) to skip when the diff touches only `packages/*/tests/**`, `*.md`, `docs/**`, and similar non-build paths. Existing image tags could be reused from the previous run's SHA via `gh run view`, or the deploy job could resolve `${{ github.event.pull_request.head.sha }}` → previous build SHA when no image-affecting paths changed.

**Estimated saving:** 5 min on test-only / docs-only PRs (which is a non-trivial fraction of all PRs).

### 2. Sequential `gcloud run deploy api` then `gcloud run deploy web`

`staging.yml:362-369` (deploy-api) and `staging.yml:389-406` (deploy-web) run sequentially in the same job. They share no outputs — `deploy-web` only needs the Terraform-emitted API URL, not the API revision. Each `gcloud run deploy` takes 1–2 min.

**Fix:** Split into two jobs (`deploy-api`, `deploy-web`) that both depend on `terraform-staging` and `build-images`, and run in parallel.

**Estimated saving:** 1–2 min on every run.

### 3. Helm `--wait --timeout=10m` opacity

`staging.yml:339` uses `--wait --timeout=10m` on Helm. If pods come up in 90s most runs, the remaining wait is Helm's readiness check polling — usually fine, but worth confirming we are not silently consuming the full timeout on some runs and succeeding at the wire.

**Fix (investigative, not a code change yet):** Add a step that reports `helm status` and pod readiness timestamps so we can see whether the bottleneck is pod startup, image pull, readiness probes, or Helm's polling cadence. If readiness is the slow part, switch to `--wait` with a tighter `--timeout=3m` plus explicit readiness probes tuned to actual startup time. If image pull is slow, evaluate GKE image-streaming or a smaller base image.

**Estimated saving:** 1–4 min on most runs once tuned; could regress if probes are too aggressive — measure first.

## Acceptance criteria

- [ ] Document baseline timings (a sample of 5 recent runs across PR types: code-only, test-only, docs-only) before changing anything.
- [ ] Implement #1 (path-filter image rebuild) — biggest win, smallest risk.
- [ ] Implement #2 (parallel api/web deploys) — straightforward refactor.
- [ ] Decide on #3 after baselining; only act if Helm polling is confirmed as bottleneck.
- [ ] Re-measure end-to-end staging wait on a test-only PR; target under 10 min.

## Out of scope

- Switching off GKE Autopilot or away from Helm for staging — orthogonal architectural decision.
- Caching Playwright browsers in the e2e job (not on the critical path here).

## Why this matters

Staging deploys gate `staging-integration-tests` which gates the green-check signal reviewers and the `/review` workflow read. Every minute on the deploy critical path is a minute the PR sits in "pending" state, blocking the human and AI loops that fire on PR push. Test-only and docs-only PRs are the worst offenders since they bear the full deploy cost for zero deploy-relevant change.

brownm09:
## Summary

Two targeted optimizations to cut end-to-end staging deploy wait from ~21min to under 10min on test-only / docs-only PRs:

1. **Path-filter `Build & Push Images`.** A `dorny/paths-filter@v3` step in the `preflight` job classifies the PR diff. If every changed file matches one of the explicit skip patterns (markdown, `docs/**`, `.claude/**`, `sessions/**`, test files, `infra/terraform/**`, `infra/kubernetes/**`), the build/push steps are skipped and `image_tag` resolves to the head SHA of the most recent successful staging run on this PR (via `gh run list`). Cold-start fallback: if there is no prior successful run on this branch, build fresh. `.github/**` is **not** in the skip list — workflow changes always rebuild to be safe.
2. **Split `deploy-staging` into parallel `deploy-api` and `deploy-web` jobs.** Each gets a per-service concurrency mutex (`staging-deploy-mutex-api`, `staging-deploy-mutex-web`), preserving cross-PR serialization on each shared Cloud Run service while letting api/web overlap within a single PR.

Downstream `staging-integration-tests` and `report-status` jobs were rewired:
- `clerk_configured` and `cloud_run_api_deployed` now come from `needs.deploy-api.outputs.*`
- `web_url` now comes from `needs.deploy-web.outputs.*`
- All Helm and `gcloud run deploy` commands now reference `needs.build-images.outputs.image_tag` instead of `${{ github.event.pull_request.head.sha }}`.

The plan covered acceptance criteria #1 (path-filter) and #2 (parallel deploys) from the issue. #3 (Helm `--wait --timeout` investigation) and the baseline-timings doc are deferred per scope.

## Trade-off note

Cross-PR mutex was previously a single `staging-deploy-mutex` covering both API and web deploys; now it is split per service. Two PRs *cannot* race the same Cloud Run service or Helm release, but PR #1's API deploy may overlap with PR #2's web deploy. Since these target disjoint services, this is safe and is the mechanism that enables parallelism within a PR.

## Acceptance Criteria

- [x] Path-filter image rebuild — biggest win, smallest risk.
- [x] Parallel api/web deploys.
- [ ] Re-measure end-to-end staging wait on a test-only PR; target under 10 min. (Will measure on the first post-merge follow-up PR — this PR itself touches `.github/**` so its own staging run will rebuild.)
- [ ] Decide on #3 (Helm timing investigation) after baselining. Deferred.

## Testing

`npm test` from the worktree (after `npm install`) — environmental failures unrelated to this workflow change:

- `@lifting-logbook/web:test`, `@lifting-logbook/types:test`, `@lifting-logbook/mobile:test`: jest preset / runner module resolution fails inside the worktree's `node_modules`. Tracked in [#365](https://github.com/brownm09/lifting-logbook/issues/365).
- `@lifting-logbook/api:build` (when run in the canonical repo): Prisma generated client missing `InputJsonValue` export — 11 TS errors. Tracked in [#366](https://github.com/brownm09/lifting-logbook/issues/366).

Tests: 4 passed, 0 skipped, multiple workspaces failed for environmental reasons unrelated to this change (workflow YAML only). All failures are pre-existing and tracked.

Workflow YAML structure verified via `js-yaml` parse:
```
YAML OK; jobs: preflight, terraform-staging, build-images, deploy-api, deploy-web, staging-integration-tests, report-status
```

The first real validation of the new behavior is the staging run this PR triggers, which exercises the build-rebuild path (workflow file is in the diff, so `image_affecting=true`).

### Test plan after merge

- Push a markdown-only follow-up commit on a separate PR. Confirm `Build & Push Images` enters the skip branch and `image_tag` resolves to the prior SHA.
- Confirm `deploy-api` and `deploy-web` jobs appear as siblings in the Actions UI and overlap in time.
- Measure end-to-end staging wait; document the result on the follow-up PR.

Closes #358

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -48,6 +48,7 @@ jobs:
     outputs:
       should_run: ${{ steps.check.outputs.should_run }}
       ar_repo: ${{ steps.check.outputs.ar_repo }}
+      image_affecting: ${{ steps.filter.outputs.image_affecting }}
     steps:
       - id: check
         env:
@@ -60,6 +61,31 @@ jobs:
             echo "should_run=false" >> "$GITHUB_OUTPUT"
             echo "GCP_STAGING_WORKLOAD_IDENTITY_PROVIDER not set — skipping staging pipeline."
           fi
+
+      # Classify the PR diff: image_affecting=false only when every changed file
+      # matches one of the skip patterns below. Anything else (including
+      # package.json, lockfiles, Dockerfiles, app/package source) forces rebuild.
+      # `.github/**` is intentionally NOT in the skip list — workflow changes
+      # can affect build context and should rebuild to be safe.
+      - id: filter
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            image_affecting:
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.claude/**'
+              - '!sessions/**'
+              - '!ROADMAP.md'
+              - '!**/__tests__/**'
+              - '!**/*.test.ts'
+              - '!**/*.test.tsx'
+              - '!**/*.spec.ts'
+              - '!**/*.spec.tsx'
+              - '!**/tests/**'
+              - '!e2e/**'
+              - '!infra/terraform/**'
+              - '!infra/kubernetes/**'
 
   # ── 1. Terraform: staging ─────────────────────────────────────────────────
   terraform-staging:
@@ -137,11 +163,42 @@ jobs:
       id-token: write
     outputs:
       ar_repo: ${{ steps.ar.outputs.repo }}
+      image_tag: ${{ steps.image-tag.outputs.image_tag }}
 
     steps:
       - uses: actions/checkout@v4
 
+      # Decide whether to rebuild or reuse a prior image.
+      # Reuse is only chosen when the diff is non-image-affecting AND a prior
+      # successful staging run exists on this PR's head ref. Otherwise we build
+      # fresh and tag the new image with the head SHA, preserving the
+      # SHA-as-image-tag invariant the rest of the pipeline relies on.
+      - name: Resolve image tag
+        id: image-tag
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          if [[ "${{ needs.preflight.outputs.image_affecting }}" == "false" ]]; then
+            PREV_SHA=$(gh run list \
+              --branch "${{ github.event.pull_request.head.ref }}" \
+              --workflow staging.yml \
+              --status success \
+              --limit 1 \
+              --json headSha \
+              --jq '.[0].headSha // ""')
+            if [[ -n "$PREV_SHA" ]]; then
+              echo "image_tag=$PREV_SHA" >> "$GITHUB_OUTPUT"
+              echo "skip_build=true" >> "$GITHUB_OUTPUT"
+              echo "Diff is non-image-affecting — reusing prior image tag: $PREV_SHA"
+              exit 0
+            fi
+            echo "Diff is non-image-affecting but no prior successful run found — building fresh."
+          fi
+          echo "image_tag=${{ github.event.pull_request.head.sha }}" >> "$GITHUB_OUTPUT"
+          echo "skip_build=false" >> "$GITHUB_OUTPUT"
+
       - name: Authenticate to GCP (staging)
+        if: steps.image-tag.outputs.skip_build != 'true'
         uses: google-github-actions/auth@v2
         with:
           workload_identity_provider: ${{ secrets.GCP_STAGING_WORKLOAD_IDENTITY_PROVIDER }}
@@ -152,13 +209,16 @@ jobs:
         run: echo "repo=${{ needs.terraform-staging.outputs.ar_repo }}" >> "$GITHUB_OUTPUT"
 
       - name: Resolve API URL for web image build
+        if: steps.image-tag.outputs.skip_build != 'true'
         id: api-url
         run: echo "url=${{ needs.terraform-staging.outputs.cloud_run_api_url }}" >> "$GITHUB_OUTPUT"
 
       - name: Configure Docker for Artifact Registry
+        if: steps.image-tag.outputs.skip_build != 'true'
         run: gcloud auth configure-docker ${{ env.REGISTRY }} --quiet
 
       - name: Fetch staging Clerk publishable key
+        if: steps.image-tag.outputs.skip_build != 'true'
         id: clerk-pub-key
         run: |
           CLERK_PUB=$(gcloud secrets versions access latest \
@@ -167,9 +227,11 @@ jobs:
           echo "key=$CLERK_PUB" >> "$GITHUB_OUTPUT"
 
       - name: Set up Docker Buildx
+        if: steps.image-tag.outputs.skip_build != 'true'
         uses: docker/setup-buildx-action@v3
 
       - name: Build and push API image
+        if: steps.image-tag.outputs.skip_build != 'true'
         uses: docker/build-push-action@v6
         with:
           context: .
@@ -182,6 +244,7 @@ jobs:
           cache-to: type=gha,mode=max
 
       - name: Build and push web image
+        if: steps.image-tag.outputs.skip_build != 'true'
         uses: docker/build-push-action@v6
         with:
           context: .
@@ -196,9 +259,13 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
-  # ── 3. Deploy to staging ──────────────────────────────────────────────────
-  deploy-staging:
-    name: Deploy (staging)
+  # ── 3a. Deploy API (staging) ──────────────────────────────────────────────
+  # Split into per-service jobs so they run in parallel within a single PR.
+  # Cross-PR serialization is preserved per-service via separate mutex groups —
+  # two PRs cannot race the same Cloud Run service or Helm release, but PR #1's
+  # API deploy may overlap with PR #2's web deploy (disjoint targets, safe).
+  deploy-api:
+    name: Deploy API (staging)
     runs-on: ubuntu-latest
     needs: [preflight, terraform-staging, build-images]
     if: |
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [infra] Skip image rebuild on non-image diffs; parallelize api/web deploys

### Summary

Needs revision before merge. The parallelization split is clean and the path-filter pattern is well-scoped, but the image-reuse logic has a latent correctness bug: after two consecutive skip-build runs on the same branch, `image_tag` resolves to a commit SHA whose image was never pushed to Artifact Registry, and the Cloud Run / Helm deploy will fail to pull the image. This bites on the second docs-only push, not the first — so the staging run that triggers this PR will look healthy and the regression surfaces later. One blocking finding, a couple of non-blocking notes, and one question on intended cold-start behavior.

---

### Blocking Findings

**[correctness]** `.github/workflows/staging.yml:177-198` — chained skip-builds resolve to a non-existent image tag

The `Resolve image tag` step uses `gh run list --status success --limit 1 --json headSha` to find the prior successful run and sets `image_tag` to that run's **commit SHA**. But on a skip-build run, no image is pushed with that commit's SHA — the deploy in that run uses the *previous* run's `image_tag`. So:

- Run 1 (image-affecting at `SHA1`): builds `:SHA1`, deploys `:SHA1`, succeeds. Last successful run's `headSha` = `SHA1`. ✅
- Run 2 (docs-only at `SHA2`): `PREV_SHA=SHA1`, `image_tag=SHA1`, deploys `:SHA1`, succeeds. Last successful run's `headSha` = `SHA2`. ✅
- Run 3 (docs-only at `SHA3`): `PREV_SHA=SHA2`, `image_tag=SHA2`, deploys `:SHA2`. But `:SHA2` was never pushed — only `:SHA1` exists. ❌ Deploy fails on `image not found`.

This means the optimization works for exactly one docs-only follow-up commit per image-affecting baseline, then silently breaks. Since this PR's own staging run rebuilds (workflow file is in the diff), the regression won't appear in this PR's run and will surface as a confusing deploy failure on the second docs-only PR after merge.

**Fix:** Resolve to the actual image tag of the prior run, not its commit SHA. Two viable options:

1. **Query the prior run's `build-images` job output.** Use `gh api /repos/{owner}/{repo}/actions/runs/<run_id>/jobs` against the prior successful run and read its `image_tag` output. By induction, that value points to a real pushed image.
2. **Walk back until finding a non-skip run.** Tag built images with both `:<sha>` and a floating `:pr-<N>-latest`, and have the skip path use the floating tag. Simpler but loses the SHA-as-tag invariant the PR description calls out.

Option 1 preserves the invariant and is the smaller code change. Recommend option 1.

---

### Non-Blocking Findings

**[maintainability]** `.github/workflows/staging.yml:166, 213, 219, 225, 230, 233, 247, 262` — every step in `build-images` after `checkout` carries `if: steps.image-tag.outputs.skip_build != 'true'`

Eight repeated guards. If a future step is added without the guard, the skip semantics break silently for that step. The current shape is readable and each guard is explicit, but the next person adding a step here will probably forget it.

**Suggestion:** Add an inline comment at the first guarded step ("Every step below this point must carry the same skip guard — see `image-tag` step") or extract the build into a reusable workflow that the job invokes only when `skip_build != 'true'`.

---

**[reliability]** `.github/workflows/staging.yml:184, 188` — `gh run list --branch <head.ref>` is not PR-scoped

`--branch` filters by ref name, not by PR. If a branch is reused after a previous PR closes, the prior PR's runs are still on that branch and will satisfy `--status success --limit 1`. Image-tag resolution would then point at a run from a different logical PR. Unlikely given the per-session-unique `claude/<adjective>-<animal>-<hash>` naming convention; worth noting because the constraint isn't enforced anywhere.

**Suggestion:** If strict PR-scoping is wanted, include the PR number in the run filter or query `gh pr view --json headRefOid`. Low priority.

---

**[maintainability]** `.github/workflows/staging.yml:443-451 and 514-522` — `Check Clerk secret key configured` is duplicated verbatim across `deploy-api` and `deploy-web`

Both jobs fetch the Clerk secret key. `deploy-web` only uses it as a gate (no consumer reads its output), so the fetch is a redundant Secret Manager call per deploy. Meaningful only if the check's logic ever changes and one copy drifts.

**Suggestion:** Either factor the check into a reusable workflow, or have `deploy-web` read `clerk_configured` from `needs.deploy-api.outputs` (at the cost of an explicit dependency that erodes some parallelism on this one secret fetch). The current shape is defensible; just keep the two copies in sync.

---

**[maintainability]** PR body — Testing summary does not match the `Tests: N passed, N skipped, N failed (duration)` format required by ADR-029

The body says "Tests: 4 passed, 0 skipped, multiple workspaces failed for environmental reasons unrelated to this change". The integrity policy expects a specific failed count and a duration. Failures are tracked in #365 / #366 per the failure-tracking rule, so the substance is right; only the format is off.

**Suggestion:** Edit the PR body to include the exact failed count and duration, or note explicitly that the test command did not emit a parseable summary.

---

### Questions for Author

**Question:** On the cold-start fallback path (skip-eligible diff, no prior successful run on the branch), do you intend the very first PR push to build fresh even when the diff is e.g. markdown-only?

**Context:** `.github/workflows/staging.yml:185-198` — when `image_affecting=false` and `PREV_SHA` is empty, the script falls through to `image_tag=${{ github.event.pull_request.head.sha }}` and `skip_build=false`, so the rebuild proceeds. For a branch cut for a docs-only change that never touches application code, every first push pays the full image-build cost even though no image change is needed.

**Tradeoffs:** (a) Keep as-is — simplest, predictable, no cross-branch reasoning; every new docs-only branch pays one ~5min rebuild. (b) Fall back to `main`'s latest staging image when no prior PR run exists — saves the rebuild but introduces a new edge case (what if `main`'s image is stale relative to the PR's base?). (c) Use the PR's *base* SHA's image when present — most correct semantically, but requires querying `main`'s staging run history and assumes images are retained for the base SHA. The plan document explicitly chose (a) ("don't try to resolve back to main's image — too many edge cases"), so this is a confirmation question rather than a request to change.

---

### Style Notes

Section-marker comments (`# ── 3a. Deploy API ──`) and inline rationale comments on the path-filter skip list and the cross-PR mutex split are well-placed and match the file's existing voice. No style findings.

---

PR #369, opened 2026-05-28
[test] Auto-provision Postgres for DB E2E suite via Testcontainers

### LINKED ISSUE #264: [test] Evaluate Testcontainers for local DB integration test runs
## Background

The DB adapter E2E suite (`programs.db.e2e.spec.ts`) requires a real Postgres connection via `DATABASE_URL`. In CI, a Postgres container is spun up automatically. Locally, the suite skips entirely when `DATABASE_URL` is unset — by design, documented in `jest.env.setup.js`.

## Problem

The skip-by-default behavior means DB-layer bugs only surface in CI, not during local development. The round-trip (push → CI → red → fix → push) is slower than catching failures locally.

## Option: Testcontainers

[`testcontainers`](https://node.testcontainers.org/) spins up a real Postgres container programmatically inside Jest's `globalSetup`, sets `DATABASE_URL` to the ephemeral instance, runs migrations, then tears it down in `globalTeardown`. No pre-configured `DATABASE_URL` needed — just Docker running locally. This mirrors exactly what CI does today but runs automatically on `npm test`.

**Pros:**
- Local and CI behavior become identical
- No `DATABASE_URL` configuration required for contributors
- DB bugs caught before push

**Cons:**
- Requires Docker Desktop running locally (not zero-dependency)
- Adds ~10–15s container startup to the DB suite
- Adds `testcontainers` as a dev dependency

## Alternatives Considered

- **SQLite in-memory** — Prisma supports it, but the schema uses Postgres-specific features (enums, etc.) that don't translate cleanly. Would require a parallel schema or migration strategy.
- **Mock the DB** — defeats the purpose; the in-memory adapter suite already covers logic paths fast. The DB suite exists specifically to test real Prisma + Postgres behavior.
- **Status quo** — keep skipping locally, rely on CI. Acceptable if the current push-to-CI feedback loop is fast enough and no bugs are slipping through.

## Acceptance Criteria

- [ ] Evaluate whether Docker is already a dev prerequisite in this repo (Kubernetes manifests suggest familiarity)
- [ ] If proceeding: add `testcontainers` dev dependency, implement `globalSetup`/`globalTeardown` in `apps/api/`
- [ ] DB suite runs end-to-end locally with no environment variable configuration
- [ ] CI behavior unchanged (still uses the existing Postgres container job)
- [ ] Document the Docker requirement in `CLAUDE.md` or `README`

brownm09:
## Summary

Adds Jest `globalSetup` / `globalTeardown` in `apps/api/` so the DB E2E suite (`programs.db.e2e.spec.ts`) auto-provisions Postgres locally via [`@testcontainers/postgresql`](https://node.testcontainers.org/modules/postgresql/). Mirrors what CI's `db-integration` service container already does — the round-trip of push → CI → red → fix → push is replaced by catching DB-layer bugs on the developer's first `npm test`.

CI behavior preserves the existing service-container path: when `DATABASE_URL` is already exported, `globalSetup` passes it through (re-exposed via a `LIFTING_TC_DATABASE_URL` sentinel so the existing per-worker env-isolation Proxy doesn't blank it) and skips container startup. No Docker-in-Docker on runners. The `postgres:16-alpine` image is pinned to a digest in both `jest.global-setup.js` and `.github/workflows/ci.yml` so local and CI runs are byte-identical.

ADR-013 already endorsed this approach; this PR is the implementation. `docker-compose.test.yml` is removed (superseded).

### Notable discovery: the DB E2E suite was silently skipping in CI

Until this PR, the gate `process.env.DATABASE_URL ? describe : describe.skip` was evaluated at module load in each worker, *after* `jest.env.setup.js`'s BLOCK list had blanked `DATABASE_URL`. Run [`26546022094`](https://github.com/brownm09/lifting-logbook/actions/runs/26546022094) (on `main` just before this PR) reports `Tests: 51 skipped, 51 total` for the `db-integration` job — confirming the suite was never actually executing against the service container. This PR's sentinel-based gate (`LIFTING_TC_DATABASE_URL`) is what enables real-DB execution in CI for the first time.

The first CI run with the suite enabled surfaced 7 distinct failures across three categories — exactly the value proposition of #264 paying off immediately:

**1 real production bug** — `UpdateSettingsDto` was missing `activeProgram`. NestJS `ValidationPipe({ whitelist: true })` silently stripped the field from every `PATCH /users/me/settings` request, so the repository upserted `{}`. Anyone trying to change their active program via the API had it dropped on the floor. Fixed by declaring the field on the DTO with `@IsString @MaxLength(64)`.

**1 catalog gap** — `DEFAULT_SLOT_MAP` was missing the bare `'Lat Raise'` alias used by `lift_records.csv`. CSV import 400'd on validation. Added the alias alongside `'C. Lat Raise'`.

**5 test-setup / over-specified-assertion bugs** in `programs.db.e2e.spec.ts` — Alice/Bob lacked `CycleDashboard` seeds (caused isolation tests to 404 on a dashboard precondition); no `TrainingMaxHistory` rows ever materialized for the PATCH-PR / GET-?isPR chain; the recalculate test expected the wrong response shape and assumed reductions would be applied (they're flagged); the re-import test asserted an exact skipped count that's brittle against seed-record collisions. All corrected to assert the real contract.

### Acceptance criteria (issue #264)

- [x] Evaluate Docker as dev prerequisite — yes, documented in `README.md`.
- [x] Add `@testcontainers/postgresql` dev dependency; implement `globalSetup`/`globalTeardown`.
- [x] DB suite runs end-to-end locally with no env-var configuration (Docker Desktop running suffices).
- [x] CI behavior preserved — `db-integration` job's pre-set `DATABASE_URL` is reused, no container started.
- [x] Document the Docker requirement (`CLAUDE.md` Testing, `docs/testing/e2e-coverage.md`, ADR-013 implementation note).

## Test plan

Both verified in CI run [`26548848090`](https://github.com/brownm09/lifting-logbook/actions/runs/26548848090) (this branch, latest commit):

- [x] CI `lint-and-test` job green — full API suite (including DB E2E via Testcontainers on the runner's Docker) passes.
- [x] CI `db-integration` job green — `globalSetup` CI-passthrough path used, DB E2E executes against the service container.
- [ ] Reviewer optional: run `npm test -w @lifting-logbook/api` locally with Docker Desktop running and confirm the DB suite executes end-to-end (no `DATABASE_URL` exported).
- [ ] Reviewer optional: run the same command with Docker stopped and confirm the DB suite skips gracefully (`describe.skip` keyed off `LIFTING_TC_DATABASE_URL`).

## Testing

```
Lint & Test (full API suite via Testcontainers):
Tests:       301 passed, 301 total

DB Integration Tests (service container):
Tests:       51 passed, 51 total
```

Pre-PR per `CLAUDE.md` Test Integrity policy Rule 2: the `Tests: N passed, N skipped, N failed` summary is present, skipped count is zero, no integrity violations.

### Suppression check

Pre-PR `git diff origin/main` grep for suppressions returned no matches — no new `!`, `?? null`, `ts-ignore`, `ts-expect-error`, or `eslint-disable` introduced.

### Test integrity check

The diff swaps one `describe.skip` gate for another (different sentinel variable name; same skip semantics). Net change: zero new skips, identical skip behavior — the suite still skips only when no Postgres URL is available. Not a Test Integrity Policy violation (no test or assertion was removed, no threshold lowered, no `--bail` / `passWithNoTests` / `testPathIgnorePatterns` added).

Closes #264


brownm09:
## Review: [test] Auto-provision Postgres for DB E2E suite via Testcontainers

### Summary
Merge with minor changes. Implementation is clean and the sentinel-variable pattern correctly threads around the existing `jest.env.setup.js` Proxy. One reliability concern (container leak on migration failure) is worth fixing before merge; the rest are non-blocking. Notable discovery: the diff appears to actually *enable* DB E2E execution in CI for the first time — the prior gate was almost certainly skipping silently because `DATABASE_URL` was in the pre-existing BLOCK list and got blanked in every worker before the gate ran. That makes this PR more impactful than the body advertises and should be confirmed against historical CI logs.

---

### Non-Blocking Findings

**[reliability]** `apps/api/jest.global-setup.js:42-48`
The container is started at line 33, but `globalThis.__LL_PG_CONTAINER__ = container` is only assigned at line 48 — after `execSync('npx prisma migrate deploy ...')`. If the migration throws (schema validation error, DB unreachable mid-handshake, prisma CLI not on PATH), `globalTeardown` runs with no handle and the started Postgres container leaks. Testcontainers' Ryuk reaper usually cleans these up, but Ryuk's behavior on Docker Desktop / Windows is not guaranteed — every aborted run can leak a container.
**Suggestion:** Stash the handle on `globalThis` *immediately* after `.start()` returns (before migrations), so `globalTeardown` can stop it on any subsequent failure. Alternatively, wrap migrations in `try/catch` that calls `await container.stop()` and rethrows.

**[reliability]** `apps/api/jest.global-setup.js:42`
`execSync` has no `timeout` option. A hung `prisma migrate deploy` (network issue resolving the npm registry on a cold `npx` cache, or a stuck DB handshake) will block globalSetup until Jest's outer timeout kills the whole process — and even then the container leaks per the finding above.
**Suggestion:** Pass `timeout: 60_000` (or similar) to bound migration runtime. Pairs with the leak fix above for full coverage.

**[maintainability]** `apps/api/jest.global-setup.js:32`
`postgres:16-alpine` is a floating tag. ADR-013's implementation note says "matches CI image", but `postgres:16-alpine` in `.github/workflows/ci.yml` can resolve to a different digest than the same string pulled locally a week later — silent drift between dev and CI is exactly what Testcontainers is supposed to eliminate.
**Suggestion:** Pin to a specific digest (`postgres:16-alpine@sha256:...`) in both this file and the CI workflow, or document the deliberate decision to accept drift.

**[documentation]** PR body — Missing test-run summary
Per `CLAUDE.md` Code Quality → Test integrity policy → Rule 2, the PR body must contain a `Tests: N passed, N skipped, N failed` line or an explicit statement that the project has no automated tests. The body has neither — the placeholder `Tests: not executed locally (sandbox limitation ...)` documents *why* but does not satisfy the literal rule. The project has tests; they should be run.
**Suggestion:** Once CI's `lint-and-test` and `db-integration` jobs complete, paste their summary lines into the PR body (or block merge on them). Alternatively, explicitly note that the rule was waived due to sandbox limitations and that CI green is the gating signal.

---

### Questions for Author

**Question:** Was the old DB E2E suite actually executing in CI prior to this PR, or was it silently skipping?
**Context:** `apps/api/jest.env.setup.js`'s pre-existing `BLOCK` list includes `DATABASE_URL` (verified on the PR branch). The setupFiles script runs in every worker before any spec loads and sets `process.env.DATABASE_URL = ''`. The old gate `const describeOrSkip = process.env.DATABASE_URL ? describe : describe.skip` is evaluated at module load — in the worker, *after* setupFiles blanks the var. That suggests the old gate always saw `''` (falsy) and always took the `describe.skip` branch, even when the CI service container exported `DATABASE_URL`.
**Tradeoffs:** (a) If the prior CI runs were silently skipping the suite, this PR is doing more than the description claims — it is actually turning on real-DB coverage in CI for the first time, which is a much bigger win and should be called out in the PR body and in #264's resolution comment. (b) If there is some mechanism I missed (e.g., a CI-specific override, a different test script that bypasses `jest.env.setup.js`), then the description is accurate and no change is needed — but it would be worth a one-line comment in `jest.env.setup.js` documenting how CI used to thread DATABASE_URL through. A quick check of the most recent `db-integration` job log on `main` for the `PASS .../programs.db.e2e.spec.ts` line will settle it.

---

Issue #373, opened 2026-05-28
[test] Node 24: nest build fails on transitive ESM resolution (iconv-lite, minimatch)


brownm09:
## Symptom

After fixing the ts-jest preset issue (#370), `nest build` in `apps/api` still fails on Node 24 with transitive CommonJS / ESM resolution errors like:

```
Error: Cannot find module './streams'
  ...iconv-lite/lib/index.js
```

and

```
Error: Cannot find module 'C:\...\node_modules\minimatch\dist\commonjs\index.js'
  at finalizeEsmResolution
```

These come from `@nestjs/cli` → `inquirer` → `external-editor` → `iconv-lite`, and from `minimatch`'s package.json `exports` map being stricter than Node 24's CJS loader.

## Why this matters

- Blocks `npm test` end-to-end on Node 24 even after #370 is merged (the `api:test` turbo task depends on `api:build`).
- Local-only — CI runs Node 20 and is unaffected.
- Independent of #370: #370 was specifically scoped to the ts-jest preset resolution.

## Acceptance criteria

- [ ] `npm test` runs end-to-end on Node 24 without `nest build` CommonJS resolution errors.
- [ ] CI (Node 20) continues to pass.
- [ ] If a clean fix is not viable in the short term, document the constraint in `CLAUDE.md` Platform section.

## References

- Follow-up to #370 (ts-jest preset fix).

PR #376, opened 2026-05-28
[fix] Wire prisma generate into Turbo build/test pipeline

brownm09:
## Summary

- Adds a dedicated `db:generate` Turbo task in `apps/api` keyed on `prisma/schema.prisma`; `build` and `test` declare it in `dependsOn`.
- Adds the matching `db:generate` npm script (`prisma generate`) to `apps/api/package.json`.
- Closes the gap from #355: the existing `postinstall` covers fresh installs, but schema drift between installs left the generated Prisma client stale. Turbo's input-keyed cache now regenerates the client only when `schema.prisma` changes.

## Acceptance Criteria

- [x] `npm test` regenerates the Prisma client when the schema is newer than the generated output
- [x] Subsequent runs are cache hits when the schema is unchanged
- [x] CI behavior unchanged — `.github/workflows/ci.yml` still calls `npx prisma generate` explicitly

## Testing

`npm test` was run from this worktree after `rm -rf node_modules/.prisma && npm install`. The `@lifting-logbook/api:db:generate` task ran and produced a fresh `node_modules/.prisma/client/` with `InputJsonValue` exported (35 occurrences). The previously-failing TS2694 errors on `apps/api/src/user-settings/user-settings.repository.ts` no longer reproduce — the build now gets past TypeScript compilation.

The full `npm test` then fails on two pre-existing, unrelated Node 24 / Windows worktree issues:

- **`Preset ts-jest not found`** on every workspace with a Jest test — tracked in [#365](https://github.com/brownm09/lifting-logbook/issues/365).
- **`Cannot find module 'external-editor'`** in `apps/api` `nest build` — same class of transitive CJS/ESM resolution failure tracked in [#373](https://github.com/brownm09/lifting-logbook/issues/373).

Both failures are pre-existing on `origin/main` in this worktree environment and are not introduced by this change. CI runs Node 20 and is unaffected; CI's explicit `npx prisma generate` step continues to work alongside the new Turbo task.

Tests: not summarizable as `N passed, N skipped, N failed` locally due to the pre-existing install-resolution failures above. The change itself is a build-pipeline wiring change with no production code or test code touched.

## Test Coverage Gate

No new behavior introduced — this is a build-orchestration change. The `db:generate` task is exercised in the Testing run above (cache miss → execute → produces client output).

## Suppression / Test-Integrity Check

```
git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|eslint-disable|!\.|!\[|!\s*[;,)]|\?\? null|it\.skip|...)'
```
Only match is the pre-existing `!.next/cache/**` Turbo output negation glob — not a suppression.

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -9,6 +9,7 @@
     "lint": "eslint src",
     "test": "jest",
     "test:db": "jest --testPathPattern=db\\.e2e",
+    "db:generate": "prisma generate",
     "postinstall": "prisma generate"
   },
   "dependencies": {
--- a/turbo.json
+++ b/turbo.json
@@ -2,8 +2,8 @@
   "$schema": "https://turbo.build/schema.json",
   "tasks": {
     "db:generate": {
-      "inputs": ["prisma/schema.prisma"],
-      "outputs": ["node_modules/.prisma/**", "../../node_modules/.prisma/**"]
+      "inputs": ["prisma/schema.prisma", "package.json", "../../package-lock.json"],
+      "outputs": ["node_modules/.prisma/**"]
     },
     "build": {
       "dependsOn": ["^build", "db:generate"],
@@ -11,7 +11,7 @@
       "inputs": ["src/**", "app/**", "!src/**/*.test.ts", "!src/**/*.spec.ts", "!src/**/__tests__/**", "tsconfig.json", "package.json", "next.config.*"]
     },
     "test": {
-      "dependsOn": ["build", "db:generate"],
+      "dependsOn": ["build"],
       "cache": false
     },
     "lint": {
--- a/turbo.json
+++ b/turbo.json
@@ -1,13 +1,17 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "tasks": {
+    "db:generate": {
+      "inputs": ["prisma/schema.prisma"],
+      "outputs": ["node_modules/.prisma/**", "../../node_modules/.prisma/**"]
+    },
     "build": {
-      "dependsOn": ["^build"],
+      "dependsOn": ["^build", "db:generate"],
       "outputs": ["dist/**", ".next/**", "!.next/cache/**"],
       "inputs": ["src/**", "app/**", "!src/**/*.test.ts", "!src/**/*.spec.ts", "!src/**/__tests__/**", "tsconfig.json", "package.json", "next.config.*"]
     },
     "test": {
-      "dependsOn": ["build"],
+      "dependsOn": ["build", "db:generate"],
       "cache": false
     },
     "lint": {


brownm09:
## Review: [fix] Wire prisma generate into Turbo build/test pipeline

### Summary

Merge with minor cleanup. The fix is correctly scoped: a `db:generate` Turbo task keyed on `prisma/schema.prisma` closes the schema-drift gap that `postinstall` alone could not. The diff is 9 lines across two config files with no production code touched, and the verification — fresh client output with `InputJsonValue` re-exported — confirms the original TS2694 class no longer reproduces. Highest-severity finding is non-blocking: a small redundancy in `test.dependsOn` and a couple of input/output edge cases worth tightening before merge or in a follow-up.

---

### Non-Blocking Findings

**[maintainability]** `turbo.json:15`
`test` declares `dependsOn: ["build", "db:generate"]`, but `build` already declares `dependsOn: ["^build", "db:generate"]`, so `test → build → db:generate` is the transitive chain Turbo will compute regardless. The explicit `db:generate` on `test` is redundant.
**Suggestion:** Drop `db:generate` from `test.dependsOn` and let it come through `build`. Keeps the dependency graph minimal and avoids drift if `build`'s deps are later restructured (e.g., someone removes `db:generate` from `build` and assumes `test` is still covered).

---

**[reliability]** `turbo.json:6`
`db:generate.inputs` is keyed only on `prisma/schema.prisma`. A `@prisma/client` or `prisma` version bump in `apps/api/package.json` changes the generator output without touching the schema, and Turbo will report a cache hit — leaving stale client code that does not match the installed runtime. This is the same class of drift this PR was created to fix, just at a different axis (package-version drift instead of schema drift).
**Suggestion:** Add `package.json` (and ideally the root `package-lock.json`) to the task's `inputs` so a Prisma version change invalidates the cache. Example: `"inputs": ["prisma/schema.prisma", "package.json", "../../package-lock.json"]`.

---

**[reliability]** `turbo.json:7`
`outputs` includes `"../../node_modules/.prisma/**"`, which escapes the package boundary. Turbo treats package-relative outputs as the cacheable artifact; outputs above the workspace root work in practice but are sensitive to hoisting decisions (if npm decides to nest the client under `apps/api/node_modules/.prisma/` instead of hoisting, the cache restore won't repopulate the root path, and vice versa). This is more brittle than it appears under the current install state.
**Suggestion:** Confirm by inspecting where the client actually lands on a clean install in CI (Node 20) vs. the local worktree (Node 24). If only one location is ever populated, drop the other glob. If both can be populated depending on environment, consider whether Turbo's cache restore handles the parent-relative path reliably on Windows — file a follow-up issue if not.

---

### Questions for Author

**Question:** Is `db:generate` intended to be invokable on workspaces other than `apps/api`?
**Context:** The task is declared at the root `turbo.json` without a workspace filter ([`turbo.json:4-7`](turbo.json)). Today only `apps/api/package.json` defines a `db:generate` script, so `turbo run db:generate` is effectively scoped. If a future workspace (e.g., a separate analytics service with its own Prisma schema) adds a `db:generate` script, it will silently get picked up under `build` and `test` dependencies for that workspace too.
**Tradeoffs:** (a) Leave as-is — minimal config; future workspaces opt in by defining the script, which is probably the desired behavior. (b) Rename to `api:db:generate` or move the task under an `apps/api`-specific `turbo.json` extension — explicit scoping, but adds boilerplate and forecloses reuse. (a) is the natural default unless multiple Prisma schemas are anticipated.

---

PR #379, opened 2026-05-28
[docs] Document main branch-protection required-check set (#378)

brownm09:
## Summary

Closes #378.

Branch protection on `main` previously required only two contexts: `Lint & Test` and `Staging Integration Tests`. Real per-PR gates (`DB Integration Tests`, `Playwright E2E`, `Observability Stack Smoke Test`) were not required — a PR with a failing DB or E2E suite could merge cleanly.

This PR:

1. **Updates branch protection via API** (already applied) to require all five real per-PR checks:
   - `Lint & Test`
   - `DB Integration Tests`
   - `Observability Stack Smoke Test`
   - `Playwright E2E`
   - `Staging Integration Tests`
2. **Documents the required-check set** in `docs/operations/branch-protection.md`, including the staging-credential dependency (skip-when-`should_run`-is-false) and the inspect/update commands.

### Issue premise correction

Issue #378 claimed `Staging Integration Tests` is never produced by CI. Investigation showed the job exists in `.github/workflows/staging.yml:578` and reports SUCCESS on PRs #376 and #377. It is gated by `preflight.outputs.should_run` — when staging GCP credentials are absent the job is skipped and the required check never arrives, which matches the `--admin`-required symptom but is conditional, not constant.

### Before / after

```
# before
contexts: ["Lint & Test", "Staging Integration Tests"]

# after
contexts: ["Lint & Test", "Staging Integration Tests", "DB Integration Tests",
           "Observability Stack Smoke Test", "Playwright E2E"]
```

Verified via `gh api repos/brownm09/lifting-logbook/branches/main/protection/required_status_checks`.

## Testing

Ran `npm test` from repo root.

- **11 of 12 workspaces passed.** `core`, `api-legacy`, `web`, `types`, etc. all green.
- **`@lifting-logbook/api` failed in this worktree** because Docker Desktop is not available in this environment; the Testcontainers global setup cannot start the Postgres container (`Could not find a working container runtime strategy`). This is documented in `CLAUDE.md` as a hard prerequisite for the API DB E2E suite and is environment-only — the failure is identical to running `npm test` on a fresh clone without Docker running, not a code regression introduced by this PR. CI (which provides `DATABASE_URL` directly without Testcontainers) is unaffected.

No code paths are changed by this PR. The only file added is a markdown doc; the branch-protection update is a GitHub API setting change with no automated test coverage.

[...]

Issue #378, opened 2026-05-28
Branch protection requires 'Staging Integration Tests' check that CI never produces


brownm09:
## Problem

Every PR merge currently requires `--admin` because the `main` branch protection rule lists `Staging Integration Tests` as a required status check, but the CI workflow (`.github/workflows/ci.yml`) does not produce a check with that exact name.

Checks actually produced by CI:
- Lint & Test
- DB Integration Tests
- Observability Stack Smoke Test
- Playwright E2E

Required checks in branch protection:
- Lint & Test ✅
- Staging Integration Tests ❌ (never reported)

Because the required check never arrives, `mergeStateStatus` stays `BLOCKED` even when all real CI checks pass, and every merge falls back to `gh pr merge --admin`.

Discovered during PR #377 merge (2026-05-28).

## Fix

Either:
1. Remove `Staging Integration Tests` from the required checks list, or
2. Add a job named exactly `Staging Integration Tests` to the CI workflow.

## Impact

Every in-session merge silently requires `--admin` — bypasses the intended gate without surfacing why.

PR #381, opened 2026-05-28
[infra] Add required-checks drift detection workflow

### LINKED ISSUE #380: Cron-based required-checks drift detection
## Problem

`main` branch protection (`required_status_checks.contexts`) and the actual set of per-PR CI jobs can drift silently in two directions:

- **Loud direction:** protection requires a context that no longer fires → every PR sits at `mergeStateStatus: BLOCKED` until someone updates protection. This was #378.
- **Silent direction:** a real CI job exists but is not required → regressions in that suite merge cleanly. This was the actual pre-#378 state (`DB Integration Tests`, `Playwright E2E`, `Observability Stack Smoke Test` all ran but did not gate).

PR #379 fixed the silent direction by broadening the required-check list and documented the directive at `docs/operations/branch-protection.md` ("If staging is intentionally decommissioned, remove `Staging Integration Tests` from the required-check list at the same time"). That directive is purely advisory — nothing enforces it. The exact failure mode that produced #378 is one workflow rename away from recurring.

## Proposal

Add a nightly (or weekly) workflow `.github/workflows/required-checks-drift.yml` that:

1. Calls `gh api repos/brownm09/lifting-logbook/branches/main/protection/required_status_checks`.
2. Diffs against a checked-in `.github/expected-required-checks.json`.
3. Opens (or updates) an issue on mismatch.

Also catches drift introduced by editing branch protection through the GitHub UI without updating the doc.

## Acceptance criteria

- [ ] `.github/expected-required-checks.json` checked in with the five current contexts.
- [ ] Workflow runs on a schedule and on push to the JSON file.
- [ ] On mismatch, opens an issue with the diff in the body. Idempotent — re-uses an existing open issue rather than spamming.
- [ ] `docs/operations/branch-protection.md` updated to point at the expected-checks file as the source of truth.

## Why this matters

Without enforcement, the silent-direction failure mode (a real CI job stops gating because it was renamed or removed) leaves the repo trusting green merges that no longer cover what they used to. The cost of mechanical detection is one ~30-line workflow and one JSON file; the cost of catching it via prod regression is much higher.

Follows up #378 / #379.

brownm09:
## Summary

Adds a scheduled workflow that diffs live `main` branch-protection required contexts against a checked-in expected list, and opens (or comments on the existing) issue on mismatch. Follows up #378 / #379, which fixed the immediate drift but left enforcement advisory.

- `.github/expected-required-checks.json` — source of truth, five contexts (sorted alphabetically; sort is normalized at diff time so order isn't load-bearing).
- `.github/workflows/required-checks-drift.yml` — runs Mondays at 13:17 UTC (~09:17 ET), on pushes to `main` that touch the JSON or the workflow itself, and on `workflow_dispatch`. Fetches `branches/main/protection/required_status_checks`, computes `missing` (expected − actual) and `extra` (actual − expected), and on drift: creates/comments the `required-checks-drift`-labeled issue **and** exits 1 so the Actions tab also goes red. The diff is printed in the run log before exit, so one click into the failed run shows the cause.
- `docs/operations/branch-protection.md` — adds a "Source of truth" section pointing at the JSON file and the workflow; updates the staging-decommission directive to reference the JSON.

## Acceptance criteria

- [x] `.github/expected-required-checks.json` checked in with the five current contexts.
- [x] Workflow runs on a schedule (weekly Mondays) and on push to the JSON file.
- [x] On mismatch, opens an issue with the diff in the body. Idempotent — re-uses an existing open `required-checks-drift` issue rather than spamming.
- [x] `docs/operations/branch-protection.md` updated to point at the expected-checks file as the source of truth.

## Test plan

- [x] JSON file parses: `node -e "JSON.parse(require('fs').readFileSync('.github/expected-required-checks.json'))"` → OK.
- [x] Workflow YAML parses with `js-yaml` → OK.
- [x] `@lifting-logbook/core` test suite — **Tests: 160 passed, 0 skipped, 0 failed (34.5s)**.
- [ ] After merge, trigger via `gh workflow run required-checks-drift.yml` from `main`; expect green run with no issue created.
- [ ] On a throwaway branch, temporarily add a fake context to the JSON and run `gh workflow run required-checks-drift.yml --ref <branch>`; expect failed run, new `Required-checks drift detected` issue labeled `required-checks-drift`. Re-run → expect a comment on the same issue, no duplicate.

## Testing summary

Local full-suite run blocked at `@lifting-logbook/api#test` global setup: Docker Desktop is not running in this worktree environment, so Testcontainers cannot provision Postgres for the DB E2E suite (`Could not find a working container runtime strategy`). This is the documented platform constraint in `CLAUDE.md` → Testing — not a regression. CI runs Docker in the service container and is unaffected. The change touches only `.github/`, `docs/`, and a new YAML — no application code or test setup is modified, and no test runner config / threshold is changed.

**Tests: 160 passed, 0 skipped, 1 errored — `@lifting-logbook/api#test` global setup (Docker not running locally; pre-existing platform constraint, CI unaffected).**

## Coverage gate

No new application behavior is added. The drift workflow itself is verified end-to-end via the post-merge `workflow_dispatch` steps in the test plan above; there is no in-repo unit test for it because it is a thin YAML wrapper around `gh api` + a `node` diff and depends on GitHub-side branch-protection state.

## Suppression / test-integrity check

`git diff origin/main` adds no `ts-ignore`, `eslint-disable`, `!`-suppression, `?? null` coercion, `it.skip`, `xit`, `passWithNoTests`, `--bail`, or `testPathIgnorePatterns` — and deletes no test files or coverage thresholds.

Closes #380

### DIFF
--- a/.github/expected-required-checks.json
+++ b/.github/expected-required-checks.json
@@ -1,0 +1,9 @@
+{
+  "contexts": [
+    "DB Integration Tests",
+    "Lint & Test",
+    "Observability Stack Smoke Test",
+    "Playwright E2E",
+    "Staging Integration Tests"
+  ]
+}
--- a/.github/workflows/required-checks-drift.yml
+++ b/.github/workflows/required-checks-drift.yml
@@ -16,16 +16,32 @@ permissions:
   contents: read
   issues: write
 
+# Serialize runs so a schedule tick and a workflow_dispatch (or push) cannot
+# both read an empty `gh issue list` and each create a duplicate drift issue.
+# cancel-in-progress: false — the second run still needs to complete and post
+# a comment / fail red; we don't want it cancelled.
+concurrency:
+  group: required-checks-drift
+  cancel-in-progress: false
+
 jobs:
   drift-check:
     name: Detect required-checks drift
     runs-on: ubuntu-latest
+    # The whole pipeline runs in <30s; 5 minutes is a generous bound on
+    # transient GitHub API hangs without burning the default 6h timeout.
+    timeout-minutes: 5
     steps:
       - uses: actions/checkout@v4
 
       - name: Fetch live required-status-checks
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Branch-protection endpoints require Administration:read, which
+          # GITHUB_TOKEN does not grant. BRANCH_PROTECTION_READ_TOKEN is a
+          # fine-grained PAT with that single scope. The fallback to
+          # GITHUB_TOKEN keeps the workflow loud-red if the secret is ever
+          # missing or expired (it 403s), rather than silently green.
+          GH_TOKEN: ${{ secrets.BRANCH_PROTECTION_READ_TOKEN || secrets.GITHUB_TOKEN }}
         run: |
           set -euo pipefail
           gh api "repos/${{ github.repository }}/branches/main/protection/required_status_checks" \
@@ -88,7 +104,18 @@ jobs:
           EOF
           echo "drift=$(cat drift.txt)" >> "$GITHUB_OUTPUT"
 
-      - name: Open or update drift issue
+      - name: Upload diagnostic artifacts
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: drift-diagnostics
+          path: |
+            actual.json
+            body.md
+          if-no-files-found: ignore
+          retention-days: 30
+
+      - name: Open or update drift issue(s)
         if: steps.diff.outputs.drift == '1'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -98,14 +125,19 @@ jobs:
             --description "Branch-protection required-check set diverges from .github/expected-required-checks.json" \
             --color B60205 \
             --force
-          EXISTING=$(gh issue list \
+          # Comment on every open drift issue (typically 0 or 1). Looping
+          # handles the rare case of multiple open issues (manual creation,
+          # reopened old issue) without silently ignoring extras.
+          mapfile -t EXISTING < <(gh issue list \
             --label required-checks-drift \
             --state open \
             --json number \
-            --jq '.[0].number // empty')
-          if [ -n "$EXISTING" ]; then
-            echo "Updating existing issue #$EXISTING"
-            gh issue comment "$EXISTING" --body-file body.md
+            --jq '.[].number')
+          if [ "${#EXISTING[@]}" -gt 0 ]; then
+            for N in "${EXISTING[@]}"; do
+              echo "Updating existing issue #$N"
+              gh issue comment "$N" --body-file body.md
+            done
           else
             echo "Creating new drift issue"
             gh issue create \
@@ -117,5 +149,5 @@ jobs:
       - name: Fail run when drift detected
         if: steps.diff.outputs.drift == '1'
         run: |
-          echo "Required-checks drift detected — see the 'Compute drift' step output and the linked issue."
+          echo "Required-checks drift detected — see the 'Compute drift' step output and the linked issue(s)."
           exit 1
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: [infra] Add required-checks drift detection workflow

### Summary

Merge with minor follow-ups. The PR is well-scoped (three files, ~140 lines, infra-only), the design choices are documented in the body, and the doc/test gates pass: no Documentation Maintenance table applies, no new application behavior, test-run summary present with the one errored suite explained as a documented platform constraint. No blocking findings. The notable gap is a TOCTOU window between `gh issue list` and `gh issue create` with no `concurrency:` group on the workflow — low likelihood in practice but the kind of thing that can produce a duplicate issue if a `schedule` tick and a `workflow_dispatch` run overlap.

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/required-checks-drift.yml:19`
The workflow has no `concurrency:` group. The triggers include `schedule`, `push` to `main` (on JSON/workflow edits), and `workflow_dispatch`. Two of these firing within ~30s of each other will race: both runs read an empty `gh issue list --label required-checks-drift --state open`, both create a new issue. The push+dispatch overlap is the most plausible — you edit the JSON, push, and manually dispatch to verify in the same minute.
**Suggestion:** Add `concurrency: { group: required-checks-drift, cancel-in-progress: false }` at workflow level so the second run queues behind the first. `cancel-in-progress: false` (not the default `true`) is intentional — you want the second run to complete and post a comment, not be cancelled.

---

**[reliability]** `.github/workflows/required-checks-drift.yml:101-105`
`gh issue list ... --jq '.[0].number // empty'` takes the first open issue with the `required-checks-drift` label. If a second drift issue ever exists (manual creation, race per above, or someone reopens an old one), the workflow silently comments on whichever sorts first and ignores the rest. The drift signal stays accurate, but the "single open tracking issue" invariant is assumed, not enforced.
**Suggestion:** Either assert at most one open issue (`jq 'length'` check, fail loudly if >1) or comment on all of them. Cheap, and it makes the invariant explicit.

---

**[maintainability]** `.github/workflows/required-checks-drift.yml:20`
No `timeout-minutes:` on the job. The default runner timeout is 6 hours — fine for a manual `gh api` + a small `node` block, but it leaves a hung step (e.g., a transient GitHub API hang) burning runner minutes until the default kicks in.
**Suggestion:** Add `timeout-minutes: 5` to the job. The whole pipeline runs in well under a minute today; 5 minutes is generous and keeps the failure mode bounded.

---

**[maintainability]** `.github/workflows/required-checks-drift.yml:33,78`
`actual.json` and `body.md` are written to the workspace but not uploaded as artifacts. On a failed scheduled run two days later, the diff is only in the run-log scrollback — fine for the current week's run, but log retention rotates and post-mortem becomes harder.
**Suggestion:** Add an `actions/upload-artifact@v4` step (always-run via `if: always()`) for `actual.json` and `body.md`. Negligible storage; saves a Slack-thread "what was the diff again?" question.

---

### Questions for Author

**Question:** Was leaving `GITHUB_TOKEN` (rather than a PAT) as the only auth path an intentional first-cut, with the 403-as-red-signal serving as the fallback?
**Context:** The plan in `~/.claude/plans/issue-380-stateless-wadler.md` notes the `BRANCH_PROTECTION_READ_TOKEN` PAT fallback but deferred wiring it. `repos/.../branches/main/protection/required_status_checks` requires admin access; the default `GITHUB_TOKEN` has admin scope on the repo it runs in only when the repo grants it via workflow permissions or org policy.
**Tradeoffs:** (a) Keep as-is — first 403 becomes the forcing function to add the PAT, no premature secret management. (b) Wire the PAT now via `${{ secrets.BRANCH_PROTECTION_READ_TOKEN || secrets.GITHUB_TOKEN }}` — survives a token-scope regression but introduces a secret that may not be needed. The plan's choice (a) is defensible; just confirming.

---

PR #383, opened 2026-05-29
[fix] Mount CLERK_SECRET_KEY in web deployments (prod auth broken)

### LINKED ISSUE #382: [bug] Production web missing CLERK_SECRET_KEY — auth middleware broken, /cycle returns 500
## Symptom

Visiting `https://liftinglogbook.com/` does not prompt for login — user lands on the marketing page and any deeper navigation (e.g. `/cycle`) returns HTTP 500. Expected behavior per PR #321: unauthenticated visitors are redirected to `/sign-in`.

Observed (clean browser, no cookies):

```
GET https://liftinglogbook.com/        -> 200 (marketing page renders; no redirect)
GET https://liftinglogbook.com/cycle   -> 500
```

## Root cause

The production web deployment never mounts `CLERK_SECRET_KEY`. Without it, `clerkMiddleware().auth.protect()` cannot verify sessions and the middleware throws — Next.js renders a 500 instead of the intended `/sign-in` redirect.

Evidence:

- `apps/web/middleware.ts:7-11` — middleware protects all non-public routes via `auth.protect()`, which requires `CLERK_SECRET_KEY` server-side.
- `infra/kubernetes/charts/web/templates/deployment.yaml:35-40` — only `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` is mounted; no `CLERK_SECRET_KEY` env entry.
- `.github/workflows/deploy.yml:464-473` — the "Sync web secrets to Kubernetes (production)" step writes only `clerk-publishable-key` into `web-secrets`. (The api step at line 454-462 does fetch `clerk-secret-key`, so the secret exists in Secret Manager.)
- `infra/cloud-run/web-service.yaml:34-45` — the 10% A/B Cloud Run web service is also missing the secret key.

Same gap exists in the staging web-secrets sync step (~line 270).

## Fix

1. `infra/kubernetes/charts/web/templates/deployment.yaml` — add `CLERK_SECRET_KEY` env entry sourced from `web-secrets` secret (`key: clerk-secret-key`).
2. `.github/workflows/deploy.yml` — extend both staging and production web-secrets sync steps to fetch `lifting-logbook-{staging,prod}-clerk-secret-key` from Secret Manager and include `--from-literal=clerk-secret-key="$CLERK_KEY"`.
3. `infra/cloud-run/web-service.yaml` — add `CLERK_SECRET_KEY` env entry sourced from the prod Secret Manager secret.

No app-code changes.

## Verification

After deploy:
- `GET /cycle` (no cookies) → 307 redirect to `/sign-in` (not 500).
- Signing in via Clerk succeeds and returns to the requested page.

brownm09:
## Summary

Production `liftinglogbook.com` does not prompt for login. Confirmed live:
- `GET /` → 200 (marketing page renders; no redirect to `/sign-in`)
- `GET /cycle` → **500**

Root cause: the web deployment never receives `CLERK_SECRET_KEY`. `apps/web/middleware.ts` calls `auth.protect()` on every non-public route, which requires the secret key server-side. Without it Clerk middleware throws — Next.js returns 500 instead of redirecting to `/sign-in`. The Clerk secret already exists in Secret Manager (the API workflow consumes it at deploy.yml:454); only the web side was missing.

## Changes

- `infra/kubernetes/charts/web/templates/deployment.yaml` — add `CLERK_SECRET_KEY` env entry sourced from the `web-secrets` Kubernetes Secret (key `clerk-secret-key`).
- `.github/workflows/deploy.yml` — staging and production "Sync web secrets to Kubernetes" steps now fetch `lifting-logbook-{stg,prod}-clerk-secret-key` and include `--from-literal=clerk-secret-key=…` in the `kubectl create secret` invocation.
- `infra/cloud-run/web-service.yaml` — add `CLERK_SECRET_KEY` env entry sourced from `lifting-logbook-prod-clerk-secret-key` Secret Manager secret (the 10% A/B Cloud Run target).

No app-code changes. Closes #382.

## Testing

Infra/YAML-only change; no application code or tests are modified. `npm test` was not run because the diff cannot affect any unit/integration suite. Pre-PR grep checks:

- Suppression policy grep against the diff: **0 matches**.
- Test-integrity grep against the diff: **0 matches**.

Verification will be live, post-deploy:

- [ ] After deploy: `curl -sI https://liftinglogbook.com/cycle` returns `307` redirecting to `/sign-in` (currently returns 500).
- [ ] After deploy: signing in via Clerk lands back on the requested page.
- [ ] `kubectl -n production get secret lifting-logbook-web-secrets -o jsonpath='{.data}'` shows both `clerk-publishable-key` and `clerk-secret-key`.
- [ ] `gcloud run services describe lifting-logbook-prod-web --region=us-central1 --format='value(spec.template.spec.containers[0].env[].name)'` includes `CLERK_SECRET_KEY`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -273,9 +273,11 @@ jobs:
           CLERK_PUB=$(gcloud secrets versions access latest \
             --secret="lifting-logbook-stg-clerk-publishable-key" \
             --project="${{ vars.GCP_STAGING_PROJECT_ID }}")
+          [ -n "$CLERK_PUB" ] || { echo "ERROR: clerk-publishable-key fetch returned empty (staging)"; exit 1; }
           CLERK_KEY=$(gcloud secrets versions access latest \
             --secret="lifting-logbook-stg-clerk-secret-key" \
             --project="${{ vars.GCP_STAGING_PROJECT_ID }}")
+          [ -n "$CLERK_KEY" ] || { echo "ERROR: clerk-secret-key fetch returned empty (staging)"; exit 1; }
           kubectl create secret generic web-secrets \
             --namespace=staging \
             --from-literal=clerk-publishable-key="$CLERK_PUB" \
@@ -471,9 +473,11 @@ jobs:
           CLERK_PUB=$(gcloud secrets versions access latest \
             --secret="lifting-logbook-prod-clerk-publishable-key" \
             --project="${{ vars.GCP_PROD_PROJECT_ID }}")
+          [ -n "$CLERK_PUB" ] || { echo "ERROR: clerk-publishable-key fetch returned empty (production)"; exit 1; }
           CLERK_KEY=$(gcloud secrets versions access latest \
             --secret="lifting-logbook-prod-clerk-secret-key" \
             --project="${{ vars.GCP_PROD_PROJECT_ID }}")
+          [ -n "$CLERK_KEY" ] || { echo "ERROR: clerk-secret-key fetch returned empty (production)"; exit 1; }
           kubectl create secret generic web-secrets \
             --namespace=production \
             --from-literal=clerk-publishable-key="$CLERK_PUB" \
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -273,9 +273,13 @@ jobs:
           CLERK_PUB=$(gcloud secrets versions access latest \
             --secret="lifting-logbook-stg-clerk-publishable-key" \
             --project="${{ vars.GCP_STAGING_PROJECT_ID }}")
+          CLERK_KEY=$(gcloud secrets versions access latest \
+            --secret="lifting-logbook-stg-clerk-secret-key" \
+            --project="${{ vars.GCP_STAGING_PROJECT_ID }}")
           kubectl create secret generic web-secrets \
             --namespace=staging \
             --from-literal=clerk-publishable-key="$CLERK_PUB" \
+            --from-literal=clerk-secret-key="$CLERK_KEY" \
             --dry-run=client -o yaml | kubectl apply -f -
 
       - name: Helm deploy API (GKE staging)
@@ -467,9 +471,13 @@ jobs:
           CLERK_PUB=$(gcloud secrets versions access latest \
             --secret="lifting-logbook-prod-clerk-publishable-key" \
             --project="${{ vars.GCP_PROD_PROJECT_ID }}")
+          CLERK_KEY=$(gcloud secrets versions access latest \
+            --secret="lifting-logbook-prod-clerk-secret-key" \
+            --project="${{ vars.GCP_PROD_PROJECT_ID }}")
           kubectl create secret generic web-secrets \
             --namespace=production \
             --from-literal=clerk-publishable-key="$CLERK_PUB" \
+            --from-literal=clerk-secret-key="$CLERK_KEY" \
             --dry-run=client -o yaml | kubectl apply -f -
 
       - name: Helm deploy API (GKE production)
--- a/infra/cloud-run/web-service.yaml
+++ b/infra/cloud-run/web-service.yaml
@@ -43,3 +43,8 @@ spec:
                 secretKeyRef:
                   name: lifting-logbook-prod-clerk-publishable-key
                   key: latest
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: [fix] Mount CLERK_SECRET_KEY in web deployments (prod auth broken)

### Summary
Merge with one verification. The diff is minimal (18 lines, three files), tightly scoped to the root cause, and consistent across all three deploy surfaces (GKE Helm, Cloud Run YAML, deploy.yml secret-sync). The only thing that could break the staging path on first deploy is whether `lifting-logbook-stg-clerk-secret-key` actually exists in the staging GCP project's Secret Manager — that should be confirmed before merge to avoid a red staging deploy. No correctness, security, or maintainability blockers in the change itself.

---

### Blocking Findings

*None.*

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/deploy.yml:276-278`
Staging now fetches `lifting-logbook-stg-clerk-secret-key`. The production path is known-good (the API workflow at line ~454 already consumes the prod variant successfully), but there is no prior consumer of the *staging* secret in this workflow — `api-secrets` and `web-secrets` for staging both only mounted the publishable key previously. If the staging secret was never created in Secret Manager, the first deploy after merge will fail with `ERROR: (gcloud.secrets.versions.access) NOT_FOUND` and abort the workflow before `kubectl apply` runs.
**Suggestion:** Before merging, run `gcloud secrets list --project=$STAGING_PROJECT | grep clerk-secret-key` to confirm the staging secret exists. If it does not, create it with the same Clerk staging instance secret key before merging — otherwise the staging job will go red and block the rollout.

**[reliability]** `.github/workflows/deploy.yml:281,479`
The `kubectl create secret … --dry-run=client -o yaml | kubectl apply -f -` pattern is the right idempotency idiom, but with a `pipefail`-aware shell a failure in `gcloud secrets versions access` still leaves `$CLERK_KEY` empty and `kubectl apply` then writes an empty `clerk-secret-key` to the Secret. The pod restart will then loop on Clerk init failure exactly as it does today — the workflow will show green while production stays broken.
**Suggestion:** Add a sanity check after the gcloud call (e.g., `[ -n "$CLERK_KEY" ] || { echo "clerk-secret-key fetch returned empty"; exit 1; }`) for both `CLERK_PUB` and `CLERK_KEY` in both staging and production steps. Cheap insurance against silent regressions of exactly the class of bug this PR is fixing.

**[maintainability]** `infra/cloud-run/web-service.yaml:46-50` / `infra/kubernetes/charts/web/templates/deployment.yaml:41-45`
Two deploy targets now mount the same Clerk env vars from two differently-shaped sources (Cloud Run pulls from Secret Manager directly via `secretKeyRef.name=lifting-logbook-prod-clerk-secret-key`; GKE pulls from a synthesized `web-secrets` k8s Secret that the workflow rebuilds on every deploy). Both are correct; they will drift independently if a future change adds, rotates, or renames a key on only one side. The bug this PR fixes is itself an instance of that drift (publishable-key was wired on both, secret-key was wired on neither).
**Suggestion:** Consider a single source-of-truth list in the workflow (or a values file) enumerating the keys both targets need, asserted at deploy time. Not for this PR — it is the kind of follow-up that prevents the next variant of #382.

---

### Questions for Author

**Question:** Should production deploys verify the env was actually injected before marking the rollout healthy?
**Context:** `infra/kubernetes/charts/web/templates/deployment.yaml:48-51` defines liveness/readiness probes (values.yaml-driven), but Next.js will serve a 200 on `/` from the static prerender even when middleware is throwing on every other route — exactly the failure mode this PR fixes. A probe that hits `/` would have stayed green through the entire outage.
**Tradeoffs:** (a) Point the readiness probe at a path that goes through middleware (e.g., `/cycle`) — would have failed-open the broken deploy and prevented promotion, but requires the probe to be tolerant of 307s. (b) Keep `/` as the probe target and add a separate post-deploy smoke check in the workflow that curls `/cycle` and asserts `307` — slower feedback but no probe semantics changes. (c) Status quo + this fix — accept that "static `/` returns 200" is sufficient signal and rely on the auth-posture follow-up to prevent the class.

---

### Style Notes

Nothing to flag. The new YAML blocks mirror the existing publishable-key blocks exactly (same indentation, same key naming style, same secretKeyRef shape) which is the right call for a fix this targeted.

---

PR #390, opened 2026-05-31
[fix] Add defense-in-depth auth checks at page/layout level (#384)

### LINKED ISSUE #384: [fix] Add defense-in-depth auth checks at page/layout level (don't rely solely on middleware)
## Problem

The entire production auth posture currently relies on a single layer: `apps/web/middleware.ts`'s `clerkMiddleware().auth.protect()` call. There is no defense-in-depth at the page or layout level:

- `apps/web/app/page.tsx` — static marketing page; no `auth()` check, no redirect for authenticated users.
- `apps/web/app/onboarding/page.tsx` — renders `<OnboardingFlow />` unconditionally; no `auth()` check.
- No protected-segment layout (e.g., `app/(authed)/layout.tsx`) that calls `auth()` server-side.

This was made concrete by #382 / #383: a missing `CLERK_SECRET_KEY` caused the middleware to throw on every protected route. The site did not fail closed — `/` returned a cached 200 from the edge and `/cycle` returned 500. From a user's perspective the marketing page was open with a "Get Started" link straight into `/onboarding`. Any future middleware misconfiguration (bad matcher, swallowed error, accidental `auth.protect()` removal, edge-cache bypass change) reopens the entire site exactly the same way.

## Proposal

Add page-level / layout-level `auth()` calls as defense-in-depth. Two complementary changes:

1. **Protected-segment layout.** Group the authenticated routes under a route group (e.g., `app/(authed)/` containing `cycle/`, `onboarding/`, and any future authenticated pages) with a `layout.tsx` that calls `auth()` server-side and redirects unauthenticated requests to `/sign-in`. A failure in middleware no longer opens these pages — the layout would still 307.

2. **Root page redirect for authenticated users.** `app/page.tsx` should call `auth()`; if the user is signed in, redirect to `/cycle` (or wherever the authenticated landing belongs). This means a signed-in user never sees the marketing card and the static prerender no longer masks failures.

## Out of scope

- Whether `/sign-in` and `/sign-up` should also have any guards (they should remain public).
- Migrating the entire app to a single auth gate that replaces middleware. This issue proposes *additional* layers, not a replacement.

## Acceptance criteria

- [ ] A route-group layout (or equivalent) guards every currently-protected page with a server-side `auth()` call.
- [ ] `app/page.tsx` redirects signed-in users to the authenticated landing page.
- [ ] A test (Jest server-component or Playwright once #259 lands) demonstrates that if `clerkMiddleware` is bypassed/disabled, the protected pages still 307 to `/sign-in`. Until #259, a documented manual verification plan suffices.

## References

- Originating outage: #382
- Originating fix: #383
- Middleware: `apps/web/middleware.ts:7-11`
- Clerk server-side auth helpers: https://clerk.com/docs/references/nextjs/auth

brownm09:
## Summary

Adds server-side `auth()` checks at the layout/page level so a middleware failure no longer opens protected pages, addressing the failure mode demonstrated by #382/#383.

- New `app/(authed)/layout.tsx` calls `auth()` and redirects unauthenticated requests to `/sign-in`. All currently-protected segments (cycle, onboarding, history, programs, settings) moved into the route group via `git mv` so URLs are unchanged and blame is preserved.
- `app/page.tsx` converted to an async server component that redirects signed-in users to `/cycle`. The `auth()` call opts the route out of static prerender, removing the cached-200 failure mode.
- Both gates honor the existing `DEV_AUTH_TOKEN` escape hatch, mirroring the skip in `apps/web/middleware.ts`.

Middleware (`auth.protect()`) is unchanged — this is defense-in-depth, not a replacement.

## Acceptance criteria

- [x] A route-group layout guards every currently-protected page with a server-side `auth()` call — `app/(authed)/layout.tsx`.
- [x] `app/page.tsx` redirects signed-in users to the authenticated landing page (`/cycle`).
- [x] Tests demonstrate that if Clerk auth fails to populate a userId, protected pages 307 to `/sign-in` — see `app/(authed)/layout.test.tsx` and `app/page.test.tsx` (3 cases each: unauth, auth, DEV_AUTH_TOKEN bypass).

## Testing

Ran `npm test -w @lifting-logbook/web` and the non-API workspaces. API DB E2E suite could not run locally due to Docker Desktop daemon returning 500 (engine not running in this worktree environment); CI runs the same suite against a service container and will execute it on this PR.

\`\`\`
Tests: 34 passed, 0 skipped, 0 failed (~30s)   # @lifting-logbook/web
\`\`\`

Other workspaces (`core`, `types`, `eslint-rules`) all green.

**Coverage gate:** New behavior covered — 6 new test cases across the two new test files. Each verifies redirect, auth-pass, and DEV_AUTH_TOKEN bypass.

**Suppression check:** clean (no `ts-ignore`, `eslint-disable`, `!` assertions, or `?? null` added).

**Test-integrity check:** clean (no skip markers, deleted tests, lowered thresholds, or bypass flags).

**Failure-tracking note:** the API DB E2E failure is environmental (local Docker daemon unhealthy) and unrelated to this change. CI will run that suite against its own Postgres service container — if it fails there it must be addressed before merge.

## Manual verification plan

Until #259 (Playwright) lands:

1. Locally unset `CLERK_SECRET_KEY` and `DEV_AUTH_TOKEN`, set `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` to any valid-looking value, `npm run dev -w @lifting-logbook/web`.
2. Visit `/cycle` while signed out. Expected: 307 to `/sign-in` from the layout even if middleware errors. Today's behavior (without this PR) was a 500.
3. Sign in, visit `/`. Expected: 307 to `/cycle`. Today: marketing card renders for signed-in users.

## Out of scope (per issue)

- Guards on `/sign-in`, `/sign-up` (remain public).
- Removing middleware (`auth.protect()` stays).

Closes #384


brownm09:
## Review: [fix] Add defense-in-depth auth checks at page/layout level (#384)

### Summary
Merge with minor changes considered. The change is small, well-scoped, and faithfully matches the issue's three acceptance criteria. The two new gates (`(authed)/layout.tsx` and `app/page.tsx`) are minimal, the route-group restructure is a pure `git mv` of 65 files with no content edits, and the six new test cases cover the unauth / auth / `DEV_AUTH_TOKEN`-bypass matrix on both gates. Highest-severity observation is non-blocking: the new layer's `DEV_AUTH_TOKEN` escape hatch is copy-paste-identical to the middleware's, so a single misconfigured env var still bypasses *both* layers — the defense-in-depth story has a shared single-point-of-failure that the PR could note explicitly.

---

### Non-Blocking Findings

**[security]** `apps/web/app/(authed)/layout.tsx:9` and `apps/web/app/page.tsx:7`
Both new gates skip `auth()` entirely when `process.env.DEV_AUTH_TOKEN` is any truthy string — the exact pattern used by `apps/web/middleware.ts`. The PR's rationale is "defense against middleware misconfiguration", but if the misconfiguration is "`DEV_AUTH_TOKEN` accidentally set in a deployed environment," middleware *and* both new gates fail open identically. The layers are not independent against this specific class of failure.
**Suggestion:** Either gate the escape hatch on `process.env.NODE_ENV !== 'production'` in the new files (so production cannot bypass auth via this var no matter what), or add a one-line comment in `(authed)/layout.tsx` acknowledging that the escape-hatch parity is intentional and that production hardening (e.g., a Cloud Run startup probe that fails when `DEV_AUTH_TOKEN` is set) is tracked separately.

---

**[maintainability]** `apps/web/app/(authed)/layout.test.tsx:48` and `apps/web/app/page.test.tsx:45`
The "auth passes" and "DEV_AUTH_TOKEN bypass" cases assert `expect(element).toBeTruthy()` on the returned React element. Any non-null return — including an accidentally empty fragment, a regression that returns the unauthenticated marketing card from the authed layout, or a no-op `<></>` — satisfies this assertion.
**Suggestion:** Tighten to a more specific check: for the layout, assert the returned element's type is `React.Fragment` and `props.children === 'protected'`; for the home page, assert a recognizable marker from the marketing card (e.g., `JSON.stringify(element).includes('Lifting Logbook')` or render with `react-test-renderer` and query for the wordmark). The mocked `redirect` throw already covers the redirect path, so the positive paths are where the assertion strength matters.

---

**[maintainability]** `apps/web/app/(authed)/layout.tsx:1-13`
No file-level comment explains why the `DEV_AUTH_TOKEN` branch exists or that the layer is intentional duplication of middleware. The next reader who notices "middleware already does this" may delete the layout to remove the duplication and reintroduce the #382/#383 failure mode.
**Suggestion:** Add a 2–3 line JSDoc above `AuthedLayout` stating: this is defense-in-depth against middleware misconfiguration (link to #384 / #382), `DEV_AUTH_TOKEN` skip mirrors middleware to keep mock-API dev mode working, do not remove without a replacement layer.

---

### Questions for Author

**Question:** Should `page.tsx`'s signed-in → `/cycle` redirect honor `DEV_AUTH_TOKEN` too, or always redirect signed-in users regardless?
**Context:** [`apps/web/app/page.tsx:7-10`](apps/web/app/page.tsx#L7-L10) wraps the signed-in redirect inside `if (!process.env.DEV_AUTH_TOKEN)`, so in mock-API dev mode a "signed-in" user still sees the marketing card. This mirrors the layout for consistency, but the layout's bypass exists so unauthenticated mock-API dev sessions can reach protected pages — the home page doesn't have the same need.
**Tradeoffs:** (a) Keep the bypass — symmetry with the layout, no special-case behavior to remember when toggling dev mode. (b) Drop the bypass on the home page only — signed-in users always go to `/cycle`, simplifying the dev-mode mental model, but introduces an asymmetry between the two new gates that needs comment. (c) Add a follow-up issue to remove `DEV_AUTH_TOKEN` entirely once mock-API mode has a cleaner replacement.

---

*Reviewed by `claude-sonnet-4.5` via Claude Code*

brownm09:
Stale CI noise: that run (26734574578) was on commit `9bf9d24` before the lint failure was fixed. The job's "Wait for staging to be ready" step had `WEB_URL=""`, so it never reached Playwright. The post-rebase run (26734795667) is currently re-deploying staging on the up-to-date head; integration tests will run against the new deploy. No action needed on the stale comment.

PR #387, opened 2026-05-29
[fix] Pass NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY to web image build

brownm09:
## Summary

Production deploys to `main` have been silently failing on the **Build & Push Images** job for at least three commits (`026f46f`, `c3c1093`, `0cbc16c`). The web Dockerfile declares `ARG NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` because Next.js needs it during `/_not-found` prerender (the root layout's `<ClerkProvider>` throws "Missing publishableKey" otherwise), but `deploy.yml`'s `Build and push web image` step only passed `NEXT_PUBLIC_API_URL` as a build-arg. The result: no new web image has been pushed to Artifact Registry recently, and the Clerk runtime fix from #383 was sitting at the gate.

This is the build-time analogue of #382/#383 — same root cause class (Clerk publishable key missing from web deploy pipeline), different layer (image build vs. runtime env).

## Changes

- `.github/workflows/deploy.yml` — new `Resolve Clerk publishable key for web image build` step that fetches `lifting-logbook-{stg,prod}-clerk-publishable-key` from Secret Manager (staging vs. prod mode same as the existing `Resolve API URL` step). Passes the value through to `Build and push web image` as a second build-arg. Includes the same empty-value guard pattern as #383. Masks the key in logs via `::add-mask::` for defense in depth, even though Clerk publishable keys are non-secret by design.

Closes #386.

## Testing

Infra/workflow-only change; no application code or tests are modified. `npm test` was not run because the diff cannot affect any unit/integration suite. Pre-PR grep checks:

- Suppression policy grep against the diff: **0 matches**.
- Test-integrity grep against the diff: **0 matches**.

The actual verification surface is post-merge — the CI build itself is the test:

- [ ] Post-merge `Build & Push Images` job succeeds on commit landing in main (previously failing for 3+ consecutive commits).
- [ ] `deploy-staging` consumes the new image and passes smoke + integration tests.
- [ ] Production deploy completes and `curl -sI https://liftinglogbook.com/cycle` returns 307→`/sign-in`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #391, opened 2026-05-31
[fix] Mount Clerk secret in Cloud Run web + add logout button

### LINKED ISSUE #389: [fix] Cloud Run web prod/staging missing CLERK_SECRET_KEY at runtime; no logout UI
## Problem

Production web does not redirect unauthenticated users to `/sign-in`, and there is no logout button anywhere in the UI.

## Root causes

**1. Cloud Run web (staging + prod) is missing `CLERK_SECRET_KEY` at runtime.**

PR #383 mounted `CLERK_SECRET_KEY` in the GKE Helm chart and synced the Secret via `kubectl create secret`. But the Cloud Run web deploys go through `gcloud run deploy ... --set-env-vars=\"API_URL=...\"` with **no** `--set-secrets` / `--update-secrets` flag (`.github/workflows/deploy.yml` ~lines 355-365 staging, 554-564 prod). `infra/cloud-run/web-service.yaml` references the secret correctly but is reference-only — never applied via `gcloud run services replace`. Without `CLERK_SECRET_KEY` in the container, `clerkMiddleware().auth.protect()` cannot initialize and requests fall through unauthenticated, so the home/cycle pages render with no login prompt.

GKE is not affected because the chart sets a `version: <image.tag>` pod label, which changes on every deploy and forces a new ReplicaSet, so pods always restart and re-read the Secret.

**2. No Clerk UI in the app shell.** `apps/web/app/cycle/layout.tsx` renders a plain header with no `<UserButton />`, `<SignOutButton>`, or `<SignedIn>` component. Even authenticated users have no way to sign out.

## Fix

- `.github/workflows/deploy.yml`: add `--update-secrets=\"CLERK_SECRET_KEY=lifting-logbook-{stg,prod}-clerk-secret-key:latest,NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=lifting-logbook-{stg,prod}-clerk-publishable-key:latest\"` to both `gcloud run deploy ... -web` steps.
- `apps/web/app/cycle/layout.tsx`: import and render `<UserButton afterSignOutUrl=\"/sign-in\" />` in the header.

## Acceptance

- [ ] Incognito visit to staging web URL redirects to `/sign-in` on both Cloud Run and GKE paths.
- [ ] After sign-in, `/cycle` renders a `<UserButton />` in the header.
- [ ] Clicking the button shows Clerk popover with Sign out; signing out returns to `/sign-in`.
- [ ] Same verified on production.

brownm09:
## Summary

Closes #389. Two distinct root causes behind "production doesn't ask me to log in" and "no logout button":

1. **Cloud Run web (staging + prod) was missing `CLERK_SECRET_KEY` at runtime.** PR #383 mounted the secret in GKE via Helm, but the Cloud Run `gcloud run deploy` steps in `deploy.yml` only set `API_URL` via `--set-env-vars` with no `--update-secrets` flag. `infra/cloud-run/web-service.yaml` references the secret correctly but is never applied via `gcloud run services replace`. Without the secret in the container, `clerkMiddleware().auth.protect()` cannot initialize and requests fall through unauthenticated, so home/cycle pages render with no login prompt.
2. **No Clerk UI in the app shell.** `apps/web/app/cycle/layout.tsx` rendered a plain header with no `<UserButton />` / `<SignOutButton>`. Authenticated users had no way to sign out.

GKE is unaffected because the Helm chart sets `version: <image.tag>` as a pod label, which changes on every deploy and forces a new ReplicaSet — pods always restart and re-read the Secret. So no Helm change is needed.

## Changes

- `.github/workflows/deploy.yml`: add `--update-secrets="CLERK_SECRET_KEY=...,NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=..."` to both `gcloud run deploy ... -web` steps (staging + prod).
- `apps/web/app/cycle/layout.tsx`: render `<UserButton afterSignOutUrl="/sign-in" />` in the header.

## Test plan

- [x] `npm test -w @lifting-logbook/web`: **Tests: 28 passed, 0 skipped, 0 failed (32.6s)**.
- [x] `npx tsc --noEmit -p apps/web/tsconfig.json`: clean.
- [x] Full repo `npm test`: 11/12 workspaces passed. `@lifting-logbook/api` DB E2E failed because Docker Desktop is not running on this machine — Testcontainers cannot start the Postgres container. Per `CLAUDE.md` this is a documented local environment prerequisite, not a regression introduced by this PR. CI runs the suite with a service container and will validate.
- [x] Pre-PR suppression grep: clean.
- [x] Pre-PR test-integrity grep: clean.
- [ ] After staging deploy: incognito visit to staging web URL redirects to `/sign-in`. Sign in, navigate to `/cycle`, confirm `<UserButton />` is rendered and signs out cleanly.
- [ ] Same verification on production after promotion.

## Notes

- `next build` does not run locally due to the documented Node 24 + Windows tarball-extraction issue ([#373](https://github.com/brownm09/lifting-logbook/issues/373)). CI runs Node 20 and is unaffected. The change is contained: one YAML flag and one React import + JSX element.
- The `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` is already build-embedded by PR #387; mounting it as a runtime env too is redundant but matches `web-service.yaml`'s intent and is harmless.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -362,6 +362,7 @@ jobs:
             --allow-unauthenticated \
             --service-account="${{ needs.terraform-staging.outputs.web_workload_sa }}" \
             --set-env-vars="API_URL=${{ needs.terraform-staging.outputs.cloud_run_api_url }}" \
+            --update-secrets="CLERK_SECRET_KEY=lifting-logbook-stg-clerk-secret-key:latest,NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=lifting-logbook-stg-clerk-publishable-key:latest" \
             --quiet
 
   # ── 4. Smoke test staging ────────────────────────────────────────────────
@@ -561,6 +562,7 @@ jobs:
             --allow-unauthenticated \
             --service-account="${{ steps.tf-prod.outputs.web_workload_sa }}" \
             --set-env-vars="API_URL=${{ steps.tf-prod.outputs.cloud_run_api_url }}" \
+            --update-secrets="CLERK_SECRET_KEY=lifting-logbook-prod-clerk-secret-key:latest,NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=lifting-logbook-prod-clerk-publishable-key:latest" \
             --quiet
 
       - name: Print deployment URLs
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -362,7 +362,7 @@ jobs:
             --allow-unauthenticated \
             --service-account="${{ needs.terraform-staging.outputs.web_workload_sa }}" \
             --set-env-vars="API_URL=${{ needs.terraform-staging.outputs.cloud_run_api_url }}" \
-            --update-secrets="CLERK_SECRET_KEY=lifting-logbook-stg-clerk-secret-key:latest,NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=lifting-logbook-stg-clerk-publishable-key:latest" \
+            --update-secrets="CLERK_SECRET_KEY=lifting-logbook-stg-clerk-secret-key:latest" \
             --quiet
 
   # ── 4. Smoke test staging ────────────────────────────────────────────────
@@ -562,7 +562,7 @@ jobs:
             --allow-unauthenticated \
             --service-account="${{ steps.tf-prod.outputs.web_workload_sa }}" \
             --set-env-vars="API_URL=${{ steps.tf-prod.outputs.cloud_run_api_url }}" \
-            --update-secrets="CLERK_SECRET_KEY=lifting-logbook-prod-clerk-secret-key:latest,NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY=lifting-logbook-prod-clerk-publishable-key:latest" \
+            --update-secrets="CLERK_SECRET_KEY=lifting-logbook-prod-clerk-secret-key:latest" \
             --quiet
 
       - name: Print deployment URLs
--- a/apps/web/app/cycle/layout.tsx
+++ b/apps/web/app/cycle/layout.tsx
@@ -1,5 +1,3 @@
-import { UserButton } from '@clerk/nextjs';
-
 export default function CycleLayout({
   children,
 }: {
@@ -14,7 +12,6 @@ export default function CycleLayout({
           <a href="/programs">Programs</a>
           <a href="/settings/training-maxes">Settings</a>
         </nav>
-        <UserButton afterSignOutUrl="/sign-in" />
       </header>
       {children}
     </main>
--- a/apps/web/app/cycle/layout.tsx
+++ b/apps/web/app/cycle/layout.tsx
@@ -1,3 +1,5 @@
+import { UserButton } from '@clerk/nextjs';
+
 export default function CycleLayout({
   children,
 }: {
@@ -12,6 +14,7 @@ export default function CycleLayout({
           <a href="/programs">Programs</a>
           <a href="/settings/training-maxes">Settings</a>
         </nav>
+        <UserButton afterSignOutUrl="/sign-in" />
       </header>
       {children}
     </main>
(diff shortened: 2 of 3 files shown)


brownm09:
## Review: [fix] Mount Clerk secret in Cloud Run web + add logout button

### Summary

Merge with minor changes. The two fixes correctly address the diagnosed root causes — `--update-secrets` on both Cloud Run web deploy steps and a `UserButton` in the cycle layout. The diff is tightly scoped (7 added lines, 2 files). One non-blocking UX gap: `UserButton` only appears on routes nested under `/cycle/`, leaving `/history`, `/programs`, `/settings/*`, `/onboarding`, and `/` (home) with no sign-out affordance. One question for the author about runtime vs. build-time publishable-key consistency.

Pre-PR gates: suppression grep clean, integrity grep clean, test summary present (`28 passed, 0 skipped, 0 failed`), `Closes #389` present, no skill/hook/script/routine touched so no doc-reconciliation finding. No new testable behavior in `apps/web` beyond a single Clerk component import — the existing `npm test -w @lifting-logbook/web` run is sufficient; deploy.yml changes are config-only and exercised by the staging-then-prod pipeline.

---

### Non-Blocking Findings

**[maintainability]** `apps/web/app/cycle/layout.tsx:17`
`UserButton` is mounted only inside `CycleLayout`, but the App Router tree shows `history`, `programs`, `settings`, `onboarding`, and the root `page.tsx` as sibling route segments that do not inherit this layout. A signed-in user who lands on `/history` or `/settings/training-maxes` (both linked from the same nav this PR edits) sees no sign-out affordance.
**Suggestion:** Either lift `UserButton` into `apps/web/app/layout.tsx` (wrapped in `<SignedIn>` so it does not render on `/sign-in` / `/sign-up`), or extract a small `<AppHeader />` component shared by the cycle, history, programs, and settings layouts. The latter avoids duplicating the nav links across four layouts as the app grows.

**[reliability]** `.github/workflows/deploy.yml:365,565`
`NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` is now provided both at build time (PR #387 build-arg, baked into the client bundle) and at runtime via `--update-secrets`. If the Secret Manager value ever drifts from the value passed as a build-arg, the server side will read the new key while the already-built client bundle keeps the old one — Clerk fails with a key-mismatch error that is hard to attribute. The PR body notes the runtime mount is "redundant but harmless"; it is only harmless while the two values stay in sync.
**Suggestion:** Source the build-arg from the same Secret Manager entry (e.g., `gcloud secrets versions access latest --secret=...` in the build step) so a single rotation updates both sides, or drop the runtime mount of the publishable key and rely solely on the build-time value (the secret key still needs the runtime mount).

---

### Questions for Author

**Question:** Have the Secret Manager entries `lifting-logbook-stg-clerk-secret-key`, `lifting-logbook-stg-clerk-publishable-key`, `lifting-logbook-prod-clerk-secret-key`, and `lifting-logbook-prod-clerk-publishable-key` already been created in the respective GCP projects, and does the Cloud Run web service account have `secretmanager.secretAccessor` on each?
**Context:** PR #383 added the same-named secrets for the GKE path, but Cloud Run runs under a different workload identity (`needs.terraform-staging.outputs.web_workload_sa` and `steps.tf-prod.outputs.web_workload_sa`). If the secret bindings only granted access to the GKE workload SA, the new `--update-secrets` flag will make the deploy fail with `PERMISSION_DENIED: Permission 'secretmanager.versions.access' denied`.
**Tradeoffs:** (a) The IAM bindings already cover both SAs — nothing to do. (b) Bindings cover only the GKE SA — Terraform needs an additional `google_secret_manager_secret_iam_member` for each web workload SA before this PR is safe to merge. (c) Use a single `web` SA shared between Cloud Run and GKE — simpler IAM but couples failure domains.

---

brownm09:
Self-answered the IAM question after posting:

- `infra/terraform/cloud-run.tf:221-231` already binds `roles/secretmanager.secretAccessor` on both `clerk_publishable_key` and `clerk_secret_key` to `google_service_account.web_workload`.
- That SA is exactly what `deploy.yml` passes as `--service-account` via `needs.terraform-staging.outputs.web_workload_sa` / `steps.tf-prod.outputs.web_workload_sa`.

So `--update-secrets` should succeed on first deploy. IAM concern resolved.

Separately noted (not blocking): `cloud-run.tf:79-86, 154-172` already declares these env vars on the Terraform-managed Cloud Run resource, but the deploy pipeline uses imperative `gcloud run deploy` instead of `terraform apply`, so the Terraform env block is never realized. Worth a follow-up to consolidate the source of truth.

PR #392, opened 2026-06-01
[infra] Build web image per-env to stop baking staging values into prod

### LINKED ISSUE #388: [design] Web image deploys to both envs but bakes env-specific NEXT_PUBLIC_* values at build time
## Problem

The web image build (`apps/web/Dockerfile`) consumes `NEXT_PUBLIC_*` build-args that are **environment-specific** but the resulting image is **deployed to both staging and production** by tag (`web:${{ github.sha }}`). Concretely, when `staging_enabled == true` in `.github/workflows/deploy.yml`:

- `NEXT_PUBLIC_API_URL` resolves to the *staging* Cloud Run API URL (line ~170).
- `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` (added in #387) resolves to the *staging* Clerk publishable key.

Both values are then embedded into the Next.js JS bundle at build time. The same image is later promoted unchanged to production by both `deploy-staging` and `deploy-production`, so the prod web pod runs a bundle that hard-codes staging values for these two variables.

If staging and prod share the same backing services (single Clerk instance, single API URL aliased internally), this is silently a no-op. The moment a real second environment exists for either — separate Clerk instance for staging, separate API host, separate publishable key — this breaks in production while staging stays green, exactly the failure mode #382 exhibited at a different layer.

## Why this matters now

PRs #383 (runtime CLERK_SECRET_KEY) and #387 (build-time NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY) both fixed instances of the broader pattern "Clerk credential not wired through deploy pipeline." Both fixes preserved the inherited "use staging value, deploy to both" shortcut. This issue tracks fixing the root pattern so future env-specific values don't repeat the same bug.

## Proposal sketches (not exhaustive)

1. **Per-env image tags.** Build twice — `web:${{ github.sha }}-staging` and `web:${{ github.sha }}-prod` — with env-specific build-args. Doubles image storage and build time but is the cleanest contract.
2. **Move to runtime config.** Refactor `apps/web` so the API URL and Clerk publishable key are read at runtime (from a server-rendered injection or `getServerSideProps`-style data) instead of embedded at build time. Eliminates the problem entirely but requires app changes.
3. **Promote-not-rebuild + staging-only build.** Keep one image per commit, build with prod values, deploy to staging too (with the prod values embedded). Inverts the current direction. Trivial to implement but couples staging to prod values.

Decision is design-level — file separately to scope properly.

## Acceptance criteria

- [ ] An ADR records the chosen approach.
- [ ] The web image build pipeline produces correct env-specific values for whichever env it deploys to.
- [ ] A deliberate dry-run validates that staging-specific values do not appear in the production bundle.

## References

- Originating runtime fix: #383
- Originating build-time fix: #387
- Inherited pattern source: PR #321 (Clerk introduction) and the prior NEXT_PUBLIC_API_URL wiring
- Failing run that surfaced this in #386: https://github.com/brownm09/lifting-logbook/actions/runs/26657923819

brownm09:
## Summary

- Build the `apps/web` image twice per pipeline run with per-env `NEXT_PUBLIC_*` build-args: `web:<sha>-staging` (staging-only) and `web:<sha>-prod` (production-only, also `:latest`).
- Update `deploy-staging` and `deploy-production` to reference the env-specific tag for both GKE (Helm) and Cloud Run deploys.
- New ADR-025 records the decision and the explicit trade (staging gate no longer validates prod-embedded public values); `docs/deploy.md` adds a checklist for new `NEXT_PUBLIC_*` variables, a bundle-grep verification procedure, and a first-time prod bootstrap recovery procedure.

Closes #388.

## Why

Root-pattern fix following the runtime (#383) and build-time (#387) Clerk credential fixes. Previously a single `web:<sha>` image was built with staging values when `staging_enabled == true` and promoted unchanged to production. The prod web pod ran a JS bundle with staging values inlined — silent today because staging/prod share infra, silently broken the moment they don't (same shape as #382).

Phase 2 (runtime public-config refactor, will supersede ADR-025) is tracked as a follow-up issue per the ADR — it touches `apps/web` source and the `<ClerkProvider>` bootstrap path, so it shouldn't block closing #388.

## Pipeline shape (build-images)

1. Resolve staging values (gated on `staging_enabled`).
2. Re-auth to prod, resolve prod values (always).
3. Re-auth to staging for the staging-AR pushes (gated on `staging_enabled`).
4. Push API image (unchanged).
5. Build+push `web:<sha>-staging` (gated on `staging_enabled`).
6. Build+push `web:<sha>-prod` + `:latest` (always).

Both web variants are pushed to the same AR repo as the API image (`steps.ar.outputs.repo`) — per-env AR routing is out of scope for this PR. The cross-project pull arrangement (prod GKE/Cloud Run pulling from staging AR when staging is enabled) is documented in ADR-025 § "Cross-project Artifact Registry note" and tracked as a follow-up.

## Review findings addressed (commit 03935c7)

`/review` posted https://github.com/brownm09/lifting-logbook/pull/392#issuecomment-4589404188 — 1 blocking + 5 non-blocking findings + 2 author questions. All resolved in commit `03935c7`:

- **(blocking documentation)** `docs/README.md` Architecture Decision Records table now includes ADR-025.
- **(reliability)** Bootstrap concern documented in ADR-025 § "First-time prod bootstrap" and `docs/deploy.md` § "First-time prod bootstrap" with the three-step manual `terraform apply` recovery procedure. Hard-fail behavior kept (loud surface > silent skip).
- **(reliability)** Auth-context invariant comment added immediately before `Configure Docker for Artifact Registry` in deploy.yml.
- **(maintainability)** `:latest` semantics comment added on the prod build step.
- **(maintainability)** Cross-project AR pull arrangement documented in ADR-025 with the explicit warning not to remove the prod-SA `artifactregistry.reader` grant on staging AR without also moving the push target.
- **(maintainability)** PR body Testing section now includes a contractual status line (below).

## Testing

`npm test` from the repo root.

Tests: not run on this branch — Jest module resolution blocked by [#373](https://github.com/brownm09/lifting-logbook/issues/373) (Node 24 / Windows `npm install` tarball corruption; `nvm` not on PATH in this shell). This PR touches only `.github/workflows/deploy.yml` and `docs/*.md` — no application or test source — so it cannot affect test outcomes.

End-to-end verification of the pipeline change happens when this branch is pushed to `main` and `deploy.yml` runs; the bundle-grep procedure documented in `docs/deploy.md` → "Verifying per-env web image build" satisfies acceptance criterion #3 of #388.

### Suppression / test-integrity / coverage

- Suppression grep: clean.
- Test-integrity grep: clean. No skip markers, deleted tests, lowered thresholds, or `passWithNoTests` introduced.
- Coverage gate: no testable behavior added (CI YAML + docs only). N/A.

## Test plan

- [ ] After merge: `deploy.yml` run on `main` shows both `Build and push web image (staging)` and `Build and push web image (production)` steps succeed and push to AR.
- [ ] `deploy-staging` pulls `web:<sha>-staging`; `deploy-production` pulls `web:<sha>-prod`.
- [ ] Run the bundle-grep procedure in `docs/deploy.md` against each env; both must report `OK:` (no cross-env values).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -163,40 +163,31 @@ jobs:
           PREFLIGHT_REPO="${{ needs.preflight.outputs.ar_repo }}"
           echo "repo=${STAGING_REPO:-$PREFLIGHT_REPO}" >> "$GITHUB_OUTPUT"
 
-      - name: Resolve API URL for web image build
-        id: api-url
-        run: |
-          if [[ "${{ needs.preflight.outputs.staging_enabled }}" == "true" ]]; then
-            echo "url=${{ needs.terraform-staging.outputs.cloud_run_api_url }}" >> "$GITHUB_OUTPUT"
-          else
-            # Production-only mode: gcloud auth is already done above.
-            URL=$(gcloud run services describe lifting-logbook-prod-api \
-              --region="${{ env.REGION }}" \
-              --project="${{ vars.GCP_PROD_PROJECT_ID }}" \
-              --format="value(status.url)")
-            echo "url=${URL}" >> "$GITHUB_OUTPUT"
-          fi
+      # ── Per-env value resolution for web image build (ADR-025) ────────────
+      # NEXT_PUBLIC_* values are inlined into the Next.js JS bundle at build
+      # time, so the web image is built once per target env with that env's
+      # values. Both image variants are pushed to the same AR repo (steps.ar)
+      # — the same pattern the API image uses today; per-env AR routing is
+      # out of scope for this change.
 
-      - name: Resolve Clerk publishable key for web image build
-        id: clerk-pub
+      - name: Resolve staging API URL (web image build)
+        id: api-url-staging
+        if: needs.preflight.outputs.staging_enabled == 'true'
+        run: |
+          echo "url=${{ needs.terraform-staging.outputs.cloud_run_api_url }}" >> "$GITHUB_OUTPUT"
+
+      - name: Resolve staging Clerk publishable key (web image build)
+        id: clerk-pub-staging
+        if: needs.preflight.outputs.staging_enabled == 'true'
         run: |
           # NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY is build-time embedded in the
           # Next.js bundle (see apps/web/Dockerfile ARG). Without it, `next
           # build` aborts on /_not-found prerender because <ClerkProvider>
-          # in the root layout throws "Missing publishableKey". The image
-          # is shared between staging and production deploys (cache reuse
-          # by tag=github.sha), so we follow the same staging-vs-prod
-          # convention as `Resolve API URL` above.
-          if [[ "${{ needs.preflight.outputs.staging_enabled }}" == "true" ]]; then
-            KEY=$(gcloud secrets versions access latest \
-              --secret="lifting-logbook-stg-clerk-publishable-key" \
-              --project="${{ vars.GCP_STAGING_PROJECT_ID }}")
-          else
-            KEY=$(gcloud secrets versions access latest \
-              --secret="lifting-logbook-prod-clerk-publishable-key" \
-              --project="${{ vars.GCP_PROD_PROJECT_ID }}")
-          fi
-          [ -n "$KEY" ] || { echo "ERROR: clerk-publishable-key fetch returned empty"; exit 1; }
+          # in the root layout throws "Missing publishableKey".
+          KEY=$(gcloud secrets versions access latest \
+            --secret="lifting-logbook-stg-clerk-publishable-key" \
+            --project="${{ vars.GCP_STAGING_PROJECT_ID }}")
+          [ -n "$KEY" ] || { echo "ERROR: staging clerk-publishable-key fetch returned empty"; exit 1; }
           # Use ::add-mask:: so the key never appears in logs even though
           # it is a publishable (non-secret) Clerk key. Defense in depth.
           echo "::add-mask::$KEY"
@@ -204,6 +195,48 @@ jobs:
           # but the Clerk publishable key is a single line so a direct echo
           # is safe.
           echo "key=${KEY}" >> "$GITHUB_OUTPUT"
+
+      # Production-value resolution requires auth to the prod project. The
+      # staging auth above is replaced for the duration of these steps; the
+      # API image build and any subsequent staging push re-authenticate to
+      # staging below. In production-only mode this is a no-op re-auth.
+      - name: Authenticate to GCP (production, for prod web build-args)
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ secrets.GCP_PROD_WORKLOAD_IDENTITY_PROVIDER }}
+          service_account: ${{ secrets.GCP_PROD_SERVICE_ACCOUNT }}
+
+      - name: Resolve production API URL (web image build)
+        id: api-url-prod
+        run: |
+          # The prod Cloud Run API service is created by terraform-production,
+          # which runs *after* this job. On first-time prod bootstrap the
+          # `describe` call fails; rerun this workflow after the prod Cloud
+          # Run service exists. Tracked as a known limitation in ADR-025.
+          URL=$(gcloud run services describe lifting-logbook-prod-api \
+            --region="${{ env.REGION }}" \
+            --project="${{ vars.GCP_PROD_PROJECT_ID }}" \
+            --format="value(status.url)")
+          [ -n "$URL" ] || { echo "ERROR: prod API URL fetch returned empty (is lifting-logbook-prod-api deployed?)"; exit 1; }
+          echo "url=${URL}" >> "$GITHUB_OUTPUT"
+
+      - name: Resolve production Clerk publishable key (web image build)
+        id: clerk-pub-prod
+        run: |
+          KEY=$(gcloud secrets versions access latest \
+            --secret="lifting-logbook-prod-clerk-publishable-key" \
+            --project="${{ vars.GCP_PROD_PROJECT_ID }}")
+          [ -n "$KEY" ] || { echo "ERROR: prod clerk-publishable-key fetch returned empty"; exit 1; }
+          echo "::add-mask::$KEY"
+          echo "key=${KEY}" >> "$GITHUB_OUTPUT"
+
+      # Restore staging auth for the image pushes that target the staging AR.
+      - name: Re-authenticate to GCP (staging, restore for pushes)
+        if: needs.preflight.outputs.staging_enabled == 'true'
+        uses: google-github-actions/auth@v2
+        with:
+          workload_identity_provider: ${{ secrets.GCP_STAGING_WORKLOAD_IDENTITY_PROVIDER }}
+          service_account: ${{ secrets.GCP_STAGING_SERVICE_ACCOUNT }}
 
       - name: Configure Docker for Artifact Registry
         run: gcloud auth configure-docker ${{ env.REGISTRY }} --quiet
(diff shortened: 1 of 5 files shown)


brownm09:
## Review: [infra] Build web image per-env to stop baking staging values into prod

### Summary

Merge with minor changes. The decision (per-env image tags) is well-justified and the ADR captures the explicit trade clearly; pipeline mechanics, auth re-shuffle, and deploy-side tag updates all line up correctly across staging-enabled and production-only modes. One blocking documentation gap (the ADR-025 row is missing from the index in `docs/README.md`) and two reliability concerns (a new always-on failure path on prod-only bootstraps, and a brittle GA auth-context replacement) should be addressed before merge.

---

### Blocking Findings

**[documentation]** `docs/README.md` — ADR index missing ADR-025
The repo's ADR table in `docs/README.md` lists ADR-001 through ADR-024 (around line 158). This PR adds `docs/adr/ADR-025-web-image-per-env-build.md` and updates `docs/adr-references.md`, but the index in `docs/README.md` is not touched. The index is the project's primary discovery surface for ADRs, so a new ADR that lands without an index row is effectively invisible to anyone orienting from the top-level docs.
**Fix:** Append a row for ADR-025 to the "Architecture Decision Records" table in `docs/README.md` matching the existing format (link, title, status).

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/deploy.yml:213-220` — first-time prod bootstrap now fails in staging-enabled mode
Before this PR, `gcloud run services describe lifting-logbook-prod-api` only ran in production-only mode (`staging_enabled == false`); when staging was enabled, the API URL came from `terraform-staging.outputs.cloud_run_api_url` and the prod URL was never resolved. After this PR, the prod-URL `describe` runs **on every pipeline execution** because the new `api-url-prod` step is unconditional. On a first-time deploy of a new prod environment with staging enabled (terraform-production has not yet applied), the `describe` will fail and the entire `build-images` job aborts — including blocking the staging deploy that previously succeeded independently. ADR-025 acknowledges this; the YAML comment says "rerun this workflow after the prod Cloud Run service exists" but `build-images` runs before `terraform-production`, so there is no single rerun that resolves it without manual ordering.
**Suggestion:** Either (a) gate the prod-API describe behind a check that the service exists and fall back to the `terraform-production` output (introduces a job dependency), or (b) catch the empty-URL case and skip the prod web build entirely on the bootstrap run, letting the next pipeline pick it up. At minimum, document the exact bootstrap sequence in `docs/deploy.md` so the recovery is mechanical.

**[reliability]** `.github/workflows/deploy.yml:201-209, 232-238` — `google-github-actions/auth@v2` replaces credentials globally for the job
The new "Authenticate to GCP (production)" step runs unconditionally and replaces whatever credentials the job had. The "Re-authenticate to GCP (staging)" step that restores them is gated on `staging_enabled == true`. In production-only mode this is fine because subsequent pushes target prod AR. In staging-enabled mode the restore runs, so subsequent staging-AR pushes also work. But this dance assumes the `auth@v2` action is the only thing that mutates `GOOGLE_APPLICATION_CREDENTIALS` / `gcloud config` state and that no later step in the job reads credentials before the restore — fragile under future edits. The `gcloud auth configure-docker` step (around line 241 in the new layout) runs after the restore, which is correct today but easy to break.
**Suggestion:** Add an inline comment immediately before the `Configure Docker for Artifact Registry` step stating "auth context must be staging-AR-capable at this point — any step inserted between Re-authenticate and here must preserve that". Long-term, consider extracting prod-value resolution into its own job that depends only on preflight and emits `url`/`key` outputs, so the auth swap is contained.

**[maintainability]** `.github/workflows/deploy.yml:282-292` — `:latest` semantics flip in staging-enabled mode
The `:latest` tag now points at the prod web image regardless of mode. When `staging_enabled == true`, that prod image is pushed to the staging AR (because `steps.ar.outputs.repo` resolves to staging when staging is enabled). Anyone reading `:latest` from the staging AR — manual ops, a future cache-warming job, an unaware deploy script — gets the prod-configured bundle. The old behavior also pushed `:latest` from a single image; the semantic confusion is new.
**Suggestion:** Either drop `:latest` from the prod build, push it from both builds (with the staging one writing `:latest-staging` and the prod one writing `:latest-prod`), or leave a comment near the prod build documenting which registry receives `:latest` in each mode.

**[maintainability]** `.github/workflows/deploy.yml:282-292, 263-280` vs deploy-production lines 591 and 610 — prod web image pushed to staging AR while prod deploy reads from prod AR
In staging-enabled mode, the prod web build pushes to `steps.ar.outputs.repo` (staging AR), but `deploy-production` Helm and Cloud Run use `steps.tf-prod.outputs.ar_repo` (prod AR). The PR body explicitly defers per-env AR routing as out of scope and notes the API image already has this shape, so this is not a regression — but the asymmetry between "build target" and "deploy source" is now load-bearing on a cross-project AR-pull arrangement that is nowhere documented in the PR or ADR.
**Suggestion:** Add a one-paragraph note to ADR-025 or `docs/deploy.md` describing the current cross-project AR access path (whatever it is — IAM grant, mirror, or coincidence that `tf-prod.outputs.ar_repo == build-images.outputs.ar_repo` in some configurations), so the next person who touches this knows what assumption they are preserving. This is also the natural place to link the follow-up "per-env AR routing" issue when it is filed.

**[maintainability]** PR body — "Tests" line does not match the `Tests: N passed, N skipped, N failed` format required by the test-integrity policy
The PR body explains tests aren't runnable due to documented #373 (Node 24 / Windows install corruption) and confirms the change is YAML + docs only. That satisfies the failure-tracking rule (failure cited to an open issue) and the spirit of the integrity gate (no degradation introduced). But the `Tests: N passed, N skipped, N failed (duration)` summary line is the contractual surface the gate inspects, and it is absent.
**Suggestion:** Add a single line under "Testing" stating `Tests: not run on this branch — Jest module resolution blocked by #373; see PR body for rationale`. This keeps the gate auditable without forcing a Node 20 reinstall on this branch.

---

### Questions for Author

**Question:** Is the cross-project AR pull (build pushes to staging AR, prod deploy reads from prod AR) actually working today, or does `tf-prod.outputs.ar_repo` happen to equal `build-images.outputs.ar_repo` in the configurations you've exercised?
**Context:** `deploy.yml` lines 591 and 610 reference `steps.tf-prod.outputs.ar_repo`, while `build-images` pushes to `steps.ar.outputs.repo`. The PR body says this matches the API image's existing pattern, but the API image has not been validated against a real prod deploy that uses a separate prod AR — it has only been validated under the staging-enabled mode where the two repos may coincide.
**Tradeoffs:** (a) The two repos coincide today (e.g., preflight + tf-prod both resolve to the same project) — no action needed but the cross-project assumption is brittle. (b) The two repos differ and prod has explicit pull access to staging AR — fine but the IAM grant should be documented in ADR-025. (c) The two repos differ and the existing API deploy is also broken on a clean prod AR — then this PR doesn't make it worse but the wider bug should be filed before the Phase 2 follow-up.

---

**Question:** Should the prod web build be skipped entirely on first-time bootstrap (when `lifting-logbook-prod-api` does not yet exist) rather than failing the whole `build-images` job?
**Context:** `deploy.yml:213-220` (api-url-prod) exits 1 if the `describe` returns empty. On bootstrap that aborts staging deploy too, even though staging is independently deployable.
**Tradeoffs:** (a) Hard-fail as written — simple, surfaces the missing service loudly, but couples staging to prod existence. (b) Soft-skip the prod build when prod-api is missing — preserves staging-only bootstrap, but lets a misconfigured prod pipeline run green for a while. (c) Move prod-value resolution into a step that depends on `terraform-production` outputs — eliminates the bootstrap chicken-and-egg but reshapes job dependencies and may slow the critical path.

---

### Style Notes

The new comment blocks (`── Per-env value resolution …`, `── Web image: per-env builds …`) are a nice readability gain in the YAML — keep that convention going for future structural changes. ADR-025's "Explicit trade" subsection is exactly the right place to call out the broken byte-identical promotion contract; this is a model for how future ADRs that override prior invariants should be structured.

---

*Reviewed by `claude-opus-4` via Claude Code*

Issue #394, opened 2026-06-01
[infra] Local Docker Desktop WSL distro missing — api E2E suite cannot run locally


brownm09:
## Problem

On this dev machine, `docker info` fails with `open //./pipe/dockerDesktopLinuxEngine: The system cannot find the file specified` and `wsl -l -v` reports no installed distributions. Docker Desktop's GUI launches but its Linux engine never comes up because the `docker-desktop` WSL distribution that backs it has been unregistered (or never installed after a reset).

Effect: `apps/api` Jest globalSetup (`apps/api/jest.global-setup.js`) cannot start the Testcontainers Postgres, so `npm test` from repo root fails on the api workspace even when the change under test is unrelated to the api.

CI is unaffected — GitHub Actions provides a Postgres service container and bypasses Testcontainers via `DATABASE_URL` (see `docs/testing/e2e-coverage.md`).

## Repro

```bash
wsl -l -v          # 'no installed distributions'
docker info         # ERROR: cannot find pipe
npm test            # @lifting-logbook/api#test fails at globalSetup
```

## Fix options

1. Reset Docker Desktop to factory defaults (Settings → Troubleshoot → Reset to factory defaults) — re-registers the `docker-desktop` WSL distro.
2. Uninstall and reinstall Docker Desktop.
3. Switch local api E2E to a non-Testcontainers fallback (out of scope here — Testcontainers is the documented path).

This is a dev-environment-only issue; no code change required. Filing per `CLAUDE.md` failure-tracking rule so it can be cited in PRs where the local Docker prerequisite was unavailable.

PR #395, opened 2026-06-01
[fix] Probe readiness via /api/healthz so Clerk-init failures fail-closed

### LINKED ISSUE #385: [fix] Web readiness probe should exercise middleware (currently passes during auth-broken rollouts)
## Problem

The current readiness/liveness probes for `apps/web` target the root path `/`, which Next.js serves as a static prerender (the marketing page in `app/page.tsx`). During the #382 outage every protected route returned 500 because Clerk middleware was throwing, yet the probes stayed green the entire time because `/` is statically optimized and the edge serves a cached prerender that bypasses the failing middleware. The bad rollout was therefore promoted to "ready" and stayed serving production traffic.

## Proposal

Move the readiness probe to a path that exercises middleware so a Clerk-init failure (or any other middleware crash) fails-closed during rollout instead of failing-open.

Two viable shapes — pick one:

1. **Probe `/cycle` (or any other protected path) and treat 307 as healthy.** Requires the probe to follow redirects or to treat `httpGet.path: /cycle` + `httpGet.httpHeaders` as healthy for 3xx. Kubernetes treats 2xx and 3xx as success by default, so this works without code changes — the route just needs to actually hit middleware (not be statically rendered).
2. **Add a dedicated `/healthz` route in `apps/web` that calls `auth()` server-side** (or any other server-side dependency) and returns `200 OK` on success, `503` on failure. Probe that. Decouples probe semantics from auth routing and gives us a single observable for "the web container is functional."

Option 2 is cleaner but heavier. Option 1 is one-line change in `values.yaml`.

## Acceptance criteria

- [ ] Readiness probe path exercises middleware (either a protected route or a dedicated `/healthz` that does).
- [ ] Reproducing the #382 failure mode (deploy with `CLERK_SECRET_KEY` removed) causes the new pod to fail readiness rather than be marked Ready.
- [ ] Liveness probe semantics reconsidered in the same PR (currently mirrors readiness — may want to keep it on `/` so a transient middleware blip doesn't restart the pod into a crash loop).

## References

- Originating outage: #382
- Originating fix: #383
- Probe definitions: `infra/kubernetes/values/production-web.yaml`, `infra/kubernetes/charts/web/values.yaml`
- Kubernetes probe success codes: https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#http-probes

brownm09:
## Summary

The current readiness probe targets `/`, which until #384 was the static marketing page (`app/page.tsx`) and was served from the prerender cache, bypassing middleware. During the #382 outage, every protected route 500'd because Clerk middleware was throwing — but `/` stayed 200 and the bad rollout was promoted to Ready.

This PR moves readiness to a new `/api/healthz` route that runs through `clerkMiddleware` and returns an explicit 200/503 contract. The route is matched by Clerk's `/(api|trpc)(.*)` middleware matcher, so a missing `CLERK_SECRET_KEY` (as in #382) now throws inside middleware → 500 → fails readiness, before the route handler ever executes. The route handler's `auth()` call is an additional, distinct guard that detects "clerkMiddleware did not run on this request" (e.g., a future matcher misconfig) by inspecting the `x-clerk-auth-status` header middleware stamps; `auth()` does not re-validate the secret key.

Liveness stays on `/`. Since #384 merged, `/` is dynamic and exercises `clerkMiddleware` too — so liveness `failureThreshold` is bumped to 9 (~3 minutes of sustained failure before pod restart) so transient Clerk blips don't crash-loop pods. Readiness keeps `failureThreshold: 3` — bad rollouts must fail fast and stay un-Ready.

### Changes

- **`apps/web/app/api/healthz/route.ts`** (new) — `force-dynamic` route. Calls `auth()` and returns `200 {ok:true}` on success, `503 {ok:false}` on `auth()` throw (with the error logged server-side, not in the body — the route is reachable via the public Ingress), and `200 {mode:'dev-auth'}` when `DEV_AUTH_TOKEN` bypasses Clerk.
- **`apps/web/middleware.ts`** — add `/api/healthz(.*)` to `isPublicRoute` so `clerkMiddleware` still runs (exercising Clerk init) but `auth.protect()` doesn't redirect probe traffic. `(.*)` suffix matches the sibling sign-in/sign-up patterns.
- **`infra/kubernetes/charts/web/values.yaml`** — `readinessProbe.httpGet.path` → `/api/healthz`. Both probes get explicit `timeoutSeconds: 3` (K8s default of 1s is too tight for routes that do any work). Liveness `failureThreshold: 9`; readiness `failureThreshold: 3`.

No template changes — `deployment.yaml` renders `.Values.{liveness,readiness}Probe` via `toYaml`. Per-env values files don't override probes, so the chart-level change applies to staging and production.

## Acceptance criteria

- [x] Readiness probe path exercises middleware (`/api/healthz` runs through `clerkMiddleware`).
- [x] Reproducing the #382 failure mode (Clerk init throws) causes the new pod to fail readiness — `clerkMiddleware` throws inside `assertKey` before the route handler runs; the route handler's `try/catch` is a second guard for matcher misconfig regressions, covered by a test that mocks `auth()` to throw.
- [x] Liveness probe semantics reconsidered: with #384 making `/` dynamic, liveness `failureThreshold` raised to 9 so transient Clerk blips don't crash-loop running pods.

## Testing

Ran from repo root:

```bash
npx turbo run test --filter='!@lifting-logbook/api'
```

```
Tasks: 9 successful, 9 total
```

Web suite (new healthz tests included):
```
PASS app/api/healthz/route.test.ts
  GET /api/healthz
    √ returns 200 with ok:true when Clerk auth() resolves
    √ returns 503 with no error body when auth() throws
    √ skips Clerk and returns ok:true in DEV_AUTH_TOKEN mode
Tests: 39 passed, 0 skipped, 0 failed (19.4 s)
```

Cross-workspace totals: web 39 / core 160 / api-legacy 1 / eslint-rules 11 — all passing, 0 skipped, 0 failed.

### Pre-existing failure note

The `@lifting-logbook/api` workspace was excluded from this run because its Jest globalSetup requires Docker (Testcontainers Postgres), and Docker Desktop's WSL distro is missing on this machine. Tracked in #394. CI is unaffected — GitHub Actions provides a Postgres service container that bypasses Testcontainers. This PR does not touch any code in `apps/api`.

### Code-quality notes

Adds one ESLint suppression at `apps/web/app/api/healthz/route.ts:33` — `eslint-disable-next-line no-console -- server-side log only; not exposed`. The probe handler logs the swallowed `auth()` error via `console.error` for operators (the error is intentionally not in the 503 response body — the route is publicly reachable via the GCE Ingress). Rationale is inline at the call site; flagged here per the Suppression policy Rule 1 (PR-body justification).

## Test plan

- [ ] Watch CI for green api E2E suite (proves no cross-workspace regression).
- [ ] After merge, in next staging deploy: `kubectl describe pod <web>` shows readiness probe hitting `/api/healthz` with the new threshold values.
- [ ] Manual regression: deploy a web pod with `CLERK_SECRET_KEY` unset; readiness should stay failing and the rollout should not promote (was the #382 failure mode).

Closes #385

### DIFF
--- a/apps/web/app/api/healthz/route.test.ts
+++ b/apps/web/app/api/healthz/route.test.ts
@@ -1,0 +1,61 @@
+/**
+ * @jest-environment node
+ */
+import { GET } from './route';
+
+jest.mock('@clerk/nextjs/server', () => ({
+  auth: jest.fn(),
+}));
+
+import { auth } from '@clerk/nextjs/server';
+
+const mockedAuth = auth as unknown as jest.Mock;
+
+describe('GET /api/healthz', () => {
+  const originalEnv = process.env.DEV_AUTH_TOKEN;
+
+  afterEach(() => {
+    if (originalEnv === undefined) {
+      delete process.env.DEV_AUTH_TOKEN;
+    } else {
+      process.env.DEV_AUTH_TOKEN = originalEnv;
+    }
+    mockedAuth.mockReset();
+  });
+
+  it('returns 200 with ok:true when Clerk auth() resolves', async () => {
+    delete process.env.DEV_AUTH_TOKEN;
+    mockedAuth.mockResolvedValueOnce({ userId: null });
+
+    const res = await GET();
+
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual({ ok: true });
+    expect(mockedAuth).toHaveBeenCalledTimes(1);
+  });
+
+  // Regression test for #382 / #385: a Clerk-init failure must fail readiness,
+  // not be hidden behind a 200 from a statically-rendered page.
+  it('returns 503 with the error message when Clerk auth() throws', async () => {
+    delete process.env.DEV_AUTH_TOKEN;
+    mockedAuth.mockRejectedValueOnce(new Error('Missing CLERK_SECRET_KEY'));
+
+    const res = await GET();
+
+    expect(res.status).toBe(503);
+    expect(await res.json()).toEqual({
+      ok: false,
+      error: 'Missing CLERK_SECRET_KEY',
+    });
+  });
+
+  it('skips Clerk and returns ok:true in DEV_AUTH_TOKEN mode', async () => {
+    process.env.DEV_AUTH_TOKEN = 'dev-token';
+
+    const res = await GET();
+
+    expect(res.status).toBe(200);
+    expect(await res.json()).toEqual({ ok: true, mode: 'dev-auth' });
+    expect(mockedAuth).not.toHaveBeenCalled();
+  });
+});
--- a/apps/web/app/api/healthz/route.test.ts
+++ b/apps/web/app/api/healthz/route.test.ts
@@ -34,19 +34,27 @@ describe('GET /api/healthz', () => {
     expect(mockedAuth).toHaveBeenCalledTimes(1);
   });
 
-  // Regression test for #382 / #385: a Clerk-init failure must fail readiness,
-  // not be hidden behind a 200 from a statically-rendered page.
-  it('returns 503 with the error message when Clerk auth() throws', async () => {
+  // Regression test for #385: when auth() throws (e.g., the route fell outside
+  // clerkMiddleware's matcher and the auth-status header is absent), the probe
+  // must return 503 so readiness fails. The error message is logged server-side
+  // but deliberately not included in the response body — see route.ts.
+  it('returns 503 with no error body when auth() throws', async () => {
     delete process.env.DEV_AUTH_TOKEN;
+    const consoleErrorSpy = jest
+      .spyOn(console, 'error')
+      .mockImplementation(() => undefined);
     mockedAuth.mockRejectedValueOnce(new Error('Missing CLERK_SECRET_KEY'));
 
     const res = await GET();
 
     expect(res.status).toBe(503);
-    expect(await res.json()).toEqual({
-      ok: false,
-      error: 'Missing CLERK_SECRET_KEY',
-    });
+    expect(await res.json()).toEqual({ ok: false });
+    // The original error is logged server-side for operators.
+    expect(consoleErrorSpy).toHaveBeenCalledWith(
+      expect.stringContaining('[healthz]'),
+      expect.objectContaining({ message: 'Missing CLERK_SECRET_KEY' }),
+    );
+    consoleErrorSpy.mockRestore();
   });
 
   it('skips Clerk and returns ok:true in DEV_AUTH_TOKEN mode', async () => {
--- a/apps/web/app/api/healthz/route.ts
+++ b/apps/web/app/api/healthz/route.ts
@@ -1,0 +1,22 @@
+import { auth } from '@clerk/nextjs/server';
+import { NextResponse } from 'next/server';
+
+// Readiness probe target. Must not be statically rendered or edge-cached —
+// otherwise the same Clerk-init failure mode (#382) that broke production
+// would again be hidden from Kubernetes during rollout.
+export const dynamic = 'force-dynamic';
+
+export async function GET() {
+  if (process.env.DEV_AUTH_TOKEN) {
+    return NextResponse.json({ ok: true, mode: 'dev-auth' });
+  }
+  try {
+    await auth();
+    return NextResponse.json({ ok: true });
+  } catch (err) {
+    return NextResponse.json(
+      { ok: false, error: err instanceof Error ? err.message : 'unknown' },
+      { status: 503 },
+    );
+  }
+}
--- a/apps/web/app/api/healthz/route.ts
+++ b/apps/web/app/api/healthz/route.ts
@@ -1,11 +1,26 @@
 import { auth } from '@clerk/nextjs/server';
 import { NextResponse } from 'next/server';
 
-// Readiness probe target. Must not be statically rendered or edge-cached —
-// otherwise the same Clerk-init failure mode (#382) that broke production
-// would again be hidden from Kubernetes during rollout.
+// Readiness probe target. Must stay force-dynamic so Next never prerenders
+// or edge-caches it — otherwise the failure mode this route exists to detect
+// (#382: clerkMiddleware throws on missing CLERK_SECRET_KEY) would be hidden
+// from Kubernetes during rollout the same way it was when the probe hit `/`.
 export const dynamic = 'force-dynamic';
 
+// Two guards in one handler:
+//   1. The route matches Clerk's `/(api|trpc)(.*)` middleware matcher, so a
+//      missing CLERK_SECRET_KEY (as in #382) throws inside clerkMiddleware
+//      → 500 → fails readiness, before this handler ever runs. That is the
+//      primary #382 regression detector.
+//   2. The auth() call below additionally verifies clerkMiddleware actually
+//      ran on this request — it inspects the `x-clerk-auth-status` header
+//      that middleware stamps. If a future matcher edit silently excludes
+//      `/api/healthz`, auth() throws "auth header missing" and we 503.
+// auth() does NOT re-validate the secret key or make a network call; that
+// work already happened in middleware. We deliberately keep the response
+// body minimal — no error.message — so a publicly-reachable probe endpoint
+// does not leak Clerk misconfiguration details. The original error is still
+// logged server-side for operators.
 export async function GET() {
   if (process.env.DEV_AUTH_TOKEN) {
     return NextResponse.json({ ok: true, mode: 'dev-auth' });
@@ -14,9 +29,8 @@ export async function GET() {
     await auth();
     return NextResponse.json({ ok: true });
   } catch (err) {
-    return NextResponse.json(
-      { ok: false, error: err instanceof Error ? err.message : 'unknown' },
-      { status: 503 },
-    );
+    // eslint-disable-next-line no-console -- server-side log only; not exposed
+    console.error('[healthz] auth() threw — middleware likely did not run:', err);
+    return NextResponse.json({ ok: false }, { status: 503 });
   }
 }
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -6,8 +6,10 @@ const isPublicRoute = createRouteMatcher([
   '/sign-in(.*)',
   '/sign-up(.*)',
   // Kubernetes readiness probe — must run through clerkMiddleware (to exercise
-  // Clerk init) but must not be redirected to /sign-in. See #385.
-  '/api/healthz',
+  // Clerk init) but must not be redirected to /sign-in. See #385. The `(.*)`
+  // suffix matches the sibling sign-in/sign-up patterns and tolerates trailing
+  // slashes or future sub-paths without re-opening the original issue.
+  '/api/healthz(.*)',
 ]);
 
 const clerkHandler = clerkMiddleware(async (auth, request) => {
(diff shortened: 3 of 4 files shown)


brownm09:
## Review: [fix] Probe readiness via /api/healthz so Clerk-init failures fail-closed

### Summary

Merge with minor changes. The implementation is tight: `force-dynamic` on the route, `isPublicRoute` (not `matcher` exclusion) to keep middleware running, liveness left alone, and a regression test that mocks `auth()` to throw. The highest-severity finding is the contradictory test-summary line in the PR body ("31 passed, 31 skipped 0 failed" vs. "0 skipped, 0 failed" elsewhere) — almost certainly a typo, but the Test Integrity policy requires resolving it before merge. The rest are non-blocking robustness suggestions on the probe config and external exposure.

---

### Blocking Findings

**[correctness]** PR body — Testing section
The summary line reads `Tests: 31 passed, 31 skipped 0 failed (16.3 s)`, but immediately below the body claims "0 skipped, 0 failed." Under the Test Integrity policy (ADR-029), a non-zero skipped count must be justified per-skip in the PR body; a contradictory line cannot stand.
**Fix:** Re-run the suite and paste the canonical `Tests: N passed, N skipped, N failed (duration)` line verbatim into the Testing section. If 31 skipped is correct, add a per-skip justification block; if it's a typo (likely "0 skipped"), correct it.

---

### Non-Blocking Findings

**[reliability]** `infra/kubernetes/charts/web/values.yaml:52-61`
`readinessProbe` sets only `httpGet`, `initialDelaySeconds`, and `periodSeconds`. The new target does real work (`auth()` call into Clerk SDK), unlike the previous static `/`. K8s defaults apply: `timeoutSeconds: 1`, `failureThreshold: 3`. A 1-second timeout against a cold-start route that initializes Clerk on first call can produce spurious readiness flaps during scale-up — and three consecutive failures (=30s with `periodSeconds: 10`) is on the aggressive side for a route that needs to do a network-bound init.
**Suggestion:** Add explicit `timeoutSeconds: 3` and `failureThreshold: 3` (or 5) so the probe semantics are documented rather than relying on the K8s defaults, and so cold starts don't flap. Consider a `startupProbe` if cold init is consistently >1s.

**[security]** `apps/web/app/api/healthz/route.ts:17`
The 503 response body includes `error: err.message`. The Ingress (`infra/kubernetes/charts/web/values.yaml:39-43`) is enabled with `className: gce` and no path exclusions, so `/api/healthz` is reachable from the public internet, not just from the kubelet. Leaking Clerk error text (e.g., `Missing CLERK_SECRET_KEY`) to an unauthenticated external caller is a small information disclosure — useful to an attacker confirming misconfiguration windows, harmless to a legitimate user.
**Suggestion:** Either (a) keep the error message but log it server-side and return `{ok: false}` only, or (b) gate the route to cluster-internal traffic by excluding `/api/healthz` from the public Ingress and exposing it only on the pod port the kubelet reaches directly. Option (a) is the smaller change.

**[maintainability]** `apps/web/middleware.ts:10`
`isPublicRoute` lists `/api/healthz` as an exact-string pattern. `createRouteMatcher` handles query strings, but a trailing slash (`/api/healthz/`) or a future versioned variant (`/api/healthz/v2`) won't match. The K8s probe path is exact so this is fine today.
**Suggestion:** If you want defense against future drift, write `/api/healthz(.*)` to match the sibling `/sign-in(.*)` pattern.

**[maintainability]** `apps/web/app/api/healthz/route.test.ts:21-22`
The `afterEach` resets `process.env.DEV_AUTH_TOKEN` and `mockedAuth.mockReset()` but the suite has no `beforeAll`/`afterAll` to capture and restore `originalEnv` at the suite boundary. If another test file in the same Jest worker mutates `DEV_AUTH_TOKEN`, the `originalEnv` captured at module-load time may be stale.
**Suggestion:** Low priority — Jest test files run in their own modules. Mention only because the pattern is slightly fragile; capturing `originalEnv` inside `beforeAll` and restoring in `afterAll` would be tighter.

---

### Questions for Author

**Question:** Does an unauthenticated `auth()` call from inside a route handler actually exercise the same Clerk-init code path that failed in #382?
**Context:** `apps/web/app/api/healthz/route.ts:13` calls `await auth()` without a userId requirement. In #382 the failure was `clerkMiddleware` itself throwing on request because `CLERK_SECRET_KEY` was missing — but `clerkMiddleware` runs before the route handler, so if middleware throws, Next returns 500 and the route handler's `try/catch` never runs anyway. The probe still fails-closed (500 fails readiness), so the bug is caught — but the route handler's `auth()` call may be redundant belt-and-braces rather than the primary detector.
**Tradeoffs:** (a) Leave as-is — belt-and-braces; defends against a future failure mode where middleware succeeds but `auth()` from a route fails. (b) Remove the `auth()` call and return 200 unconditionally — relies entirely on `clerkMiddleware` running (which it does for this path because it's in `isPublicRoute` not excluded from `matcher`). (c) Replace `auth()` with a more explicit init check if Clerk exposes one. Option (a) is what's shipped and is the safest; worth a sentence in the route file's comment confirming the design intent.

---

### Style Notes

Comments throughout (`route.ts`, `middleware.ts`, `values.yaml`) are unusually good — they explain the *why* (link back to #382/#385) rather than restating the *what*. Keep this pattern.

---

brownm09:
## Review (rev 2): [fix] Probe readiness via /api/healthz so Clerk-init failures fail-closed

### Summary

Merge-ready. All prior blocking findings addressed, and the reconciliation with #384 (now-merged) is captured: liveness `failureThreshold: 9` × `periodSeconds: 20` = ~3 min grace before pod restart, which is the right knob now that `/` is dynamic. The route file's comment is precise about what `auth()` actually does (matcher-misconfig detector, not Clerk-init detector — verified by tracing `@clerk/nextjs/dist/esm/app-router/server/auth.js` → `createGetAuth.js` → `headers-utils.js`: `auth()` only checks for the `x-clerk-auth-status` header that middleware stamps; it does not re-validate the secret key). One small non-blocking note on suppression-policy bookkeeping.

---

### Non-Blocking Findings

**[maintainability]** `apps/web/app/api/healthz/route.ts:33`
A new `eslint-disable-next-line no-console -- server-side log only; not exposed` lands in the diff. Per the Code Quality → Suppression policy (Rule 1), every suppression should be accompanied by a PR-body note explaining why a proper fix wasn't appropriate. The inline justification is good; the PR body does not separately call it out.
**Suggestion:** Add one sentence to the PR body Changes section: "Adds one `eslint-disable no-console` for a deliberate server-side operator log in `route.ts`; rationale is inline." Low priority — the inline comment is the substantive justification.

**[reliability]** `infra/kubernetes/charts/web/values.yaml:45-50`
The comment on the liveness probe says it "exercises `clerkMiddleware`" since #384 made `/` dynamic. Technically true, but it understates the actual probe behavior: for an unauthenticated request to `/`, `clerkMiddleware`'s `auth.protect()` returns a 307 redirect to `/sign-in` *before* the page handler renders. The kubelet treats 307 as success, so liveness gets a 307 on healthy middleware and a 500 on a throwing middleware — the page handler is never the failure surface for the probe. Same effective behavior as the static-prerender era, just by a different mechanism.
**Suggestion:** Optional one-line addition to the comment for future readers: "Anonymous probe requests get a 307 from `auth.protect()` → kubelet treats as success; only a middleware throw produces 500." Skip if you'd rather keep the comment short.

---

### Resolved from prior review

- **Blocking — PR-body summary line.** Now reads `Tests: 39 passed, 0 skipped, 0 failed (19.4 s)`. ✓
- **Reliability — probe timeouts/thresholds.** Both probes now have explicit `timeoutSeconds: 3` and explicit `failureThreshold`. ✓
- **Security — error message leak.** 503 body is now `{ok: false}` only; original error logged server-side via `console.error('[healthz] ...')`. ✓
- **Maintainability — matcher pattern.** Now `/api/healthz(.*)`. ✓
- **Question — does `auth()` exercise the #382 code path?** Traced through Clerk source. It does not — `auth()` only verifies the auth-status header that middleware stamps. The #382 regression is closed by `clerkMiddleware` running on the route at all (matcher covers `/(api|trpc)(.*)`); the route handler's `auth()` is a distinct matcher-misconfig detector. Route file's comment now reflects this precisely.

### Reconciliation with main (4 merges since branch cut)

Branch was rebased onto `c72af26`. Surveyed merges:
- **#383** (mount CLERK_SECRET_KEY in web): no overlap with this PR's files.
- **#387** (NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY to web build): no overlap.
- **#391** (Cloud Run Clerk secret + logout button): no overlap.
- **#390 / #384** (defense-in-depth page-level auth): made `app/page.tsx` dynamic. This invalidated the original "liveness on `/` is safe because `/` is static" rationale. Resolution: kept liveness on `/` (still the right contract — site root is the most representative health signal) and bumped `failureThreshold` to 9. Documented in the values.yaml comment.
- **#392** (per-env web image build): no overlap.

---

brownm09:
Both CI failures on this PR are pre-existing operational gaps, not regressions from the code change:

1. **Build & Push Images** — Artifact Registry returned `504` while writing a layer blob during the API image push (transient GCP infra error, surfaced in the buildx logs). Re-running the failed jobs now via `gh run rerun --failed`.

2. **Staging Integration Tests** — the bot comment above. `lifting-logbook-stg-clerk-secret-key` in GCP Secret Manager is still the `REPLACE_ME` placeholder. The guard at `.github/workflows/staging.yml:410-423` is doing what it's designed to do — skip deploys until the secret is set. This requires a manual one-time step via `docs/deploy.md` Step 4 (set the secret in Secret Manager for the staging project), which is outside the scope of this PR. Same gate would block any PR until the secret is configured; the staging-deploy lane has been gated this way since #391 mounted the secret in the Cloud Run deploy step.

No code change in `apps/web/app/api/healthz/` or `infra/kubernetes/charts/web/values.yaml` can resolve either; the registry 504 needs a retry, and the staging Clerk secret needs to be set in GCP Secret Manager by the operator. Re-running the failed jobs to clear (1); (2) will continue to skip the deploy/integration leg until the secret is configured.

Per the project's failure-tracking rule, citing this here in lieu of a separate tracking issue: the staging-Clerk-secret-placeholder gate is a known operational state, not a regression. Filing a tracking issue would be appropriate if the placeholder is left in place long-term; otherwise it self-resolves on the next operator action.

brownm09:
**Correction to my earlier reply.** I was wrong that the staging Clerk secret was still the `REPLACE_ME` placeholder — the user confirmed it has been set. The actual failure chain on the first run was:

1. `build-images` failed with the transient 504 from Artifact Registry.
2. `deploy-api` was therefore skipped (it `needs: [build-images]`).
3. `clerk_configured` is an output of `deploy-api`'s `Check Clerk secret key configured` step. Because the job was skipped, that output is the empty string, not `'false'`.
4. `integration-tests` runs `Verify deploy prerequisites` at `.github/workflows/staging.yml:597`, which checks `!= 'true'`. Empty string `!= 'true'` is true → the step emits the misleading "Clerk secret not configured" error and the bot scrapes it.

So the guard conflates two distinct upstream states ("`clerk_configured=false`" vs. "deploy-api skipped"). A small workflow fix would distinguish them — e.g., separate branches for "deploy-api did not run" vs. "clerk secret check returned false" — so transient build failures aren't reported as Clerk-config problems. Worth a separate tracking issue; not in scope for this PR.

Rerun in progress on commit `077fe8c`. With the secret actually set, the deploy/integration leg should run normally this time.

PR #403, opened 2026-06-01
[fix] Serialize deploy.yml runs to prevent Terraform state-lock races

brownm09:
## Summary

Adds a top-level `concurrency:` block to `.github/workflows/deploy.yml` so back-to-back pushes to `main` no longer race the staging Terraform state lock.

```yaml
concurrency:
  group: deploy-main
  cancel-in-progress: false
```

`cancel-in-progress: false` queues subsequent runs rather than cancelling — a production deploy in flight must not be aborted by a later push (Cloud Run revision rollout, `terraform apply` mid-flight, secret sync).

## Origin

Deploy run [26763984260](https://github.com/brownm09/lifting-logbook/actions/runs/26763984260) failed with HTTP 412 conditionNotMet on the GCS-backed TF state when two pushes to `main` arrived 11 seconds apart on 2026-05-31 (`fc4117c` then `c72af26`). Independent of #399/#400 — the smoke-URL fix does not address this.

## Acceptance Criteria

- [x] Top-level `concurrency` block added to `deploy.yml`
- [x] Group key is `deploy-main`
- [x] `cancel-in-progress: false`
- [x] PR body explains why cancellation must be disabled (above)

## Testing

Workflow-only YAML change; no test impact. Verified:
- `js-yaml` parse of `.github/workflows/deploy.yml` succeeds (OK)
- `npm test` from repo root with `--filter='!@lifting-logbook/api'`: **Tests: 211 passed, 0 skipped, 0 failed (21s)** across core, types, web, api-legacy, mobile, eslint-rules workspaces

Pre-existing failure: `@lifting-logbook/api` tests fail locally because Docker Desktop is not running in this environment (Testcontainers requires it for the Postgres DB E2E suite — see `apps/api/jest.global-setup.js`). CI runs Docker via service containers and is unaffected. The change in this PR is workflow-only YAML and does not touch any code path the API tests cover.

Functional concurrency behavior cannot be reproduced in isolation — it asserts on the next concurrent push to `main`.

## Coverage / suppression / test-integrity grep

All grep checks clean (no matches).

Closes #401

Issue #401, opened 2026-06-01
Add top-level concurrency block to deploy.yml to prevent Terraform state-lock collisions


brownm09:
## Problem

`.github/workflows/deploy.yml` has no top-level `concurrency:` block. Two pushes to `main` 11 seconds apart on 2026-05-31 (commits \`fc4117c\` then \`c72af26\`) raced for the staging Terraform state lock; the second run failed with HTTP 412 conditionNotMet (Deploy run \`26763984260\`).

This will recur on any rapid back-to-back push. It is independent of #399 — the smoke-URL fix in #400 does not address it.

## Proposed fix

Add at the top of `.github/workflows/deploy.yml`:

\`\`\`yaml
concurrency:
  group: deploy-main
  cancel-in-progress: false
\`\`\`

\`cancel-in-progress: false\` queues subsequent runs rather than cancelling — prod deploys must not be aborted mid-flight.

## Acceptance Criteria

- [ ] Top-level \`concurrency\` block added to \`deploy.yml\`
- [ ] Group key is \`deploy-main\` (or equivalent stable string)
- [ ] \`cancel-in-progress: false\`
- [ ] PR body explains why cancellation must be disabled

## Origin

Surfaced during code review of #400 (smoke-test URL fix) and in the session stub for 2026-06-01.

PR #400, opened 2026-06-01
[fix] Drop duplicate https:// from staging web smoke test URL

brownm09:
## Summary

One-line fix to `.github/workflows/deploy.yml:448`. The terraform output `cloud_run_web_url` already includes the scheme, so prepending `https://` produced `https://https://lifting-logbook-stg-web-...run.app/`. Curl returned `000` on all 18 retries, the smoke job failed, and the production deploy job was skipped — blocking every production deploy in the last 5 main pushes.

The staging API smoke test (line 464) uses the output value directly without prepending a scheme and passes; this aligns the web smoke with that pattern.

## Acceptance Criteria

- [x] Smoke test curl URL contains exactly one `https://` prefix
- [ ] Next Deploy run on `main` reaches the production approval gate (verified post-merge)

## Testing

**Not exercised by `npm test`** — this is a GitHub Actions workflow YAML change with no corresponding unit-test surface in the monorepo. The change is validated by the next Deploy workflow run on `main` after merge.

Tests: not run (workflow-only change; no unit-test coverage applies).

No suppressions added. No test-integrity violations (no skip markers, no deleted tests, no coverage threshold changes).

## Verification (post-merge)

1. Watch the Deploy workflow run triggered by the squash-merge commit.
2. Confirm `Smoke Test (staging) → Smoke test Cloud Run staging (web)` reports `Staging web returned 200` (or `307`) on attempt 1.
3. Confirm `Deploy (production)` reaches `waiting` status for environment approval.

Closes #399

## Review findings

Reviewed by Claude. One non-blocking finding: deploy.yml has no top-level `concurrency` block, allowing rapid back-to-back pushes to race for the Terraform state lock (precedent: runs `26763973774`/`26763984260` on 2026-05-31). Out of scope for this one-line URL fix; filed as #401 against the **CI/CD Foundation** epic in milestone **v0.1 — Foundation**. No blocking findings; no questions for the author (symmetric working pattern at line 464 confirms the fix).

PR #404, opened 2026-06-01
[fix] Add public /healthz route + two-probe staging smoke

### LINKED ISSUE #402: Staging web smoke fails (404) because /  is Clerk-protected; add /healthz to apps/web and probe both routes
## Problem

Deploy run [26770343171](https://github.com/brownm09/lifting-logbook/actions/runs/26770343171) (post-merge of #400) failed at the staging web smoke step. The URL bug fixed by #400 was real but masked a second, latent issue: the smoke targets `/` on the Next.js web app, but `/` is gated by Clerk's `clerkMiddleware()` and returns **404** for anonymous requests (rewritten to a sentinel `/clerk_*` path that doesn't exist):

\`\`\`
HTTP/1.1 404 Not Found
x-clerk-auth-reason: protect-rewrite, dev-browser-missing
x-clerk-auth-status: signed-out
x-middleware-rewrite: /clerk_1780335803276
\`\`\`

The app itself is healthy — `/sign-in` returns 200. Production has been blocked across five `main` deploys for an intermittent variant of this; the post-merge run finally exposed it cleanly because Cloud Run had fully shifted traffic to the new revision before the smoke probe ran (the PR-branch run on #400 happened to probe a prior revision and passed, which is why the URL fix looked complete).

## Decision

Two-probe staging smoke for the web service, with one route owned by app code and one by user-facing flow:

1. **Add `/healthz` to `apps/web`** — a public liveness endpoint owned by our code, excluded from `clerkMiddleware()`'s matcher. Contract: returns `200` iff the Next.js runtime is serving requests. Does **not** touch the database, the API, or downstream deps — a meaningful but minimal liveness signal.
2. **Update `.github/workflows/deploy.yml`** smoke step to probe **both** `/healthz` (expects 200) and `/sign-in` (expects 200). Each curl loop has its own retry budget.

## Rationale

Doing only (1) probe `/sign-in` is a one-line workflow change but couples our liveness signal to Clerk's route naming — a future provider swap or route rename silently breaks the smoke. Doing only (2) add `/healthz` requires the workflow change anyway and gives only a liveness signal, not a sign-of-life on the auth flow.

Doing **both** distinguishes two failure modes that today are indistinguishable:

| Probe result | Diagnosis |
|---|---|
| `/healthz` 200, `/sign-in` 200 | App healthy, auth wired correctly → safe to promote |
| `/healthz` 200, `/sign-in` non-200 | Clerk middleware misconfigured (matcher, env vars, route map) |
| `/healthz` non-200 | App not booting / Next.js runtime broken |

Today's incident is in the second row — five deploys in a row could not surface it because there was only one probe and the URL bug ate the signal. Adding `/healthz` as a separately-owned route makes the next regression of this class diagnosable from the workflow log alone.

The tradeoff is more surface: `/healthz` must stay excluded from Clerk's matcher; matcher drift could re-introduce the 404. This is a known-known and lives in the same file as the matcher — accepted.

This change establishes a workflow-smoke convention rather than a global rule; rationale lives in this issue and the resulting PR. No ADR warranted — the decision is reversible (drop `/healthz`, fall back to a single `/sign-in` probe) and confined to one workflow file plus one route file.

## Acceptance Criteria

- [ ] `apps/web` exposes `GET /healthz` returning `200 OK` with body `ok`
- [ ] `apps/web/middleware.ts` (or wherever `clerkMiddleware()` is configured) excludes `/healthz` from auth gating
- [ ] `.github/workflows/deploy.yml` staging smoke probes both `/healthz` and `/sign-in`, each expecting 200 within the existing retry budget
- [ ] Smoke step's failure message identifies which probe failed
- [ ] Next Deploy run on `main` reaches the production approval gate

## Out of scope

- Adding `/healthz` to the API (already covered by the existing `/health` probe).
- Deepening `/healthz` to check DB/Redis/etc. — start with runtime-only liveness; deepen iff a future incident motivates it.
- The deploy-workflow concurrency block (separate issue #401).

## Origin

Surfaced in the session continuation after merging #400. Session stub: `engineering-journal/sessions/lifting-logbook/2026-06-01_125728.stub.md`.

brownm09:
## Summary

- Adds `GET /healthz` in `apps/web` returning `200 ok` (text/plain).
- Excludes `/healthz` from `clerkMiddleware()`'s matcher so the probe never enters Clerk.
- Splits the staging `smoke-test` step into two probes (`/healthz` and `/sign-in`), each with its own retry budget and a distinct failure message identifying which probe failed.

## Origin

Deploy run [26770343171](https://github.com/brownm09/lifting-logbook/actions/runs/26770343171) (post-merge of #400) failed at the staging web smoke step. The smoke targeted `/`, but `/` is gated by `clerkMiddleware()` and Clerk rewrites anonymous requests to a sentinel `/clerk_*` path that doesn't exist → 404. The app itself was healthy (`/sign-in` returned 200) but the one-probe smoke conflated "Next runtime dead" with "Clerk auth gate firing." Five deploys in a row could not surface this.

## Failure-mode matrix

| `/healthz` | `/sign-in` | Diagnosis |
|---|---|---|
| 200 | 200 | App + auth healthy — safe to promote |
| 200 | non-200 | Clerk middleware misconfigured (matcher / env vars / route map) |
| non-200 | * | Next.js runtime not booting |

`/healthz` is distinct from the existing `/api/healthz` route (added in #395): `/api/healthz` deliberately runs *through* `clerkMiddleware` as the Clerk init-failure regression detector for K8s readiness; `/healthz` is the opposite contract — pure runtime liveness, never enters Clerk.

## Acceptance Criteria

- [x] `apps/web` exposes `GET /healthz` returning `200 OK` with body `ok`
- [x] `apps/web/middleware.ts` matcher excludes `/healthz` (`healthz` added to the matcher's negative lookahead group; comment points at #402)
- [x] `.github/workflows/deploy.yml` staging smoke probes both `/healthz` and `/sign-in`, each expecting 200 within the existing per-step retry budget (18 attempts × 10s each)
- [x] Each probe's failure message identifies which probe failed
- [ ] Next Deploy run on `main` reaches the production approval gate (verified at merge time)

## Out of scope

Per issue: API `/healthz` (covered by existing `/health` probe), DB/Redis-depth checks, and the deploy-workflow concurrency block (separate PR #403 for #401).

## Testing

`npx turbo run test --filter='!@lifting-logbook/api'` from repo root: **Tests: 212 passed, 0 skipped, 0 failed (~18s)** across core, types, web, api-legacy, mobile, eslint-rules workspaces (web suite went from 39 → 40 with the new `app/healthz/route.test.ts`).

Pre-existing failure: `@lifting-logbook/api` tests fail locally because Docker Desktop is not running in this environment (Testcontainers needs it for the Postgres DB E2E suite — see `apps/api/jest.global-setup.js`). CI runs Docker via service containers and is unaffected. This PR does not touch any code path the API tests cover.

YAML parse of `.github/workflows/deploy.yml` succeeds (js-yaml load OK).

Smoke-step behavior is verified end-to-end by the next Deploy run on `main` after merge — that is the issue's acceptance gate.

## Suppression / test-integrity grep

All checks clean (no matches).

Closes #402

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -440,20 +440,43 @@ jobs:
     if: needs.preflight.outputs.staging_enabled == 'true' && needs.deploy-staging.result == 'success'
 
     steps:
-      - name: Smoke test Cloud Run staging (web)
+      # Two-probe staging smoke (#402). One probe owned by app code (/healthz),
+      # one by the user-facing auth flow (/sign-in). They distinguish three
+      # failure modes that a single `/` probe collapsed into "404":
+      #   /healthz 200, /sign-in 200 → safe to promote
+      #   /healthz 200, /sign-in non-200 → Clerk middleware misconfigured
+      #   /healthz non-200 → Next.js runtime not booting
+      # /healthz is excluded from clerkMiddleware's matcher; /sign-in is
+      # auth-public via createRouteMatcher.
+      - name: Smoke test Cloud Run staging — /healthz (web runtime)
         run: |
+          BASE_URL="${{ needs.terraform-staging.outputs.cloud_run_web_url }}"
           # Allow up to 3 minutes for Cloud Run to become healthy
           for i in $(seq 1 18); do
-            STATUS=$(curl -s -o /dev/null -w "%{http_code}" \
-              "${{ needs.terraform-staging.outputs.cloud_run_web_url }}/" || true)
-            if [ "$STATUS" = "200" ] || [ "$STATUS" = "307" ]; then
-              echo "Staging web returned $STATUS — OK"
+            STATUS=$(curl -s -o /dev/null -w "%{http_code}" "$BASE_URL/healthz" || true)
+            if [ "$STATUS" = "200" ]; then
+              echo "Staging /healthz returned 200 — OK"
               exit 0
             fi
-            echo "Attempt $i: got $STATUS, retrying in 10s..."
+            echo "/healthz attempt $i: got $STATUS, retrying in 10s..."
             sleep 10
           done
-          echo "Staging smoke test timed out" && exit 1
+          echo "Staging /healthz smoke FAILED — Next.js runtime not serving 200 (probe owned by app code, distinct from Clerk auth flow)" && exit 1
+
+      - name: Smoke test Cloud Run staging — /sign-in (Clerk auth flow)
+        run: |
+          BASE_URL="${{ needs.terraform-staging.outputs.cloud_run_web_url }}"
+          # Allow up to 3 minutes for Clerk middleware + sign-in route to settle
+          for i in $(seq 1 18); do
+            STATUS=$(curl -s -o /dev/null -w "%{http_code}" "$BASE_URL/sign-in" || true)
+            if [ "$STATUS" = "200" ]; then
+              echo "Staging /sign-in returned 200 — OK"
+              exit 0
+            fi
+            echo "/sign-in attempt $i: got $STATUS, retrying in 10s..."
+            sleep 10
+          done
+          echo "Staging /sign-in smoke FAILED — runtime healthy but Clerk auth flow not serving 200 (check middleware matcher and Clerk env vars)" && exit 1
 
       - name: Smoke test Cloud Run staging (API auth check)
         run: |
--- a/apps/web/app/healthz/route.test.ts
+++ b/apps/web/app/healthz/route.test.ts
@@ -1,0 +1,14 @@
+/**
+ * @jest-environment node
+ */
+import { GET } from './route';
+
+describe('GET /healthz', () => {
+  it('returns 200 with text body "ok"', async () => {
+    const res = GET();
+
+    expect(res.status).toBe(200);
+    expect(res.headers.get('content-type')).toBe('text/plain');
+    expect(await res.text()).toBe('ok');
+  });
+});
--- a/apps/web/app/healthz/route.ts
+++ b/apps/web/app/healthz/route.ts
@@ -1,0 +1,20 @@
+import { NextResponse } from 'next/server';
+
+// Pure runtime liveness probe for the staging deploy smoke (#402). Returns 200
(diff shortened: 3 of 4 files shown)


brownm09:
## Review: [fix] Add public /healthz route + two-probe staging smoke

### Summary
Merge with one minor consideration noted. The two-probe split is well-scoped, the failure-mode matrix in the PR body matches the workflow's exit messages, and `/healthz` is correctly separated from the existing `/api/healthz` (which deliberately runs *through* Clerk per #395). Tests added; matcher comment explains the negation. No blocking findings.

---

### Non-Blocking Findings

**[maintainability]** `apps/web/middleware.ts:36`
The matcher exclusion `healthz` is a prefix match inside the negative lookahead, not an anchored path segment — `(?!_next|healthz|...)` would also exclude any future route starting with `healthz` (e.g., `/healthzfoo`, `/healthz-admin`). Today no such route exists, so the bypass is purely theoretical, but a contributor who adds one later would silently skip Clerk middleware without realising. The same prefix-match shape applies to `_next`, where it is intentional (Next reserves the prefix); for `healthz` the intent is the single route.
**Suggestion:** Either tighten the negation to a path-segment match (e.g., add a separate `'/((?!healthz(?:/|$)).*)'` line, or move the healthz exclusion out of the matcher and gate inside the middleware function via `request.nextUrl.pathname === '/healthz'`), or leave a comment at the matcher noting the prefix-match shape so a future `/healthzfoo` addition does not surprise.

**[reliability]** `.github/workflows/deploy.yml:455-479`
Both probes use a 3-minute (18 × 10s) budget back-to-back, so a slow staging revision could consume up to 6 minutes total before the step fails. This is a strict superset of the previous one-probe budget, not a regression — just worth knowing if the deploy job's overall timeout is tight.
**Suggestion:** None unless a tighter overall budget is needed; the additional minutes are an acceptable cost for the diagnostic value the second probe provides.

---

### Questions for Author
None — PR body, route comment, and middleware comment together resolve every intent question raised by the diff.

---

Issue #407, opened 2026-06-02
[bug] Staging /healthz returns 404 despite route being on main — blocks every main Deploy run at smoke gate


brownm09:
## Symptom

`/healthz` returns **404** on the deployed staging Cloud Run revision, even though `apps/web/app/healthz/route.ts` is committed to `main` (verified via `git show origin/main:apps/web/app/healthz/route.ts`). This blocks every `main` Deploy run at the staging smoke step → production approval gate is never reached.

Three consecutive `main` Deploy runs have failed at the same step:

| Run | SHA | Failure |
|---|---|---|
| [26797013015](https://github.com/brownm09/lifting-logbook/actions/runs/26797013015) | `9937da1` | `/healthz` 404 × 18 attempts |
| [26780207624](https://github.com/brownm09/lifting-logbook/actions/runs/26780207624) | `b9f76a2` | `/healthz` 404 × 18 attempts |
| [26779634287](https://github.com/brownm09/lifting-logbook/actions/runs/26779634287) | `1881226` | `/healthz` 404 × 18 attempts |

## Live staging surface (probed 2026-06-02)

```
GET https://lifting-logbook-stg-web-mn3pkptvma-uc.a.run.app/            → 404
GET https://lifting-logbook-stg-web-mn3pkptvma-uc.a.run.app/healthz     → 404   ← bug
GET https://lifting-logbook-stg-web-mn3pkptvma-uc.a.run.app/sign-in     → 200
GET https://lifting-logbook-stg-web-mn3pkptvma-uc.a.run.app/api/healthz → 200
```

`/sign-in` (Clerk-public) and `/api/healthz` (#395, runs *through* Clerk) both return 200 — the Next.js runtime is up and Clerk middleware is correctly configured. The Clerk-protect-rewrite that caused the original `/` 404 in #402 has been ruled out: this is `/healthz` specifically not being reachable.

## Why this is a real bug (not just stale traffic)

The PR-branch CI for #404 ([run 26779312778](https://github.com/brownm09/lifting-logbook/actions/runs/26779312778)) showed `/healthz` → 200 and `/sign-in` → 200, which is why #404 merged with confidence. Hypothesis at the time: post-merge Cloud Run was still serving the pre-#404 revision while the PR-branch deploy had pushed a fresh tagged revision. That hypothesis predicts the bug would clear once `main`'s revision rolled out.

[...]

PR #408, opened 2026-06-02
[fix] Guard /healthz in standalone build; isolate web image cache scopes

brownm09:
## Summary

Three consecutive `main` Deploy runs ([9937da1](https://github.com/brownm09/lifting-logbook/actions/runs/26797013015), [b9f76a2](https://github.com/brownm09/lifting-logbook/actions/runs/26780207624), [1881226](https://github.com/brownm09/lifting-logbook/actions/runs/26779634287)) have failed at the staging smoke gate with `/healthz` → 404, blocking production approval. The route file is committed on main and `/api/healthz` (a sibling route under `app/api/`) returns 200 — so the Next.js runtime is up, Clerk middleware is correctly configured, and the issue is specific to `/healthz`.

PR #404's PR-branch CI saw `/healthz` → 200, so the build is capable of producing the route. Something diverges between PR-branch builds (staging.yml) and main builds (deploy.yml), both of which use the same Dockerfile and previously shared an unscoped `type=gha` BuildKit cache.

This PR ships two complementary changes:

1. **Build-time assertion** in `apps/web/Dockerfile`: a `RUN test -d` step after `next build` that fails the image build if `.next/standalone/.../app/healthz` or `.../app/api/healthz` is missing. This is the regression guard required by AC #5 and serves as a deterministic diagnostic — if the assertion fails on the next deploy, the failure mode is build-time exclusion (confirmed); if it passes but the route still 404s, the failure mode is runtime routing (escalates to investigation, not a blind smoke retry).
2. **Per-flavor cache scopes** in `deploy.yml` and `staging.yml`: each web image flavor now has its own GHA cache scope (`web-staging-v2`, `web-prod-v2`, `web-pr-v2`) instead of all three sharing the unscoped namespace. The `-v2` suffix forces a one-time cache miss on merge, rebuilding the `next build` layer from clean inputs. This is the most likely root cause given the static evidence and is hypothesis #3 from the issue.

Closes #407

## Root cause (suspected, pending CI confirmation)

GHA BuildKit cache poisoning on the post-merge main runs. The three web image builds (staging.yml's `web-image`, deploy.yml's `web (staging)`, deploy.yml's `web (production)`) shared an unscoped `type=gha` cache. A stale `next build` layer cached from before `apps/web/app/healthz/route.ts` existed (#404 added it) appears to have been silently reused. The new scopes isolate per-flavor caches; the `-v2` suffix breaks lineage.

[...]

Issue #405, opened 2026-06-01
Follow-up to #401/#402: review findings from PR #403 & #404


brownm09:
Post-merge `/review` of [#403](https://github.com/brownm09/lifting-logbook/pull/403) and [#404](https://github.com/brownm09/lifting-logbook/pull/404) surfaced seven non-blocking items to roll into a single follow-up PR.

## From #403 (concurrency block)

1. **Doc accuracy in `.github/workflows/deploy.yml`** — in-file comment and PR body said `cancel-in-progress: false` "queues subsequent runs." GitHub Actions retains at most one pending run per group; intermediate commits between the running and latest-queued run are silently cancelled. Append one sentence clarifying this so a future operator debugging "why did commit X never deploy?" finds the answer in the file.
2. **Static group-key safety note** — `group: deploy-main` is correct while `push: main` is the only trigger. Add a comment line noting future triggers (workflow_dispatch, release, additional branches) would require a parameterized key.

## From #404 (/healthz + two-probe smoke)

3. **Matcher unit test** — `apps/web/middleware.ts` excludes `/healthz` from Clerk via `healthz(?:[/?]|$)`. There is no test that asserts \`/healthz\` bypasses Clerk **and** \`/healthzfoo\`, \`/healthz-admin\` still enter it. A future "simplify" of the regex back to bare \`healthz\` would silently widen the bypass without breaking any test. Add a unit test that imports \`config.matcher[0]\` and asserts both the positive and negative sets.
4. **Tighten matcher anchor** — change \`healthz(?:[/?]|$)\` to \`healthz(?:\?|$)\` so only the bare path and querystring variants bypass Clerk; any future \`/healthz/admin\` subroute then enters auth by default. (Routes under \`apps/web/app/healthz/\` would need explicit opt-out if intentionally public.) The unit test from item 3 should encode this.
5. **\`/sign-in\` retry budget** — once \`/healthz\` returns 200 the runtime is up; \`/sign-in\` should respond within seconds. Drop the second probe from 18×10s (180s) to 6×10s (60s) so failures surface faster.
6. **Normalize \`BASE_URL\`** — add \`BASE_URL=\"\${BASE_URL%/}\"\` to both smoke steps so a future Terraform module emitting a trailing slash doesn't produce \`https://.../?\` double-slash requests.
[...]

PR #410, opened 2026-06-02
[fix] Rename /healthz → /livez (Google Frontend intercepts /healthz on Cloud Run)

### LINKED ISSUE #409: Staging /healthz still 404 post-#408 — runtime route registration, not build-time exclusion
## Context

PR #408 (closed #407) shipped two changes:
1. Per-flavor GHA cache scopes (`web-staging-v2`, `web-prod-v2`, `web-pr-v2`) — hypothesis #3 from #407
2. A `RUN test -d` build-time assertion in `apps/web/Dockerfile` verifying `/healthz` and `/api/healthz` directories exist in `.next/standalone/.../app/`

Post-merge main Deploy run [26799958951](https://github.com/brownm09/lifting-logbook/actions/runs/26799958951):
- Build & Push Images: **passed** (assertion did not trip — both healthz directories present in standalone tree)
- Smoke Test (staging): **failed** — `/healthz` returned 404 on all 18 retry attempts

## New evidence (live staging headers)

```
$ curl -sI .../healthz
HTTP/1.1 404 Not Found
Content-Type: text/html; charset=UTF-8
Content-Length: 1568
(no x-middleware-rewrite, no x-clerk-auth-*, no x-powered-by)

$ curl -sI .../api/healthz
HTTP/1.1 200 OK
x-clerk-auth-status: signed-out
x-middleware-rewrite: /api/healthz

$ curl -sI .../sign-in
HTTP/1.1 200 OK
x-clerk-auth-status: signed-out
x-middleware-rewrite: /sign-in
```

The 404 on `/healthz` has no Clerk/middleware headers (matcher correctly excludes it, as designed) but ALSO no Next.js indication that the route was even attempted. Other responses carry `x-middleware-rewrite`, this one doesn't.

## Diagnosis

This is **hypothesis #2 from #407: runtime route registration**, not hypothesis #3 (cache poisoning, which is now disproven). The build copies `route.js` into `.next/standalone/.../app/healthz/route.js`, but the Next.js runtime's route lookup doesn't register it.

Most likely sub-causes:
1. Next.js standalone trace pruning excludes the route from `.next/standalone/.../app-paths-manifest.json` even though `route.js` is copied
2. The route is registered but bypassed at the runtime layer because of some `dynamic = 'force-dynamic'` interaction with the standalone server
3. Static-export pre-rendering produces a different artifact for `/healthz` (no page chunk) than the standalone server expects

The Dockerfile `test -d` assertion shipped in #408 was **insufficient** — it verifies the file exists but doesn't verify the route is in `app-paths-manifest.json`. Will be tightened as part of the fix.

## Investigation plan

1. Add a temporary diagnostic RUN step in `apps/web/Dockerfile` that dumps the relevant Next.js manifests (`app-paths-manifest.json`, `middleware-manifest.json`, `routes-manifest.json`, `prerender-manifest.json`) from both the pre-standalone tree (`/app/apps/web/.next/server/`) and the standalone tree (`/app/apps/web/.next/standalone/apps/web/.next/server/`) to the build log.
2. Open the resulting PR, push, read the GHA build log to identify which manifest is missing `/healthz`.
3. Patch based on the actual cause.

## Acceptance criteria

- [ ] Root cause identified from manifest comparison
- [ ] `/healthz` returns 200 on staging after deploy
- [ ] Build-time assertion tightened to assert `/healthz` is in `app-paths-manifest.json`, not just on disk
- [ ] Main Deploy reaches production approval gate
- [ ] Diagnostic dump step reverted before merge

Production deploys remain blocked until this is resolved.

brownm09:
## Summary

Fixes the staging `/healthz` 404 that has blocked every `main` Deploy since PR #406 merged on 2026-06-01. After three wrong hypotheses (cache poisoning in #408, HEAD const-alias and middleware matcher narrowing in the first two #410 attempts), the actual root cause was identified empirically: **Google Frontend (GFE) on Cloud Run intercepts the path `/healthz` before requests reach the container.** This PR renames the route to `/livez` (Kubernetes liveness convention) which GFE does not intercept.

## Root cause

Live probes of every sibling path against the deployed staging revision:

```
/healthz: 404 [google]          ← branded Google 404 page, request never reaches container
/livez: 404 [next]              ← Next.js 404 (route doesn't exist there yet — this PR adds it)
/readyz: 404 [next]
/health: 404 [next]
/healthy: 404 [next]
/healthcheck: 404 [next]
/healthzz: 404 [next]
/api/healthz: 200               ← under /api/ prefix, not intercepted
/sign-in: 200
```

Only `/healthz` is intercepted. The image build, app-paths-manifest.json, standalone tree, and route handler were always correct — `/healthz` requests just never reached the Next.js server. The 404 body has Google logo references (`<a href=//www.google.com/>`), not the Next.js 404 page Next.js's `_not-found/page.js` would generate.

PR #404's PR-branch CI saw `/healthz` → 200 — either GFE behavior on Cloud Run direct URLs changed between then and #406's merge, or that probe ran against a different URL.

## Disproven hypotheses (preserved in git history)

| PR / commit | Hypothesis | Evidence that disproved it |
|---|---|---|
| #408 `8c8f24b` | GHA BuildKit cache poisoning across web flavors | Dockerfile `test -d` assertion passed on the post-merge main Deploy; smoke still 404'd |
| #410 `c2fbf02` | `export const HEAD = GET` breaks Next.js 16 runtime route dispatch | Live `/healthz` curl still 404 after `9578b06` deployed with HEAD export removed |
| #410 `492de57` | Middleware matcher narrowing from `healthz(?:[/?]|$)` to `healthz(?:\?|$)` | Live `/healthz` curl still 404 after matcher reverted |

## Changes

1. **`apps/web/app/healthz/` → `apps/web/app/livez/`** (git mv; route contents updated to reflect new name; in-file comments preserve historical context for future readers).
2. **`apps/web/middleware.ts`** — matcher anchor now `livez(?:\?|$)`, restoring the #405 path-segment lockdown (nested `/livez/*` enters Clerk).
3. **`apps/web/middleware.matcher.test.ts`** — ENTER/BYPASS cases switched to `/livez`.
4. **`.github/workflows/deploy.yml`** — smoke probe path `/healthz` → `/livez`; comments document the GFE-intercept finding.
5. **`apps/web/Dockerfile`** — directory `test -d` assertion + manifest `node -e` check updated for `/livez/route` and `/api/healthz/route`.
6. **`.gitignore`** — `.claude/scheduled_tasks.lock` (Claude harness scratch).

The two-probe staging smoke (#402) still distinguishes Next-runtime-up (`/livez`) from Clerk-init-up (`/api/healthz`); only the runtime probe path changes.

## Closes

Closes #409
Refs #407

## Testing

```
```

Result: `Tests: 56 passed, 0 skipped, 0 failed` across 8 suites (11.3s).

### Coverage gate
- Renamed route's test (`apps/web/app/livez/route.test.ts`) covers GET behavior and the HEAD-undefined regression guard (conservative — see file comment).
- Matcher tests cover all path-segment cases for `/livez`.

### Suppression / test-integrity greps
```
git diff origin/main -- . | grep -E '(ts-ignore|ts-expect-error|...)' → no matches
git diff origin/main -- . | grep -E '(it\.skip|xit\(|...)'           → no matches
git diff --diff-filter=D --name-only origin/main -- '*.test.*' '*.spec.*' → apps/web/app/healthz/route.test.ts (renamed to apps/web/app/livez/route.test.ts; replacement covers same surface and adds /livez-named regression guards)
git diff origin/main -- jest.config.* .nycrc vitest.config.*         → no matches
```

The `git mv` shows as a delete+add in the diff; the replacement test covers the same surface area plus the HEAD-undefined regression guard. No coverage degradation.

## Verification

Will be verified by the post-merge `main` Deploy run's smoke probe — that's the only path that exercises the new `/livez` route from outside the cluster.

## Note on Next.js empirical guards

The HEAD-undefined regression test (in `apps/web/app/livez/route.test.ts`) is kept as a conservative guard. The HEAD const-alias hypothesis was disproven by live curl (the real bug was the GFE-intercepted path), but Next.js 16.2.4's behavior under `export const HEAD = GET` was never independently verified — prefer `export async function HEAD()` if HEAD is ever needed.


brownm09:
## Review: [fix] Remove HEAD const-alias from /healthz route (Next.js 16 runtime dispatch bug)

### Summary
Merge with minor process change. The root-cause analysis is thorough and the empirical evidence chain (#404 worked with only GET → #406 added `HEAD = GET` → main started 404'ing → manifest diagnostic ruled out every build-time hypothesis → removing the one export restores GET) is sound. Diff is small (71 lines), focused, includes a regression guard test, and honestly acknowledges what the fix does and doesn't prove. The one substantive concern is process: the actual fix commit (`c2fbf02`) hasn't been verified against the live staging deploy yet — the PR-branch staging deploy is still serving the diagnostic-only image (`58b984d`). The staging.yml workflow on this PR will deploy `c2fbf02` and re-run the smoke probe before merge — wait for that to return `/healthz` → 200 before merging.

---

### Non-Blocking Findings

**[reliability]** `apps/web/Dockerfile` (new `RUN node -e` step)
The new manifest-presence assertion would have passed for the bug this PR fixes — the PR body's diagnostic dump confirmed `/healthz/route` appeared in `app-paths-manifest.json` (both pre-standalone and standalone) while runtime still 404'd. The guard is honest about being necessary-but-not-sufficient, but it does not protect against the specific failure mode that motivated #409 (a route exported via a const-alias pattern that passes static analysis but fails runtime dispatch). The Jest test in `route.test.ts` is the actual defense for this class.
**Suggestion:** Either accept that the Jest test owns this class (current PR posture) and leave the Dockerfile guard as-is for the more general trace-pruning regressions, OR add a complementary grep that flags `export const HEAD = ` in any route handler under `apps/web/app/`. The grep is narrow but cheap and catches the exact pattern in one more place.

---

**[maintainability]** `apps/web/app/healthz/route.test.ts:18`
The regression guard asserts `route.HEAD === undefined`. The `route.ts` comment correctly tells future developers that HEAD can be re-added as a function declaration (`export async function HEAD()`) — but if someone follows that guidance, this test will fail and there's no cross-reference telling them they also need to update it. The test comment mentions the issue numbers but not its own coupling to the future-change path.
**Suggestion:** Add one line to the `route.ts` HEAD comment: `// If re-added, update apps/web/app/healthz/route.test.ts:18 (the regression guard asserts HEAD is undefined).`

---

**[reliability]** Verification gate (process, not code)
The PR body says "verification pending post-merge." But the staging.yml workflow on this PR will deploy `c2fbf02` to staging and rerun the smoke probe before merge — that's the actual pre-merge verification. Merging without confirming the new image serves `/healthz` → 200 on staging risks shipping a fix that doesn't fix (we already saw the cache-poisoning fix on #408 sail through a green PR-branch smoke and fail post-merge — though that was the cancellation/branch-vs-main divergence, not the same failure mode).
**Suggestion:** Treat the PR-branch `Deploy web (staging)` + `Smoke Test (staging)` jobs on commit `c2fbf02` as the pre-merge gate. If they're green and a live `curl /healthz` from the staging URL returns 200, merge. If they fail, the diagnostic-tightened Dockerfile assertion + the live smoke output will point at whatever remains.

---

### Style Notes
The Dockerfile node script duplicates loop scaffolding for a small fixed list — `paths × required` is a 2×2 matrix and is fine inline, no need to refactor. The Jest file imports `route` two ways (`import * as route` AND `import { GET }`) which is mild duplication; `route.GET()` would work, but the dual-import keeps the GET test readable without `route.` prefix everywhere.

---

brownm09:
Not acted on — same cancellation artifact as [PR #408's instance](https://github.com/brownm09/lifting-logbook/pull/408#issuecomment-4598843385). The bot comment references commit `c2fbf02`, whose Staging run was cancelled by my comment-nit push of `7af2e7a` minutes earlier. Per the Error Message Diligence rule (CLAUDE.md / lifting-logbook#395 incident), the integration-tests guard's `clerk_configured != 'true'` check fires identically when an upstream job in `build-images → deploy-api → integration-tests` doesn't run; cancellation produces the same message as real misconfiguration. The Clerk secret is configured — confirmed minutes ago when the previous (uncancelled) Staging run's `Check Clerk secret key configured` step passed. The new Staging run on `7af2e7a` is in progress; will wait for its real verdict before any action.

Issue #421, opened 2026-06-02
[fix] apps/web jest suites fail to compile — missing jest-dom and Node process.env type augmentations


brownm09:
## Problem

`npm test -w @lifting-logbook/web` fails at the TypeScript compile step for 6 of 8 test suites with two distinct categories of TS2339:

1. `Property 'toBeInTheDocument'/'toBeDisabled'/'toHaveTextContent' does not exist on type 'JestMatchers<HTMLElement>'` — the `@testing-library/jest-dom` matcher augmentations are not visible to the test compiler.
2. `Property 'env' does not exist on type 'Process'` (in `app/page.test.tsx`) — the global Node `process` type is the minimal browser-like one, not the one provided by `@types/node`.

The 23 tests that do compile pass. Verified pre-existing on \`origin/main\` (HEAD \`73f89a1\`) by stashing the working tree and re-running.

## Symptom

\`\`\`
Test Suites: 6 failed, 2 passed, 8 total
Tests:       23 passed, 23 total
\`\`\`

Failing files include:
- \`app/page.test.tsx\` (process.env)
- \`app/(authed)/onboarding/steps/StepProgram.test.tsx\` (jest-dom matchers)
- 4 others in the same category

## Likely cause

Either the \`tsconfig\` \`types\` array for \`apps/web\` no longer includes \`@testing-library/jest-dom\` and \`node\`, or the jest setup file is not importing \`@testing-library/jest-dom\` (which provides the type augmentation as a side effect). Worth checking whether [#373](https://github.com/brownm09/lifting-logbook/issues/373) (Node 24 truncated \`.d.ts\` extraction) caused this on the developer's machine specifically — but reproduction on a freshly-stashed working tree against \`origin/main\` rules out the local-only branch hypothesis.

## Impact

**Medium.** Blocks \`npm test\` from going green locally, which forces every PR to claim "pre-existing failure" in the PR body and erodes the failure-tracking rule. Does not block CI if CI's Node version produces a clean install — but if CI is also red, this is a higher-severity blocker.

## Acceptance criteria

- [ ] \`npm test -w @lifting-logbook/web\` compiles all 8 suites against \`origin/main\`.
- [ ] Root cause identified (tsconfig vs setup file vs install corruption).
- [ ] PR notes whether CI was masking this and how.

brownm09:
[...]

PR #422, opened 2026-06-02
[fix] Disambiguate staging verify-prereqs guard (skipped vs misconfigured)

### LINKED ISSUE #398: [fix] staging.yml verify-prereqs guard conflates upstream-skipped with clerk-misconfig
## Problem

The `integration-tests` job in `.github/workflows/staging.yml` has a "Verify deploy prerequisites" step (`staging.yml:595-604`) that conflates two distinct upstream states into a single misleading error message.

```yaml
- name: Verify deploy prerequisites
  run: |
    if [[ "${{ needs.deploy-api.outputs.clerk_configured }}" != "true" ]]; then
      echo "::error::Staging Clerk secret key is not configured. Follow docs/deploy.md Step 4 to set lifting-logbook-stg-clerk-secret-key in GCP Secret Manager, then re-run CI."
      exit 1
    fi
```

`needs.deploy-api.outputs.clerk_configured` is set to `'true'` or `'false'` by the `Check Clerk secret key configured` step at `staging.yml:410-423` — **only when `deploy-api` actually runs**. When `deploy-api` is skipped (because its upstream `build-images` failed, e.g., a transient Artifact Registry 504), the output is the empty string, not `'false'`. The guard's `!= 'true'` check fires identically in both cases, but reports the skip as a Clerk-secret misconfiguration.

The `github-actions[bot]` review-comment template in the same file scrapes this message and emits it to the PR as a configuration warning. Users (and the assistant) then chase a non-existent secret problem instead of the real upstream build failure.

## Incident

PR #395 (2026-06-01). First CI run: `build-images` 504'd on layer-blob push to Artifact Registry; `deploy-api` was skipped; integration-tests printed the "Staging Clerk secret key is not configured" error and the bot posted a config warning on the PR. The staging Clerk secret was actually correctly set. Full chain: https://github.com/brownm09/lifting-logbook/pull/395#issuecomment-4594434736

## Proposed fix

Distinguish the two upstream states in the guard. One sketch:

```yaml
- name: Verify deploy prerequisites
  run: |
    CLERK_OUT='${{ needs.deploy-api.outputs.clerk_configured }}'
    if [[ -z "$CLERK_OUT" ]]; then
      echo "::error::deploy-api was skipped — likely because build-images failed. Check the 'Build & Push Images' job; re-run failed jobs if the failure was transient (e.g., Artifact Registry 504)."
      exit 1
    fi
    if [[ "$CLERK_OUT" != "true" ]]; then
      echo "::error::Staging Clerk secret key is not configured. Follow docs/deploy.md Step 4 to set lifting-logbook-stg-clerk-secret-key in GCP Secret Manager, then re-run CI."
      exit 1
    fi
    # same split for cloud_run_api_deployed if relevant
```

Also update the github-actions bot review-comment template (search `staging.yml` for the bot block) to mirror the same distinction so it doesn't auto-post a Clerk-config warning when the actual cause is upstream-skip.

## Acceptance criteria

- [ ] `Verify deploy prerequisites` step distinguishes empty (upstream skipped) from `'false'` (real misconfig) and emits separate error messages with distinct, actionable guidance.
- [ ] Same split applied to `cloud_run_api_deployed` check if it suffers the same conflation.
- [ ] Bot review-comment template updated so a transient-build cascade does not post a Clerk-config warning on the PR.
- [ ] Tested by deliberately failing `build-images` (or by reading the next transient build-failure run) and confirming the new error message names the upstream skip.

## Impact

**Medium.** Recurs every time a transient build failure occurs — wastes user time chasing a phantom Clerk-config issue, pollutes PRs with misleading bot warnings, and silently teaches operators to ignore the bot when it might be reporting a real misconfig. Token-cheap to fix.

brownm09:
## Summary

Closes #398.

The `Verify deploy prerequisites` step in `staging-integration-tests` and the bot comment template in `report-status` both used `needs.deploy-api.outputs.clerk_configured != 'true'` as a proxy for "Clerk misconfigured." When `deploy-api` is skipped (because `build-images` failed), the output comes through as the empty string, not `'false'`, so the check fired identically in both cases and mis-reported transient build failures (e.g., Artifact Registry 504) as a phantom Clerk-secret misconfiguration.

Both call sites now consult `needs.deploy-api.result` first:
- `result !== 'success'` (covers `failure`, `skipped`, `cancelled`) → distinct error naming the upstream cause. The bot template further refines the message with an `upstreamHint` ternary so `skipped`/`cancelled` points at `build-images` while `failure` points at a step error inside `deploy-api`.
- `success` with `clerk_configured == 'false'` → unchanged Clerk-config message.
- `success` with `cloud_run_api_deployed == 'false'` → unchanged Cloud Run failure message.

Because `staging-integration-tests` runs under `if: always() && needs.preflight.outputs.should_run == 'true'` (see [#345](https://github.com/brownm09/lifting-logbook/issues/345)), a hard-fail (`deploy-api.result === 'failure'`) is also reachable here, not just `skipped`/`cancelled`. The v1 patch only branched on `skipped`/`cancelled` — it was extended to `!= 'success'` after `/review` flagged the `failure` gap.

Motivating incident: [PR #395](https://github.com/brownm09/lifting-logbook/pull/395) (2026-06-01).

## Acceptance criteria

- [x] `Verify deploy prerequisites` step distinguishes any non-`success` `deploy-api.result` (empty outputs) from `'false'` outputs (real misconfig) and emits separate error messages with distinct, actionable guidance.
- [x] Same split applied to `cloud_run_api_deployed` (now gated behind the `deploy-api.result` check, so the empty-string conflation can no longer reach it).
- [x] Bot review-comment template (`report-status` job) updated so a transient-build cascade does not post a Clerk-config warning on the PR; the message names the right cause for `failure` vs `skipped`/`cancelled`.
- [ ] Behavior verification deferred to the next real transient `build-images` failure — there is no clean way to deliberately fail Artifact Registry in CI without invasive scaffolding. See Test plan below.

## Test plan

- [x] **Unit / integration suite** — `npm test` from the repo root.
  - Tests: 23 passed, 0 skipped, 0 failed (132 s) in the suites that compile. 6 of 8 `apps/web` test suites fail to *compile* with `TS2339` for `@testing-library/jest-dom` matchers and `process.env` — these are TypeScript compile errors, not test skips/failures. **Pre-existing on `origin/main`** (verified by stashing this diff and re-running). Tracked in [#421](https://github.com/brownm09/lifting-logbook/issues/421). Outside the scope of this YAML-only change.
  - Workspace summary: 10 of 12 turbo tasks succeeded; only `@lifting-logbook/web#test` failed, for the reason above. `@lifting-logbook/core`, `@lifting-logbook/api`, and remaining workspaces all green.
- [ ] **CI staging workflow** — exercised by this PR's own staging run; if `build-images` succeeds, the new branches are not hit but parsing/structure is validated. If `build-images` fails, the new "deploy-api did not succeed" message should appear.
- [ ] **Real behavior verification** — confirm against the next transient `build-images` failure (or deploy-api step failure) that the new guard message names the upstream state and that the bot posts the corresponding comment.

## Compliance checks

- **Suppression policy grep**: clean (YAML-only edit).
- **Test integrity grep**: clean.
- **Test integrity skipped count**: 0 skipped in the runs that completed; pre-existing failures are compile-time TS errors, not skip markers.

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -599,11 +599,24 @@ jobs:
     steps:
       - name: Verify deploy prerequisites
         run: |
-          if [[ "${{ needs.deploy-api.outputs.clerk_configured }}" != "true" ]]; then
+          # deploy-api.result is the source of truth for "did the job run at all".
+          # When build-images fails, deploy-api is skipped (per its `if:` guard) and
+          # its outputs come through as empty strings, not 'false' — so a plain
+          # `clerk_configured != 'true'` check would mis-report the skip as a Clerk
+          # misconfiguration. See #398 / PR #395 for the motivating incident.
+          DEPLOY_RESULT='${{ needs.deploy-api.result }}'
+          CLERK_OUT='${{ needs.deploy-api.outputs.clerk_configured }}'
+          API_DEPLOYED='${{ needs.deploy-api.outputs.cloud_run_api_deployed }}'
+
+          if [[ "$DEPLOY_RESULT" == "skipped" || "$DEPLOY_RESULT" == "cancelled" ]]; then
+            echo "::error::deploy-api did not run (result: $DEPLOY_RESULT) — likely because build-images failed. Check the 'Build & Push Images' job; re-run failed jobs if the failure was transient (e.g., Artifact Registry 504 on layer-blob push)."
+            exit 1
+          fi
+          if [[ "$CLERK_OUT" != "true" ]]; then
             echo "::error::Staging Clerk secret key is not configured. Follow docs/deploy.md Step 4 to set lifting-logbook-stg-clerk-secret-key in GCP Secret Manager, then re-run CI."
             exit 1
           fi
-          if [[ "${{ needs.deploy-api.outputs.cloud_run_api_deployed }}" != "true" ]]; then
+          if [[ "$API_DEPLOYED" != "true" ]]; then
             echo "::error::Cloud Run API deploy failed — integration tests cannot run. Check the 'Fetch Cloud Run API logs on failure' step in the Deploy API (staging) job. Tracked in #344."
             exit 1
           fi
@@ -714,13 +727,19 @@ jobs:
           script: |
             const shouldRun = '${{ needs.preflight.outputs.should_run }}' === 'true';
             const testResult = '${{ needs.staging-integration-tests.result }}';
+            const deployApiResult = '${{ needs.deploy-api.result }}';
             const webUrl = '${{ needs.deploy-web.outputs.web_url }}';
 
             let body;
             const clerkConfigured = '${{ needs.deploy-api.outputs.clerk_configured }}' === 'true';
             const cloudRunApiDeployed = '${{ needs.deploy-api.outputs.cloud_run_api_deployed }}' === 'true';
+            // deploy-api skipped/cancelled means an upstream job (typically build-images)
+            // failed. Surface that explicitly so the bot doesn't mis-report the cascade
+            // as a Clerk-secret misconfiguration — see #398 / PR #395.
             if (!shouldRun) {
               body = '⚪ **Staging skipped** — `GCP_STAGING_WORKLOAD_IDENTITY_PROVIDER` is not configured.';
+            } else if (deployApiResult === 'skipped' || deployApiResult === 'cancelled') {
+              body = `⚠️ **Staging integration tests did not run** — \`deploy-api\` was ${deployApiResult}, likely because \`build-images\` failed. Re-run failed jobs in the [workflow run](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}) if the failure was transient (e.g., Artifact Registry 504 on layer-blob push).`;
             } else if (!clerkConfigured) {
               body = '⚠️ **Staging integration tests skipped** — `lifting-logbook-stg-clerk-secret-key` in GCP Secret Manager is still the placeholder value. Follow [docs/deploy.md Step 4](https://github.com/${{ github.repository }}/blob/${{ github.event.pull_request.head.sha }}/docs/deploy.md#step-4--sign-up-for-clerk-and-create-two-apps) to configure the staging Clerk secret key, then re-run CI.';
             } else if (!cloudRunApiDeployed) {
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -599,17 +599,20 @@ jobs:
     steps:
       - name: Verify deploy prerequisites
         run: |
-          # deploy-api.result is the source of truth for "did the job run at all".
-          # When build-images fails, deploy-api is skipped (per its `if:` guard) and
-          # its outputs come through as empty strings, not 'false' — so a plain
-          # `clerk_configured != 'true'` check would mis-report the skip as a Clerk
-          # misconfiguration. See #398 / PR #395 for the motivating incident.
+          # deploy-api.result is the source of truth for "did the job deliver its outputs".
+          # The integration-tests job runs under `if: always()` (see #345), so deploy-api
+          # can reach this step in any of three non-success states:
+          #   - skipped/cancelled: upstream build-images failed (per deploy-api's `if:` guard).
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [fix] Disambiguate staging verify-prereqs guard (skipped vs misconfigured)

### Summary

Needs revision before merge. The fix correctly handles `skipped` and `cancelled`, but `deploy-api.result === 'failure'` — which is reachable here because `staging-integration-tests` runs under `if: always() && needs.preflight.outputs.should_run == 'true'` (see [#345](https://github.com/brownm09/lifting-logbook/issues/345)) — falls through to the same `clerk_configured != 'true'` check the PR set out to disambiguate, reintroducing the exact phantom-Clerk message for the hard-fail case. PR scope is appropriate (one YAML file, 21/2), so the fix is a small extension rather than a re-split.

---

### Blocking Findings

**[correctness]** `.github/workflows/staging.yml:610-617` and `.github/workflows/staging.yml:741-744`

The new guard branches on `DEPLOY_RESULT == "skipped" || "cancelled"` and otherwise falls through to `"$CLERK_OUT" != "true"`. `needs.deploy-api.result` has four values: `success | failure | skipped | cancelled`. When `deploy-api` hard-fails *before* its `Check Clerk secret key configured` step runs (e.g., the GCP auth step fails, or the job is canceled mid-run, or a step earlier than the Clerk check errors), the job sets `result = 'failure'` with `clerk_configured` empty. Empty `!= "true"` → the guard fires the "Staging Clerk secret key is not configured" error, and the bot posts the Clerk-config warning. That is the exact PR #395 incident shape, just reached through a different upstream state.

The surrounding job's own block comment makes the `failure` case explicit: *"always() ensures the job runs even when a needs job hard-fails (e.g. GCP auth failure in deploy-api/deploy-web)"* — so this is not a hypothetical state, it is the documented contract.

**Fix:** Treat any non-`success` `DEPLOY_RESULT` as "the job didn't deliver its outputs," then disambiguate inside that branch. One shape:

```bash
if [[ "$DEPLOY_RESULT" != "success" ]]; then
  echo "::error::deploy-api did not succeed (result: $DEPLOY_RESULT). If 'skipped' or 'cancelled', build-images likely failed — check the 'Build & Push Images' job (e.g., Artifact Registry 504). If 'failure', open the deploy-api job log for the step that errored."
  exit 1
fi
```

Apply the same split to the bot template: replace `else if (deployApiResult === 'skipped' || deployApiResult === 'cancelled')` with `else if (deployApiResult !== 'success')` and adjust the message to cover failure too. Without this, the acceptance-criteria claim *"distinguishes empty (upstream skipped) from `'false'` (real misconfig)"* is only true for two of the three non-success states.

---

### Non-Blocking Findings

**[maintainability]** PR body — Test summary line

The Test integrity policy (ADR-029, Rule 2) requires the line `Tests: N passed, N skipped, N failed (duration)` verbatim. The PR body has `Tests: 23 passed, 23 total`, which is the Jest per-suite line, not the policy's summary line, and elides the skipped/failed counts. The body does flag the `apps/web` compile failures and cites [#421](https://github.com/brownm09/lifting-logbook/issues/421), so the intent is clear — only the line shape is non-conforming.

**Suggestion:** Replace the per-suite line with the rolled-up summary (`Tests: 23 passed, 0 skipped, 0 failed (Xs) — 6 of 8 apps/web suites failed to compile pre-existingly; see #421`).

**[maintainability]** `.github/workflows/staging.yml:741` — bot template ignores `deploy-web.result`

The bot template now consults `deployApiResult` but still treats `deploy-web` as implicit. If `deploy-web` is skipped because of a `build-images` failure but `deploy-api` somehow proceeded (rare given they share an upstream, but possible if `deploy-web` gains a stricter `if:` later), the success branch posts `[Staging environment](${webUrl})` with an empty `webUrl`. Not introduced by this PR, but the same disambiguation pattern would apply.

**Suggestion:** Out of scope for this PR; track in a follow-up alongside any future change to `deploy-web`'s `if:` guard.

---

### Style Notes

YAML and inline shell are consistent with the surrounding workflow. Inline `# deploy-api.result is the source of truth…` comment block correctly cites #398/PR #395 — good provenance for the next reader.

---

PR #423, opened 2026-06-02
[fix] Surface actionable error when Docker is unreachable for api E2E

### LINKED ISSUE #394: [infra] Local Docker Desktop WSL distro missing — api E2E suite cannot run locally

brownm09:
## Summary

When `apps/api` Jest `globalSetup` cannot reach Docker, the Testcontainers stack trace was the only signal — opaque unless you'd read the file. This PR wraps `container.start()` and `prisma migrate deploy` in try/catch and re-throws a multi-line message that names the root cause, links #394, and lists three recovery options.

- **Default behavior remains hard-fail** so DB-touching changes cannot ship without local verification.
- **`LIFTING_SKIP_DB_E2E=1` escape hatch** is documented as opt-in only: it skips DB provisioning entirely and lets the spec's existing `describeOrSkip` leave its blocks pending. Valid only when the diff under test touches no DB code.
- **CI passthrough branch** (`DATABASE_URL` pre-set) is untouched.
- Updated `CLAUDE.md` and `docs/testing/e2e-coverage.md` to document the new error contract and escape-hatch contributor rule.

Closes #394

## Testing

`npm test -w @lifting-logbook/api` (full api workspace, Docker up):

```
Test Suites: 22 passed, 22 total
Tests:       301 passed, 301 total
Snapshots:   0 total
Time:        11.506 s
```

`LIFTING_SKIP_DB_E2E=1 npx jest --testPathPatterns=programs.db.e2e` (escape-hatch path):

```
[jest.global-setup] LIFTING_SKIP_DB_E2E=1 — skipping DB E2E provisioning. DB-touching changes will not be locally verified. See issue #394.
Test Suites: 1 skipped, 0 of 1 total
Tests:       51 skipped, 51 total
```

The hard-fail message path was exercised by inspection — the original `Testcontainers/.start()` failure is now caught and re-thrown via `buildDockerUnavailableMessage(err)`. Cannot be exercised live in this PR because Docker came back up on this machine between the issue being filed and now (which is the third state covered by the test summary above).

### Coverage gate

No new product behavior — this is test-infrastructure ergonomics. Verified manually across all three paths (Docker up, opt-out, and by-inspection for hard-fail). No new automated tests required.

### Suppression / test-integrity

No new suppression markers added. No new `.skip`/`xit`/`describe.skip` markers introduced — the existing `describeOrSkip` in `programs.db.e2e.spec.ts` is untouched. The skipped count in the opt-out path is environment-gated by a documented contributor rule (escape hatch only valid for non-DB diffs).

## Test plan

- [x] `npm test -w @lifting-logbook/api` — 301 passed / 0 failed / 0 skipped with Docker up
- [x] `LIFTING_SKIP_DB_E2E=1 npx jest --testPathPatterns=programs.db.e2e` — 51 skipped, warning printed
- [ ] CI: api workspace passes against the GitHub Actions Postgres service container

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -327,7 +327,8 @@ npm test
 
 **Prerequisites:**
 - Run `npm run build` first if you have touched compiled output (e.g., API controllers, shared types in `packages/types`).
-- The API DB E2E suite (`apps/api/src/programs/programs.db.e2e.spec.ts`) auto-provisions Postgres via Testcontainers in `apps/api/jest.global-setup.js`. Docker Desktop must be running; no `DATABASE_URL` configuration is required locally. In CI, the existing service container provides `DATABASE_URL` and globalSetup uses it directly. See [`docs/testing/e2e-coverage.md`](docs/testing/e2e-coverage.md) for details.
+- The API DB E2E suite (`apps/api/src/programs/programs.db.e2e.spec.ts`) auto-provisions Postgres via Testcontainers in `apps/api/jest.global-setup.js`. Docker Desktop must be running; no `DATABASE_URL` configuration is required locally. In CI, the existing service container provides `DATABASE_URL` and globalSetup uses it directly.
+- **When Docker is down:** globalSetup hard-fails with a multi-line actionable message naming three recovery options (fix Docker, use `docker-compose.test.yml`, or set `LIFTING_SKIP_DB_E2E=1`). The escape hatch is only valid when the diff under test touches no DB code (no changes under `apps/api/prisma/`, no repository changes); cite [issue #394](https://github.com/brownm09/lifting-logbook/issues/394) in the PR body when it is used. See [`docs/testing/e2e-coverage.md`](docs/testing/e2e-coverage.md) for the full recovery procedure.
 
 Individual workspaces can be verified independently:
 
--- a/apps/api/jest.global-setup.js
+++ b/apps/api/jest.global-setup.js
@@ -1,18 +1,25 @@
 // Jest globalSetup — provisions a Postgres instance for the DB E2E suite.
 //
-// Two paths:
+// Three paths:
 //   1. CI (or any env that already exports DATABASE_URL): passthrough — re-expose
 //      the existing URL via LIFTING_TC_DATABASE_URL so the spec can pick it up
 //      without bypassing the env-isolation Proxy in jest.env.setup.js.
-//   2. Local: start an ephemeral postgres:16-alpine container via Testcontainers
+//   2. Explicit local opt-out (LIFTING_SKIP_DB_E2E=1): skip provisioning entirely;
+//      the DB E2E spec's describeOrSkip will leave its blocks pending. Contract:
+//      this escape hatch is only valid when the diff under test touches no DB
+//      code (prisma schema, migrations, repositories). See issue #394.
+//   3. Local: start an ephemeral postgres:16-alpine container via Testcontainers
 //      (matches CI image), run `prisma migrate deploy` against it, and stash the
-//      container handle on globalThis for teardown.
+//      container handle on globalThis for teardown. On any failure (Docker
+//      unreachable, image pull error, migration error) we throw an actionable
+//      multi-line error so the api workspace test run aborts with a clear cause
+//      — never a raw Testcontainers stack trace.
 //
-// In both cases the spec reads LIFTING_TC_DATABASE_URL (not DATABASE_URL) because
-// jest.env.setup.js force-blanks DATABASE_URL in every worker to keep the
-// in-memory e2e suite wiring InMemoryRepositoryFactory. The DB E2E spec restores
-// DATABASE_URL from the sentinel inside its own beforeAll; jest.env.setup.js
-// allows that one specific write through its Proxy.
+// In paths 1 and 3 the spec reads LIFTING_TC_DATABASE_URL (not DATABASE_URL)
+// because jest.env.setup.js force-blanks DATABASE_URL in every worker to keep
+// the in-memory e2e suite wiring InMemoryRepositoryFactory. The DB E2E spec
+// restores DATABASE_URL from the sentinel inside its own beforeAll;
+// jest.env.setup.js allows that one specific write through its Proxy.
 
 const path = require('path');
 const { execSync } = require('child_process');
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: [fix] Surface actionable error when Docker is unreachable for api E2E

### Summary
Merge with one minor revision. The change cleanly converts a raw Testcontainers stack into a self-explanatory failure with a documented escape hatch, and the docs/CLAUDE.md updates are well-reconciled. The one substantive issue is that the actionable error message is reused for failures that are *not* Docker-unreachability (the `prisma migrate deploy` catch path) — exactly the kind of composite-state-collapsed-into-one-label anti-pattern the global Error Message Diligence rule warns against. Worth a one-line fix before merge.

---

### Blocking Findings

*(none — the message-reuse issue below is non-blocking; fixable in a follow-up if you'd rather merge as-is.)*

---

### Non-Blocking Findings

**[correctness]** `apps/api/jest.global-setup.js:103`
The `prisma migrate deploy` catch re-throws via `buildDockerUnavailableMessage(err)`, which produces the headline "Cannot start Postgres via Testcontainers — Docker appears unreachable." But by the time this catch runs, the container has already started successfully (line 77) and the handle was stashed on `globalThis`. A failure here is almost certainly *not* Docker being unreachable — it's a migration SQL error, a Prisma schema mismatch, an image/schema version drift, or a timeout. Reporting "Docker appears unreachable" as the headline misdirects the contributor to fix the wrong thing (reset WSL, switch to docker-compose) when the real problem is in the migration. The underlying error appears at the bottom of the message, but the headline is what gets read first and quoted in chat. This is the exact pattern called out in global `CLAUDE.md` → Error Message Diligence (the staging.yml incident).
**Suggestion:** Split into two builders — `buildDockerUnavailableMessage` for the `container.start()` catch and a separate `buildMigrationFailedMessage` for the migration catch. The migration message should headline "prisma migrate deploy failed against the started Testcontainers Postgres" and list different recovery options (inspect the migration, run with `DEBUG=prisma:*`, check schema drift). Both can share a footer that prints the underlying error.

**[maintainability]** `apps/api/jest.global-setup.js:58`
The opt-out check is `process.env.LIFTING_SKIP_DB_E2E === '1'`. A contributor who sets `LIFTING_SKIP_DB_E2E=true`, `=yes`, or just `LIFTING_SKIP_DB_E2E=` (empty) will get the hard-fail with no warning that their env var was present-but-not-recognized — easy to misread as a code bug.
**Suggestion:** Either broaden to a truthy check (`['1','true','yes'].includes(...)`) or, if you want to keep the contract strict, log a one-line warning when the var is set to anything other than `'1'` so the contributor sees their value was ignored.

**[maintainability]** `apps/api/jest.global-setup.js:33-46`
The "three recovery options" are duplicated verbatim in three places: this builder, `docs/testing/e2e-coverage.md` ("Running locally when Docker is unavailable"), and the new `CLAUDE.md` Testing bullet. The escape-hatch path detail (e.g., the "no changes under `apps/api/prisma/`, no repository changes" rule) lives only in the docs; if it drifts, the inline error and the docs diverge silently. Not worth refactoring now, but worth a note.
**Suggestion:** Either trim the inline message to a one-line pointer (`see docs/testing/e2e-coverage.md § Running locally when Docker is unavailable`) plus the underlying error, or accept the duplication and add a `// keep in sync with docs/testing/e2e-coverage.md` comment above `buildDockerUnavailableMessage` so future edits surface the coupling.

---

### Questions for Author

**Question:** Is leaving `globalThis.__LL_PG_CONTAINER__` set on the migration-failure path intentional, or should it be cleared like the local `container` variable?
**Context:** Lines 101-102 do `await container.stop().catch(() => {})` and `globalThis.__LL_PG_CONTAINER__ = undefined;` — good. The intent is clearly to leave teardown a no-op since the container is already stopped. That looks correct. The question is only whether the subsequent re-throw (now wrapped via `buildDockerUnavailableMessage`) plays nicely with Jest's globalTeardown contract: Jest still calls globalTeardown after a thrown globalSetup, and `jest.global-teardown.js:6` already guards on the handle being undefined, so this should be fine. Flagging only because I want to confirm you've exercised the throw-after-migration-failure path mentally — none of the verification states in the PR body actually walked through it.
**Tradeoffs:** (a) Leave as-is — relies on the existing teardown guard, which is correct today but couples this file's correctness to the teardown's no-op behavior. (b) Add an explicit `try/finally` around the migrate block — more defensive but adds nesting for a path that already works.

---

### Style Notes

The new comment block at the top of `jest.global-setup.js` is well-written — the explicit numbering of the three paths and the contract note for the escape hatch will save the next reader a lot of time. Migration message split (above) would extend that same clarity to the failure paths. No other style notes.

---

brownm09:
Addressed all three findings in commit `1c3f0c0`.

**[correctness] Misleading headline on migration failures (the substantive one)** — fixed by splitting into a separate `buildMigrationFailedMessage`. The migration-catch branch now headlines `` `prisma migrate deploy` failed against the started Testcontainers Postgres `` and explicitly states "Docker IS reachable — the container started cleanly. The failure is in the migration step itself." then lists migration-specific causes (schema drift, SQL error, version mismatch, timeout). Updated the top-of-file comment block to describe the two-distinct-messages contract and cross-link Error Message Diligence.

**[maintainability] Strict `=== '1'` opt-out** — broadened to a `SKIP_DB_E2E_TRUTHY` set of `{1, true, yes}` (case-insensitive via `toLowerCase`). Anything set-but-not-in-the-set now emits a near-miss warning ("LIFTING_SKIP_DB_E2E=\"on\" is not a recognized truthy value (accepted: 1, true, yes). Treating as unset — falling through to Testcontainers provisioning.") before falling through, so a typoed value can't silently skip-fail.

**[maintainability] Doc-coupling for recovery copy** — added a `keep in sync with docs/testing/e2e-coverage.md (...) and CLAUDE.md` comment above `buildDockerUnavailableMessage`.

**Teardown question (self-answered without your intervention)** — wrote `apps/api/scripts/check-globalsetup-failure-paths.js`, a standalone Node script that monkey-patches `@testcontainers/postgresql` and `child_process.execSync` to exercise all three failure paths (start throws, migrate throws, typoed skip var) without real Docker. 17/17 checks pass. Confirms on the migration-failure path: `container.stop()` called exactly once, `globalThis.__LL_PG_CONTAINER__` cleared to `undefined`, `globalTeardown` no-ops without throwing and without a second stop call. Repeatable via `node apps/api/scripts/check-globalsetup-failure-paths.js` — committed so future regressions in the failure-path contract are catchable without real Docker.

**Verification.**
- `npm test -w @lifting-logbook/api`: 22 suites, 301 tests, 15.058s — green.
- `node apps/api/scripts/check-globalsetup-failure-paths.js`: 17/17 checks pass.

Re: the "Staging Integration Tests" failure on this push — that's the workflow PR #422 is fixing; the failure mode is unrelated to anything in this diff (no changes here touch `staging.yml`, the build pipeline, or any deploy step). Not chasing per Error Message Diligence.

PR #424, opened 2026-06-02
[fix] Resolve Node 24 local test failures (OOM + NextResponse type drift)

### LINKED ISSUE #419: [fix] Local test failures on Node 24 — core Jest worker OOM and web route.test TS2576 NextResponse.json
## Symptom

Running `npm test` locally on Node 24 (Windows) against `origin/main` (commit `73f89a1`) produces the following pre-existing failures, unrelated to any code under test:

### packages/core — 6 suites OOM

```
FAIL tests/core/utils/parser/parseCycleDashboard.test.ts
FAIL tests/core/utils/jsUtil.test.ts
FAIL tests/core/services/workout/createGridV2.test.ts
FAIL tests/core/services/workout/updateLiftDates.test.ts
FAIL tests/core/services/bodyWeight.test.ts
FAIL tests/core/utils/import/validateLiftImport.test.ts
```

Each fails with:
```
Test suite failed to run
  Jest worker ran out of memory and crashed
    at ChildProcessWorker._onExit (../../node_modules/jest-worker/build/index.js:928:26)
```

Recurrence of the failure class previously fixed in #89 — appears to have regressed (or returned in a different form) under Node 24.

### apps/web — TS type errors in route handler tests

```
app/api/healthz/route.test.ts:65,66 — error TS2576:
  Property 'json' does not exist on type 'NextResponse<{ ok: boolean; }>'.
  Did you mean to access the static member 'NextResponse<{ ok: boolean; }>.json' instead?
```
Same class of error also surfaces in `app/livez/route.test.ts` and `lib/__tests__/programPlan.test.ts`.

This appears to be a `@types/next` / `next` type definition issue — `NextResponse` instances now disallow the instance `.json()` access these tests use.

## Suspected root cause

Local Node version is 24.13.0; CI runs Node 20 (per `.nvmrc`). The Node 24 caveat documented in [CLAUDE.md](https://github.com/brownm09/lifting-logbook/blob/main/CLAUDE.md) warns about install-time corruption but does not cover Jest worker memory pressure or `next` type drift under Node 24. CI is currently green for both packages, so this is a local-only failure mode at the moment.

## Why this matters

Blocks the pre-PR `npm test` gate for any local change. Currently being unblocked per CLAUDE.md failure-tracking rule by citing this issue in PR bodies.

## Acceptance criteria

- [ ] Identify whether Node 24 worker heap exhaustion is the root cause for the `packages/core` OOMs; if so, document the workaround (e.g., `--maxWorkers=2` or `NODE_OPTIONS=--max-old-space-size=4096`) or pin Jest config to handle it
- [ ] Investigate the `NextResponse.json` instance-vs-static-member type drift in the route handler tests; either update the test patterns or pin the relevant `@types/next` / `next` versions
- [ ] Confirm `npm test` passes locally on Node 24

## Workaround

Test suite passes in CI (Node 20). Local runs on Node 20 (`nvm use 20.11.1`) are expected to pass.

brownm09:
## Summary

Two changes that clear the pre-existing local failures documented in #419 without touching CI behaviour.

- **`jest.config.base.js`** — adds `workerIdleMemoryLimit: '512MB'` and `maxWorkers: '50%'`. Idle workers above 512MB RSS are restarted (reclaims heap between fixture-heavy suites); cap at 50% of CPU drops concurrent CSV-fixture loading on multi-core dev boxes enough to clear the 6 OOMs. Node 20 CI runners are already constrained, so the effective behaviour there doesn't change.
- **`apps/web/app/api/healthz/route.test.ts`** — three `await res.json()` sites now go through `(res as Response).json()`. `NextResponse` extends `Response`, so the cast is type-safe and runtime-equivalent; the Next 16.2.4 type declares `json` as a static factory which shadows the inherited instance method in `NextResponse<T>`, producing TS2576. Three lines, no production code, no `next` version pinning, no eslint suppression.

The issue body also lists `apps/web/app/livez/route.test.ts` and `apps/web/lib/__tests__/programPlan.test.ts` as affected, but neither uses `res.json()` on a `NextResponse` (`livez` uses `res.text()`; `programPlan` has no `NextResponse` usage at all). Only the three `healthz` call sites actually emit TS2576.

## Testing

Run on Node 24.13.0 locally after a clean `npm ci`:

- `npm test -w @lifting-logbook/web -- app/api/healthz/route.test.ts`
  - `Tests: 3 passed, 0 skipped, 0 failed (5.268s)`
- `npm test -w @lifting-logbook/core`
  - `Tests: 160 passed, 0 skipped, 0 failed (8.014s)` — all 26 suites green, including the 6 previously OOMing ones (`parseCycleDashboard`, `jsUtil`, `createGridV2`, `updateLiftDates`, `bodyWeight`, `validateLiftImport`)
- `npm test` (full monorepo via turbo)
  - 12/12 tasks successful (49.7s)
  - api: `Tests: 301 passed, 0 skipped, 0 failed (11.063s)`
  - web: `Tests: 56 passed, 0 skipped, 0 failed (8.495s)`
  - core: `Tests: 160 passed, 0 skipped, 0 failed (8.014s)`
- **Combined: `Tests: 517 passed, 0 skipped, 0 failed`** across the three workspaces with test suites.

Coverage-gate check: no new behaviour added — this is a Jest-config and test-cast change. Existing tests fully exercise the affected paths (healthz had three cases before and still does).

Suppression check (`git diff origin/main | grep -E '(ts-ignore|ts-expect-error|eslint-disable|!\.|!\[|!\s*[;,)]|\?\? null)'`): empty.

Test-integrity check (skip/xdescribe/passWithNoTests/etc. grep): empty. The `--passWithNoTests` flag in npm output is a pre-existing workspace `test` script setting, unchanged here.

## Side observation — install corruption recurrence

Reaching the green state above required two rounds of `rm -rf node_modules && npm ci` because the initial install on Node 24 left several packages with truncated extractions (`std-env/dist/index.mjs` missing; `jest-config/build/` containing only `index.js`, which made Jest's preset loader fail with `Preset ts-jest not found relative to rootDir`). This is the same failure class already documented in `CLAUDE.md`'s "Node 24 caveat" section. Worth tracking separately if it recurs — not in scope for #419.

Closes #419

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -16,6 +16,7 @@ Include in your opening brief only: the issue you are working on, current branch
     - Malformed native binaries: `node_modules/@turbo/windows-64/bin/turbo.exe` failing with `EFTYPE`
 
     Downstream failures: `nest build` CJS resolution errors (`iconv-lite/lib/streams`, `minimatch/dist/commonjs/index.js`), `TS1110 Type expected` from `light-my-request`, or `spawnSync ... EFTYPE` from turbo. Fix: `rm -rf node_modules/<package> && npm install <package> --no-save` to re-extract a single package, or `rm -rf node_modules && npm ci` for a full reset. CI runs Node 20 and is unaffected. Original investigation: [#373](https://github.com/brownm09/lifting-logbook/issues/373).
+  - **Node 24 Jest worker OOM (Windows only):** Several `packages/core` CSV-fixture-heavy suites exhaust per-worker heap when run in parallel on Node 24. Codified workaround: `jest.config.base.js` applies `workerIdleMemoryLimit: '512MB'` + `maxWorkers: '50%'` when `process.platform === 'win32'`. Linux Node 20 CI is unaffected by both the failure and the setting. Investigation: [#419](https://github.com/brownm09/lifting-logbook/issues/419).
 - **Package manager:** npm (workspaces)
 - **`jq` is NOT available.** Use `node -e` with a temp file in the working directory for JSON parsing:
   ```bash
--- a/apps/web/app/api/healthz/route.test.ts
+++ b/apps/web/app/api/healthz/route.test.ts
@@ -30,7 +30,7 @@ describe('GET /api/healthz', () => {
     const res = await GET();
 
     expect(res.status).toBe(200);
-    expect(await res.json()).toEqual({ ok: true });
+    expect(await (res as Response).json()).toEqual({ ok: true });
     expect(mockedAuth).toHaveBeenCalledTimes(1);
   });
 
@@ -48,7 +48,7 @@ describe('GET /api/healthz', () => {
     const res = await GET();
 
     expect(res.status).toBe(503);
-    expect(await res.json()).toEqual({ ok: false });
+    expect(await (res as Response).json()).toEqual({ ok: false });
     // The original error is logged server-side for operators.
     expect(consoleErrorSpy).toHaveBeenCalledWith(
       expect.stringContaining('[healthz]'),
@@ -63,7 +63,7 @@ describe('GET /api/healthz', () => {
     const res = await GET();
 
     expect(res.status).toBe(200);
-    expect(await res.json()).toEqual({ ok: true, mode: 'dev-auth' });
+    expect(await (res as Response).json()).toEqual({ ok: true, mode: 'dev-auth' });
     expect(mockedAuth).not.toHaveBeenCalled();
   });
 });
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -7,10 +7,13 @@ module.exports = {
     '^.+\\.tsx?$': ['ts-jest', { tsconfig: '<rootDir>/tsconfig.json' }],
   },
   transformIgnorePatterns: ['/node_modules/'],
-  // Local Node 24 (Windows) workers exhaust heap when several CSV-fixture-heavy
+  // Node 24 (Windows) workers exhaust heap when several CSV-fixture-heavy
   // packages/core suites run in parallel. Restarting idle workers above 512MB
-  // RSS and capping parallelism at 50% of CPU keeps Node 20 CI behaviour
-  // effectively unchanged while clearing the OOMs locally. See #419.
-  workerIdleMemoryLimit: '512MB',
-  maxWorkers: '50%',
+  // RSS and capping parallelism at 50% of CPU clears the OOMs locally. Scoped
+  // to win32 so Node 20 Linux CI keeps Jest's default parallelism (cpus - 1)
+  // and never pays the worker-restart cost. See #419 and CLAUDE.md's Node 24
+  // caveat for the codified workaround.
+  ...(process.platform === 'win32'
+    ? { workerIdleMemoryLimit: '512MB', maxWorkers: '50%' }
+    : {}),
 };
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -7,4 +7,10 @@ module.exports = {
     '^.+\\.tsx?$': ['ts-jest', { tsconfig: '<rootDir>/tsconfig.json' }],
   },
   transformIgnorePatterns: ['/node_modules/'],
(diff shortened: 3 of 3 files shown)


brownm09:
## Review: [fix] Resolve Node 24 local test failures (OOM + NextResponse type drift)

### Summary
Ready to merge with one item worth discussing before doing so. The diff is 12 lines across two files, the four pre-PR gates (suppression, integrity, coverage, summary line) are all clean, and 517/0/0 verifies end-to-end on Node 24. The highest-severity item is a scope question on `maxWorkers: '50%'` — it ships to CI too, which the PR body says is fine but does not show measurements for. Everything else is non-blocking or stylistic.

---

### Non-Blocking Findings

**[performance]** `jest.config.base.js:15`
`maxWorkers: '50%'` lands in the shared base config, so it applies to GitHub Actions runners as well as local. The PR body argues "Node 20 CI runners are already constrained, so the effective behaviour there doesn't change," but Jest's default is `cpus - 1`. On the current `ubuntu-latest` (4 vCPU), 50% = 2 workers vs. the default 3 — a real ~33% drop in parallelism for every test job, every PR. The OOM problem only manifests on Node 24 + Windows + the CSV-fixture core suites; tying the cap to that environment would keep CI throughput unchanged.
**Suggestion:** Either (a) gate `maxWorkers` on `process.platform === 'win32'` or an env flag set only locally, or (b) measure CI wall-clock before/after on a representative PR and add the numbers to the PR body so the "effectively unchanged" claim is grounded. Option (a) is the lower-risk default.

**[maintainability]** `jest.config.base.js:10-13`
The justification comment names Node 24, Windows, CSV fixtures, and `packages/core` — but the setting itself is applied globally with no link from `apps/web` or `apps/api` configs back to this rationale. Six months from now, someone debugging slow web tests will find a `maxWorkers: '50%'` in the base and have to read #419 to understand it.
**Suggestion:** Add the issue number into the inline comment ("See #419") — already done, good. Consider also noting in `CLAUDE.md`'s Node 24 caveat row that this is the codified workaround so the link is bidirectional. The stub explicitly deferred this; flagging here so it isn't forgotten.

**[maintainability]** `apps/web/app/api/healthz/route.test.ts:33,51,66`
Three identical `(res as Response).json()` casts. A single `const res: Response = await GET()` at each call site (or a small `readJson(res)` helper) would centralize the workaround. Not urgent — 3 lines is below the threshold where duplication actually hurts — but if Next 16.x keeps the static-shadowing type for other handlers (`/livez`, future routes), the pattern will spread.
**Suggestion:** Leave as-is for this PR. If the same cast appears in a fourth place, fold into a helper at that point.

---

### Questions for Author

**Question:** Is the `workerIdleMemoryLimit: '512MB'` also intended to apply to CI, or only locally?
**Context:** The inline comment at `jest.config.base.js:10` says "Local Node 24 (Windows) workers exhaust heap..." but the setting is placed in the unconditional export, so Linux Node 20 CI workers will also recycle at 512MB RSS. On a healthy run that may never trigger; on slower runs (cold caches, larger fixtures landing later) it could add restart overhead.
**Tradeoffs:** (a) Leave global — simplest, the threshold is well above typical idle RSS and the cost is bounded. (b) Gate on `process.platform === 'win32'` next to the `maxWorkers` gate above — keeps the workaround scoped to the platform it was diagnosed on. (c) Raise the limit (e.g., 1GB) — keeps it global but with more headroom.

---

### Style Notes

The two-line block comment above the new settings is well-pitched for someone landing on this file cold — names the OS, Node version, suite class, and links the issue. Match that pattern in future config workarounds.

---

brownm09:
Stale run — `26847205613` was for SHA `2e92063` and was **cancelled** when commit `8001734` (review follow-up) was pushed. The cancellation killed `Deploy web (staging)`, leaving `WEB_URL=""` empty in the integration job, so `Wait for staging to be ready` looped 18× against an empty URL and emitted "Staging did not become ready in time" — that message describes the wait-loop outcome, not the actual cause. The new Staging run on `8001734` is still in progress; no code action required on this report. (Per ADR-034 — error-message diligence on automation guards.)

Issue #432, opened 2026-06-03
CI red on main: package-lock.json out of sync with package.json (npm ci fails)


brownm09:
## Problem

Every CI job on `main` fails at the `npm ci` step — both `Lint & Test` and `DB Integration Tests` die before any test runs (and `Playwright E2E` is consequently skipped). main is currently red ([latest run](https://github.com/brownm09/lifting-logbook/actions/runs/26895549682)).

```
npm ERR! code EUSAGE
npm ERR! `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR! Invalid: lock file's @clerk/backend@3.4.14 does not satisfy @clerk/backend@3.5.0
npm ERR! Invalid: lock file's @clerk/shared@4.14.0 does not satisfy @clerk/shared@4.15.0
npm ERR! Missing: react@19.2.7 from lock file
```

## Root cause

`package.json` declares caret ranges (e.g. `@clerk/backend: ^3.4.13`, `@clerk/nextjs: ^6.12.0`) that now resolve to newer published versions (clerk-backend 3.5.0, clerk-shared 4.15.0, with react 19.2.7 pulled in transitively) than the versions pinned in the committed `package-lock.json`. `npm ci` requires an exact, in-sync lockfile and refuses to install. The lockfile was not regenerated after the relevant dependency range bump.

## Fix

Regenerate the lockfile (`npm install`) so it is back in sync, and commit it. After the fix `npm ci` succeeds locally.

## Verification

`npm ci` → succeeds. Auth-touching suites still green after the resulting clerk 3.4→3.5 / 4.14→4.15 minor bump: api in-memory 250 passed, web 56 passed, core 160 passed; DB E2E 51 passed via Testcontainers.

PR #436, opened 2026-06-04
[infra] Add CI lockfile-sync guard with clear drift message

### LINKED ISSUE #435: [infra] CI guard: fail fast with a clear message when package-lock.json drifts from package.json
## Problem

`package-lock.json` drifted out of sync with the caret ranges in `package.json` (clerk-backend, clerk-shared, react), which turned `main` CI red at the `npm ci` step. `npm ci` refuses an out-of-sync lockfile, but the resulting `EUSAGE` error is cryptic and does not name the lockfile-drift root cause. It was fixed reactively in #432/#433, but nothing surfaces the drift early or explains it.

## Goal

Add an early step to `.github/workflows/ci.yml` (in the `Lint & Test` job, before `npm ci`) that detects lockfile drift and fails with a clear, actionable message instead of the cryptic `npm ci` EUSAGE:

```bash
npm install --package-lock-only --ignore-scripts
git diff --exit-code package-lock.json
```

## Acceptance Criteria

- [ ] Early CI step regenerates the lockfile metadata and `git diff --exit-code`s it
- [ ] On drift, the step prints a clear remediation message (run `npm install`, commit the regenerated lockfile)
- [ ] Verified: passes on the current in-sync lockfile
- [ ] Verified: fails on a deliberately drifted lockfile
- [ ] This is layer 2 of a 4-layer lockfile-drift defense (branch protection, CI guard, global pre-push hook, global CLAUDE.md rule)

brownm09:
## Summary

Layer 2 of the 4-layer lockfile-drift defense. Adds an early step to the `Lint & Test` job in `ci.yml` that regenerates the lockfile metadata and fails with an actionable message if `package-lock.json` has drifted from `package.json` — surfacing the root cause before the cryptic `npm ci` EUSAGE error that turned `main` red (fixed reactively in #432/#433).

```yaml
npm install --package-lock-only --ignore-scripts
git diff --exit-code package-lock.json   # clear ::error:: + remediation on drift
```

## Acceptance Criteria

- [x] Early CI step regenerates lockfile metadata and `git diff --exit-code`s it
- [x] On drift, prints a clear remediation message (run `npm install`, commit the regenerated lockfile)
- [x] Verified: passes on the current in-sync lockfile
- [x] Verified: fails on a deliberately drifted lockfile (`@types/jest` range change)

## Testing

This is a **CI-workflow-only change** (single addition to `.github/workflows/ci.yml`); it touches no application, package, or test code, so the monorepo test suite cannot validate it. Verification was done by directly exercising the guard's exact commands:

- **Synced lockfile:** `npm install --package-lock-only --ignore-scripts` → `git diff --exit-code package-lock.json` exits 0 (PASS, no drift).
- **Deliberately drifted lockfile:** changed `@types/jest` range `^30.0.0` → `^29.5.0`, re-ran the guard → diff non-empty, would `exit 1` (guard correctly FAILS). Working tree restored clean afterward.
- Pre-commit lint hook: `Tasks: 5 successful, 5 total`.

Full `npm test` not run: the change is CI-config only and the suite (which requires Docker/Playwright) yields no signal on workflow YAML. No suppressions or test-integrity violations introduced (pre-PR greps clean).

Closes #435

_Test-integrity note: no suite run — this change is CI-workflow-only and the test suite cannot validate workflow YAML; verification is the synced/drifted simulation above (Tests: not applicable)._

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,20 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-turbo-
 
+      - name: Verify package-lock.json is in sync with package.json
+        run: |
+          # Layer 2 of the lockfile-drift defense (see issue #435). Regenerate the
+          # lockfile metadata from package.json without touching node_modules, then
+          # fail with a clear message if it differs from the committed lockfile.
+          # This surfaces drift before the cryptic `npm ci` EUSAGE error fires below.
+          npm install --package-lock-only --ignore-scripts
+          if ! git diff --exit-code package-lock.json; then
+            echo "::error::package-lock.json is out of sync with package.json."
+            echo "A dependency was changed without regenerating the lockfile (the diff above shows the delta)."
+            echo "Fix: run 'npm install' locally and commit the updated package-lock.json in the same change."
+            exit 1
+          fi
+
       - name: Install dependencies
         run: npm ci
 


brownm09:
## Review: [infra] Add CI lockfile-sync guard with clear drift message

### Summary
**Merge.** This is a tight, correct, single-file CI change that does exactly what it claims: it regenerates lockfile metadata and fails early with an actionable message before the opaque `npm ci` EUSAGE error. The guard's failure direction is sound — if `npm install --package-lock-only` produces a diff, the committed lockfile genuinely does not satisfy `package.json`, which is precisely the state `npm ci` rejects, so there is no false-positive path relative to the error it front-runs. No blocking findings. A few minor non-blocking notes below; none should hold the merge.

---

### Non-Blocking Findings

**[performance]** `.github/workflows/ci.yml:37`
The guard runs on **every** CI run, including PRs that touch no dependencies, adding an `npm install --package-lock-only` (a dependency-resolution pass, potentially a registry round-trip) to each run.
This is the right tradeoff — always-correct beats conditional — and `setup-node`'s `cache: 'npm'` (already configured above) keeps it cheap. Noting it only so the cost is a conscious choice.
**Suggestion:** Leave as-is. If CI minutes ever become a concern, gate the step on a `package*.json` path filter, but that adds drift risk (a step that doesn't run can't catch anything) for marginal savings — not worth it now.

**[reliability]** `.github/workflows/ci.yml:41`
`npm install --package-lock-only` can reach the registry to resolve versions; a transient registry 5xx would fail this step.
The blast radius is identical to the existing `npm ci` step that follows, so this adds no new failure surface — but it does move the first network-dependent step earlier in the job. Acceptable.
**Suggestion:** No change needed. If registry flakiness shows up in practice, the same retry strategy you'd apply to `npm ci` covers both.

**[maintainability]** PR body — Testing section
The PR body documents the CI-config-only rationale and the manual synced/drifted verification thoroughly, but omits the literal `Tests: N passed, N skipped, N failed` summary line the test-integrity policy asks for.
For a workflow-YAML-only change the suite yields no signal, so this is the known false-positive class rather than a real gap — flagging only for completeness against the policy checklist.
**Suggestion:** Optionally add one line to the Testing section stating "No suite run — change is CI-workflow-only; suite cannot validate workflow YAML" so the policy box is explicitly ticked.

---

### Style Notes
Clean. The inline comment block explaining "Layer 2 of the lockfile-drift defense" with the `#435` cross-reference is exactly the right amount of context for a future reader encountering this step — keep that habit.

---

PR #429, opened 2026-06-03
feat: custom user-created lifts (persistence, ownership, resolution)

brownm09:
Closes #425

## Summary

Adds a per-user `CustomLift` entity that persists, resolves alongside the built-in catalog (custom takes precedence), and is selectable everywhere lifts are picked. Follows the hexagonal boundary (ADR-002): catalog data stays pure in `packages/core`; persistence lives in the API adapter layer.

### Design decisions
- **`Lift.id` is a stable uuid** (the DB row PK, surfaced directly as the REST key) — *not* derived from the name. A lift can be renamed via PATCH without breaking references. DB keeps `@@unique([userId, name])` to block duplicate names per user.
- **`resolveLift` gains an optional `customLifts` arg** searched before the catalog — pure, source-compatible with existing 3-arg call sites, unknown id still throws.
- **Consumption wiring:** since the production system keys lifts by free `LiftName` string (catalog `id`/`resolveLift` are not on the production path), the single touch-point is `GET /programs/:program/lifts`, which now merges the user's custom-lift names. All downstream consumers (training maxes, lift records, workout generation, lift-metadata) already accept arbitrary lift strings.

### Changes
- **types:** `CustomLift extends Lift`, `Lift.isCustom?`; `CustomLiftResponse`/`CreateCustomLiftRequest`/`UpdateCustomLiftRequest`
- **core:** `resolveLift` custom-first lookup + unit tests
- **prisma:** `CustomLift` model + migration `20260603000000_add_custom_lift`, `@@unique([userId, name])`, `@@index([userId])`
- **ports/adapters:** `ICustomLiftRepository` (full CRUD) with in-memory + Prisma adapters, threaded through all three repository factories. Prisma `update`/`delete` are **userId-guarded** (PK is `id` alone) to preserve ownership isolation.
- **api:** `GET/POST/PATCH/DELETE /lifts/custom` (`CustomLiftController` in `LiftsModule`); 404 `CustomLiftNotFoundError` (added to `DomainNotFoundFilter`), new 409 `DomainConflictFilter` for `CustomLiftConflictError`, registered in `main.ts`.

## Acceptance criteria
- [x] `CustomLift` type exported; `Lift.isCustom?` added
- [x] `CustomLift` Prisma model + migration, unique per `(userId, name)`
- [x] `ICustomLiftRepository` with in-memory + Prisma adapters via the per-user factory
- [x] `resolveLift` prefers custom over catalog; unknown id still errors
[...]

Issue #426, opened 2026-06-03
[feat] Onboarding — estimate maxes for any catalog or custom lift


brownm09:
**Priority: High** · **Depends on #425** (custom lifts) · Proposal: [`docs/proposals/2026-06-03-onboarding-any-lift-max-estimation.md`](https://github.com/brownm09/lifting-logbook/blob/claude/onboarding-movement-config-yPPAR/docs/proposals/2026-06-03-onboarding-any-lift-max-estimation.md)

## Problem

Onboarding's "Enter Lifts" step is hardcoded to Bench/Squat/Deadlift via `LiftKey = 'bench' | 'squat' | 'deadlift'` and `LIFT_LABELS` in `apps/web/app/(authed)/onboarding/lib.ts`. A lifter whose program centers on other movements can't seed those maxes during setup, and there's no path for a lift outside the catalog.

Second, latent defect in the same flow: **the computed maxes are never persisted.** `OnboardingFlow.tsx` computes `computedMaxes` and shows them on Confirm, but `handleConfirm` calls `createFirstCycle(selectedProgramId)` (`onboarding/actions.ts`) with only the program id — the entered maxes are discarded, so the user lands on `/cycle/1` with no training maxes from onboarding.

## Proposed solution

- Replace the fixed `LiftKey` union + `LIFT_LABELS` record with a **dynamic add/remove list** of lift rows. Seed with the big three (zero-config users see no change); allow adding any catalog **or custom (#425)** lift. Reuse the existing manage-lifts lift-picker rather than building a new selector.
- Keep estimation on the client `brzycki1RM` (`onboarding/lib.ts`), mirroring core `estimateTrainingMax` (`packages/core/src/services/maxes/estimateTrainingMax.ts`). Rep-range guidance unchanged.
- **Persist confirmed maxes**: extend `createFirstCycle` (`onboarding/actions.ts`) to accept `{ lift, oneRm }[]` and write them via the existing `PATCH /programs/:program/training-maxes` before redirecting to `/cycle/1` — closing the discard gap.

## Acceptance criteria

- [ ] "Enter Lifts" supports adding/removing lift rows; any catalog or custom lift is selectable
- [ ] Big-three defaults preserved for zero-config users
- [ ] Confirmed maxes are persisted to training maxes on completion (discard gap closed)
- [ ] Estimation continues to use the Brzycki path; rep-range guidance unchanged
[...]

Issue #425, opened 2026-06-03
[feat] Custom user-created lifts — persistence, ownership, and resolution


brownm09:
**Priority: High** · Proposal: [`docs/proposals/2026-06-03-custom-lifts.md`](https://github.com/brownm09/lifting-logbook/blob/claude/onboarding-movement-config-yPPAR/docs/proposals/2026-06-03-custom-lifts.md)

> Foundational issue. Blocks #B (onboarding any-lift) and #C (movement profiles) — both need a user-owned lift entity to attach to.

## Problem

Lifts are referenced as plain strings (`LiftRecord.lift`, `TrainingMax.lift`, `LiftMetadata.lift` are unconstrained `String` columns), and the catalog is a fixed 25-entry in-code array in `packages/core/src/catalog/lifts.ts` with zero persistence. A user who trains a movement the catalog doesn't contain has no first-class way to add it. This blocks onboarding-for-any-lift and per-lift movement profiles, both of which need a user-owned lift entity.

The [original lift-library proposal](https://github.com/brownm09/lifting-logbook/blob/main/docs/proposals/2026-04-13-lift-library-exercise-tagging.md) explicitly left custom lifts out of scope and flagged the persistence/ownership boundary as a spike to do before client work. This is that resolution.

## Proposed solution

- **Types** (`packages/types/src/domain.ts`): add `CustomLift extends Lift { userId; isCustom: true; createdAt }`; add `Lift.isCustom?: boolean`.
- **Persistence** (`apps/api/prisma/schema.prisma`): `CustomLift` model, `@@unique([userId, name])`, `@@index([userId])`, `@@map("custom_lift")` — following the `LiftMetadata` / `TrainingMax` / `StrengthGoal` per-user pattern.
- **Port + adapters**: `ICustomLiftRepository` (`apps/api/src/ports/`) with in-memory + Prisma adapters, threaded through the existing `IRepositoryFactory.forUser(...)` (#144 pattern).
- **Resolution**: extend `resolveLift` (`packages/core/src/catalog/resolve.ts`) to consult the user's custom catalog **before** the built-in `LIFT_CATALOG`. Core stays pure — custom lifts are passed in as data; no DB access in `packages/core` (ADR-002).
- **API**: `GET/POST/PATCH/DELETE /lifts/custom` returning the same `Lift`-shaped contract the catalog uses, so existing lift UI works unchanged.

## Acceptance criteria

[...]

PR #437, opened 2026-06-04
[feat] Lift movement profiles — joint action and complexity metadata

### LINKED ISSUE #427: [feat] Lift movement profiles — joint action and complexity metadata
**Priority: High** · **Depends on #425** (custom lifts, for the editable-for-custom half) · Proposal: [`docs/proposals/2026-06-03-lift-movement-profiles.md`](https://github.com/brownm09/lifting-logbook/blob/claude/onboarding-movement-config-yPPAR/docs/proposals/2026-06-03-lift-movement-profiles.md)

## Problem

`Lift` (`packages/types/src/domain.ts`) describes a lift via `classification` (`compound | accessory`, a training **role**) and `movementTags` (`push | pull | vertical | horizontal | hinge | carry | squat`, a movement **pattern**). Neither captures anatomical **joint actions** (internal/external rotation, flexion, extension, abduction, adduction) nor movement **complexity** (simple vs compound). These should be preconfigured for in-catalog lifts and editable for custom lifts.

Important distinction to preserve: movement *complexity* (`simple | compound` = single- vs multi-joint) is **not** the existing `classification` (`compound | accessory` = training role). A Goblet Squat is movement-`compound` but role-`accessory`. Both axes are kept.

## Proposed solution

Per the chosen modeling approach, fold pattern + joint action + complexity into **one combined `MovementProfile` structure** (`packages/types/src/domain.ts`):

```ts
export type JointAction =
  | 'flexion' | 'extension'
  | 'internal-rotation' | 'external-rotation'
  | 'abduction' | 'adduction';
export type MovementComplexity = 'simple' | 'compound';
export interface MovementProfile {
  patterns: MovementTag[];        // existing movementTags migrate here
  jointActions: JointAction[];    // new anatomical axis
  complexity: MovementComplexity; // new simple/compound axis
}
```

- Extend `Lift` with `movementProfile: MovementProfile`; migrate `movementTags` data into `.patterns` (keep `movementTags` as a one-release deprecated alias, or migrate all call sites in the same PR). Keep `classification` unchanged; document the role-vs-complexity distinction in the type doc comment.
- Preconfigure all 25 catalog entries in `packages/core/src/catalog/lifts.ts` (e.g. Face Pull → external-rotation + horizontal pull + simple; Back Squat → flexion/extension + squat + compound).
- Custom-lift create/edit + `POST/PATCH /lifts/custom` (#425) accept a `movementProfile`; catalog values are read-only defaults.

## Acceptance criteria

- [ ] `JointAction`, `MovementComplexity`, `MovementProfile` exported from `packages/types`
- [ ] `Lift.movementProfile` added; existing `movementTags` data preserved under `.patterns`
- [ ] All 25 catalog entries carry a preconfigured `movementProfile`; `packages/core` unit tests assert each
- [ ] Custom-lift create/edit accepts and persists a `movementProfile` (depends on #425)
- [ ] Role-vs-complexity distinction documented in the type doc comment
- [ ] Strict TS compilation passes across `packages/types`, `packages/core`, and updated call sites

## Out of scope

Volume/analytics aggregation by joint action; a movement-pattern browse/filter UI; auto-inferring profiles from lift names.

brownm09:
## Summary

Folds the existing `movementTags` pattern axis into a combined **`MovementProfile`** structure on `Lift`, adding two new axes — anatomical **joint actions** and movement **complexity** — preconfigured for catalog lifts and editable for custom lifts.

```ts
export interface MovementProfile {
  patterns: MovementTag[];        // existing movementTags data migrated here
  jointActions: JointAction[];    // new anatomical axis
  complexity: MovementComplexity; // simple (single-joint) | compound (multi-joint)
}
```

This is a **full migration (clean break)**: `movementTags` is removed everywhere and its data moves to `movementProfile.patterns` — no deprecated alias. (Custom lifts shipped the same day in #429, so there is no production data at risk.)

`complexity` (`simple|compound`, single/multi-joint mechanics) is deliberately kept **distinct** from `classification` (`compound|accessory`, training role) — documented in the type doc comment. A Goblet Squat is movement-`compound` yet role-`accessory`.

### Storage
The Prisma `CustomLift` model stores the profile as **separate columns**: `patterns String[]` (renamed from `movementTags`), `jointActions String[]`, `complexity String @default("simple")`. Migration `20260603120000_add_movement_profile` renames the column (preserving data) and adds the two new ones.

## Acceptance criteria

- [x] `JointAction`, `MovementComplexity`, `MovementProfile` exported from `packages/types`
- [x] `Lift.movementProfile` added; existing `movementTags` data preserved under `.patterns`
- [x] All catalog entries carry a preconfigured `movementProfile`; `packages/core` unit tests assert each
- [x] Custom-lift create/edit accepts and persists a `movementProfile` (#425/#429)
- [x] Role-vs-complexity distinction documented in the type doc comment
- [x] Strict TS compilation passes across `packages/types`, `packages/core`, and updated call sites

## Notes / decisions

- **Catalog count:** the issue says "all 25 catalog entries"; the catalog actually has **23** — all 23 are covered.
- **Default-on-omit:** when a custom-lift create omits `movementProfile`, it defaults to `{ patterns: [], jointActions: [], complexity: 'simple' }`, preserving the prior behavior where `movementTags` could be omitted. `simple` is the conservative single-joint default.
- **Nested validation:** `movementProfile` is validated via a `MovementProfileDto` with `@ValidateNested()` + `@Type()`, guarded by `@IsObject()` (matching the existing `update-settings.dto.ts` pattern) so primitives can't bypass the nested no-op under the `whitelist`-only ValidationPipe. E2E tests assert 400 for invalid pattern, joint action, complexity, and a primitive profile — proving validation actually fires.
- **`apps/web` untouched:** no web code references `movementTags` (this issue is independent of #426).

## Testing

`npm test` across all workspaces (Docker up; DB e2e via Testcontainers applied the new migration via `prisma migrate deploy`):

- **core:** Tests: 168 passed, 0 skipped, 0 failed
- **api:** Tests: 326 passed, 0 skipped, 0 failed (includes the DB e2e asserting `patterns`/`jointActions`/`complexity` columns round-trip)
- **web:** Tests: 56 passed, 0 skipped, 0 failed
- **Combined: 550 passed, 0 skipped, 0 failed**

`git grep movementTags` returns only a single intentional doc-comment reference after migration.

### Suppression policy note
The pre-PR suppression grep flags `!` non-null assertions in `liftCatalog.test.ts` (e.g. `LIFT_CATALOG.find((l) => l.id === 'face-pull')!`). These are **not new suppressions of a real error** — they are the pre-existing test idiom already used throughout this file (deadlift, bench-press, etc.) for asserting a catalog entry known to exist by id; a missing entry would surface as an obvious test failure. The four added instances (face-pull, lateral-raise, back-squat, goblet-squat) follow that established pattern.

Closes #427


brownm09:
## Review: [feat] Lift movement profiles — joint action and complexity metadata

### Summary
**Merge with minor changes.** This is a clean, well-tested full migration — types, catalog, DTOs, ports, both adapters, and the Prisma layer all move in lockstep, and the test additions (default-on-omit, round-trip, four validation-failure cases, DB-column round-trip) genuinely exercise the new behavior rather than just renaming assertions. The one finding worth resolving before merge is a nullability asymmetry in the migration (`jointActions` added bare while `complexity` got `NOT NULL DEFAULT`), which leaves a latent NULL-read path for any row created in the gap between today's two custom-lift migrations. The rest are documentation-sync nits.

---

### Blocking Findings

**[reliability]** `apps/api/prisma/migrations/20260603120000_add_movement_profile/migration.sql:13`
The migration adds `"jointActions" TEXT[]` with no `NOT NULL` and no `DEFAULT`, while the very next line adds `"complexity" TEXT NOT NULL DEFAULT 'simple'`. The Prisma schema types the field as `jointActions String[]` (non-nullable). Any `custom_lift` row created by `20260603000000_add_custom_lift` *before* this migration runs will have `jointActions = NULL`. Prisma Client reads of a NULL scalar-list column into a non-nullable `String[]` field throw `Inconsistent column data`, which would surface as a 500 on `GET /lifts/custom` for that owner. The create path always writes `[]`, so new rows are safe and the DB e2e (fresh container, new rows only) never exercises the gap — meaning the test pass does **not** clear this. The original plan specified `NOT NULL DEFAULT '{}'`; the implementation dropped it for `jointActions` only.
**Fix:** `ADD COLUMN "jointActions" TEXT[] NOT NULL DEFAULT '{}';` to match `complexity` and the Prisma non-nullable type. (Pre-existing caveat: the renamed `patterns`/former `movementTags` column was also bare `TEXT[]`, so this convention is inherited — but `complexity` shows the stricter idiom was intended here, making the asymmetry look unintentional.)

---

### Non-Blocking Findings

**[documentation]** `apps/api/prisma/schema.prisma:155`
The new comment reads `// MovementProfile, stored as separate columns (see ADR / issue #427).` but no ADR exists — `grep` of `docs/adr/` for movement profile / #427 returns nothing. The design record that *does* exist is `docs/proposals/2026-06-03-lift-movement-profiles.md`. A dangling "see ADR" pointer sends a future reader looking for a document that was never written.
**Suggestion:** Either write the ADR (the role-vs-complexity separate-axis decision is arguably ADR-worthy) and reference it by number, or change the comment to point at the proposal/issue only.

**[documentation]** `docs/proposals/2026-06-03-lift-movement-profiles.md:3,50`
The proposal is still `**Status:** draft` and its design text says `movementTags` "may be retained as a deprecated alias … or every call site migrated in the same PR — implementer's call." This PR made the decision (full clean-break migration, no alias) and shipped it, but the proposal still reads as undecided. It also still says "all **25** catalog entries" where the PR correctly notes 23.
**Suggestion:** Update the proposal `Status` to `accepted`/`shipped`, replace the alias-optional language with the chosen full-migration decision, and fix the 25→23 count, so the proposal reflects what actually shipped.

---

### Questions for Author

**Question:** Is `farmers-carry` having `jointActions: []` with `complexity: 'compound'` the intended model, rather than `'simple'` or a stabilization joint action?
**Context:** `packages/core/src/catalog/lifts.ts:531` — every other `compound` entry carries flexion/extension joint actions, but farmer's carry has an empty `jointActions` array yet is marked movement-`compound`. The catalog test only asserts the array is valid (empty is allowed), so nothing pins the intent.
**Tradeoffs:** (a) Keep `compound` + `[]` — defensible since a loaded carry is a multi-segment stabilization task with no prime mover through a ROM, but it's the only compound-with-no-joint-actions entry, which a future reader may "fix" as a mistake. (b) Switch to `complexity: 'simple'` — aligns "no joint action driven" with single-joint mechanics, but mislabels the whole-body demand. (c) Keep `compound` and add a one-line comment on that entry explaining the empty `jointActions` is deliberate — preserves intent and pre-empts the false-positive fix.

---

### Style Notes
None beyond the above — naming (`patterns`/`jointActions`/`complexity`), the shared `MovementProfileDto` extraction, and the `@IsObject()` guard comment mirroring `update-settings.dto.ts` are all consistent with surrounding code. The `!` non-null assertions flagged by the suppression grep are correctly justified in the PR body as the pre-existing catalog-test idiom.

---

Issue #441, opened 2026-06-04
[fix] Remove stale lint exclusions from pre-commit hook


brownm09:
## Problem

`.husky/pre-commit` runs `turbo run lint` but filters out two workspaces:

```sh
turbo run lint \
  --filter=!@lifting-logbook/web \
  --filter=!@lifting-logbook/core
```

Both exclusions are now stale:

- **`@lifting-logbook/web`** was excluded during early scaffolding when `apps/web` had no `src/` (issue #9). `apps/web` is now fully built (Next.js App Router under `apps/web/app/`), defines a `lint` script (`"lint": "eslint app"`), and lints clean (0 errors). Because pre-commit skips it, lint errors in `apps/web` bypass the local `git commit` gate and surface only in CI's "Lint & Test" job. This already caused a real miss — an unused import committed in PR #438, caught only on a manual lint run.
- **`@lifting-logbook/core`** was excluded for 51 pre-existing GAS-migration lint errors (issue #51). Issue #51 is now CLOSED and `packages/core` lints clean (`npm run lint -w @lifting-logbook/core` → 0 errors). The exclusion no longer reflects reality.

## Impact

The local pre-commit lint gate is weaker than CI for two of the most actively developed workspaces. Lint errors slip through `git commit` and are caught late (CI) or never (local). This defeats the purpose of the pre-commit hook for those workspaces.

## Expected

The pre-commit hook should lint all workspaces (no stale filters), matching the CI "Lint & Test" job. `turbo run lint` with no filters passes clean today across all 7 workspaces.

Issue #418, opened 2026-06-02
[infra] Follow-ups from PR #416: next-wave action bumps + digest-pin evaluation


brownm09:
**Context:** Spun off from [PR #416](https://github.com/brownm09/lifting-logbook/pull/416) review.

PR #416 cleared the 2026-06-16 Node 20 deprecation wave by bumping `actions/checkout`, `google-github-actions/auth`, and `hashicorp/setup-terraform` to their next majors. Two adjacent hygiene items were deliberately left out of scope:

## 1. Audit remaining workflow actions for the next runtime cycle

Other major-tag pins not named in the current deprecation banner — these will be in some future Node-runtime wave and are easier to clear ahead of a deadline than under one:

- `actions/setup-node@v4`
- `actions/cache@v4`
- `actions/upload-artifact@v4`
- `google-github-actions/setup-gcloud` (deploy.yml, staging.yml)
- `docker/setup-buildx-action`
- `docker/build-push-action`

**Acceptance:** confirm the latest major of each is on the Node runtime GitHub currently ships on its runners, and bump in a single tightly-scoped PR (or document why staying back is preferred). Verify input compatibility against the target action.yml before bumping, same protocol as #416.

## 2. Evaluate digest-pinning for privileged workflows

`deploy.yml` and `staging.yml` authenticate to GCP via OIDC and push container images, holding workload-identity-federated credentials for both `GCP_STAGING_SERVICE_ACCOUNT` and `GCP_PROD_SERVICE_ACCOUNT`. The repo convention is floating major-tag pins (`@v3`, `@v5`, etc.), which automatically picks up patch releases but exposes the workflow to a compromised tag move on the upstream action repository — a single tag-force-push on `actions/checkout` or `google-github-actions/auth` could exfiltrate the OIDC token during the auth step.

**Options:**
- (a) Keep `@v<major>` everywhere — matches existing convention, automatic patch tracking, standard GitHub guidance for first-party actions.
- (b) Digest-pin only `deploy.yml`/`staging.yml` — meaningfully reduces supply-chain risk on the workflows that hold prod credentials, at the cost of manual upgrades and a slightly bifurcated convention.
- (c) Digest-pin everywhere — strongest posture, biggest maintenance tax, warrants Dependabot config to keep digests current.

**Acceptance:** pick an option (a/b/c), document the decision in `docs/operations/` or an ADR, and if (b) or (c) is selected, open the corresponding PR with the digest pins + Dependabot config.

[...]

PR #446, opened 2026-06-04
[fix] Correctness audit — cross-user spec isolation, domain guards, schedule fallback

### LINKED ISSUE #434: Deep correctness audit of application logic (latent bug hunt)
## Goal

A deliberate, multi-area audit of application logic for latent correctness bugs — beyond broken tooling and failing tests (both of which are currently addressed: see #430/#431 for the `test:db` flag and #432/#433 for the `npm ci` lockfile drift; all test suites pass on main).

Filed per a request to defer deep correctness investigation to its own tracked effort rather than inline it. Best run as a focused, multi-agent review pass.

## Suggested scope

- Domain/core logic in `packages/core` (cycle generation, max progression, resolution) — edge cases around deload/test weeks, rounding, and lift resolution.
- API adapter ↔ port contract correctness and per-user isolation invariants across all Prisma repositories (PK-only `where` clauses that should also guard `userId`).
- Next.js data-fetching / static-vs-dynamic correctness in `apps/web`.

## Concrete lead to verify first

During the bug hunt, a clean local `turbo run build --filter=@lifting-logbook/web` (after clearing a stale `.next`) failed prerendering `/settings/schedule` because the build fetched `http://localhost:3004/users/me/settings` with no API running:

```
Error occurred prerendering page "/settings/schedule"
Export encountered an error on /(authed)/settings/schedule/page: /settings/schedule, exiting the build.
```

Notably `/cycle` and `/history` hit the same fetch failure but **catch it and fall back to a default** (via `getActiveProgram`), whereas `/settings/schedule` does not appear to. **Needs verification**: determine whether CI's web build (Playwright E2E job) provisions an API or sets build-time config that masks this — i.e. whether this is a genuine missing-fallback bug in the schedule page or purely a local no-API artifact. The npm-ci failure (#432) previously masked this job, so it has not actually run green recently. Confirm via the next CI run on the #433 branch before treating it as a bug.

brownm09:
## Summary

Acts on the `#434` deep correctness audit. Three findings, all fixed here; the housekeeping items (#441, #421, #418) ride in a separate PR so this security fix is reviewable on its own.

### CRITICAL — cross-user custom-program-spec leak
`HybridLiftingProgramSpecRepository.getCustomSpec()` queried `customProgramSpec` by `programId` alone. Any authenticated user could read another user's custom program spec by UUID. The repo was also a **startup singleton** in `PrismaRepositoryFactory` (constructed with no `userId`), so it could not be scoped per-request.

- Added a `userId` constructor param; the query now guards via the relation: `where: { programId, program: { userId } }`.
- Moved instantiation into `forUser(user)` alongside every other per-user repo.

### HIGH — domain-logic guards (boundary + defense-in-depth)
`CustomProgramSpecRowDto` had no bounds on `increment` or `wtDecrementPct`, so nonsensical specs persisted:
- `increment: 0` → `MROUND` divided by zero → `NaN` weights.
- large `wtDecrementPct` → later sets' work percentage went negative → negative prescribed weight.

Fixes:
- **Boundary (root cause):** `@IsPositive()` on `increment`; `@Min(0) @Max(1)` on `wtDecrementPct`.
- **Core (defense-in-depth):** `MROUND` returns the unrounded value when `multiple === 0`; `PROG_SPEC_WORK_PCTS` throws `RangeError` when the decrement drives the final set's percentage negative.

### HIGH — `/settings/schedule` missing fetch fallback
The page awaited `fetchUserSettings()` with no `.catch()`, unlike `/programs` and `/cycle`. It crashed the build-time prerender (and runtime) when the API was unreachable. Added `.catch(() => DEFAULT_SETTINGS)`, annotated `fallback-covered-by:` per the error-fallback-test-coverage standard.

## Testing

```
core: Test Suites: 27 passed, 27 total · Tests: 174 passed, 174 total
web:  Test Suites: 12 passed, 12 total · Tests: 71 passed, 71 total
api:  Test Suites: 25 passed, 25 total · Tests: 334 passed, 334 total
```
Combined: **Tests: 579 passed, 0 skipped, 0 failed**. API suite ran the full Testcontainers Postgres DB E2E (Docker up locally).

New tests:
- `hybrid-program-spec.repository.spec.ts` — built-in routing + cross-user isolation where-clause + non-owner returns `[]`.
- `create-custom-program.dto.spec.ts` — rejects `increment` ≤ 0 and `wtDecrementPct` outside [0,1].
- `config.test.ts` — `MROUND(N, 0)` no longer `NaN`; `PROG_SPEC_WORK_PCTS` boundary + negative-throw.
- `schedule/page.test.tsx` — success path passes fetched schedule; failure path falls back to `null` without throwing (data-level assertions, not structure-only).

Pre-PR checks: suppression grep clean, test-integrity grep clean, no deleted tests, lockfile unchanged. Lint clean for `api`/`core`/`web` (one pre-existing unused-`eslint-disable` warning in `apps/web/app/api/healthz/route.ts:32`, untouched here).

## Acceptance criteria (#434)
- [x] Domain/core edge cases around rounding audited and guarded.
- [x] Per-user isolation invariant restored on the custom-program-spec read path.
- [x] `apps/web` static-vs-dynamic fetch correctness — `/settings/schedule` no longer crashes without an API (matches the `getActiveProgram` fallback pattern called out in the issue).

Closes #434

---

`/review` (claude-opus-4-8) reported 1 blocking + 2 non-blocking; all fixed in `b503f3a`:

- **[blocking · reliability]** DTO `@Max(1)` admitted a cross-field-invalid `wtDecrementPct` (e.g. `0.6`/3 sets) that passed validation, persisted, then threw an uncaught `RangeError` → 500 at plan generation. **Fixed** — added the `WtDecrementWithinSetBound` cross-field validator (400 at creation); the core `RangeError` is now a true backstop. Corrected the e2e fixture (`0.9` → `0.1`), which was itself out-of-range.
- **[non-blocking · performance]** `HybridLiftingProgramSpecRepository` rebuilt the global in-memory seed map per request. **Fixed** — the factory now shares one `InMemoryLiftingProgramSpecRepository` and injects it; the per-user wrapper binds only `prisma` + `userId`.
- **[non-blocking · maintainability]** Isolation unit test asserted query shape, not Prisma's actual filtering. **Fixed** — added a Testcontainers DB E2E: owner creates a custom program; a different user requesting the same UUID's `/spec` gets `[]`.
- **[question · existing data]** No production concern — custom programs shipped in #425 with no real user data yet; the only out-of-range instance was the e2e's own fixture, corrected here.

Updated test totals: api **25 suites, 337 passed** (+2 DTO cross-field, +1 DB isolation E2E). Full suite remains green (core 174, web 71).

---

### Additional change (not a review finding)

`8f2543b` removes an unused `eslint-disable-next-line no-console` in `apps/web/app/api/healthz/route.ts` (`no-console` is inactive for `apps/web`, so the directive only tripped ESLint's "unused directive" warning). This was the last warning blocking a fully clean `apps/web` lint (now 0/0). Folded in here at the author's request as part of the #434 quality pass; it is the only other `apps/web` change on this branch.

### DIFF
--- a/apps/api/src/adapters/prisma/hybrid-program-spec.repository.spec.ts
+++ b/apps/api/src/adapters/prisma/hybrid-program-spec.repository.spec.ts
@@ -1,0 +1,53 @@
+import { PrismaClient } from '@prisma/client';
+import { HybridLiftingProgramSpecRepository } from './hybrid-program-spec.repository';
+
+const CUSTOM_PROGRAM_ID = '11111111-1111-4111-8111-111111111111';
+
+function makePrisma(): PrismaClient {
+  return {
+    customProgramSpec: {
+      findMany: jest.fn().mockResolvedValue([]),
+    },
+  } as unknown as PrismaClient;
+}
+
+describe('HybridLiftingProgramSpecRepository', () => {
+  it('routes built-in (non-UUID) program names to the in-memory adapter', async () => {
+    const prisma = makePrisma();
+    const repo = new HybridLiftingProgramSpecRepository(prisma, 'user-1');
+
+    const spec = await repo.getProgramSpec('5-3-1');
+
+    // In-memory built-in program resolves without touching the DB.
+    expect(spec.length).toBeGreaterThan(0);
+    expect(prisma.customProgramSpec.findMany).not.toHaveBeenCalled();
+  });
+
+  it('scopes custom-program lookups to the requesting user (isolation guard)', async () => {
+    const prisma = makePrisma();
+    const repo = new HybridLiftingProgramSpecRepository(prisma, 'user-1');
+
+    await repo.getProgramSpec(CUSTOM_PROGRAM_ID);
+
+    expect(prisma.customProgramSpec.findMany).toHaveBeenCalledWith(
+      expect.objectContaining({
+        where: { programId: CUSTOM_PROGRAM_ID, program: { userId: 'user-1' } },
+      }),
+    );
+  });
+
+  it("returns [] when another user's id yields no owned rows", async () => {
+    const prisma = makePrisma();
+    // findMany already resolves to [] for a non-owner because of the userId guard.
+    const repo = new HybridLiftingProgramSpecRepository(prisma, 'attacker');
+
+    const spec = await repo.getProgramSpec(CUSTOM_PROGRAM_ID);
+
+    expect(spec).toEqual([]);
+    expect(prisma.customProgramSpec.findMany).toHaveBeenCalledWith(
+      expect.objectContaining({
+        where: { programId: CUSTOM_PROGRAM_ID, program: { userId: 'attacker' } },
+      }),
+    );
+  });
+});
--- a/apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts
+++ b/apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts
@@ -6,11 +6,13 @@ import { PrismaService } from './prisma.service';
 const UUID_PATTERN = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
 
 export class HybridLiftingProgramSpecRepository implements ILiftingProgramSpecRepository {
-  private readonly inMemory = new InMemoryLiftingProgramSpecRepository();
-
+  // The built-in spec data is user-independent. It defaults to a fresh in-memory
+  // repo for standalone construction (tests), but the factory injects a shared
+  // instance so the seed map is built once per process rather than per request.
   constructor(
     private readonly prisma: PrismaService,
     private readonly userId: string,
+    private readonly inMemory: ILiftingProgramSpecRepository = new InMemoryLiftingProgramSpecRepository(),
   ) {}
 
   async getProgramSpec(program: string): Promise<LiftingProgramSpec[]> {
--- a/apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts
+++ b/apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts
@@ -8,7 +8,10 @@ const UUID_PATTERN = /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{
 export class HybridLiftingProgramSpecRepository implements ILiftingProgramSpecRepository {
   private readonly inMemory = new InMemoryLiftingProgramSpecRepository();
 
-  constructor(private readonly prisma: PrismaService) {}
+  constructor(
+    private readonly prisma: PrismaService,
+    private readonly userId: string,
+  ) {}
 
   async getProgramSpec(program: string): Promise<LiftingProgramSpec[]> {
     if (UUID_PATTERN.test(program)) {
@@ -18,8 +21,10 @@ export class HybridLiftingProgramSpecRepository implements ILiftingProgramSpecRe
   }
 
   private async getCustomSpec(id: string): Promise<LiftingProgramSpec[]> {
+    // Guard on the owning program's userId — a bare where:{ programId } would let
+    // any authenticated user read another user's custom program spec by UUID.
     const rows = await this.prisma.customProgramSpec.findMany({
-      where: { programId: id },
+      where: { programId: id, program: { userId: this.userId } },
       orderBy: [{ week: 'asc' }, { order: 'asc' }],
     });
     return rows.map((s) => ({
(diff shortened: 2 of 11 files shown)


brownm09:
## Review: [fix] Correctness audit — cross-user spec isolation, domain guards, schedule fallback

### Summary
Merge with minor changes. The core of the PR — the cross-user isolation guard, the per-request factory refactor, the MROUND zero-guard, and the schedule fallback — is correct, well-tested, and the highest-value change (a genuine multi-tenant data leak) is solid. The one finding worth resolving before merge is a **reliability gap**: the new `RangeError` in `PROG_SPEC_WORK_PCTS` is reachable through DTO-valid input and surfaces as an uncaught 500 at cycle-generation time. Gate checks all pass (coverage present for every behavior change; integrity clean; test-summary line in PR body). Appropriately scoped at ~247 lines.

Gate results: 2b — no doc-reconciliation table in this repo (skipped). 2c — no README coverage gaps (test files only). 2d — new behavior is tested (isolation, DTO, core guards, schedule fallback). 2e — no skip markers / deletions / lowered thresholds; `Tests: 579 passed, 0 skipped, 0 failed` present.

---

### Blocking Findings

**[reliability]** `apps/api/src/custom-programs/create-custom-program.dto.ts:31` (+ `packages/core/src/constants/config.ts:62`)
The DTO admits `wtDecrementPct` up to `1` (`@Max(1)`), but `PROG_SPEC_WORK_PCTS` now throws `RangeError` whenever `1 - (sets-1)*wtDecrementPct < 0`. Those two bounds disagree: `wtDecrementPct: 0.6` with `sets: 3` passes validation and persists, then throws at generation time via `createGridV2 → generateLiftPlan → PROG_SPEC_WORK_PCTS` ([generateLiftPlan.ts:46](packages/core/src/services/workout/generateLiftPlan.ts:46)). Neither `not-found.filter.ts` (`@Catch` not-found errors) nor `conflict.filter.ts` (`@Catch(CustomLiftConflictError)`) catches a bare `RangeError`, so it falls through to NestJS's default handler as a **500 on cycle view**. Note this is a behavior change in kind: before this PR the same input produced (wrong) negative weights but did not crash — the guard trades silent-bad-data for an uncaught exception on a critical read path.
**Fix:** Enforce the cross-field bound at the boundary so it's a 400 at creation, not a 500 at read — a class-validator custom constraint on `CustomProgramSpecRowDto` checking `wtDecrementPct <= 1/(sets-1)` (with the existing core `RangeError` remaining as a true unreachable backstop). Alternatively, map `RangeError` to a 4xx/handled response at the API boundary. If you judge the trigger too rare to fix now, file a tracking issue and link it per ADR-028 rather than leaving it undispositioned.

---

### Non-Blocking Findings

**[performance]** `apps/api/src/adapters/prisma/prisma-repository-factory.ts:46`
`new HybridLiftingProgramSpecRepository(this.prisma, user.id)` is now built on every `forUser()` call, and its constructor instantiates a fresh `InMemoryLiftingProgramSpecRepository`, which rebuilds the seed `Map` (`new Map([[SEED_PROGRAM, seedProgramSpec()]])`) each time. The in-memory adapter's own comment states "Program specs are global (not per-user)", so this is repeated work for data that never varies by request. The cost is small today (one seeded program), but it grows with the built-in catalog and is on the hot path of authenticated requests.
**Suggestion:** Keep a single shared `InMemoryLiftingProgramSpecRepository` (or the factory's existing `philosophyRepo` pattern) and pass it into the per-request hybrid repo, so only the `userId`/`prisma` binding is per-request and the global seed is constructed once.

**[maintainability]** `apps/api/src/adapters/prisma/hybrid-program-spec.repository.spec.ts:30`
The isolation test asserts the *shape* of the `findMany` argument (`where: { programId, program: { userId } }`) against a mock that always resolves `[]`. It confirms the guard is written, but it would pass even if Prisma's relation-filter semantics were wrong (e.g., if `program: { userId }` did not actually constrain the join) — the literal object is asserted, not the filtering behavior. Given this is the fix for a real cross-tenant leak, the query-shape assertion is the weakest link.
**Suggestion:** Add a Testcontainers DB E2E (the suite already provisions Postgres) that seeds a custom program for user A and asserts user B's repo returns `[]` for that program's UUID — end-to-end proof the relation filter isolates, not just that the object is passed.

---

### Questions for Author

**Question:** Should existing persisted custom programs with an out-of-range `wtDecrementPct` be migrated or surfaced, or is the data guaranteed clean?
**Context:** The new `RangeError` only matters for specs where `1 - (sets-1)*wtDecrementPct < 0`. If any such rows already exist (the field had no bound before this PR), they will start 500-ing on cycle view the moment this ships — the blocking finding's trigger isn't only future input.
**Tradeoffs:** (a) One-off query to check for existing out-of-range rows before merge — cheap, confirms whether the blocking finding is latent-only or live. (b) Assume clean (no programs created yet / builder never allowed it) — fine if true, but unverified. (c) Pair the cross-field DTO fix with a data backfill/clamp — only warranted if (a) finds offenders.

---

### Findings Disposition
Per ADR-028, before `gh pr merge` record each finding's disposition (fixed in `<sha>` | filed `#N`) in a "Review findings disposition" section of the PR body:
- [reliability] cross-field `wtDecrementPct` / `RangeError` → 500 — **fix or file**
- [performance] per-request InMemory reconstruction — **fix or file**
- [maintainability] isolation test asserts shape not behavior — **fix or file**

---

Issue #417, opened 2026-06-02
[infra] Production environment has no approval rule — Deploy (production) runs straight through


brownm09:
## Symptom

Main Deploy run [26839874601](https://github.com/brownm09/lifting-logbook/actions/runs/26839874601) for the #412 squash commit `e6ba23a` reached the `Deploy (production)` job and **executed immediately** without pausing for human approval. The job's status went straight from queued → `in_progress` → `success`, no `waiting` state.

The job in `.github/workflows/deploy.yml` declares:

```yaml
deploy-production:
  name: Deploy (production)
  ...
  environment: production
```

GitHub Environment-level protection rules (required reviewers, wait timer, branch restrictions) are configured in the **repo Settings → Environments → production** UI, not in the workflow YAML. Whatever is there today does not require approval.

## Why this matters

The pipeline header comment (`deploy.yml:4-5`) describes the design as:

> Pipeline: [preflight] → terraform staging → build images → deploy staging → smoke test → **[manual approval]** → terraform production → deploy production

The actual behavior is missing that manual approval step. Every push to `main` that passes staging integration tests will auto-deploy to production. That removes the human gate intended as a final guard against regressions that staging tests do not catch, and means a bad merge (e.g., a regression in `apps/web` that staging Playwright misses but real users hit) ships to prod with no review window.

This was masked for weeks by the same upstream chain that masked #411 — staging smoke (`/healthz` → `/livez`) failed every run since 2026-06-01, so production deploys never reached the gate. With #412 + #414 merged today, the first run that actually exercised `Deploy (production)` did so without the gate.

## Fix

Add a Required reviewers rule (and optionally a wait timer) to the `production` environment via repo Settings:

1. **Settings → Environments → production**
2. Check **Required reviewers** and add at least `brownm09`.
3. Optionally check **Wait timer** with a small value (e.g., 5–10 minutes) as a circuit-breaker against accidental merges.
4. Leave **Deployment branches** restricted to `main`.

Verification: trigger a no-op main push (e.g., a docs change) and confirm `Deploy (production)` enters `waiting` and surfaces an approval prompt before executing.

## Out of scope

This issue is about the environment-protection rule only. Whether to also restructure the workflow to make the gate explicit in YAML (e.g., a dedicated `approve-production` job with `environment: production` and no `run:` steps) is a separate design question — would let the gate be reviewed in code rather than in repo Settings.

## Acceptance criteria

- [ ] `production` environment has Required reviewers configured
[...]

PR #452, opened 2026-06-04
[fix] Consolidate production approval gate onto deploy-production

### LINKED ISSUE #451: [fix] Production deploy requires two approvals instead of one
## Problem

Both `approve-production` and `deploy-production` jobs in `.github/workflows/deploy.yml` reference `environment: production`. GitHub issues one approval prompt per job that references a protected environment, so reviewers are prompted twice for a single production deploy.

The comment on `deploy-production` (lines 636–638) already acknowledged this correctly — the gate is "fully expressed by `approve-production`" — but the `environment: production` declaration on `deploy-production` was left in place, causing the duplicate prompt.

## Fix

Remove `environment: production` from `deploy-production`. The upstream `needs: [approve-production]` dependency and `if: always() && needs.approve-production.result == 'success'` condition already prevent the job from running without approval.

brownm09:
## Summary

Fixes the double approval prompt on production deploys by **consolidating the `environment: production` gate onto the `deploy-production` job** and removing the redundant `approve-production` job.

**Root cause:** `approve-production` was added (#417) on the assumption — stated in its own comment — that "both this job and deploy-production share the same environment, so a single approval releases both." That assumption is wrong: GitHub raises one required-reviewer prompt **per job** that targets a protected environment, not one per environment. With both jobs referencing `environment: production`, reviewers were prompted twice.

**Fix:** Delete `approve-production` and put `environment: production` back on `deploy-production`, inheriting the build/smoke gating `if` that previously lived on `approve-production`. This yields:
- A single approval prompt.
- The GitHub-enforced approval backstop on the job that actually deploys (not just a workflow `if`).
- An accurate GitHub deployment record (attached to the real deploy, not a no-op job).
- `docs/deploy.md` Step 6 ("the `deploy-production` job will pause and wait for your approval") is correct again with no doc change.

## Testing

YAML-only change. Validated by parsing the workflow with `js-yaml` and asserting: `approve-production` is removed, `deploy-production.environment == 'production'`, and `deploy-production.needs` no longer references the deleted job. Pre-commit lint passed (7/7 packages). No application code touched.

Tests: N/A — no testable source changes (CI workflow definition only).

Review by `claude-opus-4-8` on the initial approach (0 blocking, 2 non-blocking + 1 design question): https://github.com/brownm09/lifting-logbook/pull/452#issuecomment-4626024717

- **[documentation] `docs/deploy.md:318-320` stale** — Fixed in `d5d47d1`. Consolidating the gate back onto `deploy-production` makes the existing doc sentence accurate again; no edit required.
- **[reliability] deployment record on no-op job** — Fixed in `d5d47d1`. The deployment record now attaches to `deploy-production` (the job that actually deploys), so production environment status reflects real deploy outcome.
- **[question] gate placement design** — Resolved by adopting alternative (b) from the review: deleted `approve-production`, gate lives on `deploy-production`. Chosen because the separate approval job's entire rationale rested on the disproven "one prompt per environment" assumption.

Closes #451

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -594,25 +594,28 @@ jobs:
             echo "API smoke test failed: expected 200 or 403, got $STATUS" && exit 1
           fi
 
-  # ── 5. Manual approval gate ──────────────────────────────────────────────
-  # Surfaces the production approval gate in the workflow itself, in addition
-  # to (not instead of) the `environment: production` reference on
-  # deploy-production below. Required reviewers are configured on the
-  # production environment in repo Settings (see #417); this job has no
-  # `run:` of substance — it exists to:
-  #   1. Make the gate visible in code review of this file rather than only
-  #      in repo Settings → Environments → production.
-  #   2. Act as a tripwire — if a future refactor strips `environment:
-  #      production` from deploy-production, this job still pauses the
-  #      pipeline at the same point.
-  # The reviewer clicks Approve once; both this job and deploy-production
-  # share the same environment, so a single approval releases both.
-  approve-production:
-    name: Approve production deploy
+  # ── 5. Deploy to production (requires manual approval) ───────────────────
+  # The `environment: production` reference below is the manual approval gate.
+  # GitHub raises one required-reviewer prompt per job that targets a protected
+  # environment, so the gate lives on this job alone — the job that actually
+  # deploys — giving a single approval, the GitHub-enforced backstop on the
+  # real deploy, and an accurate deployment record against the production
+  # environment.
+  #
+  # History: a separate `approve-production` job previously carried this
+  # environment "in addition to" deploy-production (#417), on the assumption
+  # that two jobs sharing one environment would prompt only once. They don't —
+  # GitHub prompts per-job — which produced a double approval (#451). The gate
+  # is now consolidated onto the deploy job, where the environment protection
+  # belongs.
+  deploy-production:
+    name: Deploy (production)
     runs-on: ubuntu-latest
     needs: [preflight, build-images, deploy-staging, smoke-test]
-    # Same gating logic as deploy-production below — only request approval
-    # when the pipeline would actually be ready to deploy.
+    # Only deploy when the pipeline is actually ready: build succeeded and,
+    # when staging is enabled, the staging smoke test passed. `always()` lets
+    # this run even though deploy-staging/smoke-test are skipped in
+    # production-only mode (their `skipped` result is not `success`).
     if: |
       always() &&
       needs.build-images.result == 'success' &&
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [fix] Remove duplicate production approval gate

### Summary
**Merge with minor changes.** The fix correctly resolves the double-approval — both `approve-production` and `deploy-production` declared `environment: production`, and GitHub raises one reviewer prompt per environment-referencing job. Removing it from `deploy-production` is safe because the GCP prod secrets are repository-scoped (confirmed in `docs/deploy.md` Step 5), so secret resolution is unaffected, and the `needs: [approve-production]` + `if` condition still gates the deploy. The highest-priority follow-up is a stale sentence in `docs/deploy.md` that still names `deploy-production` as the job that pauses for approval.

---

### Non-Blocking Findings

**[documentation]** `docs/deploy.md:318-320`
The setup guide states: *"The `deploy-production` job in `.github/workflows/deploy.yml` will pause and wait for your approval before proceeding."* After this PR, `deploy-production` no longer carries `environment: production` — `approve-production` is now the job that pauses. An operator following Step 6 and cross-checking the workflow will find no environment reference on `deploy-production` and may conclude the gate is missing.
**Suggestion:** Update the sentence to name `approve-production` as the pausing job (e.g., "The `approve-production` job will pause and wait for your approval; `deploy-production` then runs once approval is granted."). This is a one-line edit and fits naturally in this PR.

**[reliability]** `.github/workflows/deploy.yml:610` (approve-production) vs `:629` (deploy-production)
GitHub records a *deployment* against an environment from the job that references it. With `environment: production` now on the no-op `approve-production` job, the production environment's deployment history and the Deployments API will mark production as "deployed" the moment approval is granted — before `terraform apply` and the Cloud Run/Helm rollout in `deploy-production` even start. If the actual deploy fails, the production environment still shows a green deployment, decoupling environment status from real deploy outcome.
**Suggestion:** Accept this as a known tradeoff, or consider the alternative in the question below which keeps the deployment record on the job that actually deploys.

---

### Questions for Author

**Question:** Was hosting the gate on `approve-production` (rather than deleting `approve-production` and keeping `environment: production` on `deploy-production`) the intended design choice?
**Context:** Before this PR, `deploy-production` had an independent, GitHub-enforced approval gate via its own `environment: production`. After this PR, the *only* thing preventing an unapproved prod deploy is the workflow expression `if: always() && needs.approve-production.result == 'success'` (deploy.yml:638). A future refactor that edits the `needs:` array or the `if` logic could silently remove the gate, since the GitHub-level backstop on the deploy job is now gone — the inverse of the "tripwire" rationale in the comment at lines 597-609.
**Tradeoffs:** (a) Keep env on `approve-production` (this PR) — gate is visible as a dedicated job, but the real deploy has no independent GitHub-enforced backstop and the deployment record tracks the no-op job. (b) Delete `approve-production` and keep `environment: production` on `deploy-production` — single prompt, accurate deployment tracking, and the gate sits on the job that actually deploys, all enforced by GitHub independent of `if` logic; cost is losing the in-file "tripwire" job that surfaces the gate in code review.

---

### Style Notes
The updated comment reads clearly and the inline rationale ("No `environment: production` here — that would trigger a second required-reviewer prompt") is exactly the kind of note that prevents this regression from being reintroduced. Good.

---

### Findings Disposition
Per ADR-028, each finding must be fixed in this PR or filed and linked before merge. Both non-blocking findings are trivially fixable in-PR (the doc edit is one line; the reliability item is either a one-line acceptance note or the design change from the question). Recommend resolving the doc finding in this PR at minimum.

---

PR #455, opened 2026-06-05
[fix] Hide custom-lift option when partial catalog match exists

### LINKED ISSUE #454: [fix] Hide custom-lift option when partial catalog match exists
## Problem

When the user types a partial match of a catalog lift name (e.g. "overhead" for "Overhead Press"), the picker shows both the catalog match and an "Add 'overhead' as a custom lift" option. Clicking the custom option creates a misspelled/partial lift instead of the intended catalog lift.

## Root cause

`canAddCustom` is true whenever `trimmed.length > 0 && !alreadySelected && !matchesCatalog`. It does not check whether any catalog lifts partially match the query (`available.length`), so the custom-add option appears alongside catalog results.

## Expected behaviour

The "Add as custom lift" option should only appear when the query produces **no catalog matches at all** — neither exact (`matchesCatalog`) nor partial (`available.length === 0`). If any catalog lift contains the query as a substring, the custom option is hidden.

## Fix

In `apps/web/app/(authed)/onboarding/steps/StepLifts.tsx`, add `available.length === 0` to the `canAddCustom` guard:

```tsx
const canAddCustom = trimmed.length > 0 && !alreadySelected && !matchesCatalog && available.length === 0;
```

brownm09:
## Summary

Fixes #454.

Typing a partial lift name (e.g. `overhead`) now shows only the matching catalog option (`Overhead Press`) without also displaying an "Add as a custom lift" offer. Previously, `canAddCustom` only excluded exact catalog matches — partial matches set `available.length > 0` and `canAddCustom = true` simultaneously, so both the catalog option and the custom-add option appeared.

**Root cause:** `canAddCustom` checked `!matchesCatalog` (exact match only) but not whether any catalog lifts already appeared in `available`. Fix: add `available.length === 0` to the guard.

## Changes

- `StepLifts.tsx`: `canAddCustom` gains `&& available.length === 0` — custom-add only offered when no catalog results are present
- `StepLifts.test.tsx`: regression test covering the partial-match case ("type 'overhead' → only catalog option appears, no custom-add")

## Testing

```
Tests: 74 passed, 74 total (12 suites) — 1.353s
```

Suppressions: none added.
Test integrity: no skip markers, no deleted tests, no lowered thresholds.

TS compile warnings in `StepLifts.test.tsx` are pre-existing (jest-dom augmentation gap tracked in #421, mitigated by `diagnostics: { warnOnly: true }` in `jest.config.js`).

### DIFF
--- a/apps/web/app/(authed)/onboarding/steps/StepLifts.test.tsx
+++ b/apps/web/app/(authed)/onboarding/steps/StepLifts.test.tsx
@@ -129,6 +129,19 @@ describe('StepLifts — add a lift', () => {
       screen.queryByRole('option', { name: /as a custom lift/i }),
     ).not.toBeInTheDocument();
   });
+
+  it('does not offer custom-add when the query partially matches a catalog lift', async () => {
+    const user = userEvent.setup();
+    render(<Harness />);
+
+    // 'overhead' partially matches 'Overhead Press' — the catalog option appears
+    // but the custom-add option must not, since there are catalog results.
+    await user.type(screen.getByLabelText('Add a lift'), 'overhead');
+    expect(screen.getByRole('option', { name: 'Overhead Press' })).toBeInTheDocument();
+    expect(
+      screen.queryByRole('option', { name: /as a custom lift/i }),
+    ).not.toBeInTheDocument();
+  });
 });
 
 describe('StepLifts — remove a lift', () => {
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: [fix] Hide custom-lift option when partial catalog match exists

### Summary

Ready to merge. The one-line guard change is logically correct for every case (partial match, exact match, no match, already-selected), the `aria-expanded` condition remains sound under the new guard, and the regression test is tight and specific. No findings.

---

PR #459, opened 2026-06-08
[fix] Fall back to full lift catalog when onboarding fetch fails

### LINKED ISSUE #458: [fix] Onboarding lift search only offers 'custom lift' when catalog fetch fails
## Problem

On the onboarding **Enter Lifts** step, the add-a-lift search box only ever offers *\"Add ⟨text⟩ as a custom lift\"* — it never surfaces matching built-in catalog lifts (Overhead Press, Barbell Row, etc.). This makes the search feel broken.

## Root cause

The picker dropdown is built from a `catalog` prop in [`StepLifts.tsx`](apps/web/app/(authed)/onboarding/steps/StepLifts.tsx). The custom-lift option only appears when `available.length === 0` (no catalog match that isn't already selected) — that logic is correct.

The `catalog` is fetched server-side in [`onboarding/page.tsx`](apps/web/app/(authed)/onboarding/page.tsx):

```ts
try {
  const program = await getActiveProgram();
  catalog = await fetchLiftCatalog(program);   // GET /programs/:program/lifts
} catch {
  catalog = [...DEFAULT_LIFTS];                 // ['Bench Press','Squat','Deadlift']
}
```

Two compounding bugs:

1. **The fallback catalog is identical to the pre-selected rows.** `OnboardingFlow` seeds the same `DEFAULT_LIFTS` as the initial selected lifts. So when the fetch fails, every fallback entry is already selected → `available` is empty for **every** query → the only option the picker can ever render is the custom-lift add. Search is functionally dead in the fallback path.
2. **The fetch failure is silently swallowed** by a bare `catch {}` with no log, so the underlying catalog-fetch failure (auth/token handshake on Cloud Run) is invisible in production.

The healthy path (fetch succeeds → full `LIFT_NAMES` catalog) works correctly. The symptom indicates the catalog fetch is failing in production and falling back to the broken three-lift list.

## Fix

- Change the fallback catalog from `DEFAULT_LIFTS` to the full built-in `LIFT_NAMES` so search remains functional even when the API is unreachable.
- Surface the swallowed fetch error via `console.error` so the failure is observable.

## Acceptance criteria

- [ ] When the catalog fetch fails, the onboarding picker still offers all built-in lifts not already selected (e.g. Overhead Press, Barbell Row).
- [ ] The catalog-fetch failure is logged server-side rather than silently swallowed.
- [ ] Test coverage asserts the fallback catalog is the full built-in list, distinct from the pre-selected defaults.

## Notes

The separate question of *why* the catalog fetch fails in production (the Clerk + GCP identity-token handshake during onboarding) is upstream of this fix; this issue makes the fallback graceful and the failure observable so that root cause can be diagnosed from logs.

brownm09:
## Summary

The onboarding **Enter Lifts** search box only ever offered *"Add ⟨text⟩ as a custom lift"* and never surfaced built-in catalog matches (Overhead Press, Barbell Row, etc.), making the search appear broken.

**Root cause:** the picker's custom-lift option correctly only appears when `available.length === 0`. But on a catalog-fetch failure, [`onboarding/page.tsx`](apps/web/app/(authed)/onboarding/page.tsx) fell back to `DEFAULT_LIFTS` (`Bench Press`, `Squat`, `Deadlift`) — the exact three lifts `OnboardingFlow` pre-selects as the starting rows. Every fallback entry was therefore already selected, so `available` was empty for **every** query and the picker collapsed to the custom-lift option. The failure was also silently swallowed by a bare `catch {}`, hiding the upstream fetch failure (Clerk + GCP identity-token handshake on Cloud Run) in production.

## Changes

- Fall back to the full built-in `LIFT_NAMES` instead of `DEFAULT_LIFTS`, so catalog search stays functional even when the API is unreachable.
- Log the swallowed fetch error via `console.error` so the upstream failure is observable.
- Add `apps/web/app/(authed)/onboarding/page.test.tsx` covering the success path and asserting the fallback catalog is the full built-in list (distinct from the pre-selected defaults), plus that the error is logged.

## Acceptance Criteria

- [x] When the catalog fetch fails, the picker still offers built-in lifts not already selected (e.g. Overhead Press, Barbell Row).
- [x] The catalog-fetch failure is logged server-side rather than silently swallowed.
- [x] Test coverage asserts the fallback catalog is the full built-in list, distinct from the pre-selected defaults.

## Testing

Run from the repo root:

```
npx turbo run lint --filter=@lifting-logbook/web
```

- Full web suite: **Tests: 76 passed, 0 skipped, 0 failed (17.7s)** — includes the new `page.test.tsx` (success + fallback assertions). Re-run after the review-finding fix below.
- Lint: clean.
- Pre-PR suppression and test-integrity greps: clean (no suppressions, skips, or integrity violations added).
- No `package.json` changed → lockfile-sync check N/A.
- DB E2E suite not run (Docker): change is web-only, touches no `apps/api/prisma/` or repository code.

## Notes

This makes the fallback graceful and the failure observable. The separate question of *why* the catalog fetch fails in production (the onboarding-time auth-token handshake) is upstream of this fix and can now be diagnosed from the logged error.

`/review` ([comment](https://github.com/brownm09/lifting-logbook/pull/459#issuecomment-4653206103)): 0 blocking, 1 non-blocking.

- **[maintainability]** `page.test.tsx` — fallback assertion matched React's hand-escaped serialized attribute, coupling the test to renderer escaping rules. **Fixed in `e35af06`** — now parses the echoed `data-catalog` attribute and compares arrays (`toEqual([...LIFT_NAMES])`).

Closes #458

### DIFF
--- a/apps/web/app/(authed)/onboarding/page.test.tsx
+++ b/apps/web/app/(authed)/onboarding/page.test.tsx
@@ -1,0 +1,71 @@
+import { renderToStaticMarkup } from 'react-dom/server';
+import type { ReactElement } from 'react';
+import { LIFT_NAMES } from '@lifting-logbook/types';
+
+jest.mock('@/lib/api', () => ({
+  fetchLiftCatalog: jest.fn(),
+}));
+
+jest.mock('@/lib/active-program', () => ({
+  getActiveProgram: jest.fn().mockResolvedValue('5-3-1'),
+}));
+
+// Stub the client flow so the test isolates the page's catalog-fetch fallback.
+// It echoes the resolved catalog so we can assert what the page passed down.
+jest.mock('./OnboardingFlow', () => ({
+  OnboardingFlow: ({ catalog }: { catalog: string[] }) => (
+    <div data-catalog={JSON.stringify(catalog)}>onboarding-flow</div>
+  ),
+}));
+
+import { fetchLiftCatalog } from '@/lib/api';
+import OnboardingPage from './page';
+import { DEFAULT_LIFTS } from './lib';
+
+const mockedFetch = fetchLiftCatalog as unknown as jest.Mock;
+
+describe('OnboardingPage — catalog fetch fallback', () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  it('passes the fetched catalog to the flow on success', async () => {
+    const fetched = ['Squat', 'Bench Press', 'My Custom Lift'];
+    mockedFetch.mockResolvedValue(fetched);
+
+    const element = (await OnboardingPage()) as ReactElement;
+    const html = renderToStaticMarkup(element);
+
+    expect(html).toContain('onboarding-flow');
+    // Data-level assertion: the real fetched catalog — including the custom
+    // lift that no fallback would produce — reached the flow.
+    expect(html).toContain('My Custom Lift');
+  });
+
+  it('falls back to the full built-in LIFT_NAMES when the fetch fails', async () => {
+    mockedFetch.mockRejectedValue(new Error('API down'));
+    // Silence the intentional error log this path emits.
+    const errSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+
+    const element = (await OnboardingPage()) as ReactElement;
+    const html = renderToStaticMarkup(element);
+
+    expect(html).toContain('onboarding-flow');
+
+    // The fallback must be the full built-in catalog, NOT the three seeded
+    // DEFAULT_LIFTS. Overhead Press is in LIFT_NAMES but not DEFAULT_LIFTS, so
+    // its presence proves the picker stays searchable when the API is down (#458).
+    const overheadInDefaults = (DEFAULT_LIFTS as readonly string[]).includes('Overhead Press');
+    expect(overheadInDefaults).toBe(false);
+    expect(LIFT_NAMES).toContain('Overhead Press');
+    expect(html).toContain('Overhead Press');
+
+    // The catalog handed down is exactly the built-in list.
+    expect(html).toContain(`data-catalog="${JSON.stringify([...LIFT_NAMES]).replace(/"/g, '&quot;')}"`);
+
+    // The failure is surfaced, not silently swallowed.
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: [fix] Fall back to full lift catalog when onboarding fetch fails

### Summary
**Merge with one minor change.** This is a well-scoped, correctly diagnosed fix: the catch-block fallback was previously `DEFAULT_LIFTS` — byte-for-byte the three lifts `OnboardingFlow` pre-selects — so `available` was empty for every query and the picker could only ever offer the custom-lift row. Falling back to the full built-in `LIFT_NAMES` restores search, and replacing the bare `catch {}` with a `console.error` makes the upstream failure observable (and matches the sibling `getActiveProgram` pattern in `active-program.ts:9-10`). Tests cover both paths with real data-level assertions. The only finding is a brittle test assertion; everything else is solid.

---

### Non-Blocking Findings

**[maintainability]** `apps/web/app/(authed)/onboarding/page.test.tsx:62`
The exact-equality check builds the expected attribute by hand-replicating React's HTML escaping: `data-catalog="${JSON.stringify([...LIFT_NAMES]).replace(/"/g, '&quot;')}"`. This couples the test to React's server serialization format and its specific attribute-escaping rules. A React upgrade that changes attribute serialization (or any lift name that introduces a character React escapes differently, e.g. `&`, `<`) would break this assertion for reasons unrelated to the fallback contract it's meant to guard.
**Suggestion:** Extract the attribute value and compare parsed arrays instead of matching the escaped string — e.g. `const m = html.match(/data-catalog="([^"]*)"/); expect(JSON.parse(m[1].replace(/&quot;/g, '"'))).toEqual([...LIFT_NAMES]);` — or, since lines 58-59 already prove `Overhead Press` is present-and-not-a-default, drop the exact-match line and add a negative assertion (e.g. a known custom lift is absent) to pin "exactly the built-ins" without the escaping coupling.

---

### Findings Disposition
- **[maintainability]** `page.test.tsx:62` brittle exact-match assertion — _to be dispositioned (fixed in <sha> | filed #N) before merge._

---

PR #496, opened 2026-06-10
[fix] Run prisma generate before `turbo run dev` (Prisma dev-path gap, #490)

brownm09:
## Summary

Closes the residual gap in **issue #490 item 1 — Stale Prisma client paper cut**.

The paper cut (recurrences #224/#235/#242/#366) was structurally fixed for the **build/test** paths by:
- `postinstall: prisma generate` in `apps/api/package.json` ([#355](https://github.com/brownm09/lifting-logbook/pull/355))
- Turbo `build → db:generate` wiring in `turbo.json` ([#376](https://github.com/brownm09/lifting-logbook/pull/376))

The `dev` task was the one remaining path: `turbo run dev` had no `db:generate` dependency, so a **schema-only `git pull` followed by `npm run dev`** (no reinstall, so `postinstall` never fires) could still boot the API against a stale generated client.

This adds `db:generate` to the `dev` task's `dependsOn`. For `apps/api` this runs `prisma generate` (cached on `schema.prisma` / lockfile inputs, so it only re-runs when the schema actually changes) before the persistent dev server starts. For packages without a `db:generate` script (web) it is the same no-op the `build` task already relies on.

> Note: the retro cited `#405` as a recurrence, but `#405` is a concurrency-block review follow-up, not a Prisma issue — the last genuine recurrence was `#366` (2026-05-28), the same day the Turbo build wiring landed.

## Acceptance criteria
- [x] `npm run dev` regenerates the Prisma client after a schema change without a manual `prisma generate` or reinstall
- [x] No regression for packages without a `db:generate` task
- [x] Verified via `turbo run dev --dry-run`: `@lifting-logbook/api#dev` now depends on `@lifting-logbook/api#db:generate`

## Testing
Full suite via `turbo run test` (Windows, Node 20.11.1, Docker up for the DB E2E testcontainer):

`Tests: api 26 suites / 357 passed; all 12 workspace test tasks successful, 0 skipped, 0 failed (2m4s)`

The `dev`-task graph change cannot affect `test`/`build` outcomes (neither depends on `dev`); the suite run confirms no incidental breakage. Verified the new dependency edge with `turbo run dev --dry-run=json` for both `@lifting-logbook/api` (dependency present) and `@lifting-logbook/web` (no-op, unchanged).

Coverage: this is build-tooling config (no runtime behavior); per the project coverage table, refactor/tooling changes require only that existing tests pass — they do.

Suppression / test-integrity greps against `origin/main`: clean. Lockfile: unchanged.

## Scope
[...]

Issue #490, opened 2026-06-10
Biweekly retro — still-open action items (windows 04-19 .. 06-09)


brownm09:
Still-open findings surfaced by the backfilled biweekly retrospectives (engineering-journal `sessions/meta/retro/`). Deduped against existing open issues; only genuinely-untracked items are raised here.

> **Reconciliation (2026-06-10).** Traced each item against current `main`. The ~50 PRs since the retro windows already shipped most of this; only small residual gaps remained. Addressed below.

## New / untracked
- [x] **Stale Prisma client paper cut** — _already structurally fixed_ by `postinstall: prisma generate` (#355) + Turbo `build→db:generate` (#376); the last genuine recurrence was #366 (2026-05-28), and the retro's "#405" cite was a misattribution (a concurrency-block follow-up, not Prisma). Residual `dev`-path gap closed in **#496** (`turbo run dev` now depends on `db:generate`). _Known residual:_ direct workspace invocation (`npm run dev -w @lifting-logbook/api`) bypasses Turbo and is not covered — accepted (canonical `npm run dev` + the reinstall/`postinstall` path are covered; a script-level fix would double-run generation on the canonical path).
- [x] **Single source of truth for deploy secrets** — GCP Secret Manager is already the SSOT (every surface fetches from it; no manual copies). Residual fail-fast gap closed in **#497**: a pre-promote auth-secret preflight in `deploy-production` aborts before any prod mutation if `clerk-secret-key`/`clerk-publishable-key` is missing/placeholder (the #382 failure point), with transient-vs-missing error classification.
- [x] **Pre-prod served-surface smoke** — staging already smokes `/livez` + `/sign-in` through the real Cloud Run ingress before the promote gate, and prod has the DB-backed `/readyz` smoke (#460). Residual gap closed in **#497**: prod now also smokes the web served surface (`/livez` + `/sign-in`) after the prod web deploy, catching the #382 (auth outage) and #407 (`/healthz` GFE intercept) classes in prod.
- [x] **Staging integration-test CI flakiness** — root cause (Cloud SQL connectivity) needs GCP-runtime investigation, so it is now **tracked in #498** (with the specific GCP root-cause steps). The 7 `staging.yml` references that pointed at the *closed, off-topic* #344 are re-pointed to #498 in **#499**. The root-cause fix itself remains open in #498.

## Already tracked (linked, not refiled)
- Prod error alerting / thresholds → #487, #468, #472.

[...]

PR #481, opened 2026-06-09
[infra] Wire OTel Collector DaemonSet into the deploy pipeline (#474)

brownm09:
## Summary

Wires the OTel Collector **GKE DaemonSet** into `deploy.yml` so a push-to-main deploy lands the collector in staging + production with **no manual `helm install`**. Today the chart exists but nothing deploys it, so traces/logs/metrics never reach Grafana Cloud in prod. This is the primary topology per [ADR-018](../blob/main/docs/adr/ADR-018-observability-stack.md) (GKE carries the bulk of traffic; Cloud Run is the 10% A/B replica).

Closes #474. The Cloud Run sidecar is split to a focused follow-up — **#480** (see "Scope" below).

## ⚠️ One-time secret provisioning — required before merge/deploy

The Grafana token is **never committed** and the secrets are **not** Terraform-managed. The Secret Manager containers are created **and** populated by [`scripts/bootstrap-otel-secrets.sh`](../blob/main/scripts/bootstrap-otel-secrets.sh) (idempotent — re-running just adds a newer version). The deploy's sync step **fails loudly** if a secret is absent or empty. Populate the real tokens **once per environment** before the deploy that needs them (full walkthrough in `docs/deploy.md` → *OTel Collector / Grafana Cloud telemetry*):

**Easiest path — guided script:**
```bash
./scripts/bootstrap-otel-secrets.sh        # both envs; prompts for instance IDs + token (token hidden)
```
It creates the secret containers if they don't exist yet and writes both envs' versions.

**Equivalent by hand:** from the Grafana Cloud portal get the OTLP + Loki endpoints/instance-IDs and an API token (send metrics + logs + traces), then:
```bash
OTLP_HEADER="Basic $(printf '%s:%s' "$OTLP_INSTANCE_ID" "$GRAFANA_TOKEN" | base64 -w0)"
LOKI_HEADER="Basic $(printf '%s:%s' "$LOKI_INSTANCE_ID" "$GRAFANA_TOKEN" | base64 -w0)"
# create the container first if it doesn't exist: gcloud secrets create <name> --replication-policy=automatic --project=<proj>
printf '%s' "$OTLP_HEADER" | gcloud secrets versions add lifting-logbook-stg-otel-otlp-auth-header --data-file=- --project=<STG>
printf '%s' "$LOKI_HEADER" | gcloud secrets versions add lifting-logbook-stg-otel-loki-auth-header --data-file=- --project=<STG>
# …repeat with the prod secret names + prod project
```
[...]

Issue #460, opened 2026-06-09
[fix] Prod /programs/:program/lifts 500s — custom_lift table missing; deploy pipeline has no migrate step


brownm09:
## Summary

The onboarding lift-catalog symptom investigated in #458 / #459 has a confirmed production root cause, and it is **not** the Clerk + GCP identity-token handshake hypothesized there. Production `GET /programs/:program/lifts` (the catalog endpoint `fetchLiftCatalog` calls) returns **HTTP 500** on every request because the `custom_lift` table does not exist in the production database. The migration that creates it exists in the repo but was never applied to prod, and **the deploy pipeline has no migration step**.

## Evidence (production logs, `lifting-logbook-prod`)

**Web service** (`lifting-logbook-prod-web`) — fetch failures are API 5xx, not 401/403:

```
getActiveProgram: failed to fetch user settings, falling back to default Error: API 500 Internal Server Error for /users/me/settings
Error: API 503 Service Unavailable for /programs/5-3-1/training-maxes/history
Error: API 500 Internal Server Error for /programs/5-3-1/cycles/current
```

**API service** (`lifting-logbook-prod-api`) — the actual cause, recurring on 2026-06-04, 06-05 (×2), 06-07, 06-08:

```
ERROR  GET /programs/5-3-1/lifts
Invalid `prisma.customLift.findMany()` invocation:
The table `public.custom_lift` does not exist in the current database.
```

The API's `getLifts` (`apps/api/src/programs/manage-lifts.controller.ts`) runs `prisma.customLift.findMany()` and merges the result with `LIFT_NAMES`. With the table missing, Prisma throws → 500 → web `fetchLiftCatalog` throws → onboarding falls back. (PR #459 made that fallback graceful and observable; it did not address — and could not — the upstream 500.)

## Root cause

1. Migration `apps/api/prisma/migrations/20260603000000_add_custom_lift/migration.sql` exists in the repo (added 2026-06-03) and `model CustomLift` is in `schema.prisma`. The API code that depends on the table shipped to prod, but the table was never created there — errors begin 2026-06-04, immediately after.
[...]

PR #497, opened 2026-06-10
[infra] Harden prod promote: auth-secret preflight + web served-surface smoke (#490)

brownm09:
## Summary

Closes the residual gaps for **issue #490 items 2 and 3** with two fail-fast guards in `deploy.yml`'s `deploy-production` job. Both mirror patterns already reviewed and shipped elsewhere in the pipeline (the staging `clerk-check` gate and the staging smoke job).

> **Reconciliation note (#490 was largely already shipped).** The ~50 PRs since the retro windows already addressed most of #490. This PR adds only the genuine residual gaps; #481 (merged mid-session) further moved the prod web-deploy anchor, so this is rebased on current `main`.

### Item 2 — Validate production auth secrets (pre-promote)
GCP Secret Manager is already the single source of truth — every surface (Cloud Run `--update-secrets`, the k8s Secret sync, the web build-arg) *fetches* from it rather than holding its own copy, so the only failure mode left is **the source value was never populated**. [#382](https://github.com/brownm09/lifting-logbook/issues/382) was exactly that: a missing prod `clerk-secret-key`, caught only after the web service was already serving auth 500s in prod.

The new step runs **right after prod auth and before any prod mutation** (terraform apply / migrate / deploy), asserting `clerk-secret-key` and `clerk-publishable-key` are present and not the `REPLACE_ME` placeholder. It **aggregates** all failures (doesn't stop at the first) and **never echoes** a secret value. DB-url and OTel-header secrets keep their own per-step `[ -n ]` guards later in the job.

### Item 3 — Production web served-surface smoke
Prod previously had only the DB-backed API `/readyz` smoke ([#460](https://github.com/brownm09/lifting-logbook/issues/460)); the **web served surface was never re-checked in prod**. This adds `/livez` (Next.js runtime) + `/sign-in` (Clerk auth flow) probes after the prod web Cloud Run deploy, the same two-probe shape the staging smoke uses. It catches both classes that previously reached prod undetected:
- **#382** — missing prod `clerk-secret-key` → web cannot serve auth (`/sign-in` fails)
- **#407** — GFE intercepts `/healthz` before the container → probe `/livez`

The prod web service is `--allow-unauthenticated`, so no ID token is needed (unlike the IAM-gated API `/readyz`). Runs before the OTel steps so telemetry rollout never gates it.

## Acceptance criteria
- [x] A missing or `REPLACE_ME` prod auth secret aborts the deploy **before** any prod mutation, with an aggregated, value-free error
- [x] Prod web `/livez` + `/sign-in` are probed through the real Cloud Run ingress after the web deploy; a non-200 fails the deploy
- [x] New steps ordered correctly in `deploy-production` (preflight after auth; smokes after web deploy, before OTel) — verified by parsing the rendered workflow
- [x] No change to staging behavior or to production-only (non-staging) mode

## Testing
This is a **CI-workflow-only** change (`.github/workflows/deploy.yml`); it touches no application, build, or test code, so the Jest suite does not exercise it. Verification performed:

1. **YAML parse + step-order inspection** — loaded the rendered workflow and confirmed `deploy-production` step order: `Authenticate → Validate production auth secrets (pre-promote) → Set up Terraform → … → Deploy web to Cloud Run → /livez smoke → /sign-in smoke → OTel → Print URLs`.
2. **Pattern parity** — both guards are line-for-line consistent with the already-shipped staging `clerk-check` gate and the staging `smoke-test` job (`/livez` + `/sign-in`, `BASE_URL%/` trailing-slash strip, retry loops).
3. **Full suite (no incidental breakage)** via `turbo run test` (Windows, Node 20.11.1, Docker up): `Tests: api-client 21, core 199, api 357, web 78 — 12/12 workspace test tasks passed, 0 skipped, 0 failed`.

The behavioral guards run only inside the real prod deploy (require GCP + Cloud Run) and cannot be exercised locally; they fail loudly by design and roll-back guidance points to `docs/deploy.md`.

Suppression / test-integrity greps against `origin/main`: clean. No `package.json`/lockfile changes.

## Scope
Part of umbrella **#490** (items 2 and 3 of 4). Item 1 is [#496](https://github.com/brownm09/lifting-logbook/pull/496); item 4 (staging Cloud SQL flakiness, #344) needs GCP-runtime root-cause work and is being refined as tracking, not code. #490 stays open until item 4 is dispositioned.

---

`/review` on this PR surfaced 0 blocking, 3 non-blocking findings + 1 question ([review comment](https://github.com/brownm09/lifting-logbook/pull/497#issuecomment-4671126086)):

1. **[reliability] Preflight mis-diagnosed transient/IAM gcloud errors as "secret absent"** — **fixed in `ceaaf9d`.** stdout (secret) now captured to a shell var and stderr to a file; failure is classified — only `NOT_FOUND` reports "missing", anything else reports "inaccessible (transient/IAM), re-run if transient". Secret value never touches disk.
2. **[documentation] docs/deploy.md not updated for the new gate + smoke** — **fixed in `ceaaf9d`.** Added a note to *Ongoing operations → Approving a production deploy* describing both new self-guards.
3. **[maintainability] `clerk-publishable-key` check redundant with build-images** — **accepted (intentional).** Kept deliberately so the preflight is a single, complete "auth secrets present" assertion at the promote boundary; `build-images` runs in a separate job under different auth, so this is defense-in-depth, not dead weight. No code change.
4. **[question] Smoke targets `cloud_run_web_url`, not a custom-domain GLB** — **accepted (option a).** The Cloud Run URL exercises Google Frontend (so it catches the #407 `/healthz`-intercept class); single-user prod has no fronting GLB today. Adding a custom-domain probe is noted for if/when a GLB is introduced. No code change.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -690,6 +690,42 @@ jobs:
           workload_identity_provider: ${{ secrets.GCP_PROD_WORKLOAD_IDENTITY_PROVIDER }}
           service_account: ${{ secrets.GCP_PROD_SERVICE_ACCOUNT }}
 
+      # ── Pre-promote auth-secret validation (#490 item 2 / #382) ────────────
+      # Fail fast BEFORE any production mutation (terraform apply, migrate, Cloud
+      # Run / Helm deploy) if a serve-gating auth secret is absent or still the
+      # Terraform placeholder. GCP Secret Manager is already the single source of
+      # truth — every surface (Cloud Run --update-secrets, the k8s Secret sync,
+      # the web build-arg) fetches from it rather than holding its own copy — so
+      # the only failure mode left is "the source value was never populated".
+      # #382 was exactly that: a missing prod clerk-secret-key, caught only after
+      # the web service was already serving auth 500s in prod. This surfaces it at
+      # promote time instead. It aggregates all failures (does not stop at the
+      # first) and never echoes a secret value. Scope is the two auth secrets
+      # whose absence breaks the served surface; the DB-url and OTel-header
+      # secrets keep their own per-step `[ -n ]` guards later in the job.
+      - name: Validate production auth secrets (pre-promote)
+        run: |
+          MISSING=0
+          for SECRET in \
+            lifting-logbook-prod-clerk-secret-key \
+            lifting-logbook-prod-clerk-publishable-key; do
+            VALUE=$(gcloud secrets versions access latest \
+              --secret="$SECRET" \
+              --project="${{ vars.GCP_PROD_PROJECT_ID }}" 2>/dev/null || true)
+            if [ -z "$VALUE" ]; then
+              echo "::error title=Missing production secret::$SECRET is absent or has no accessible version. Populate it in GCP Secret Manager (docs/deploy.md Step 4) before promoting."
+              MISSING=1
+            elif [ "$VALUE" = "REPLACE_ME" ]; then
+              echo "::error title=Placeholder production secret::$SECRET is still the Terraform placeholder (REPLACE_ME). Set the real value in GCP Secret Manager (docs/deploy.md Step 4) before promoting."
+              MISSING=1
+            fi
+          done
+          if [ "$MISSING" -ne 0 ]; then
+            echo "Production auth-secret preflight failed — aborting before any production mutation. See the errors above." >&2
+            exit 1
+          fi
+          echo "Production auth secrets present and non-placeholder — proceeding."
+
       - name: Set up Terraform
         uses: hashicorp/setup-terraform@v4
         with:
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: [infra] Harden prod promote: auth-secret preflight + web served-surface smoke (#490)

### Summary
Merge with minor changes. Both guards are sound, fail in the safe (closed) direction, and faithfully mirror already-shipped patterns (the staging `clerk-check` gate and the staging `smoke-test` job). The one finding worth fixing before merge is an **error-message-diligence** issue in the preflight: a transient/permission failure from `gcloud secrets versions access` is currently swallowed and re-reported as "secret absent" — the exact misdiagnosis class the repo documented in #395. No blocking findings.

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/deploy.yml` — preflight (`Validate production auth secrets`)
`VALUE=$(gcloud secrets versions access latest ... 2>/dev/null || true)` swallows *all* stderr and exit codes. A transient Secret Manager 5xx, an expired/insufficient IAM grant, or a network blip all produce an empty `VALUE`, which the next line reports as `::error title=Missing production secret::… is absent or has no accessible version` and aborts the promote. Per CLAUDE.md → Error Message Diligence and the #395 incident (a transient 504 mis-reported as a Clerk misconfiguration), this guard's message asserts a root cause it hasn't verified. Failing closed is correct; the *message* is what misleads.
**Suggestion:** Capture stderr and distinguish "definitely absent" from "could not access". Minimal version: drop `2>/dev/null`, capture `err=$(... 2>&1 >/dev/null)` or check `$?`, and word the empty-value branch as "absent **or inaccessible** (could be a transient Secret Manager / IAM error — re-run if so)" rather than asserting absence.

**[documentation]** `docs/deploy.md` (cross-tree, not in diff)
The PR adds two operator-visible deploy behaviors — the promote now fails fast if a prod auth secret is missing/placeholder, and prod web is smoke-tested (`/livez` + `/sign-in`) after deploy. `docs/deploy.md` documents the deploy pipeline and is where the new error messages point operators (Step 4). An operator reading it would not know the promote now self-aborts on a missing secret or that a web smoke gates "success."
**Suggestion:** Add a sentence to `docs/deploy.md` noting the pre-promote auth-secret gate and the prod web smoke, so the runbook matches pipeline behavior.

**[maintainability]** `.github/workflows/deploy.yml` — preflight, `lifting-logbook-prod-clerk-publishable-key`
The publishable key is already validated (`[ -n "$KEY" ]`) in the earlier `build-images` job, where it is baked into the web image as a `NEXT_PUBLIC_*` build-arg; by the time `deploy-production` runs, the image is built. Re-checking it here gates nothing the image build didn't already gate — it is belt-and-suspenders, not load-bearing (the load-bearing check is `clerk-secret-key`, consumed by the web `--update-secrets` at deploy time).
**Suggestion:** Keep it if a single complete "auth secrets present" assertion is the intent (reasonable), or drop it to keep the gate scoped to what `deploy-production` actually consumes. Either is fine — just confirm it's intentional.

---

### Questions for Author

**Question:** Should the smoke target the real customer-facing ingress, or is the Cloud Run URL the intended target?
**Context:** Both probes hit `steps.tf-prod.outputs.cloud_run_web_url` (the `*.run.app` URL). That exercises Google Frontend — so it does catch the #407 `/healthz`-GFE-intercept class — but if prod is fronted by a custom-domain GLB / Cloud Armor, that layer is bypassed. The retro item phrased this as "through the real ingress."
**Tradeoffs:** (a) Keep `cloud_run_web_url` — consistent with the staging smoke and sufficient for the #382/#407 classes (both at the Cloud Run/GFE layer); simplest. (b) Add a probe against the custom domain — closer to "real ingress" and would catch GLB/DNS/cert regressions, but only meaningful if a GLB exists for prod (single-user Cloud-Run-only mode has none) and adds a Terraform-output dependency. For single-user prod today, (a) is correct; (b) becomes worthwhile if/when a fronting GLB is added.

---

### Findings Disposition
- **[reliability] preflight transient-error misdiagnosis:** fix-in-PR (tighten message + capture stderr).
- **[documentation] docs/deploy.md:** fix-in-PR (add a note) or file — to be dispositioned.
- **[maintainability] publishable-key redundancy:** accept (intentional complete-gate) — disposition in PR body.
- **[question] real-ingress target:** accept (a) for single-user prod; note for future GLB.

---

PR #501, opened 2026-06-10
[fix] Resync package-lock.json: Clerk drift reddening main (#500)

brownm09:
## Summary

Closes #500. `main` is red for every open PR. Upstream published `@clerk/backend@3.6.0` and `@clerk/shared@4.16.0`, both accepted by the existing caret ranges in `package.json` (`^3.5.0` / `^4.15.0`), but the committed lockfile still pinned `3.5.0` / `4.15.0`. `npm ci` therefore refuses to install (lock out of sync), failing every PR's required checks at install/setup:

- **Lint & Test** — the lockfile-sync verify step
- **DB Integration Tests** — `npm ci`
- **Build & Push Images** — `npm ci` in the Docker builder
- **Staging Integration Tests** — cascades (build-images skipped → deploy-prereq guard)

This is the [ADR-036](https://github.com/brownm09/dev-env/blob/main/docs/adr/036-lockfile-drift-prevention.md) drift class (same as #432/#433), triggered purely by an upstream publish — no local dependency edit.

## Change
Pure lockfile resync (`npm install`) — bumps the two transitive Clerk packages to the latest their existing caret ranges already allow. **No `package.json` change.**

## Testing
```
Tests: api-client 21, core 199, web 78, api 357 — 12/12 workspace test tasks passed, 0 skipped, 0 failed
```
Run via `turbo run test` (Windows, Node 20.11.1, Docker up) after `npm install` brought `node_modules` to the new versions — confirms the bump is semver-safe.

Additional verification:
- Regenerated lockfile is a **stable fixed point**: a second `npm install --package-lock-only` produces an identical file (md5 match), so the CI drift check (`npm install --package-lock-only && git diff --exit-code`) will pass.
- `package.json` confirmed unchanged (`git diff --exit-code package.json …/package.json`).

Suppression / test-integrity greps against `origin/main`: clean. Lockfile is the only changed file.

## Sequencing
This must merge first — it unblocks the four required checks for every other open PR (#496, #497, #499). Those PRs then update-branch to pick up the synced lockfile.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #474, opened 2026-06-09
[infra] Wire OTel Collector deployment + prod values into the deploy pipeline


brownm09:
## Problem

The OTel Collector chart (`infra/kubernetes/charts/otel-collector/`) exists but **nothing deploys it**: there are no `otel`/`OTEL_COLLECTOR` references in `.github/workflows/deploy.yml`, no `infra/kubernetes/values/{staging,production}.yaml` for the collector (the chart's `env.OTEL_COLLECTOR_OTLP_ENDPOINT` / `OTEL_COLLECTOR_LOKI_ENDPOINT` default to `""`), and the auth Secret is documented as a manual pre-`helm install` step (`templates/NOTES.txt`).

Consequence: after a prod deploy + `mimirtool` alert load, **traces and logs still do not flow to Grafana Cloud** until someone manually creates the `otel-collector-secrets` K8s Secret and runs `helm install` with endpoints supplied. The observability stack is effectively half-wired in prod. Surfaced while answering "where do the Grafana Cloud creds go?" after [#462](https://github.com/brownm09/lifting-logbook/issues/462)/[PR #467](https://github.com/brownm09/lifting-logbook/pull/467).

## Proposed work

1. Add a `deploy` step (GKE path) that `helm upgrade --install`s the otel-collector chart against staging + production namespaces, after the API/web deploys.
2. Add `infra/kubernetes/values/{staging,production}.yaml` (or extend existing) with the per-env `OTEL_COLLECTOR_OTLP_ENDPOINT` (Tempo OTLP) and `OTEL_COLLECTOR_LOKI_ENDPOINT` (Loki push) values, `gcp.projectId`, and the workload-identity SA.
3. Wire the auth-header Secret via External Secrets Operator pulling `otlp-auth-header` / `loki-auth-header` from GCP Secret Manager (Workload Identity), so CI never holds the Grafana token — or, minimally, document the manual `kubectl create secret` bootstrap as the supported path with a CI guard that fails loudly if the Secret is absent.
4. Decide + document the Cloud Run path: the sidecar template `infra/cloud-run/otel-collector-sidecar.yaml` is the single-user/Cloud-Run-only equivalent; clarify which mode wires telemetry how.
5. Confirm the metrics path to Mimir (collector exposes `:8889` for scrape — document what scrapes it / remote_writes in prod, since `APIRouteHighErrorRate` depends on those metrics reaching Mimir).

## Acceptance criteria

[...]

PR #502, opened 2026-06-10
[infra] Scope production API alerts via deployment.environment.name (#487)

### LINKED ISSUE #487: [infra] Add deployment.environment to telemetry + scope alerts (shared Grafana Cloud stack)
## Problem

The free Grafana Cloud tier gives **one stack**, so staging and production both export to the *same* Tempo/Loki/Mimir tenant (the shared-stack path wired in #474 / PR #481). But the API tags telemetry with only `service.name=lifting-logbook-api` and **no environment attribute** (`apps/api/src/otel.ts` sets `serviceName` only). Consequences once #481 deploys:

- **Alerts fire on staging.** `APIRouteHighErrorRate` (and the other rules in `infra/observability/alerts/api.yaml`) group by `http_route`, not environment, so a staging 5xx trips the same rule that pages for prod. This starts the moment #481's **staging** deploy sends metrics — not just at prod.
- **Dashboards + logs intermix** staging and prod with no way to filter.

## Proposed work

1. **Add a `deployment.environment.name` resource attribute** in `apps/api/src/otel.ts`, sourced from `NODE_ENV` (e.g. `staging` / `production`). Consider the web app's `@vercel/otel` instrumentation (`apps/web/instrumentation.ts`) too. Note: OTLP resource attribute `deployment.environment.name` becomes the Mimir label `deployment_environment_name` (dots → underscores); the older `deployment.environment` is deprecated but still what many Grafana dashboards key on — pick one and use it consistently.
2. **Scope the prod alert rules** in `infra/observability/alerts/api.yaml` to the production environment label so staging errors don't page. Verify against the mimirtool load path added in #462/#467.
3. **Optionally** add an environment template variable to the API RED dashboard so staging/prod can be viewed separately.

## Acceptance criteria

- [ ] API (and web, if in scope) traces/logs/metrics carry an environment attribute distinguishing staging from production in the shared stack.
- [ ] Prod paging rules fire only on production telemetry; a staging 5xx does not page.
- [ ] Regression test / assertion that the resource attribute is set (the otel.ts change is testable).

## Context

Follow-up to #474 / PR #481, which deliberately kept the single shared stack (free tier) and reused one token for both envs. The environment-discrimination work was split out so the infra-wiring PR stayed focused. Decided with the user on 2026-06-09.

## References
- #474, PR #481
- [OTel — deployment.environment.name](https://opentelemetry.io/docs/specs/semconv/resource/deployment-environment/)
- [ADR-018 observability stack](../blob/main/docs/adr/ADR-018-observability-stack.md), [ADR-020 tail sampling](../blob/main/docs/adr/ADR-020-tail-based-sampling-policy.md)

brownm09:
## Summary

Staging and production share **one** free-tier Grafana Cloud stack. Now that the OTel collector wired in [#474](https://github.com/brownm09/lifting-logbook/issues/474) / PR #481 is merged, the next staging deploy starts exporting metrics — and the `http_route`-grouped production alert rules (`APIRouteHighErrorRate`, etc.) would page on a **staging** 5xx because nothing distinguished the two environments. This adds an environment discriminator and scopes the alerts to production, end-to-end.

The load-bearing subtlety: OTLP→Prometheus puts **resource** attributes on `target_info`, not on each metric series, so an alert label-matcher on a resource attribute alone would silently match nothing. The collector therefore promotes the attribute to a real metric label.

### Changes
- **`apps/api/src/otel.ts`** — attach a `deployment.environment.name` resource attribute from `NODE_ENV` (`production`/`staging`; `development` locally) via a pure, unit-tested `resolveDeploymentEnvironment()` helper. Adds `@opentelemetry/resources` (^2.7.1, matches the SDK).
- **`apps/web/instrumentation.ts`** — same attribute on `@vercel/otel` `registerOTel({ attributes })`.
- **`infra/kubernetes/charts/otel-collector/templates/configmap.yaml`** — a `transform/env_label` processor promotes the resource attribute to a `deployment_environment_name` metric datapoint label in the **metrics** pipeline. Traces/logs keep the resource attribute natively in Tempo/Loki.
- **`infra/observability/alerts/api.yaml`** — all four rules match `deployment_environment_name="production"`.
- **`infra/observability/alerts/api.test.yaml`** — labelled existing series + a new **staging-does-not-page** scenario.
- **`apps/api/jest.env.setup.js`** — disable OTel SDK autostart in tests ([ADR-021](docs/adr/ADR-021-no-test-tracing.md)) so importing `otel.ts` from a spec doesn't spin up exporters / a metric-reader interval.
- **`docs/runbooks/observability.md`** — document the env-scoping flow; the alert table is now marked production-scoped.

## Acceptance criteria
- [x] API (and web) traces/logs/metrics carry an environment attribute distinguishing staging from production in the shared stack.
- [x] Prod paging rules fire only on production telemetry; a staging 5xx does not page (locked by the new promtool scenario).
- [x] Regression test asserting the resource attribute is set (`apps/api/src/otel.spec.ts`).

## Testing
- `npm test -w @lifting-logbook/api` (Docker up, DB E2E via Testcontainers ran for real) — **Tests: 363 passed, 0 skipped, 0 failed (29.6s)**, 27 suites. Includes new `otel.spec.ts` (6) and Clerk-auth E2E paths.
- `npm test -w @lifting-logbook/web` — **Tests: 78 passed, 0 skipped, 0 failed (39.8s)**, 14 suites.
- `promtool test rules infra/observability/alerts/api.test.yaml` (via `prom/prometheus`) — **SUCCESS** (incl. the staging-does-not-page scenario).
- `helm template infra/kubernetes/charts/otel-collector` — renders the `transform/env_label` processor and metrics pipeline cleanly.
- `apps/web/instrumentation.ts` typechecks cleanly under `apps/web/tsconfig.json`; confirmed `@vercel/otel` `Configuration.attributes?: ResourceAttributes` accepts the option (and the runtime spreads it into the resource).

### Coverage
New testable behavior (`resolveDeploymentEnvironment`) covered by `otel.spec.ts`; alert env-scoping covered by the promtool staging scenario. Collector `transform` and web `registerOTel` attributes verified via helm render + typecheck (no unit-test harness for those layers).

### Suppression / test-integrity / lockfile gates
- Suppression grep: **clean**. Test-integrity grep: **clean**. No test files deleted, no thresholds lowered.
- **Lockfile note:** the diff adds `@opentelemetry/resources` and incidentally re-resolves `@clerk/backend` 3.5.0→3.6.0 and `@clerk/shared` 4.15.0→4.16.0 — **minor** bumps within the existing `^` ranges in `package.json` (no manifest change), pulled in by npm's natural re-resolution (latent drift already on `main`). The pre-push lockfile-drift hook requires the fully-resolved lockfile, so these ride along; both Clerk-auth E2E paths pass in the 363-test run. ([ADR-036](docs/adr/ADR-036-lockfile-drift-prevention.md))

Closes #487

---

`/review` (claude-opus-4-8) returned **0 blocking, 4 non-blocking**. Verified the two silent-failure risks first: `otel.ts` autostart is gated on `OTEL_SDK_AUTOSTART !== 'false'` (jest setup line accurate), and `NODE_ENV` is set to `production`/`staging` in the live GKE values (`production-api.yaml:11`, `staging-api.yaml:15`) so prod metrics carry the label and the rules won't no-op. All four findings addressed:

- **[observability] RED dashboard unscoped** — fixed in `2014338`. Added an `environment` template var (default `production`) to `api-red.json` and scoped all five panels with `deployment_environment_name="$environment"`; same root cause as the alert scoping (staging now exports to the shared Mimir).
- **[maintainability] duplicated `resolveDeploymentEnvironment`** — fixed in `2014338`. Added reciprocal sync-note comments in `apps/api/src/otel.ts` (canonical) and `apps/web/instrumentation.ts` (mirror). Kept the intentional duplication rather than introduce a cross-package dependency for a three-line env read.
- **[maintainability] APIHighP95Latency had no promtool coverage** — fixed in `2014338`. Added a latency scenario: production buckets resolve p95 to 2s (fires), identical staging buckets stay silent. `promtool test rules` → SUCCESS.
- **[reliability] incidental Clerk lockfile bump (merge-order hazard with #501)** — operational, no code change. **Merge [#501](https://github.com/brownm09/lifting-logbook/pull/501) first**, then `gh pr update-branch` on this PR so its lockfile delta collapses to just the `@opentelemetry/resources` addition and the Clerk lines stop conflicting.
- **[style] misleading APINoRequests comment** — fixed in `2014338`. Reworded: with `for: 10m` it is pending (not firing) at the scenario's 6m horizon.

Re-verification after fixes: `promtool test rules api.test.yaml` → **SUCCESS** (now incl. the latency scenario); dashboard JSON validates; `turbo run lint` clean across 7 packages on commit. otel.ts / instrumentation.ts changes are comment-only.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -690,6 +690,60 @@ jobs:
           workload_identity_provider: ${{ secrets.GCP_PROD_WORKLOAD_IDENTITY_PROVIDER }}
           service_account: ${{ secrets.GCP_PROD_SERVICE_ACCOUNT }}
 
+      # ── Pre-promote auth-secret validation (#490 item 2 / #382) ────────────
+      # Fail fast BEFORE any production mutation (terraform apply, migrate, Cloud
+      # Run / Helm deploy) if a serve-gating auth secret is absent or still the
+      # Terraform placeholder. GCP Secret Manager is already the single source of
+      # truth — every surface (Cloud Run --update-secrets, the k8s Secret sync,
+      # the web build-arg) fetches from it rather than holding its own copy — so
+      # the only failure mode left is "the source value was never populated".
+      # #382 was exactly that: a missing prod clerk-secret-key, caught only after
+      # the web service was already serving auth 500s in prod. This surfaces it at
+      # promote time instead. It aggregates all failures (does not stop at the
+      # first) and never echoes a secret value. Scope is the two auth secrets
+      # whose absence breaks the served surface; the DB-url and OTel-header
+      # secrets keep their own per-step `[ -n ]` guards later in the job.
+      #
+      # Error-message diligence (CLAUDE.md; the #395 transient-504 incident): a
+      # non-zero gcloud exit is NOT assumed to mean "secret absent". stdout (the
+      # secret) is captured to a shell var and stderr to a file, then the failure
+      # is classified — only a NOT_FOUND error is reported as missing; anything
+      # else is reported as "inaccessible (transient or IAM), re-run if transient".
+      # The secret value itself never touches disk; only the error text does.
+      - name: Validate production auth secrets (pre-promote)
+        run: |
+          MISSING=0
+          ERR_FILE="$RUNNER_TEMP/secret_stderr"
+          for SECRET in \
+            lifting-logbook-prod-clerk-secret-key \
+            lifting-logbook-prod-clerk-publishable-key; do
+            if VALUE=$(gcloud secrets versions access latest \
+              --secret="$SECRET" \
+              --project="${{ vars.GCP_PROD_PROJECT_ID }}" 2>"$ERR_FILE"); then
+              if [ -z "$VALUE" ]; then
+                echo "::error title=Empty production secret::$SECRET resolved to an empty value. Set the real value in GCP Secret Manager (docs/deploy.md Step 4) before promoting."
+                MISSING=1
+              elif [ "$VALUE" = "REPLACE_ME" ]; then
+                echo "::error title=Placeholder production secret::$SECRET is still the Terraform placeholder (REPLACE_ME). Set the real value in GCP Secret Manager (docs/deploy.md Step 4) before promoting."
+                MISSING=1
+              fi
+            else
+              ERR=$(cat "$ERR_FILE" 2>/dev/null || true)
+              if printf '%s' "$ERR" | grep -qi "NOT_FOUND"; then
+                echo "::error title=Missing production secret::$SECRET does not exist in GCP Secret Manager. Create and populate it (docs/deploy.md Step 4) before promoting."
+              else
+                echo "::error title=Production secret inaccessible::Could not read $SECRET — possibly a transient Secret Manager error or an IAM/permission problem, NOT necessarily a missing secret. Re-run if transient; otherwise verify the prod CI/CD service account's secretmanager.versions.access grant. Detail: $ERR"
+              fi
+              MISSING=1
+            fi
+          done
+          rm -f "$ERR_FILE"
+          if [ "$MISSING" -ne 0 ]; then
+            echo "Production auth-secret preflight failed — aborting before any production mutation. See the errors above." >&2
+            exit 1
+          fi
+          echo "Production auth secrets present and non-placeholder — proceeding."
+
       - name: Set up Terraform
         uses: hashicorp/setup-terraform@v4
         with:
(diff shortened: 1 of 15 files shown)


brownm09:
## Review: [infra] Scope production API alerts via deployment.environment.name (#487)

### Summary
**Merge with minor changes.** The change is correct and well-scoped: the end-to-end chain (SDK resource attr → collector `transform` promotes it to a metric label → alert rules match `deployment_environment_name="production"`) is sound, and the two failure modes that would have made it *silently* wrong are both safe — `otel.ts:74` genuinely gates import-time autostart on `OTEL_SDK_AUTOSTART !== 'false'` (so the new jest setup line is accurate), and `NODE_ENV` is actually set to `production`/`staging` in the live GKE values (`production-api.yaml:11`, `staging-api.yaml:15`), so prod metrics will carry the label and the rules won't no-op. No blocking findings. The highest-value follow-up is that this PR scopes the **alert rules** to production but leaves the **RED dashboard** (`api-red.json`) querying the same metrics unscoped — so dashboards will now blend staging + prod traffic, which is the same root cause #487 addresses.

---

### Non-Blocking Findings

**[observability]** `infra/observability/dashboards/api-red.json:35,54,73,92,111`
All five RED panels query `http_server_request_duration_seconds_count` / `_bucket` with **no** `deployment_environment_name` filter. The whole premise of this PR is that staging now exports to the same Mimir; the moment it does, the "production" RED dashboard's request-rate, error-rate, and p50/p95/p99 panels silently fold in staging traffic. An operator reading these during a prod incident would see contaminated numbers — the alerts would be correct but the dashboard they click through to would not.
**Suggestion:** Add `deployment_environment_name="production"` to the five panel `expr`s in this PR (it's the same one-token edit as the alert rules and the same root cause), or — if you'd rather keep this PR alert-only — file a follow-up issue and link it in the PR body so it's tracked per ADR-028. A `$environment` template variable would be the more flexible version if you go in-PR.

**[maintainability]** `apps/web/instrumentation.ts:6-9`
The web `resolveDeploymentEnvironment()` is a verbatim reimplementation of the API helper in `apps/api/src/otel.ts:26-30`, but inline and with no regression test, while the API copy is exported and covered by `otel.spec.ts`. If the fallback semantics ever change (e.g. treat `test` specially, or change the default), the two will drift silently and web telemetry will be mislabelled without any test catching it.
**Suggestion:** Duplicating a 3-line env read across two workspaces is defensible (a shared util would mean a new cross-package dependency), so this is a judgment call — but at minimum note in the PR/code that the two must stay in sync, or add a one-line web test mirroring the API assertion.

**[maintainability]** `infra/observability/alerts/api.yaml:44-47`
`APIHighP95Latency` received the new `deployment_environment_name="production"` matcher on its `_bucket` query, but `api.test.yaml` has no scenario exercising this rule at all (it covers the two error-rate rules and `APINoRequests`). The label addition relies on the collector promoting the attribute onto histogram datapoints, which is reasoned-about but not test-locked here.
**Suggestion:** This is a pre-existing coverage gap (no latency scenario existed before), so it's fair to defer — but adding a small promtool scenario for the latency rule (prod-labelled buckets fire, staging-labelled don't) would close it and lock the histogram-label assumption.

**[reliability]** `package-lock.json:2600-2620`
The PR incidentally re-resolves `@clerk/backend` 3.5.0→3.6.0 and `@clerk/shared` 4.15.0→4.16.0 (latent drift on `main`, pulled in because the pre-push hook requires a fully-resolved lockfile). This is the same delta as the dedicated drift-fix PR #501, so whichever lands second hits a `package-lock.json` conflict on these lines. The PR body already documents this.
**Suggestion:** Merge #501 first, then `gh pr update-branch` on this PR — after rebasing onto a `main` that already has 3.6.0/4.16.0, this PR's lockfile delta collapses to just the `@opentelemetry/resources` addition. Record this in the findings disposition so the sequencing isn't lost.

---

### Style Notes
The scenario-3 comment in `api.test.yaml:73-74` states "APINoRequests does fire (info, non-paging)" but no assertion verifies it — the prose describes behaviour the test doesn't lock. Either add the `exp_alerts` assertion for `APINoRequests` in that scenario or soften the comment to "is not asserted here" so the comment and the test agree.

---

### Findings Disposition
Per ADR-028, each finding above must be fixed in this PR or filed-and-linked before merge. Suggested dispositions: (1) dashboard — fix in-PR (trivial, same root cause) or file-and-link; (2) duplicate helper — add sync note / web test in-PR, or accept with a code comment; (3) latency test — defer with a tracking note (pre-existing gap) or add the scenario; (4) lockfile/#501 sequencing — operational, record the merge order. Record each as `fixed in <sha>` or `filed #N` in a "Review findings disposition" PR-body section before `gh pr merge`.

---

Issue #504, opened 2026-06-10
[infra] Apply AR-504 / terraform-apply bounded retries to deploy.yml (prod parity with #498)


brownm09:
## Problem

The #498 diagnosis (see staging.yml fix) established that the real driver of staging CI flakiness
is **transient Artifact Registry 504s on image layer-blob push** and **Terraform apply
transients** — not Cloud SQL connectivity. The staging fix wrapped those steps in bounded
retries (`Wandalen/wretry.action` on the two `docker/build-push-action` steps; a shell retry loop
on `terraform apply`).

`deploy.yml` (the push-to-`main` prod + staging promote pipeline) has the **same**
`docker/build-push-action` image build/push steps and the same `terraform apply` step, so it is
exposed to the identical transient failure classes — but did not receive the retry treatment in
the #498 PR (scoped to `staging.yml` per the ~75% scope guard).

## Proposed change

Apply the same bounded-retry pattern to `deploy.yml`:

- Wrap the API and web `docker/build-push-action@v6` build/push steps with
  `Wandalen/wretry.action` (pinned to the same SHA used in `staging.yml`,
  `e68c23e6309f2871ca8ae4763e7629b9c258e1ea` / v3.8.0), preserving each step's existing `with`
  config and per-image GHA cache scoping (#407).
- Wrap `terraform apply` (prod and/or staging promote) in a bounded shell retry.

## Acceptance criteria

- [ ] `deploy.yml` build/push steps retry transient AR 504s (bounded, cached layers keep retries cheap).
- [ ] `deploy.yml` `terraform apply` retries transient GCP API 5xx.
- [ ] No change to prod deploy *failure* semantics for genuine errors (retries exhaust → hard-fail).

## References

- #498 (staging diagnosis + fix)
- `docs/runbooks/staging-ci-flakiness.md`
- #395 (the misleading-error-chain incident this hardens against)

PR #499, opened 2026-06-10
[infra] Re-point staging Cloud SQL flakiness refs from closed #344 to #498 (#490)

brownm09:
## Summary

Closes the tracking gap for **issue #490 item 4 — staging integration-test Cloud SQL flakiness**.

`staging.yml` had 7 references describing the staging Cloud SQL CI flakiness as "tracked in #344" / "Cloud SQL flakiness (#344)". But [#344](https://github.com/brownm09/lifting-logbook/issues/344) is **closed** and was about a *different* problem — GKE Autopilot pod readiness (fixed via the Recreate strategy). An operator following those references reached the wrong, closed issue. This re-points all 7 to the correctly-scoped open tracker, [#498](https://github.com/brownm09/lifting-logbook/issues/498), which carries the specific GCP root-cause investigation (instance tier, VPC connector, connection-pool limits, log correlation) that requires runtime access not available in a code session.

This is the "refine the tracking" half of item 4; the actual root-cause fix lives in #498 and needs GCP runtime.

## Acceptance criteria
- [x] All 7 `#344` references in `staging.yml` now point to `#498`
- [x] No remaining `#344` / `issues/344` references in the workflow
- [x] No pipeline behavior change (comment + echo-string text only)

## Testing
**Comment/echo-string-only change** to a CI workflow file — it alters issue-number references inside comments and `echo`/`body` strings, touching no job logic, no application/build/test code. The Jest suite does not exercise CI YAML. Verification:
- `grep` confirms 0 remaining `#344`/`issues/344` and 7 `#498` references.
- The workflow still parses as valid YAML (loaded via js-yaml).

No automated test is applicable; there is no behavior to cover.

Suppression / test-integrity greps against `origin/main`: clean. No `package.json`/lockfile changes.

## Scope
Part of umbrella **#490** (item 4 of 4 — the tracking half). The root-cause investigation/fix is tracked in #498.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #498, opened 2026-06-10
[infra] Staging integration-test flakiness: diagnose & fix Cloud SQL connectivity root cause


brownm09:
## Problem

The staging pipeline (`.github/workflows/staging.yml`) marks the in-VPC Prisma migrate job and the Cloud Run API deploy `continue-on-error: true` because staging **Cloud SQL connectivity is intermittently flaky during CI** — the migrate job and API container startup sometimes can't reach the Cloud SQL instance, producing transient failures the pipeline tolerates (the Staging Integration Tests are the authoritative gate). The backfilled biweekly retro (2026-06-09, engineering-journal) flagged that while the *symptom* is tolerated, the *root cause is untracked* (~27 CI runs to green in one window).

The code currently cites **#344** for this ("Cloud SQL connectivity issues tracked in #344", 6 occurrences in `staging.yml`), but [#344](https://github.com/brownm09/lifting-logbook/issues/344) is **closed** and was about a *different* problem — GKE Autopilot pod readiness, fixed via the Recreate strategy. So the Cloud SQL CI flakiness has had **no correctly-scoped open tracker**. This issue is that tracker.

## Why it matters

- `continue-on-error` on the migrate + deploy steps means a *genuine* migration/connectivity failure can be masked. It is mitigated by the loud warning annotation + the integration-test gate, but remains a real risk (see [ADR-027](docs/adr/ADR-027-deploy-pipeline-migrations.md) and the #395 error-message-diligence incident).
- Flaky CI erodes signal and wastes runner minutes on re-runs.

## Root-cause investigation (needs GCP runtime access — not reproducible locally)

1. **Instance tier / activation policy** — confirm the staging Cloud SQL instance is not auto-pausing or under-provisioned (shared-core cold-start latency): `gcloud sql instances describe <stg-instance> --project=$STG`.
2. **Connector path** — staging API + migrate job reach Cloud SQL via the in-VPC Serverless VPC Access connector. Check connector throughput / min-instances and whether connector cold-starts coincide with failures: `gcloud compute networks vpc-access connectors describe ...`.
3. **Connection limits / pool** — inspect Cloud SQL `max_connections` vs concurrent Cloud Run instances + the migrate job (Prisma opens its own pool). Look for `too many connections` / `connection reset` in Cloud SQL logs at failure timestamps.
[...]

PR #505, opened 2026-06-10
[infra] Diagnose staging CI flakiness; tighten migrate, retry AR/terraform (#498)

brownm09:
## Summary

Issue #498 was filed to track an *assumed* "intermittent staging Cloud SQL connectivity flakiness" — the rationale for `continue-on-error` on the staging migrate + Cloud Run deploy steps. A read-only GCP diagnosis (`gcloud`, `lifting-logbook-staging`) **does not corroborate that premise**, and this PR re-scopes the fix accordingly (Error Message Diligence; the issue's own AC allowed "an explicit decision … with documented rationale").

### Diagnosis (full evidence: [#498 comment](https://github.com/brownm09/lifting-logbook/issues/498#issuecomment-4673917094))
- **Cloud SQL is healthy**: instance `db-f1-micro` ALWAYS-on/RUNNABLE; connector `min-instances=2` (always warm); **0** migrate task-failures across the last 46 executions; all 10 recent API revisions healthy; **no** connection-exhaustion / auth / connect-timeout errors in 7 days of logs. Only benign INFO disconnects, one planned 06-07 maintenance restart, and `relation … does not exist` **schema drift** (the class the migrate `continue-on-error` could *mask*).
- **Real red-run driver**: transient **Artifact Registry 504s on layer-blob push** + **Terraform apply** transients, cascading into the `Verify deploy prerequisites` guard (the #395 misleading-error chain). The workflow's own guard messages already named AR-504.

### Changes
- **Migrate step tightened** (`staging.yml`): removed `continue-on-error`; added a bounded **3-attempt retry** (covers both the job-image `update` and the `execute`, both idempotent) — transient blips absorb, genuine migration/schema failures **hard-fail**. Removed the now-moot "Surface swallowed migration failure" step.
- **Bounded retries on the real driver**: `Wandalen/wretry.action` (pinned to commit `e68c23e…`, v3.8.0) wrapping both `docker/build-push-action@v6` steps (config + #407 cache scoping preserved); a 2-attempt shell loop on `terraform apply`.
- **Reworded** residual `#498` narrative refs to reflect the diagnosis (image/startup + AR, not connectivity). Kept `continue-on-error` on the Cloud Run API deploy step solely for crash-log capture.
- **Docs**: ADR-027 addendum (migrate `continue-on-error` reversal) + new `docs/runbooks/staging-ci-flakiness.md`.

### Acceptance criteria
- [x] Root cause identified & documented — Cloud SQL is *not* the cause; real driver is AR-504 + terraform transients ([evidence](https://github.com/brownm09/lifting-logbook/issues/498#issuecomment-4673917094), runbook).
- [x] Fix/mitigation applied — migrate hard-fails after bounded retry; AR/terraform retries added.
- [x] `#344` refs re-pointed to `#498` — already shipped in 8786bc8 (#499); zero `#344` refs remain.

### Out of scope (follow-up)
- #504 — apply the same AR-504 / terraform-apply retries to `deploy.yml` (prod parity), held out per the ~75% scope guard.

## Testing
Change is **workflow YAML + docs only** (no app code) → per the project coverage table, refactor/docs-only requires no new tests; existing tests must pass.
- `turbo run lint`: **7 successful, 7 total** (FULL TURBO) — passes.
- `staging.yml` validated: parses as YAML (python `yaml.safe_load`) and the `wretry` nested `with:` blocks parse to the correct forwarded `docker/build-push-action` inputs (incl. web `build-args`).
- Full `npm test` (Jest, Testcontainers/Docker) not run: the diff touches no TypeScript/JS and no test files; the suite covers none of the changed surface. **Authoritative end-to-end verification is this PR's own `staging.yml` run** (the workflow runs on PRs, exercising every edited step).
- Pre-PR suppression + test-integrity greps: clean. No `package.json`/lockfile change (lockfile-sync check N/A). Rebased onto #507 (clerk lockfile resync) so CI install is unblocked.

- **[reliability] migrate `update` outside the retry loop** ([#505 review](https://github.com/brownm09/lifting-logbook/pull/505#issuecomment-4674015525)) — **Fixed in `234d91c`**: moved `gcloud run jobs update` inside the bounded retry loop (chained with `execute` via `&&`; both idempotent), so a transient blip on the update is also retried.
- **[question] `wretry` + GHA `type=gha` cache export under the wrapper** — **Verified via this PR's own staging `Build & Push Images` log** (checked for cache export working / no `ACTIONS_CACHE_URL` warning). Disposition: confirmed working — no code change needed. (If the log had shown broken cache export, fallback was a raw `docker buildx --push` retry loop.)
- *Behavior-change note (not a finding):* removing the migrate `continue-on-error` means a genuine migrate failure now fails the `deploy-api` job (previously masked); intended — `Staging Integration Tests` retains `if: always()` and still gates. Documented in the step comment + ADR-027 addendum.

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -131,10 +131,24 @@ jobs:
 
       - name: Terraform apply (staging)
         working-directory: ${{ env.TF_DIR }}
+        # Bounded retry absorbs transient GCP API 5xx / eventual-consistency
+        # during apply (a real staging red-run driver per the #498 diagnosis —
+        # see docs/runbooks/staging-ci-flakiness.md). terraform apply is
+        # re-runnable; a genuine config error fails both attempts and hard-fails.
         run: |
-          terraform apply -auto-approve \
-            -var-file=terraform.tfvars.staging \
-            -var="billing_account=${{ secrets.GCP_BILLING_ACCOUNT }}"
+          for attempt in 1 2; do
+            if terraform apply -auto-approve \
+                 -var-file=terraform.tfvars.staging \
+                 -var="billing_account=${{ secrets.GCP_BILLING_ACCOUNT }}"; then
+              exit 0
+            fi
+            if [ "${attempt}" -lt 2 ]; then
+              echo "Terraform apply attempt ${attempt} failed; retrying in 20s..."
+              sleep 20
+            fi
+          done
+          echo "::error title=Terraform apply failed::terraform apply (staging) failed on both attempts."
+          exit 1
 
       - name: Capture Terraform outputs
         id: tf-outputs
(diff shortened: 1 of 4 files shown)


brownm09:
## Review: [infra] Diagnose staging CI flakiness; tighten migrate, retry AR/terraform (#498)

### Summary
**Merge with minor changes.** Well-scoped (258 lines, workflow + docs only; no app code). The diagnosis-driven re-scoping is sound: removing the migrate `continue-on-error` stops masking schema-drift while a bounded idempotent retry preserves resilience, and the AR-504/terraform retries target the actual red-run driver. Highest-severity item is non-blocking: the `gcloud run jobs update` line sits *outside* the migrate retry loop, so a transient blip on that call still hard-fails without a retry. One verification question on `wretry` + GHA cache env propagation that the PR's own CI run will settle.

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/staging.yml` (migrate step, ~358–392)
The bounded retry wraps only `gcloud run jobs execute --wait`; the preceding `gcloud run jobs update … --image=…` runs once, outside the loop, under the step's default `set -eo pipefail`. If a transient GCP API blip (the same class this PR hardens against) hits the `update` call, the step hard-fails with no retry — the masking is gone, so that transient would now red the run.
**Suggestion:** Move the `gcloud run jobs update` inside the retry loop (it is idempotent — same image ref each attempt), so both calls get the bounded retry. Fixed in this PR.

---

### Questions for Author

**Question:** Does `docker/build-push-action`'s `type=gha` cache export/import still work when the action is wrapped by `Wandalen/wretry.action`?
**Context:** `.github/workflows/staging.yml` (both build steps) now invoke `docker/build-push-action@v6` indirectly via `wretry.action`. GHA cache (`cache-to: type=gha`) relies on `ACTIONS_RUNTIME_TOKEN` / `ACTIONS_CACHE_URL` being present in the action's runtime env; a wrapper layer occasionally drops these, which fails cache *export* with only a warning (the build still succeeds but uncached — silently slower, and a retry then rebuilds from scratch, undercutting the "cached layers make retries cheap" rationale).
**Tradeoffs:** (a) Keep `wretry` — preserves the exact build-push-action config and the #407 cache scoping; if CI confirms cache export works, this is the lowest-churn option. (b) Fall back to a raw `docker buildx build --push` retry loop — full control over env, but must re-inject `ACTIONS_RUNTIME_TOKEN`/`ACTIONS_CACHE_URL` (a known `type=gha` gotcha) and re-express the cache flags. Recommend (a), verified by inspecting this PR's staging build log for an `exporting cache` line and the absence of an `ACTIONS_CACHE_URL` warning; switch to (b) only if export is broken.

---

### Style Notes
Backoff intervals differ across the three retried steps (migrate: 15s→30s linear; terraform: fixed 20s; wretry: 15000ms). Each is reasonable for its step, so no change needed — noting only for consistency awareness.

---

### Findings Disposition
- **[reliability] migrate `update` outside retry** — to be **fixed in this PR** (move `update` inside the loop); sha recorded in PR body.
- **[question] wretry + GHA cache** — **verified via this PR's staging CI build log** (cache-export check); disposition recorded in PR body. No code change unless CI shows export broken.

Behavior-change note (not a finding): removing the migrate `continue-on-error` means a genuine migrate failure now fails the `deploy-api` job (previously masked). This is intended — the `Staging Integration Tests` job retains `if: always()` so it still runs and gates — and is documented in the step comment + the ADR-027 addendum.

---

PR #515, opened 2026-06-11
[feat] Inject apps/web public config at runtime (supersedes ADR-025)

brownm09:
## Summary

Phase 2 of the ADR-025 work (#388). Eliminates build-time `NEXT_PUBLIC_*` embedding in `apps/web` and injects public config at **runtime**, so the web image builds **once** and promotes everywhere — restoring the byte-identical promotion contract ADR-025's per-env builds gave up.

**Design (new ADR-028, supersedes ADR-025):** the root layout (a Server Component, `force-dynamic`) reads runtime env and injects it two ways:
- An inline `<head>` `<script>` sets `window.__PUBLIC_CONFIG__` before hydration (consumed by the non-React `lib/client-api.ts`), also passed to a `PublicConfigProvider` for `usePublicConfig()`.
- `<ClerkProvider publishableKey={process.env.CLERK_PUBLISHABLE_KEY}>` — Clerk's runtime-key prop pattern (unblocks the synchronous client mount).

`packages/api-client` `baseUrl` now accepts a `() => string` thunk; the browser client resolves it lazily per request.

**Env-var renames** (drop `NEXT_PUBLIC_` so Next.js does not inline them):
`NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` → `CLERK_PUBLISHABLE_KEY`, `NEXT_PUBLIC_API_URL` → `PUBLIC_API_URL` (browser-facing; distinct from server-side `API_URL`), `NEXT_PUBLIC_DEFAULT_PROGRAM` → `DEFAULT_PROGRAM`, and `NEXT_PUBLIC_DEV_AUTH_TOKEN` folded into the existing `DEV_AUTH_TOKEN`.

**Pipeline:** `deploy.yml` collapses the two per-env web builds + per-env value-resolution steps into a single `web:<sha>` build; both `deploy.yml` and `staging.yml` inject the values at runtime on the Cloud Run/GKE deploy (alongside the existing `CLERK_SECRET_KEY`). Declarative infra (Cloud Run YAML, Terraform, GKE web chart) updated for parity. `apps/web/Dockerfile` `ARG NEXT_PUBLIC_*` removed.

> **Local dev note:** untracked `apps/web/.env` / `.env.local` files must rename `NEXT_PUBLIC_API_URL`→`PUBLIC_API_URL`, `NEXT_PUBLIC_DEFAULT_PROGRAM`→`DEFAULT_PROGRAM`, `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY`→`CLERK_PUBLISHABLE_KEY`, and drop `NEXT_PUBLIC_DEV_AUTH_TOKEN` (`.env.example` is already updated).

## Acceptance criteria

1. ✅ Deploy pipeline builds the web image exactly once per commit (single `web:<sha>` build).
2. ✅ `web:<sha>` deployed to staging and production serves env-specific values, injected at runtime via `--set-env-vars`/`--update-secrets` (Cloud Run) and ConfigMap/Secret (GKE).
3. ✅ New ADR-028 records the runtime-injection design and explicitly supersedes ADR-025 (status flipped, README/adr-references updated).
4. ✅ Verification rewritten in `docs/deploy.md` → "Verifying runtime public config": the bundle now carries **no** embedded public values for **either** env (it's injected at runtime), plus a `window.__PUBLIC_CONFIG__` runtime check.

## Testing

`npm run build --filter=@lifting-logbook/web` succeeds with **no** build-args and no Clerk key; every route reports `ƒ (Dynamic)`, confirming `force-dynamic` removed the build-time prerender dependency.

Per-workspace results (Docker up; full API DB E2E ran):
- `@lifting-logbook/api-client`: **Tests: 22 passed, 0 skipped, 0 failed** — adds thunk-`baseUrl` resolution test.
- `@lifting-logbook/core`: **Tests: 199 passed, 0 skipped, 0 failed**.
- `@lifting-logbook/web`: **Tests: 88 passed, 0 skipped, 0 failed** — adds `public-config` (server assembly, script escaping, window read, provider/hook) and `client-api` (runtime base URL + dev-token gating) suites.
- `@lifting-logbook/api`: **Tests: 366 passed, 0 skipped, 0 failed** (Testcontainers Postgres).

New testable behavior is covered (coverage gate satisfied). Pre-PR suppression and test-integrity greps: clean. No `package.json`/lockfile changes. Playwright `apps/web/e2e` not run locally in this session — env-var renames are mirrored in `playwright.config.ts`; CI's Playwright job exercises it.

## Risk audit

- **Security:** `CLERK_PUBLISHABLE_KEY`/`PUBLIC_API_URL` are non-secret public values; `devAuthToken` is emitted only when `DEV_AUTH_TOKEN` is set (never in deployed envs). No secret enters `window.__PUBLIC_CONFIG__`. Inline-script JSON escapes `<` to prevent `</script>` breakout (tested).
- **Resilience/rollback:** promote-everywhere makes the staging gate exercise the exact prod artifact. Revert = restore two builds. Prod pre-promote gate already validates the publishable-key secret.
- **Data integrity/migrations:** N/A — no schema/data changes.

## Review findings disposition (`/review`, 2026-06-11)

Posted review: blocking=0, non_blocking=6. All addressed — none left as-is.

- **[reliability] localhost/undefined config fallback masks a deploy misconfig** — Fixed in `2d18945`: `readServerPublicConfig` now throws in a deployed runtime (`NODE_ENV==='production'`, not the `next build` phase) when `PUBLIC_API_URL` is unset/empty, refusing the localhost fallback (the silent-misconfig class behind #395/#458). 3 guard tests added (throws unset, throws empty-string, build-phase exempt).
- **[reliability] `staging.yml` `--update-env-vars` vs `deploy.yml` `--set-env-vars`** — Fixed in `2d18945`: `staging.yml` web deploy now uses `--set-env-vars` (replace), so the deploy command is the single source of truth for the env set; comment added.
- **[maintainability] `DEFAULT_PROGRAM` documented as runtime var but wired nowhere** — Fixed in `2d18945`: annotated as code-default-only in deployed envs in ADR-028 and both `.env.example` files (matches prior `NEXT_PUBLIC_DEFAULT_PROGRAM`, never set in build-args).
- **[reliability] `docs/deploy.md` verification grep used a `run.app` literal** — Fixed in `2d18945`: greps the resolved API hostnames from `gcloud run services describe` (a custom-domain/GKE host would otherwise slip past); replaced `/tmp` with a workdir `mktemp`.
- **[performance] `force-dynamic` root blast radius** — Verified, no code change needed: no route uses `generateStaticParams` and the only other `dynamic` exports are the already-dynamic `/livez` and `/api/healthz` routes, so no previously-static route loses optimization. Documented in ADR-028 Consequences. CI staging deploy + Playwright + integration are green on per-request SSR.
- **[maintainability] `PublicConfigProvider` has no product consumer yet + CI retains a `pk_test_` placeholder** — By design, no change: `PublicConfigProvider`/`usePublicConfig` is the ADR-028-documented React delivery channel (covered by a test); the CI/Playwright `pk_test_` placeholder is a harmless valid-format value kept belt-and-suspanders.

Fix commit: `2d18945`. Local re-verification: `@lifting-logbook/web` **91 passed, 0 skipped, 0 failed** (+3 guard tests); `npm run build --filter=@lifting-logbook/web` green (all routes `ƒ (Dynamic)`); lint clean; suppression + test-integrity greps clean.


brownm09:
## Review: [feat] Inject apps/web public config at runtime (supersedes ADR-025)

### Summary
**Merge with minor changes.** This is a clean, well-scoped refactor that restores build-once / promote-everywhere by moving the three public config values off build-time `NEXT_PUBLIC_*` inlining onto runtime injection. Two independent analysis passes (correctness/security and reliability/perf/maintainability) found **no blocking defects**: the inline-script escaping is sufficient for the `<script>` context, no secret can reach `window.__PUBLIC_CONFIG__`, `devAuthToken` is correctly gated to `DEV_AUTH_TOKEN`-set envs only, no residual `NEXT_PUBLIC_*` runtime reads remain, and the GKE `PUBLIC_API_URL` is correctly the external endpoint (not the cluster-internal `API_URL`). CI is fully green — **Build & Push Images** (the keyless web Docker build), **Deploy web (staging)**, **Staging Integration Tests**, and **Playwright E2E** all pass — which empirically validates the load-bearing assumption that `force-dynamic` suppresses the `/_not-found` build-time Clerk prerender. The highest-value follow-ups are reliability hardening (fail-loud on deploy-time misconfig) and one confirmed workflow inconsistency. All are non-blocking but should be fixed-in-PR or filed per ADR-028.

---

### Non-Blocking Findings

**[reliability]** `apps/web/lib/public-config.ts` (`PUBLIC_CONFIG_FALLBACK` / `readServerPublicConfig`)
The server read falls back to `http://localhost:3004` whenever `PUBLIC_API_URL` is unset/empty, and the layout passes a possibly-`undefined` `CLERK_PUBLISHABLE_KEY` straight to `<ClerkProvider>`. The deploy path mitigates this well (`deploy.yml:678` pre-promote secret check; GKE `[ -n "$CLERK_PUB" ] || exit 1`; `PUBLIC_API_URL` sourced from the terraform `cloud_run_api_url` output), so this is not a live gap today. But the localhost fallback is a *new* way to ship a "healthy" image silently pointed at localhost in production — the exact silent-misconfig class this repo has been bitten by (#395, #458). Build-time inlining at least failed loud.
**Suggestion:** In `readServerPublicConfig`, when `process.env.NODE_ENV !== 'development'` and `PUBLIC_API_URL` is unset/empty, throw (or emit an error-level log) instead of returning the localhost default. Keep the localhost fallback for the SSR-guard/client path and local dev only.

**[reliability]** `.github/workflows/staging.yml:641` vs `.github/workflows/deploy.yml:495,918`
`staging.yml` uses `--update-env-vars` (merge) for the web service env block while `deploy.yml` uses `--set-env-vars` (replace) for the same block. With merge semantics a variable later removed from the deploy command persists on the revision, and the two pipelines can diverge in what the running revision actually carries.
**Suggestion:** Align `staging.yml` on `--set-env-vars` so the deploy command is the single source of truth for the env set (or add a comment if merge semantics are deliberate for staging).

**[maintainability]** `docs/adr/ADR-028-web-runtime-public-config.md` env table + `apps/web/lib/active-program.ts`
`DEFAULT_PROGRAM` is listed in the ADR rename table as a managed runtime var "read in `lib/active-program.ts`", but it is wired in **no** deploy surface — not the deploy/staging workflow web steps, Cloud Run YAML, Terraform, or the GKE configmap/values. Every deployed env silently uses the hardcoded `'5-3-1'` default. (This matches `origin/main`'s `NEXT_PUBLIC_DEFAULT_PROGRAM`, so it is not a regression — but the doc now presents it as a first-class wired runtime value, which will mislead the next person trying to change the default via env.)
**Suggestion:** Either wire `DEFAULT_PROGRAM` through the deploy steps + manifests to match the doc, or annotate the ADR table and `.env.example` that it is intentionally code-default-only in deployed envs (local-dev override only).

**[reliability]** `docs/deploy.md` "Verifying runtime public config"
The embed-absence grep matches `run\.app` literally. That only matches `*.run.app` API hostnames; a custom-domain or GKE-ingress API URL would not match, so the gate can pass while a value is in fact embedded. The `window.__PUBLIC_CONFIG__` grep is also brittle (depends on no `<` in the value).
**Suggestion:** Capture the resolved API hostnames from `gcloud run services describe` (as the prior ADR-025 check did) and grep for those, so the verification can't pass on a custom-domain miss.

**[performance]** `apps/web/app/layout.tsx` — `export const dynamic = 'force-dynamic'`
Forcing the **root** layout dynamic opts every route out of static generation/ISR; all pages become per-request SSR. This is intentional and necessary (the layout's `process.env` reads must be request-time), and CI's green staging deploy + Playwright + integration runs confirm it functions — but the blast radius (root = all routes, including `/_not-found` and any error pages) is broader than the minimal need, and the ADR states the tradeoff without confirming no route relied on static optimization.
**Suggestion:** Confirm no route uses `generateStaticParams`/static rendering today (if so, the tradeoff is free), and have ADR-028 note the route set was verified. No code change needed if all routes are already dynamic.

**[maintainability]** `apps/web/components/PublicConfigProvider.tsx` + `.github/workflows/ci.yml` / `playwright.config.ts`
Two minor items: (1) `PublicConfigProvider` / `usePublicConfig()` has no consumer in the diff outside its own test — it ships the SSR-safe context channel ahead of need (harmless given all three channels derive from one server read, but currently YAGNI). (2) CI and Playwright retain the `pk_test_...` placeholder under `CLERK_PUBLISHABLE_KEY` "belt-and-suspenders", which slightly tensions the ADR's claim that the key is no longer needed at build.
**Suggestion:** Acceptable as-is; optionally drop the unused provider until a component needs it, and reconcile the ci.yml comment with the ADR wording (keep the key and soften the ADR, or confirm it's droppable and file a follow-up).

---

### Findings Disposition
Per ADR-028, each finding above is to be fixed in this PR or filed-and-linked before merge — none left as-is. Disposition (fixed in `<sha>` | filed `#N`) will be recorded in a "Review findings disposition" section of the PR body prior to `gh pr merge`.

---

Issue #464, opened 2026-06-09
Architecture review (2026-06-08): findings + six follow-up flags


brownm09:
## Context

A senior-architect review of the whole app, conducted 2026-06-08 via a four-way parallel exploration (docs/ADRs, `apps/api`, web/mobile/shared packages, infra/CI). This issue preserves the full commentary and tracks the concrete follow-ups.

The review surfaced **two categories of work**:
- **Six concrete flags** (backlog items below) — the actionable findings.
- **Three doc/code drift items** to *verify* before acting — the ADRs describe an intended system that the code may not fully match.

---

### The defining tension: enterprise architecture on a single-user fitness tracker

The PRD is explicit that the app serves two audiences: lifters *and* "technical evaluators assessing architectural depth for director-level roles." So the codebase is deliberately a portfolio artifact, and it shows — hexagonal ports/adapters, multi-tenancy with RLS, dual REST+GraphQL transports, GKE-vs-Cloud-Run A/B, React-Native-vs-Kotlin A/B, OpenTelemetry + burn-rate SLOs, KMS envelope encryption.

For the stated user problem (log a workout in under 2 minutes, tell me next week's weights), almost all of this is over-built. That's a legitimate choice for a portfolio piece, and the execution is genuinely strong. But the cost is real: every "compare two approaches" decision **doubles the surface area** to build, test, secure, and operate. Three simultaneous A/B comparisons (transport, infra, mobile) is a lot of permanent complexity for a solo project that will likely never see the traffic to make any comparison statistically meaningful. Worth owning as an explicit decision rather than absorbing silently.

### Documentation has drifted ahead of the code

The ADRs are excellent — but they describe an intended system, and in places the code hasn't caught up. That's the most dangerous failure mode for good docs: they become confidently wrong.

- **GraphQL (ADR-006, ADR-015) appears not to exist.** No `@nestjs/graphql`, no resolvers, no schema — the API is REST-only. ADR-015 (request-scoped DataLoaders for cross-user isolation) is written as if live. **Biggest doc/code gap.**
- **The 90/10 GKE<->Cloud Run traffic split (ADR-009) looks aspirational.** No Ingress canary rules or load-balancer backend weights surfaced; both targets deploy but the split mechanism isn't in the repo.
[...]

PR #516, opened 2026-06-11
[feat] Implement Postgres RLS defense-in-depth for multi-tenant isolation (#511)

### LINKED ISSUE #511: [security] Implement (or formally defer) Postgres RLS defense-in-depth for multi-tenant isolation (ADR-010 gap)
## Summary

The 2026-06-08 architecture review (#464) and its doc-reconciliation follow-up (#495) found that
[ADR-010 (Multi-Tenancy Data Isolation)](https://github.com/brownm09/lifting-logbook/blob/main/docs/adr/ADR-010-multi-tenancy-data-isolation.md)
describes a **two-layer** isolation model — application-level `user_id` scoping **plus** Postgres
Row-Level Security (RLS) as database-enforced defense-in-depth — but the **RLS layer does not exist
in the code**. Isolation is currently **single-layer** (application scoping only).

This is a potential multi-tenant isolation gap, not merely a doc gap: ADR-010 explicitly justifies
RLS as protecting against "application bugs that might otherwise leak cross-user data." Without it, a
single missing `where: { userId }` clause in any repository would leak data across tenants with no
database-level backstop.

## Verification evidence (#495, 2026-06-10)

- **No RLS policies:** zero `ENABLE ROW LEVEL SECURITY` / `CREATE POLICY` /
  `current_setting('app.current_user_id')` across all 17 migrations in
  `apps/api/prisma/migrations/`.
- **No session-variable middleware:** `apps/api/src/adapters/prisma/prisma.service.ts` is a bare
  `PrismaClient` subclass with lifecycle hooks only — no `$use`/`$extends`/`SET LOCAL`, nothing sets
  `app.current_user_id` per request/session. The second `PrismaClient` in
  `system-db-repository-factory.ts` also sets no session variable.
- **Live isolation boundary:** every repository filters by `userId` in the `where` clause at the
  application layer (e.g. `PrismaLiftRecordRepository.getLiftRecords` →
  `prisma.liftRecord.findMany({ where: { userId: this.userId, ... } })`). This is the real and only
  isolation enforcement today.

ADR-010's prose ("Row-Level Security (RLS) is enabled in Postgres as a defense-in-depth measure";
"the application sets `app.current_user_id` ... via Prisma middleware") is therefore **confidently
wrong** — the most dangerous failure mode the #464 review called out. #495 corrects the ADR's
implementation-status text and links here for the remediation decision.

## Decision required — implement or formally defer

**Option A — Implement RLS (close the gap):**
1. Add a migration enabling RLS and creating per-table `user_isolation` policies on every user-data
   table: `ALTER TABLE <t> ENABLE ROW LEVEL SECURITY;` + `CREATE POLICY user_isolation ON <t> USING
   (user_id = current_setting('app.current_user_id', true))`. (Note `FORCE ROW LEVEL SECURITY` so the
   table owner is also constrained, and that the app DB role must not be `BYPASSRLS`.)
2. Add a Prisma client extension / middleware that, per request, runs `SET LOCAL "app.current_user_id"
   = $userId` inside the same transaction/connection as the query (RLS reads a session/transaction
   GUC — it must be set on the **same** pooled connection that executes the query; `SET LOCAL` inside
   an interactive transaction is the safe shape).
3. Verify connection-pooling interaction (the Cloud SQL private-IP setup + any PgBouncer/transaction
   pooling) does not leak the GUC across pooled sessions.
4. Add an integration test proving cross-user reads return zero rows even when the application-layer
   `userId` filter is deliberately bypassed (i.e. the DB layer enforces isolation independently).

**Option B — Formally defer (accept single-layer scoping):**
- Write an ADR (or amend ADR-010) recording the explicit decision to rely on application-layer
  `user_id` scoping only for the current single-user / portfolio phase, with the RLS layer as a
  documented future enhancement — so the docs stop claiming a layer that isn't there.

Either resolution is acceptable; the unacceptable state is the current one (docs claim a security
layer that doesn't exist).

## References

- Parent: #464 (architecture review), #495 (ADR-drift reconcile that surfaced this)
- [ADR-010](https://github.com/brownm09/lifting-logbook/blob/main/docs/adr/ADR-010-multi-tenancy-data-isolation.md)
- [PostgreSQL — Row Security Policies](https://www.postgresql.org/docs/current/ddl-rowsecurity.html)

brownm09:
## Summary

Implements the Postgres Row-Level Security (RLS) defense-in-depth layer that [ADR-010](docs/adr/ADR-010-multi-tenancy-data-isolation.md) described but that did not exist — surfaced by the 2026-06-08 architecture review ([#464](https://github.com/brownm09/lifting-logbook/issues/464)) and filed as the security gap [#511](https://github.com/brownm09/lifting-logbook/issues/511). Isolation was **single-layer** (per-query `userId` scoping). This adds a database-enforced second layer so a query that forgets the `userId` filter — or an injection that bypasses the ORM — still cannot cross tenant boundaries.

### How it works
- **Migration `20260611000000_enable_rls`** — `ENABLE` + `FORCE ROW LEVEL SECURITY` and a `USING`/`WITH CHECK` policy on all 13 `userId` tables, plus an `EXISTS`-subquery policy on `custom_program_spec` (no `userId` — isolated through its parent program's FK). Creates a non-superuser, non-`BYPASSRLS` `lifting_app` role. **Fail-closed**: an unset GUC resolves to `NULL`, matching no rows.
- **`RlsInterceptor`** (global) — per authenticated request, opens one interactive transaction, runs `set_config('app.current_user_id', $userId, true)` (wrapped in a manual OTel span per [ADR-024](docs/adr/ADR-024-prisma-otel-sdk-override.md)), and stores the tx client in CLS (`nestjs-cls`). `PrismaRepositoryFactory` routes every repository through it, so all of a request's queries see the GUC.
- **`prisma-tx.util.ts`** — Prisma's interactive tx client has no `$transaction`, so the three repos that batch/interactive-transaction now use `runBatch`/`runInteractive`, which reuse the request transaction instead of nesting.
- **`cycle-plan`** raises its RLS tx timeout (it calls an LLM mid-request); the connection-hold trade-off is documented with a follow-up.

### Role requirement (why this is partly deployment-coupled)
RLS is ignored by superusers/`BYPASSRLS` roles. The app connects as the bootstrap **superuser locally and in CI**, so RLS is **dormant** there (the existing per-query scoping still applies). Deployed environments must point `DATABASE_URL` at `lifting_app` for the policies to bite. Provisioning `lifting_app` on Cloud SQL (Terraform user + secret cutover) is a **deployment-coupled follow-up** — out of scope here. This is also the **instant-rollback** path: repoint `DATABASE_URL` at the superuser role and RLS goes dormant with no schema change.

## Acceptance criteria (#511)
- [x] RLS policies on every user-data table (`ENABLE`/`FORCE` + `CREATE POLICY`), using `current_setting('app.current_user_id')`.
- [x] Application sets `app.current_user_id` per request (interceptor + `set_config` inside the request transaction).
- [x] Non-superuser application role so policies are actually enforced.
- [x] DB-E2E tests proving cross-tenant isolation, fail-closed, and `WITH CHECK`.
- [x] ADR-010 reconciled to reality.

## Testing
`npm test -w @lifting-logbook/api` (Docker up for Testcontainers):

```
Test Suites: 28 passed, 28 total
Tests:       373 passed, 373 total
Snapshots:   0 total
```

No regression from wrapping each request in a transaction or from the `runBatch`/`runInteractive` repo refactor. New `apps/api/src/adapters/prisma/rls.db.e2e.spec.ts` (7 tests) connects **as `lifting_app`** and proves: unscoped-read isolation (RLS, not app filtering), fail-closed on unset GUC, `WITH CHECK` rejection of a mismatched-`userId` insert, the `custom_program_spec` FK policy, that the test role is genuinely non-superuser/non-`BYPASSRLS` (regression guard against silently reconnecting as a superuser), and end-to-end interceptor→factory GUC wiring.

**Coverage gate:** `rls.db.e2e.spec.ts` is the new-behavior coverage (DB-E2E tier).

## Pre-PR gate checks
- **Suppression grep** — clean (no new `ts-ignore`/`eslint-disable`/`!`/`?? null`).
- **Test-integrity grep** — one match: `const describeOrSkip = TC_DATABASE_URL ? describe : describe.skip;` in `rls.db.e2e.spec.ts`. **Justification:** this is the established DB-E2E skip-guard pattern, identical to [`programs.db.e2e.spec.ts`](apps/api/src/programs/programs.db.e2e.spec.ts) — it skips **only** when `jest.global-setup.js` provisions no database (Docker down *and* not in CI, the documented [#394](https://github.com/brownm09/lifting-logbook/issues/394) escape hatch). The suite runs fully in CI and locally with Docker; it does not unconditionally skip. No deleted tests, no lowered thresholds, no bypass flags.
- **Lockfile (ADR-036)** — `nestjs-cls@^4.5.0` added to `apps/api`; `package-lock.json` regenerated and verified stable/in-sync (idempotent `--package-lock-only`; only `nestjs-cls` paths added).
- **ADR-warrant** — ADR-010 substantially revised (Implementation section + corrected Decision/status); `docs/README.md` index and `docs/adr-references.md` updated in this PR.

## Risk / rollback
- **Load-bearing assumption (verified):** the request-tx client cannot nest transactions; the `runBatch`/`runInteractive` helper handles this and is proven by the `runBatch`-under-RLS test plus the unchanged passing of `programs.db.e2e`.
- **Connection pinning:** each authenticated request holds one connection for its transaction; `cycle-plan` holds it across the LLM call (raised timeout). Low-volume endpoint; follow-up to move the LLM call outside the transaction.
- **Rollback:** repoint `DATABASE_URL` at the superuser role (RLS dormant, no schema change); full revert is a `DROP POLICY` + `DISABLE ROW LEVEL SECURITY` migration.

## Follow-ups (out of scope)
- Cloud SQL `lifting_app` provisioning + staging/prod secret cutover (infra/Terraform).
- Move the `cycle-plan` LLM call outside its RLS transaction.

Closes #511

---

`/review 516` ran on commit `a05344f`; all findings resolved below.

### Blocking
- **Factory-bypass: three controllers ran on the base client (no GUC) → fail-closed under `lifting_app`** — `CustomProgramsController`, `UserSettingsController`, and `SwitchProgramController` built repositories with the bare `PrismaService` instead of the per-request RLS transaction client, so `/programs/custom`, `/users/me/settings`, and program-switch would have returned zero rows at the `lifting_app` cutover (undetected by superuser CI). **Fixed in `717ec5e`** — added `PrismaService.clientForRequest()` (returns the CLS request tx client when active, else base) and routed all three controllers through it; widened the two affected repo constructors to `PrismaExecutor` and converted `CustomProgramsRepository`'s direct `$transaction` to `runInteractive`.

### Non-blocking — fixed in this PR (`717ec5e`)
- **Generic policy-coverage gap** (superuser CI can't catch a forgot-a-policy regression) — added a DB-E2E guard asserting every `userId` table (and `custom_program_spec`) has `FORCE` RLS + ≥1 policy, plus a `clientForRequest()` routing test.
- **GraphQL `getType() !== 'http'` skip** — GraphQL is not wired in `apps/api` today (no `GraphQLModule`), so the skip is safe and fail-closed; added an in-code note that the guard must broaden to `'graphql'` if resolvers touching `userId` tables are ever added.
- **`ALTER DEFAULT PRIVILEGES` single-owner assumption** — added a migration comment documenting that the no-`FOR ROLE` form is correct under the ADR-027 single-owner-runs-migrations contract.

### Non-blocking — filed and linked (genuinely deployment-coupled / out of scope)
- **Connection-pool sizing + `maxWait` + load test before the `lifting_app` cutover** (per-request transaction now holds a connection for the whole request; default pool ≈5/pod) — filed **#517** (folded into the cutover-prerequisites issue).
- **`cycle-plan` 120s connection-hold across the LLM call** — filed **#518** (move the LLM call outside its RLS transaction).
- **Per-request transaction observability** (span/metric on the tx envelope) **+ ESLint guard banning direct `.$transaction(` in repos** — filed **#519**.

### Confirmed correct, no action
- `lastValueFrom(next.handle().pipe(defaultIfEmpty(undefined)))` correctly guards against `EmptyError` on void/204 handlers; `set_config` is parameterized (no injection) and transaction-local (no GUC leak across pooled connections).

### Updated testing
Full API suite after the fix: **Tests: 375 passed, 0 skipped, 0 failed (28 suites)** — adds the generic policy-coverage and `clientForRequest()` routing tests.


brownm09:
## Review: [feat] Implement Postgres RLS defense-in-depth for multi-tenant isolation

### Summary
**Needs revision → resolved; ready to merge once CI is green.** The RLS mechanism is sound — parameterized GUC, transaction-local fail-closed semantics, full 14-table `USING`+`WITH CHECK`+`FORCE` coverage, and correct `runBatch`/`runInteractive` atomicity. Review (766-line diff, two parallel opus passes) surfaced one **blocking** correctness hole: three controllers (`CustomProgramsController`, `UserSettingsController`, `SwitchProgramController`) built repositories with the bare `PrismaService` rather than the per-request RLS transaction client, so under the non-superuser `lifting_app` role their queries would run with `app.current_user_id` unset and fail closed (zero rows) — breaking `/programs/custom`, `/users/me/settings`, and program-switch, and invisible to superuser-connected CI. Fixed in `717ec5e` via a `PrismaService.clientForRequest()` router. All other findings are fixed-in-PR or filed (#517/#518/#519). See the **Review findings disposition** section in the PR body.

---

### Blocking Findings

**[correctness]** `custom-programs.controller.ts:29`, `user-settings.controller.ts:15,24`, `switch-program.controller.ts:35,41`
Three controllers constructed their repositories with the bare `PrismaService` singleton, bypassing the factory's CLS-tx routing. The factory routes its 14 repos through the per-request transaction (where `set_config('app.current_user_id', …)` was applied), but these controllers do not — so their queries execute on the base connection with no GUC.
Under `lifting_app` (RLS active), `current_setting('app.current_user_id', true)` is `NULL` → every `"userId" = NULL` predicate is false → zero rows / rejected writes. The endpoints silently break at the cutover, and superuser CI (RLS dormant) never catches it.
**Fix (done, `717ec5e`):** added `PrismaService.clientForRequest()` returning the CLS request-tx client when active (else the base client); routed all three controllers through it; widened `CustomProgramsRepository`/`UserSettingsRepository` constructors to `PrismaExecutor` and converted the repo's direct `$transaction` to `runInteractive`.

---

### Non-Blocking Findings

**[correctness]** `rls.db.e2e.spec.ts` — superuser CI cannot catch a "added a `userId` table, forgot the policy" regression, since the main suite runs with RLS dormant. **Done (`717ec5e`):** added a generic guard asserting every `userId` table (and `custom_program_spec`) has `FORCE` RLS + ≥1 policy, plus a `clientForRequest()` routing test.

**[maintainability]** `rls.interceptor.ts:45` — the `getType() !== 'http'` skip would also skip GraphQL resolvers, which fail-close under `lifting_app`. GraphQL is not wired today (no `GraphQLModule`), so not a live hole. **Done (`717ec5e`):** in-code note to broaden the guard if resolvers touching `userId` tables are added.

**[correctness]** `migration.sql:35` — `ALTER DEFAULT PRIVILEGES` without `FOR ROLE` only covers tables created by the role running this migration; a future migration run by a different owner would leave new tables un-granted (hard `permission denied`, not fail-closed). Correct under the single-owner contract (ADR-027). **Done (`717ec5e`):** migration comment documenting the assumption.

**[performance]** Per-request interactive transaction now holds a DB connection for the whole request (default pool ≈5/pod, no explicit `connection_limit`/`maxWait`); `cycle-plan` holds it 120s across the LLM call. Deployment-coupled (needs load data; bites only under `lifting_app` + load). **Filed #517** (pool sizing + cutover prereqs) and **#518** (move the `cycle-plan` LLM call outside its transaction).

**[reliability/maintainability]** The per-request `$transaction` envelope is not itself observable (only `set_config` is spanned), and `runBatch`/`runInteractive` are opt-in (a future direct `.$transaction(` would break GUC scope). **Filed #519** (tx-envelope span/metric + ESLint guard).

---

### Confirmed correct, no action
`lastValueFrom(next.handle().pipe(defaultIfEmpty(undefined)))` correctly avoids `EmptyError` on void/204 handlers; the `set_config` raw SQL is a real bind parameter (no injection) and transaction-local (`SET LOCAL` — no GUC leakage across pooled connections); `runBatch` sequential path preserves all-or-nothing inside the enclosing request transaction.

---

### Findings Disposition
Recorded in full in the PR body's **Review findings disposition** section: 1 blocking + 3 non-blocking fixed in `717ec5e`; 3 non-blocking filed as #517/#518/#519. Updated suite: **375 passed, 0 skipped, 0 failed (28 suites)**.

---

PR #485, opened 2026-06-09
[feat] Smart File Import wizard — Phase 1 (classify, preview & commit any CSV)

brownm09:
## Summary

Generalizes the lift-records-only importer ([#225](https://github.com/brownm09/lifting-logbook/issues/225)) into a unified **Smart Import wizard** that accepts any CSV, auto-detects its type, previews the before→after effect, and commits — across all four data types (lift history, training maxes, strength goals, program spec).

Design of record: `docs/proposals/2026-06-09-smart-file-import.md` + the committed prototype. This PR covers **Phase 1 (MVP)**; Phase 2 and Phase 3 are filed as follow-ups (below).

### What's included
- **`packages/core`** — `classifyImport` (signal-based, not exact-header), `IMPORT_THRESHOLDS` (per-type buckets/auto-accept), `parseStrengthGoals`, per-kind validators, and `buildImportPreview` diff builders (`{ creates, updates, skips, deltas }`).
- **`apps/api`** — `saveProgramSpec` added to `ILiftingProgramSpecRepository` (+ Hybrid/in-memory impls); unified `POST /programs/:program/import?mode=preview|commit` (`ImportController`). Commit re-parses server-side and never trusts the client payload. Shared multipart helper extracted from the legacy importer.
- **`apps/web`** — wizard under `app/(authed)/import/`: Source (program picker + file) → Analyzing → Classify → Map (Phase-1 placeholder) → Review → Preview → Done; `previewImport`/`commitImport` client API.

### Key decisions
- **Program scoping:** explicit program-picker in the Source step (custom programs).
- **Program-spec writes:** custom (UUID) programs only via the existing `customProgramSpec` table — built-in templates are immutable seed data and are rejected. **No schema migration** — idempotency uses find-then-write inside a transaction.
- **Strength_Goals format:** **locked to the user's real export** — a tier ladder (Weight/dates metadata → `Lift,Current TM,Intermediate,Advanced,Elite` header → one row per lift with absolute target weights per tier → Goal Date/Note footer). Phase 1 imports the **next milestone** (lowest tier above each lift's Current TM, fallback top tier). The classifier scans all cells for the strength-goals ladder so the real file classifies confidently.

## Acceptance Criteria (Phase 1)
- [x] `parseStrengthGoals` parses the real tier-ladder export into `StrengthGoalEntry[]` (next-milestone target, lbs).
- [x] File-level classifier → `{ type, confidence, reasons[], alternatives[] }` via weighted signals.
[...]

PR #531, opened 2026-06-13
[fix] Make Smart Import commit atomic + concurrency-safe (TM/goal/program-spec)

### LINKED ISSUE #488: [reliability] Smart Import commit: atomicity + concurrent-import safety (TM/goal/program-spec)
## Problem

The Smart Import commit path (#477, PR #485) writes training-maxes and strength-goals **non-atomically**, and the program-spec write has a known concurrent-import gap:

1. **Non-atomic read+write (reporting + races).** `ImportController.commit` for `training-maxes` and `strength-goals` does `getTrainingMaxes`/`getGoals` → `buildPreview` (for counts) → write in separate awaits. A concurrent import/edit between the read and the write makes the reported `created/updated/skipped` wrong. (`apps/api/src/programs/import.controller.ts`)
2. **Partial-failure, no rollback (strength-goals).** The strength-goal commit loops `await upsertGoal(...)` per row. If a row mid-loop throws, earlier goals are already written and nothing is returned to the client — a partial, unrecoverable commit.
3. **Deferred `customProgramSpec` unique index (concurrent dupes).** `saveProgramSpec` uses idempotent find-then-write in a transaction with **no unique index** on `customProgramSpec (programId, week, offset, lift, order)`. Two simultaneous imports of the same file can both `findFirst → null → create`, producing duplicate spec rows the natural key was meant to prevent.

## Proposed fix

- Add batch write methods that return their own `{created, updated, skipped}` (mirroring `saveProgramSpec`) for training-max and strength-goal, so commit counts come from the write result, not a pre-read — and wrap the strength-goal multi-row write in a transaction so partial failure rolls back.
- Add the `customProgramSpec` unique index via a Prisma migration (applied through the in-VPC Cloud Run Job pipeline) and switch `saveProgramSpec` to an `upsert` on that constraint.

## Context

Surfaced by the `/code-review` of PR #485 (reliability pass). Phase 1 accepts these for the single-user, non-concurrent MVP path; this issue tracks the hardening.

brownm09:
## Summary

Hardens the Smart Import **commit** path, fixing the three reliability gaps the `/code-review` of PR #485 surfaced (issue #488). `program-spec` was already transactional via `saveProgramSpec`; this brings training-maxes and strength-goals to the same bar and closes a concurrent-duplicate gap on `customProgramSpec`.

1. **Non-atomic read→preview→write (TM + goals).** Commit did `getX → buildXPreview (for counts) → write` in separate awaits; a concurrent edit between read and write desynced the reported counts. **Fix:** new transactional batch methods `ITrainingMaxRepository.importTrainingMaxes` and `IStrengthGoalRepository.importGoals` read + classify + write in one transaction and return their own `{created,updated,skipped}` — counts come from the write itself.
2. **Partial-failure, no rollback (strength-goals).** The commit looped `await upsertGoal(...)` per row with no surrounding transaction; a mid-loop throw left earlier rows written. **Fix:** `importGoals` runs the whole batch in one transaction, so a failure rolls back.
3. **Missing `customProgramSpec` unique index.** `saveProgramSpec` did find-then-create with no DB constraint; two concurrent imports could both `findFirst → null → create` a duplicate. **Fix:** `@@unique([programId, week, offset, lift, order])` migration + `saveProgramSpec` insert switched to an `upsert` on that constraint.

Existing pure-write methods (`saveTrainingMaxes`, `upsertGoal`) are **untouched** — they have non-import callers (cycle-generation, the TM/goal controllers) that need neither counts nor find-then-write. The create/update/skip decision is extracted into shared classifiers (`trainingMaxRowKind`, `strengthGoalRowKind`) so preview and commit can never disagree.

## Acceptance criteria
- [x] Batch write methods return their own `{created,updated,skipped}` for TM + strength-goal (mirroring `saveProgramSpec`)
- [x] Strength-goal multi-row write wrapped in a transaction (rolls back on partial failure)
- [x] `customProgramSpec` unique index added via migration; `saveProgramSpec` switched to `upsert` on it

## Testing
`npm run build` ✅ · `npm run lint` ✅ (7/7) · Docker up for the Testcontainers DB e2e.

- `npm test -w @lifting-logbook/api` → **Tests: 385 passed, 0 skipped, 0 failed (33.2s)** — includes `programs.db.e2e.spec.ts` (migrate-deploy applied the new migration; new unique-constraint enforcement test passed) and the new Prisma repo unit specs.
- `npm test -w @lifting-logbook/core` → **Tests: 203 passed, 0 skipped, 0 failed (18.9s)** — includes new classifier tests.

New tests:
- `training-max.repository.spec.ts` / `strength-goal.repository.spec.ts` — counts, in-batch dedupe, read-in-transaction, and **mid-batch failure propagation** (proving the batch runs inside `$transaction`, so a real DB rolls back).
- `buildImportPreview.test.ts` — direct `trainingMaxRowKind` / `strengthGoalRowKind` create/update/skip tests (the shared decision used by both preview and the in-memory adapters).
- `programs.db.e2e.spec.ts` — DB-level `P2002` on a duplicate `customProgramSpec` natural key.
- In-memory import e2e unchanged — asserts identical commit counts after the rewrite.

## Suppression note (ADR-026)
`apps/api/src/adapters/prisma/strength-goal.repository.ts` `entryToData()` uses `goal.target ?? null` and `goal.ratio ?? null`. This is not silencing a type error — it is the domain→persistence mapping: `StrengthGoalEntry.target/ratio` are optional (`number | undefined`) while the Prisma columns are nullable (`Float?`), so an absent optional must become SQL `NULL`. The lines are relocated verbatim from the pre-existing `upsertGoal` (now shared by `upsertGoal` and `importGoals`); no new suppression is introduced.

## Migration / data-integrity note
The migration de-duplicates (`DELETE` keeping the lowest `id` per natural key) **before** `CREATE UNIQUE INDEX`, so it is safe on a table that already holds duplicates. `offset`/`order` are reserved words and are quoted. Validated by the Testcontainers `prisma migrate deploy` in `jest.global-setup.js` + the enforcement test. Reversible by dropping the index.

Closes #488

---

`/review` (claude-opus-4-8) — **0 blocking, 6 non-blocking**. Per ADR-028, each is fixed-in-PR or filed-and-linked:

| # | Finding | Disposition |
|---|---|---|
| 1 | `saveProgramSpec` over-counts `created` when the loser's upsert updates under a same-program concurrent race | **Fixed in `fcdaa30`** — documented as best-effort counts (data outcome correct; exact split needs an extra round-trip, over-engineering for a count display) |
| 2 | `trainingMaxRowKind` compared weights by string coercion (`300.10` vs `300.1` footgun) | **Fixed in `fcdaa30`** — numeric comparison |
| 3 | Migration de-dupe `DELETE` is silently destructive (oldest row wins) | **Fixed in `fcdaa30`** — comment flags destructiveness + adds a pre-deploy duplicate-check query |
| 4 | `saveProgramSpec` keeps a now-redundant per-row `findFirst` + duplicated where-clause; not aligned with the two new `import*` methods | **Filed [#532](https://github.com/brownm09/lifting-logbook/issues/532)** — separable refactor |
| 5 | Tally loop (seen-set + counters) copy-pasted across 4 adapters | **Filed [#532](https://github.com/brownm09/lifting-logbook/issues/532)** — shared `classifyAndCount` helper |
| 6 | P2028 (tx timeout) on the import path falls through to a generic 500 in non-RLS paths (bounded in prod) | **Filed [#532](https://github.com/brownm09/lifting-logbook/issues/532)** — map P2028 if batch sizes grow |

Post-review verification: `npm run build` ✅ (6/6) · `npm run lint` ✅ (7/7) · `npm test -w @lifting-logbook/core` → **203 passed, 0 skipped, 0 failed** · `npm test -w @lifting-logbook/api -- --testPathPatterns repository.spec` → **13 passed** (Testcontainers re-applied the edited migration cleanly). The numeric-compare change is decision-only (identical for integer/clean-float weights); the api/migration edits are comments only.

### DIFF
--- a/apps/api/prisma/migrations/20260613000000_add_custom_program_spec_unique/migration.sql
+++ b/apps/api/prisma/migrations/20260613000000_add_custom_program_spec_unique/migration.sql
@@ -4,6 +4,12 @@
 
 -- De-duplicate first: a UNIQUE INDEX fails if duplicates already exist. Keep the
 -- lowest id per natural key and drop the rest. No-op on a clean table.
+-- DESTRUCTIVE on a table that already holds duplicates: the *oldest* (lowest-id) row
+-- wins, so a duplicate carrying newer config is dropped, and Prisma has no down path.
+-- Verify production has zero duplicates before deploy:
+--   SELECT "programId","week","offset","lift","order", count(*)
+--   FROM "custom_program_spec"
+--   GROUP BY 1,2,3,4,5 HAVING count(*) > 1;
 DELETE FROM "custom_program_spec" a
 USING "custom_program_spec" b
 WHERE a."id" > b."id"
--- a/apps/api/prisma/migrations/20260613000000_add_custom_program_spec_unique/migration.sql
+++ b/apps/api/prisma/migrations/20260613000000_add_custom_program_spec_unique/migration.sql
@@ -1,0 +1,17 @@
+-- Enforce the custom-program-spec natural key (programId, week, offset, lift, order)
+-- so two concurrent Smart Imports of the same file cannot both find-then-create a
+-- duplicate row (issue #488). `offset` and `order` are reserved words — quoted.
+
+-- De-duplicate first: a UNIQUE INDEX fails if duplicates already exist. Keep the
+-- lowest id per natural key and drop the rest. No-op on a clean table.
+DELETE FROM "custom_program_spec" a
+USING "custom_program_spec" b
+WHERE a."id" > b."id"
+  AND a."programId" = b."programId"
+  AND a."week" = b."week"
+  AND a."offset" = b."offset"
+  AND a."lift" = b."lift"
+  AND a."order" = b."order";
+
+-- CreateIndex
+CREATE UNIQUE INDEX "custom_program_spec_programId_week_offset_lift_order_key" ON "custom_program_spec"("programId", "week", "offset", "lift", "order");
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -219,6 +219,9 @@ model CustomProgramSpec {
   weekType       String?
   program        CustomProgram @relation(fields: [programId], references: [id], onDelete: Cascade)
 
+  // Enforces the natural key the import path keys on, so two concurrent imports of
+  // the same file cannot both find-then-create a duplicate row (issue #488).
+  @@unique([programId, week, offset, lift, order])
   @@index([programId])
   @@map("custom_program_spec")
 }
--- a/apps/api/src/adapters/in-memory/strength-goal.adapter.ts
+++ b/apps/api/src/adapters/in-memory/strength-goal.adapter.ts
@@ -1,4 +1,5 @@
-import { StrengthGoalEntry } from '@lifting-logbook/core';
+import { StrengthGoalEntry, strengthGoalRowKind } from '@lifting-logbook/core';
+import { ImportWriteResult } from '@lifting-logbook/types';
 import { IStrengthGoalRepository } from '../../ports/IStrengthGoalRepository';
 import { StrengthGoalNotFoundError } from '../../ports/errors';
 
@@ -21,6 +22,35 @@ export class InMemoryStrengthGoalRepository implements IStrengthGoalRepository {
   async upsertGoal(program: string, goal: StrengthGoalEntry): Promise<StrengthGoalEntry> {
     this.programMap(program).set(goal.lift, goal);
     return goal;
+  }
+
+  async importGoals(
+    program: string,
+    goals: StrengthGoalEntry[],
+  ): Promise<ImportWriteResult> {
+    const map = this.programMap(program);
+    const existingByLift = new Map(map);
+
+    let created = 0;
+    let updated = 0;
+    let skipped = 0;
+    const seen = new Set<string>();
+
+    for (const g of goals) {
+      if (seen.has(g.lift)) continue; // collapse duplicate lifts within the file
+      seen.add(g.lift);
+
+      const kind = strengthGoalRowKind(g, existingByLift);
+      if (kind === 'skip') {
+        skipped++;
+        continue;
+      }
+      map.set(g.lift, g);
+      if (kind === 'create') created++;
+      else updated++;
+    }
+
+    return { created, updated, skipped };
   }
 
   async deleteGoal(program: string, lift: string): Promise<void> {
--- a/apps/api/src/adapters/in-memory/training-max.adapter.ts
+++ b/apps/api/src/adapters/in-memory/training-max.adapter.ts
@@ -1,4 +1,5 @@
-import { TrainingMax } from '@lifting-logbook/core';
+import { TrainingMax, trainingMaxRowKind } from '@lifting-logbook/core';
+import { ImportWriteResult } from '@lifting-logbook/types';
 import { ITrainingMaxRepository } from '../../ports/ITrainingMaxRepository';
 import { SEED_PROGRAM, seedTrainingMaxes } from './fixtures';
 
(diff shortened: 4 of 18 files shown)


brownm09:
## Review: [fix] Make Smart Import commit atomic + concurrency-safe (TM/goal/program-spec)

### Summary
**Merge with minor changes.** The PR correctly meets its #488 goals: `importTrainingMaxes`/`importGoals` read+classify+write in one transaction via `runInteractive` (reusing the per-request RLS tx — verified correct), the strength-goals partial-commit bug is genuinely fixed, and the `customProgramSpec` unique index + `upsert` make the program-spec path race-safe at the data layer. Coverage and test-integrity gates pass. **No blocking findings.** The highest-severity item is a count-accuracy skew in `saveProgramSpec` under a same-program concurrent-import race (data is correct; only the displayed `created`/`updated` split is off in that window).

---

### Non-Blocking Findings

**[correctness]** `apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts:135-157`
On `findFirst → null` the code runs `upsert(...)` then unconditionally `created++`. In the exact concurrent race this PR was built to survive (two imports of the same program racing past `findFirst → null`), the loser's upsert performs an UPDATE but is still counted as a create. The data outcome is right (no P2002 crash); the returned `{created}` over-reports and `{updated}` under-reports in that window. The same best-effort-under-race property applies to `importTrainingMaxes`/`importGoals` (classification is taken from the in-tx read snapshot).
**Suggestion:** Deriving the exact insert-vs-update from a Prisma `upsert` needs an extra round-trip (or `xmax` inspection) — over-engineering for a count display under a rare manual-UI race. Fix-in-PR: document the known skew on the method and note in the PR body that commit counts are best-effort under concurrent same-program imports.

**[correctness]** `packages/core/src/utils/import/buildImportPreview.ts:79`
`trainingMaxRowKind` decides skip-vs-update via string coercion `` `${before}` === `${m.weight}` ``. Both sides are `Float` → JS `number`, so a value arriving as `300.10` vs a stored `300.1` would stringify-differ (`"300.10"` ≠ `"300.1"`) and misclassify as an update — an unnecessary write + inflated `updated`. Preview and commit stay consistent with each other (the #488 goal holds); the risk is spurious updates only.
**Suggestion:** Compare numerically (`before === m.weight`) for the decision; keep the string form only for the `before`/`after` display values. Fix-in-PR (one-line, strictly safer).

**[reliability]** `apps/api/prisma/migrations/20260613000000_add_custom_program_spec_unique/migration.sql:13-20`
The `DELETE ... USING ... WHERE a.id > b.id` permanently drops duplicate `custom_program_spec` rows (keeping lowest id) before the unique index. No-op on a clean table, but any pre-existing duplicate with differing config loses the newer row silently, and Prisma has no down-migration.
**Suggestion:** Make the destructiveness explicit — clarify in the migration comment that the *oldest* row wins, and add a PR-body / deploy note recommending a pre-deploy `SELECT count(*)` duplicate check against production. Fix-in-PR (docs).

**[maintainability]** `hybrid-program-spec.repository.ts:127-158`
With the unique constraint in place, `saveProgramSpec`'s per-row `findFirst` is no longer needed for race-safety (the upsert's `where` handles it); it now serves only skip/count classification, and its where-clause is hand-duplicated against the upsert's. The other two commit methods classify from a single up-front `findMany`. This leaves program-spec structurally asymmetric (and `SaveProgramSpecResult` is now just an alias of `ImportWriteResult`, yet the method keeps the `save*` name while the others are `import*`).
**Suggestion:** Align `saveProgramSpec` on the `findMany` → classify-in-memory → `upsert` shape the other two use (removes the duplicated where-clause + a round-trip/row). Separable from #488's reliability goal → **file a follow-up**.

**[maintainability]** in-memory + Prisma adapters (×4) + the two preview builders
The PR correctly extracted the create/update/skip *decision* into shared `*RowKind` classifiers, but the surrounding *tally loop* (seen-set dedupe + created/updated/skipped counters) is still copy-pasted across four adapters.
**Suggestion:** Extract a shared generic `classifyAndCount<T>(rows, keyOf, rowKind, applyWrite)` so a future dedupe-semantics change can't desync counts across adapters. Separable refactor → **file a follow-up**.

**[reliability]** `prisma-tx.util.ts` interactive path (via the new methods)
When `runInteractive` opens its own `$transaction` (non-RLS / unit paths), a large import whose N upserts exceed Prisma's default 5s interactive-tx timeout throws **P2028**, which the per-feature exception filters don't map → generic 500. In production the import-commit path runs inside the request RLS tx, so this is bounded, but it's unmapped.
**Suggestion:** Note in the PR body that P2028/P2002 fall through to the default handler; **file a follow-up** to map P2028 to a 503/timeout response if batch sizes can grow.

---

### Findings Disposition
Recorded in the PR body before merge (per ADR-028/039): the three correctness/reliability items above are fixed in-PR (numeric compare; count-skew + best-effort-counts doc; migration destructiveness note); the structural-refactor and P2028-mapping items are filed as a tracked follow-up and linked.

---

PR #533, opened 2026-06-14
[infra] RLS production cutover: runtime lifting_app role + pooled DATABASE_URL

### LINKED ISSUE #517: [infra] RLS production cutover prerequisites: provision lifting_app on Cloud SQL + size connection pool
Follow-up from the Postgres RLS PR #516 (#511). RLS policies and the per-request `set_config` interceptor are merged, but RLS is **dormant** in every environment because the app connects as a superuser (which bypasses RLS). Cutting over requires provisioning the non-superuser `lifting_app` role and addressing the connection-resource change the per-request transaction introduces.

### Deployment provisioning
- Provision the `lifting_app` role on Cloud SQL (Terraform `google_sql_user` + password in Secret Manager).
- Point each deployed env's `DATABASE_URL` secret at `lifting_app` (staging first, then prod). `migrate deploy` keeps using the owner/superuser URL (ADR-027).
- This is the activation step for RLS in prod; it is also the instant-rollback lever (repoint `DATABASE_URL` at the superuser role → RLS dormant, no schema change).

### Connection-pool sizing (surfaced in the PR #516 review)
The RLS interceptor wraps each authenticated request in one interactive transaction, so a connection is now held for the **whole request** (previously per-query). `PrismaService` sets no `connection_limit`, so Prisma's default pool is `num_physical_cpus * 2 + 1` (≈5 on a 2-vCPU pod). Before the `lifting_app` cutover under load:
- Set an explicit `connection_limit` in `DATABASE_URL` sized to expected per-pod concurrency.
- Set an explicit Prisma `maxWait` (default 2000ms governs connection-acquisition wait) and map `P2028` to a sensible HTTP status + distinct log so pool starvation is distinguishable from a slow query.
- Load-test concurrent-request throughput under the per-request-transaction model; document the pool-size/replica math in the deploy runbook.

Blocking prerequisite of the cutover — must land before any env connects as `lifting_app`.

brownm09:
## Summary

Activates the Row-Level Security policies shipped in #511 (ADR-010) in staging and production by making the **runtime** connect as a `NOBYPASSRLS` role. Today both environments connect as `lifting-logbook-app` — a Cloud SQL built-in user and `cloudsqlsuperuser` member that **bypasses RLS** — so the per-tenant policies are inert (confirmed live: both `*-database-url` secrets resolve to `postgresql://lifting-logbook-app:…`, no `connection_limit`). Also sizes the Prisma connection pool for the small Cloud SQL tiers.

> ⚠️ **Production cutover is human-gated.** Merging performs the **staging** cutover via the normal pipeline. Production requires the manual deploy approval **and** explicit go-ahead after the staging validation gate + ≥24h soak. Do **not** enable auto-merge.

## Design — two roles, two secrets

The migrate Cloud Run Job and the runtime services currently share one `database_url` secret/role. Migrations must run as an owner/superuser (DDL + data migrations; under `FORCE ROW LEVEL SECURITY` a non-bypass role with the GUC unset filters every row), while the runtime must be `NOBYPASSRLS`. So the single role is split, not renamed:

- **Runtime** — new `google_sql_user.app_rls` (`lifting_app`, `NOSUPERUSER NOBYPASSRLS`). The Cloud Run services + GKE API pods keep reading the existing `…-database-url` secret, whose value now connects as `lifting_app` and carries `?connection_limit=N`.
- **Migrator** — unchanged `google_sql_user.app` (`lifting-logbook-app`, owner). The migrate Job now reads a new `…-migrator-database-url` secret.

A rejected alternative — renaming the one shared user (the original plan) or `ALTER ROLE … NOBYPASSRLS` — would point the **migrate Job** at the non-privileged role and break migrations.

## Changes (infra + docs only — no app code)

- `infra/terraform/main.tf` — `random_password.app_rls_password` + `google_sql_user.app_rls`; repoint `database_url` version to `lifting_app` + `?connection_limit`; new `migrator_database_url` secret + version; `local.db_connection_limit` sizing.
- `infra/terraform/cloud-run.tf` — migrate Job `DATABASE_URL` → `migrator_database_url`.
- `docs/deploy.md` — two-role split, import contingency, staging validation gate, cutover rollback.

`api_workload` already holds **project-level** `secretmanager.secretAccessor` (gke.tf), so the migrator secret needs no new IAM grant.

### Connection-pool sizing
`local.db_connection_limit = floor(max_connections × 0.8 / (api_max_instances × consumer_factor))`. `max_connections` is the tier default (not a flag): db-f1-micro ≈ 25 (staging), db-g1-small ≈ 50 (prod). `consumer_factor` is 2 on staging (Cloud Run **and** the GKE A/B share the pool, ADR-009) and 1 on prod (`enable_gke=false`). Computed: **staging 3, prod 4**. Verified against `SHOW max_connections` at the staging gate before prod.

## First-apply contingency (operational)

`lifting_app` is created idempotently by the enable_rls migration (no password). The first `terraform apply` per workspace that adds `google_sql_user.app_rls` may need a one-time import so Terraform adopts the existing role rather than failing to recreate it:

```
terraform import google_sql_user.app_rls "<project>/<instance>/lifting_app"   # per workspace
```

If `terraform plan` (staging) shows a clean create with no conflict, no import is needed. **This is the one thing to watch on the staging apply.**

## Staging validation gate (post-merge, before prod)

1. `SELECT current_user, pg_has_role('lifting_app','cloudsqlsuperuser','member')` → `lifting_app | f`
2. With the GUC unset, a `userId` table returns **zero** rows (fail-closed proof)
3. Authenticated GET to a user-data endpoint returns the caller's rows
4. `pg_stat_activity` count stays within the pool budget under load

Then **≥24h soak**, then the manual production gate.

## Testing

Infra-only; no Jest-testable code. `terraform fmt` clean and **`terraform validate` → Success** (offline `-backend=false`). Per-workspace `terraform plan` review (must show **no destroy** of `google_sql_user.app` and no instance replacement) is part of the gated apply. Pre-PR greps (suppression / test-integrity) clean; no `package.json` change (lockfile N/A).

**Tests: N/A — infrastructure-only change; validation is `terraform validate` + the staging validation gate above.**

## Risk

- **Resilience/rollback** — non-destructive apply (no user replace, existing password untouched → live connections undisturbed). Rollback = re-point `database-url` to the owner URL + redeploy (documented in `docs/deploy.md`). Migrate Job unaffected.
- **Data integrity** — migrator/runtime split keeps data migrations running with BYPASSRLS (e.g. #531-style row DELETEs); runtime enforces isolation. No schema change.
- **Security** — this *is* the fix: DB-level tenant isolation now enforced. Passwords stay in Secret Manager.

Closes #517

---

`/review` (claude-opus-4-8) on this PR surfaced **1 blocking, 3 non-blocking**. All addressed:

- **[blocking][reliability]** First `terraform apply` would 409 because `lifting_app` already exists (enable_rls ran in both envs) and the CI apply can't self-import — **fixed in `34d379f`**: `docs/deploy.md` reworded from "may need" → required, with the explicit `init` + `workspace select` + `import` sequence against the shared remote state to run before the first apply.
- **[non-blocking][security]** Runtime still carries owner/superuser creds via `SYSTEM_DATABASE_URL` (unused on the Prisma path, but bypasses RLS) — **filed #534** and deferred to keep this gated cutover non-destructive (removing it spans Cloud Run + GKE helm + api-service.yaml).
- **[non-blocking][maintainability]** `api_max_instances` duplicated the Cloud Run maxScale literal → drift risk vs the pool formula — **fixed in `34d379f`**: `cloud-run.tf` `max_instance_count` now references `local.api_max_instances` (single source of truth).
- **[non-blocking][documentation]** `$STG_RUNTIME_URL` undefined / would point at an unreachable private IP — **fixed in `34d379f`**: `docs/deploy.md` now defines it (proxy host `127.0.0.1:5434`, `lifting_app`, `sslmode=disable`).

`terraform fmt` clean and `terraform validate` → Success after the fixes.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -522,6 +522,58 @@ The non-Prisma `user_data_source` table (managed outside Prisma) lives in
 `infra/migrations/`; both scripts apply `infra/migrations/001_create_user_data_source.sql`
 after the Prisma migrations.
 
+### Row-Level Security cutover — two-role split (#517)
+
+The app database enforces per-tenant isolation with Postgres Row-Level Security (migration
+`20260611000000_enable_rls`, ADR-010). RLS is **bypassed by `cloudsqlsuperuser` members**, so the
+runtime and the migrator connect as **different roles**:
+
+- **Runtime** (`google_sql_user.app_rls`, `lifting_app`) — `NOSUPERUSER NOBYPASSRLS`. The Cloud
+  Run services and the GKE API pods read `lifting-logbook-<env>-database-url`, which connects as
+  this role (so the policies enforce) and carries `?connection_limit=N` to cap the Prisma pool.
+- **Migrator** (`google_sql_user.app`, `lifting-logbook-app`) — the owner/superuser. The migrate
+  Cloud Run Job reads the separate `lifting-logbook-<env>-migrator-database-url` so it can run DDL
+  and data migrations (which `FORCE ROW LEVEL SECURITY` would otherwise filter to zero rows).
+
+`connection_limit` is derived in `infra/terraform/main.tf` (`local.db_connection_limit`) from the
+tier's `max_connections` and the Cloud Run maxScale; staging divides by an extra factor because the
+ADR-009 GKE A/B deployment shares the same pool. `max_connections` is the tier default (not a flag)
+— verify it before prod with `SELECT setting FROM pg_settings WHERE name='max_connections';`.
+
+**First-apply contingency.** The `lifting_app` role is created idempotently by the enable_rls
+migration (with no password). The first `terraform apply` per workspace that adds
+`google_sql_user.app_rls` may need a one-time import so Terraform adopts the existing role instead
+of failing to recreate it:
+
+```bash
+terraform import google_sql_user.app_rls "<project>/<instance>/lifting_app"   # per workspace
+```
+
+**Staging validation gate (run after the staging deploy, before any prod cutover):**
+
+```bash
+# 1. App connects as the NOBYPASSRLS role (must print: lifting_app | f)
+#    via the Cloud SQL Auth Proxy (scripts/migrate-staging-db.sh sets one up on :5434)
+psql "$STG_RUNTIME_URL" -c "SELECT current_user, pg_has_role('lifting_app','cloudsqlsuperuser','member');"
+# 2. With the GUC unset, a userId table returns ZERO rows (fail-closed proof)
+psql "$STG_RUNTIME_URL" -c "SELECT count(*) FROM training_max;"   # expect 0
+# 3. Connection count stays within the pool budget under load
+psql "$STG_RUNTIME_URL" -c "SELECT count(*) FROM pg_stat_activity;"
+```
+
+Then soak staging ≥24h before approving the production deploy.
+
+**Rolling back the cutover.** Re-point the runtime back to the owner role by adding a new version of
+the `database-url` secret with the old `postgresql://lifting-logbook-app:…` value, then redeploy the
+API (Cloud Run resolves `:latest` at deploy time):
+
+```bash
+gcloud secrets versions add lifting-logbook-prod-database-url --data-file=- --project=lifting-logbook-prod   # paste owner URL
+# then redeploy the API revision (see Rolling back, above)
+```
+
+The migrate Job is unaffected by a rollback (it uses the separate migrator secret).
+
 ### OTel Collector / Grafana Cloud telemetry
 
 **The deploy pipeline deploys the OTel Collector DaemonSet automatically (#474).** On every
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -540,20 +540,37 @@ tier's `max_connections` and the Cloud Run maxScale; staging divides by an extra
 ADR-009 GKE A/B deployment shares the same pool. `max_connections` is the tier default (not a flag)
 — verify it before prod with `SELECT setting FROM pg_settings WHERE name='max_connections';`.
 
-**First-apply contingency.** The `lifting_app` role is created idempotently by the enable_rls
-migration (with no password). The first `terraform apply` per workspace that adds
-`google_sql_user.app_rls` may need a one-time import so Terraform adopts the existing role instead
-of failing to recreate it:
+**First-apply: import is required, not optional.** The `lifting_app` role was already created by the
+`enable_rls` migration (with no password) in both staging and production — confirmed live before this
+change. So `terraform plan` will show a `create` for `google_sql_user.app_rls`, and an unguarded
+`terraform apply` will **fail** with `409 ALREADY_EXISTS` (the Cloud SQL API does not upsert users).
+The deploy pipeline's CI `terraform apply` has no path to recover from this on its own, so the import
+must be run **before** the first apply that introduces `app_rls`, once per workspace, against the same
+GCS-backed remote state the pipeline uses (not a throwaway local state):
 
 ```bash
-terraform import google_sql_user.app_rls "<project>/<instance>/lifting_app"   # per workspace
+cd infra/terraform
+terraform init -backend-config="bucket=lifting-logbook-tfstate" -backend-config="prefix=terraform/state"
+terraform workspace select staging   # then repeat for production before the prod cutover
+terraform import google_sql_user.app_rls "<project>/<instance>/lifting_app"
 ```
+
+After the import, `terraform plan` shows only a password set on the adopted role (no create), and the
+pipeline apply proceeds cleanly. (If a future fresh environment has *not* run `enable_rls` yet, the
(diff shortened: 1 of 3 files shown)


brownm09:
## Review: [infra] RLS production cutover: runtime lifting_app role + pooled DATABASE_URL

### Summary
**Merge with changes — the design is sound, but one reliability gap in the cutover path should be addressed before the staging apply.** The two-role split (runtime `lifting_app` NOBYPASSRLS via the existing `database-url` secret + `?connection_limit`; migrator stays owner via a new `migrator-database-url` secret) is the correct shape, and keeping the migrate Job on the owner role is right — renaming the single shared user would have broken migrations. The connection-pool math checks out and is non-destructive (no user replace). The highest-severity item is that the first `terraform apply` will *deterministically* fail (not "may"), because the live investigation in this PR already confirmed `lifting_app` exists — and the CI pipeline has no path to perform the documented `terraform import`. Gate checks: doc-reconciliation N/A (repo has no Documentation Maintenance table); coverage gate N/A (infra-only, PR body states so); test-integrity N/A (no test files); no README coverage gap (`docs/deploy.md` updated; `infra/terraform/` has no README).

---

### Blocking Findings

**[reliability]** `infra/terraform/main.tf:235` (and `docs/deploy.md` first-apply contingency)
`google_sql_user.app_rls` is a new resource, so `terraform plan` emits a `create` for the `lifting_app` Cloud SQL user. But this PR's own live investigation confirms the role **already exists** (created by the `20260611000000_enable_rls` migration, which has run in both envs). `google_sql_user` create against an existing user returns `409 ALREADY_EXISTS` — the apply hard-fails. Merging triggers the staging pipeline's `terraform apply`, which runs in CI against the GCS-backed remote state; there is no automated step that runs `terraform import`, so the staging cutover will fail on first apply and block until a human runs the import against the CI workspace. The docs frame this as "may need … if `terraform plan` shows a clean create" — given the confirmed live state, it *will* need it, not *may*.
**Fix:** Before merge, run `terraform import google_sql_user.app_rls "<project>/<instance>/lifting_app"` against the staging (then prod) remote-state workspace so the apply is clean — or add a guarded pre-apply import step to the pipeline. Reword the PR body and `docs/deploy.md` from "may need" to "requires" and state explicitly *where* the import is run (which backend config + `terraform workspace select`), since a local import must target the same remote state the CI apply uses.

---

### Non-Blocking Findings

**[security]** `infra/terraform/main.tf:268` — `system_database_url` left on the owner role
The runtime Cloud Run/GKE API still receives `SYSTEM_DATABASE_URL` pointing at the owner/superuser role (`lifting-logbook-app`), which bypasses RLS on *every* database in the instance — including the main app DB. With `DATABASE_URL` set, the active factory is `PrismaRepositoryFactory` and `SystemDbRepositoryFactory` is never constructed (`repository-factory.module.ts:27-32`), so this owner credential is injected into the runtime but unused on the deployed path. That means a compromised runtime could connect with the owner URL and defeat the tenant isolation #517 exists to enforce. Pre-existing (this PR doesn't add the line), but it directly caps the security win.
**Suggestion:** Confirm the deployed runtime needs `SYSTEM_DATABASE_URL` at all (it doesn't on the Prisma path); if not, drop it from the API service env, or grant `lifting_app` least-privilege access to the `_system` DB and repoint `SYSTEM_DATABASE_URL` at `lifting_app`. File as a follow-up if out of scope for the cutover.

**[maintainability]** `infra/terraform/main.tf:53` — `api_max_instances` duplicates the Cloud Run literal
`local.api_max_instances = production ? 10 : 3` is a hand-maintained mirror of `max_instance_count` in `cloud-run.tf:31`. The comment says "mirrors the maxScale literal," but nothing enforces it — bump maxScale in `cloud-run.tf` without updating this local and the pool sizing silently under-counts, over-subscribing `max_connections`. On these small tiers that's exactly the failure mode the cap is meant to prevent.
**Suggestion:** Derive both from one source (a shared `local` or variable referenced by both `scaling.max_instance_count` and the sizing formula) so they can't drift.

**[documentation]** `docs/deploy.md` staging validation gate — `$STG_RUNTIME_URL` is undefined
The gate's `psql "$STG_RUNTIME_URL" …` commands point at the proxy on `:5434`, but the `database-url` secret value carries a **private-IP** host that won't resolve from a local machine through the Auth Proxy. An operator who exports the raw secret value into `$STG_RUNTIME_URL` will hang.
**Suggestion:** Define `$STG_RUNTIME_URL` explicitly (host `127.0.0.1:5434`, user `lifting_app`, `sslmode=disable`) or note that the secret's host must be rewritten to the proxy address before use.

---

### Findings Disposition
Per ADR-028, each finding above must be fixed in this PR or filed-and-linked before merge, recorded in a "Review findings disposition" section of the PR body. Note the blocking finding is procedural (sequence the import before the staging apply) rather than a code change — its disposition should state how/where the import runs.

---

Issue #489, opened 2026-06-10
[maintainability] Smart Import: consolidate per-kind dispatch, amrap normalization, wizard step constants


brownm09:
## Problem

Maintainability follow-ups surfaced by the `/code-review` of PR #485 (#477 Smart Import Phase 1):

1. **Four-way `switch (destination)` repeated across five methods.** `ImportController` enumerates the four `ImportKind`s in `preview`, `commit`, `parseAndValidate`, and `parseFor`; the three validators share a near-identical `forEach + rowErrors + {valid, errors}` shape. Adding a fifth kind means editing five parallel switches. (`apps/api/src/programs/import.controller.ts`)
2. **`amrap` normalization duplicated.** `r.amrap === true || r.amrap === 'TRUE'` and `weekType ?? null` coercion is repeated across the program-spec write path, `specValue`, and `programSpecComparable`. A future site that forgets the `=== 'TRUE'` branch silently desyncs preview from commit. (`apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts`, `packages/core/src/utils/import/buildImportPreview.ts`)
3. **Wizard step magic numbers.** `ImportWizard.tsx` drives a 7-step flow with bare integer `setStep(n)` / `step === n` across ~10 sites and a `STEP_LABELS` array indexed by raw ints; inserting/removing a step (the Phase-1 placeholder Map step) means hunting every literal. (`apps/web/app/(authed)/import/ImportWizard.tsx`)

## Proposed fix

- Introduce a per-kind handler table (`Record<ImportKind, { parse, validate, preview, commit }>`) so each kind is defined once.
- Extract `normalizeAmrap(row): boolean` (ideally normalize at the parse boundary so the model carries a real boolean) and call it everywhere.
- Add a `Step` const/enum (`SOURCE=0, ANALYZING=1, …`) and reference it throughout the wizard.

## Context

`default: throw` guards were added to the controller switches in PR #485 so an unhandled kind now fails loudly; this issue tracks the structural consolidation.

brownm09:
**Sequencing:** consolidation/cleanup. Build **LAST** — after #488 and the Phase 2/3 feature PRs (#483/#484) — to avoid merge churn (it touches the same switches/files). Full epic build-order: brownm09/lifting-logbook#488 (comment).

PR #536, opened 2026-06-14
[chore] Unify import-commit methods + classify-and-count helper + P2028 mapping (#532)

### LINKED ISSUE #532: [chore] Unify import-commit methods + classify-and-count helper + P2028 mapping (follow-up from #488)
Follow-up from the `/review` of #531 (Smart Import commit atomicity, #488). These are non-blocking maintainability/reliability improvements that are cleanly separable from #488's reliability goal (which shipped). Source: PR #531 review comment.

### 1. Unify the three import-commit methods on one shape
`importTrainingMaxes` / `importGoals` classify from a single up-front `findMany` then upsert; `saveProgramSpec` still does a per-row `findFirst` (no longer needed for race-safety now that `@@unique([programId,week,offset,lift,order])` exists — the upsert's `where` handles the race) plus a hand-duplicated where-clause.

- [ ] Refactor `saveProgramSpec` (`apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts`) to the `findMany` → classify-in-memory → `upsert` shape, removing the per-row `findFirst` and the duplicated natural-key where-clause (saves a round-trip/row).
- [ ] Consider a `programSpecRowKind` shared classifier mirroring `trainingMaxRowKind`/`strengthGoalRowKind`.
- [ ] Optional: rename `saveProgramSpec` → `importProgramSpec` for symmetry (`SaveProgramSpecResult` is already an alias of `ImportWriteResult`), or document the intentional asymmetry on the port.

### 2. Extract a shared classify-and-count helper
The seen-set dedupe + create/update/skip tally loop is copy-pasted across four adapters (in-memory + Prisma × TM/goal). The *decision* is already shared via `*RowKind`; the surrounding loop is not.

- [ ] Extract `classifyAndCount<T>(rows, keyOf, rowKind, applyWrite)` (core or a shared adapter util) so a future dedupe-semantics change can't desync counts across adapters.
- [ ] Add a test asserting in-memory and Prisma adapters return identical counts for the same input.

### 3. Map P2028 (transaction timeout) on the import-commit path
When `runInteractive` opens its own `$transaction` (non-RLS / unit paths), a large import exceeding Prisma's default 5s interactive-tx timeout throws P2028, which the per-feature exception filters don't map → generic 500. In production the path runs inside the request RLS tx (bounded), so this is low-severity.

- [ ] Map P2028 to a 503/timeout response (or pass an explicit `{ timeout, maxWait }` for batch imports) if batch sizes can grow.

_Filed from the #531 review per ADR-028 (fix-or-file). No production impact today._

brownm09:
## Summary

Follow-up from the `/code-review` of #531 (Smart Import commit atomicity, #488). Three non-blocking maintainability/reliability cleanups, cleanly separable from #531's shipped reliability goal.

**1. Unify the program-spec commit on the `findMany → classify → upsert` shape.**
`saveProgramSpec` did a per-row `findFirst` plus a hand-duplicated natural-key where-clause. The `@@unique([programId,week,offset,lift,order])` index (#488) already makes the `upsert` race-safe, so the per-row find-then-write is no longer needed. It now takes **one up-front snapshot read** and classifies in memory — N round-trips → 1.

**2. Extract a shared classify-and-count helper.**
`classifyAndCount<T>(rows, keyOf, rowKind, applyWrite)` (in `packages/core`) centralises the seen-set dedupe + create/update/skip tally loop that was copy-pasted across the four TM/goal adapters — now six sites (in-memory + Prisma × TM/goal/spec) plus the preview path route through it. A new `programSpecRowKind` classifier mirrors `trainingMaxRowKind`/`strengthGoalRowKind`, so the create-vs-update *decision* and the surrounding *loop* are both shared. A dedupe/tally change can no longer desync counts across adapters, or between preview and commit.

**3. Map Prisma P2028 (interactive-tx timeout) on the import path.**
A global `@Catch(PrismaClientKnownRequestError)` filter maps **P2028 → HTTP 503** (actionable "try a smaller file" instead of a generic 500), and delegates every other Prisma code to `BaseExceptionFilter` so the framework's default 500 **and Nest's error logging** are preserved. Batch imports also get a 30 s self-opened-tx timeout (`IMPORT_BATCH_TX_OPTIONS`), applied only on the non-RLS / system-DB path where `runInteractive` opens its own transaction (on the request path it reuses the RLS tx, whose timeout governs and whose P2028 the RLS interceptor already re-throws — now mapped by this filter too).

### Naming asymmetry decision
`saveProgramSpec` keeps its name rather than renaming to `importProgramSpec`. The issue lists the rename as optional; renaming would churn ~10 unrelated controller specs that mock the method. The **shape** is fully unified (returns the shared `ImportWriteResult`, classifies via the shared helpers); only the name differs, and the asymmetry is now documented on the port interface.

## Acceptance Criteria (#532)

- [x] Refactor `saveProgramSpec` to `findMany → classify-in-memory → upsert`; remove the per-row `findFirst` and the duplicated where-clause
- [x] Add a `programSpecRowKind` shared classifier mirroring `trainingMaxRowKind`/`strengthGoalRowKind`
- [x] `saveProgramSpec` rename → documented intentional asymmetry on the port (rename declined per scope)
- [x] Extract `classifyAndCount<T>(rows, keyOf, rowKind, applyWrite)` (core) used by all adapters
- [x] Test asserting in-memory and Prisma adapters return identical counts for the same input
- [x] Map P2028 to a 503/timeout response + explicit batch-import `{ timeout, maxWait }`

## Testing

`npm test` (full suite, Docker up for the API DB E2E via Testcontainers) + `npm run build` + `npm run lint` — all green.

- **core** — Tests: 203 passed, 0 skipped, 0 failed (31 suites)
- **api** — Tests: 393 passed, 0 skipped, 0 failed (33 suites)
- **web** — Tests: 91 passed, 0 skipped, 0 failed (16 suites)

New coverage: `classifyAndCount` unit spec (dedupe, skip, tally, mid-batch-throw propagation); `import-count-parity.spec.ts` (in-memory vs Prisma identical counts for TM/goal/spec across create/update/skip/dedupe); `saveProgramSpec` mock-tx tests (owner guard, NotFound, single-snapshot classification); `transaction-timeout.filter.spec.ts` (P2028 → 503, other codes delegate to base filter). Existing TM/goal repo specs and the in-memory import E2E stay green unchanged.

## Suppression / test-integrity / lockfile

- Suppression grep: two **pre-existing** matches only — `weekType: r.weekType ?? null` (a domain→SQL-NULL column mapping, re-indented inside the rewritten method, substance unchanged; not error-silencing) and the `eslint-disable ... no-explicit-any` shown as unchanged context from `main.ts`'s `@fastify/multipart` registration. **No new suppressions.**
- Test-integrity grep: clean. No skips added, no test files deleted, no thresholds lowered — tests **added**.
- No `package.json` changed → lockfile-sync N/A.

## Risk audit (Pass 3)

1. **Testing** — new unit + parity + filter specs above; full suite green.
2. **Observability** — the P2028 filter delegates non-P2028 Prisma errors to `BaseExceptionFilter`, preserving Nest's default error logging (no regression). No new raw SQL; writes still run inside the existing RLS span.
3. **Security** — no new endpoints/inputs; the program owner guard in `saveProgramSpec` is retained; RLS GUC handling unchanged.
4. **Resilience** — P2028 now mapped to 503; transaction atomicity unchanged (`runInteractive`); revert = standard PR revert.
5. **Performance** — `saveProgramSpec`: N per-row `findFirst` → 1 `findMany` + upserts (fewer round-trips). Not a hot path.
6. **Data integrity & migrations** — **no schema change** (the unique index shipped in #531); count parity guarded by the shared helper + the parity test.

Closes #532

---

## Review findings disposition (`/review` #536)

`/review` ran high-effort parallel correctness/security + reliability/perf/maintainability passes. Correctness/security pass: no findings.

**Blocking:** none.

**Non-blocking:**
- **[reliability] Import tx budget didn't reach the production path** — `IMPORT_BATCH_TX_OPTIONS` (30s) only governed the self-opened-tx path; the real HTTP import runs inside the per-request RLS transaction (15s `DEFAULT_RLS_TX_TIMEOUT_MS`), so a large within-limit import could P2028 at 15s despite the intended 30s. → **Fixed in `14d33df`**: unified both paths on `IMPORT_TX_TIMEOUT_MS` and annotated `ImportController.import` with `@RlsTxTimeout(IMPORT_TX_TIMEOUT_MS)`; added a metadata regression test.
- **[maintainability] Two timeout values coexisted with no single source of truth** → **Fixed in `14d33df`** (folded into the `IMPORT_TX_TIMEOUT_MS` constant shared by the decorator and `IMPORT_BATCH_TX_OPTIONS`).
- **[maintainability] Preview builders still hand-roll the dedupe/tally loop** that the commit adapters now share via `classifyAndCount` → **Filed #537** (non-trivial: preview emits per-row deltas, not just counts; fits the #489 consolidation epic).
- **[performance] Sequential per-row `tx.upsert` inside one import transaction** → **Filed #537** (the raw-SQL batch alternative trades round-trips for an untraced-SQL observability obligation per ADR-024; mitigated for now by the 30s budget; import is a manual non-hot path).

**Accepted by design (no change):**
- `classifyAndCount`'s `applyWrite(row, kind)` second param is unused by the six callsites but kept as a deliberate, tested affordance; the `void | Promise<unknown>` return union is intentional so terse Prisma upsert callbacks type-check. The "skip rows never call applyWrite" + "applyWrite must return its promise" contract is covered by the unit spec.
- The P2028 filter is instance-registered (not `APP_FILTER`) with a manually-passed `httpAdapter`, matching the two sibling domain filters — verifier confirmed no defect.

Tests after the fix: **api 394 passed, 0 skipped, 0 failed** (was 393; +1 regression test); lint 7/7; suppression/integrity greps clean; no `package.json` change.

### DIFF
--- a/apps/api/src/adapters/import-count-parity.spec.ts
+++ b/apps/api/src/adapters/import-count-parity.spec.ts
@@ -1,0 +1,154 @@
+import { PrismaClient } from '@prisma/client';
+import {
+  LiftingProgramSpec,
+  StrengthGoalEntry,
+  TrainingMax,
+} from '@lifting-logbook/core';
+import { InMemoryTrainingMaxRepository } from './in-memory/training-max.adapter';
+import { InMemoryStrengthGoalRepository } from './in-memory/strength-goal.adapter';
+import { InMemoryLiftingProgramSpecRepository } from './in-memory/lifting-program-spec.adapter';
+import { PrismaTrainingMaxRepository } from './prisma/training-max.repository';
+import { PrismaStrengthGoalRepository } from './prisma/strength-goal.repository';
+import { HybridLiftingProgramSpecRepository } from './prisma/hybrid-program-spec.repository';
+
+/**
+ * Cross-adapter parity (#532): the in-memory and Prisma import-commit methods must
+ * return identical {created, updated, skipped} for the same input. Both now route
+ * through the shared `classifyAndCount` loop + the shared `*RowKind` classifiers,
+ * so a future change to dedupe/tally semantics in one place can't desync the other.
+ *
+ * The in-memory adapter is seeded by importing the "existing" rows first (all
+ * creates); the Prisma adapter is given the same existing snapshot via a mock
+ * `findMany` whose `$transaction` runs the callback against an in-tx stub.
+ */
+
+const USER = 'user-1';
+const PROGRAM = '5-3-1';
+const SPEC_PROGRAM = '11111111-1111-4111-8111-111111111111';
+const at = new Date('2026-01-01T00:00:00.000Z');
+
+describe('import-commit count parity (in-memory vs Prisma)', () => {
+  it('training maxes: identical counts for create/update/skip/dedupe', async () => {
+    const existing: TrainingMax[] = [
+      { lift: 'Squat', weight: 300, dateUpdated: at },
+      { lift: 'Deadlift', weight: 350, dateUpdated: at },
+    ];
+    const incoming: TrainingMax[] = [
+      { lift: 'Squat', weight: 300, dateUpdated: at }, // identical → skip
+      { lift: 'Deadlift', weight: 400, dateUpdated: at }, // differs → update
+      { lift: 'Bench', weight: 200, dateUpdated: at }, // absent → create
+      { lift: 'Squat', weight: 999, dateUpdated: at }, // duplicate lift → collapsed
+    ];
+
+    const mem = new InMemoryTrainingMaxRepository();
+    await mem.importTrainingMaxes(PROGRAM, existing);
+    const memResult = await mem.importTrainingMaxes(PROGRAM, incoming);
+
+    const tx = {
+      trainingMax: {
+        findMany: jest.fn().mockResolvedValue(existing.map((m) => ({ lift: m.lift, weight: m.weight }))),
+        upsert: jest.fn().mockResolvedValue({}),
+      },
+    };
+    const prisma = {
+      $transaction: jest.fn((cb: (t: typeof tx) => unknown) => cb(tx)),
+    } as unknown as PrismaClient;
+    const prismaResult = await new PrismaTrainingMaxRepository(prisma, USER).importTrainingMaxes(
+      PROGRAM,
+      incoming,
+    );
+
+    expect(memResult).toEqual({ created: 1, updated: 1, skipped: 1 });
+    expect(prismaResult).toEqual(memResult);
+  });
+
+  it('strength goals: identical counts for create/update/skip/dedupe', async () => {
+    const existing: StrengthGoalEntry[] = [
+      { lift: 'Squat', goalType: 'absolute', unit: 'lbs', target: 400, updatedAt: at },
+      { lift: 'Deadlift', goalType: 'absolute', unit: 'lbs', target: 500, updatedAt: at },
+    ];
+    const incoming: StrengthGoalEntry[] = [
+      { lift: 'Squat', goalType: 'absolute', unit: 'lbs', target: 400, updatedAt: at }, // skip
+      { lift: 'Deadlift', goalType: 'absolute', unit: 'lbs', target: 505, updatedAt: at }, // update
+      { lift: 'Bench', goalType: 'relative', unit: 'lbs', ratio: 1.5, updatedAt: at }, // create
+      { lift: 'Squat', goalType: 'absolute', unit: 'lbs', target: 999, updatedAt: at }, // collapsed
+    ];
+
+    const mem = new InMemoryStrengthGoalRepository();
+    await mem.importGoals(PROGRAM, existing);
+    const memResult = await mem.importGoals(PROGRAM, incoming);
+
+    const tx = {
+      strengthGoal: {
+        findMany: jest.fn().mockResolvedValue(
+          existing.map((g) => ({
+            lift: g.lift,
+            goalType: g.goalType,
+            unit: g.unit,
+            target: g.target ?? null,
+            ratio: g.ratio ?? null,
+            updatedAt: g.updatedAt,
+          })),
+        ),
+        upsert: jest.fn().mockResolvedValue({}),
+      },
+    };
+    const prisma = {
+      $transaction: jest.fn((cb: (t: typeof tx) => unknown) => cb(tx)),
+    } as unknown as PrismaClient;
+    const prismaResult = await new PrismaStrengthGoalRepository(prisma, USER).importGoals(
+      PROGRAM,
+      incoming,
+    );
+
+    expect(memResult).toEqual({ created: 1, updated: 1, skipped: 1 });
+    expect(prismaResult).toEqual(memResult);
+  });
+
+  it('program spec: identical counts for create/update/skip/dedupe', async () => {
+    const spec = (lift: string, sets: number): LiftingProgramSpec => ({
+      week: 1,
+      offset: 0,
+      lift,
+      order: 1,
+      increment: 5,
+      sets,
+      reps: 5,
+      amrap: false,
+      warmUpPct: '0.65',
+      wtDecrementPct: 0,
+      activation: 'main',
+    });
+    const toRow = (s: LiftingProgramSpec) => ({ ...s, weekType: null });
+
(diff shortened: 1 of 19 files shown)


brownm09:
## Review: [chore] Unify import-commit methods + classify-and-count helper + P2028 mapping (#532)

### Summary
**Merge-ready** (findings addressed). High-effort parallel review of a 906-line, 17-file refactor. The correctness/security pass found nothing — dedupe is consistently first-wins (no behavior change vs. the loops it replaces), every adapter classifies against a pre-write snapshot, the program-spec compound-key upsert is equivalent and race-safe given the #531 unique index, the owner/RLS guards are preserved, and the P2028 filter delegates non-P2028 codes correctly with no double-send. The single highest-severity finding was reliability: the new 30s import transaction budget did not actually reach the production import path. It is now fixed in this PR. Scope is appropriate — a coherent follow-up, not over-broad.

---

### Non-Blocking Findings

**[reliability]** `apps/api/src/adapters/prisma/prisma-tx.util.ts:50` / `apps/api/src/programs/import.controller.ts:76`
`IMPORT_BATCH_TX_OPTIONS` (30s) is only applied when `runInteractive` opens its own transaction. The production HTTP import (`ImportController.import`) carries no `@SkipRlsTransaction`/`@RlsTxTimeout`, so its commit runs **inside the per-request RLS transaction**, whose timeout is `DEFAULT_RLS_TX_TIMEOUT_MS` (15s) — `IMPORT_BATCH_TX_OPTIONS` is ignored on that path. So a large within-limit import (`MAX_IMPORT_ROWS=5000`, sequential upserts) could P2028 at 15s and return the "split the file" 503, despite the PR's intended 30s budget. The effective import budget was half the advertised value and set by an unrelated RLS constant.
**Fix (done, `14d33df`):** Unified both paths on a single exported `IMPORT_TX_TIMEOUT_MS` and annotated `ImportController.import` with `@RlsTxTimeout(IMPORT_TX_TIMEOUT_MS)`; the RLS interceptor reads that metadata as the request-tx timeout, so the production path now gets 30s. Added `import.controller.spec.ts` asserting the metadata so the budget can't silently revert.

**[maintainability]** `apps/api/src/adapters/prisma/prisma-tx.util.ts:50`
Two timeout values (15s RLS, 30s batch) coexisted with no single source of truth — the kind of drift that lands a future timeout change in the wrong place.
**Fix (done, `14d33df`):** Both the decorator and `IMPORT_BATCH_TX_OPTIONS` now reference one `IMPORT_TX_TIMEOUT_MS` constant.

**[maintainability]** `packages/core/src/utils/import/buildImportPreview.ts`
The PR routes the six commit adapters through `classifyAndCount`, but the three preview builders still hand-roll the identical `seen`-set dedupe + `*RowKind` + tally loop — the exact divergence #488/#532 set out to close. Counts can't diverge today (shared classifier), but the loop lives in two places.
**Suggestion:** Filed **#537** — non-trivial because preview emits per-row deltas, not just counts; fits the #489 consolidation epic.

**[performance]** `apps/api/src/adapters/prisma/{training-max,strength-goal,hybrid-program-spec}.repository.ts`
Import commit awaits `tx.upsert` sequentially per row inside one transaction; 5000 rows = thousands of serial round-trips holding a pooled connection on small Cloud SQL tiers.
**Suggestion:** Filed **#537**. The raw-SQL batch alternative trades round-trips for an untraced-SQL observability obligation (ADR-024 — Prisma raw SQL isn't auto-traced, needs a manual span); mitigated for now by the 30s budget on a manual, non-hot path.

---

### Style Notes
None — the diff is clean and idiomatic; the new code matches surrounding repository/adapter conventions.

---

### Findings Disposition
- **[reliability] request-path import budget** — fixed in `14d33df`
- **[maintainability] two coexisting timeouts** — fixed in `14d33df`
- **[maintainability] preview/commit loop duplication** — filed #537
- **[performance] sequential per-row upserts** — filed #537
- **[maintainability] `classifyAndCount` unused `kind` param / return union** — accepted by design (tested affordance; intentional union)
- **[reliability] instance-registered P2028 filter** — no defect (verifier-confirmed; matches sibling filters)

---

PR #544, opened 2026-06-16
[infra] Surface prod Terraform plan before the production approval gate (#542)

brownm09:
## Summary

Closes #542.

On 2026-06-15 an unrelated CI-fix merge (#540) made `deploy-production` reachable for the first time in days, and a one-click approval of the `production` gate executed the **entire RLS prod cutover** (#517) — a prod-affecting Terraform diff (#533) that had sat dormant in `main`, applied to staging only because `deploy-production` had been continuously skipped. The approval gate shows *that* a deploy is pending, not *what* `terraform apply (production)` will change, so the blast radius was invisible at the moment of approval.

This implements **Layer A** (the settled design in the issue comment): a **read-only** `plan-production` job in `deploy.yml` that surfaces the prod blast radius in the run *before* the approval gate.

### Changes

1. **New `plan-production` job** (`.github/workflows/deploy.yml`)
   - `needs: [preflight, build-images]`, un-gated (no `environment:`), `if: always() && needs.build-images.result == 'success'` — runs in both staging-enabled and production-only modes.
   - Mirrors `deploy-production`'s prod auth (`google-github-actions/auth@v3` with the prod WIF provider/SA) + `setup-terraform` + the bounded-retry `init` / `workspace select production` loop.
   - Runs `terraform plan -no-color -input=false` against the **production** workspace (read-only — cannot mutate prod) and writes the `Plan: N to add, N to change, N to destroy` line + resource addresses to `$GITHUB_STEP_SUMMARY`. A non-empty plan is flagged with the #542 incident context; an empty plan prints "No changes".
   - Runs only on `push: main` (deploy.yml's sole trigger), so prod creds are **never** exposed on untrusted PR runs.

2. **`plan-production` added to `deploy-production.needs`** — a job with an `environment:` gate does not prompt for approval until all `needs` complete, so this guarantees the plan summary exists before the approver is prompted. It stays **advisory**: `deploy-production`'s `if` (already `always()`-prefixed, unchanged) does not require `plan-production` to succeed, so a transient plan failure removes that run's summary but never blocks the deploy.

3. **`docs/deploy.md`** — the "Approving a production deploy" section now instructs the approver to read the `plan-production` summary before approving, with the #517/#540 incident context.

### Acceptance criteria
- [x] The `production` deploy approval has the `terraform plan` summary available to the approver (the gate cannot prompt until `plan-production` finishes).
- [x] Documented in `docs/deploy.md`, referencing the incident.
[...]

PR #551, opened 2026-06-17
[infra] Cut plan-production over to the read-only plan SA (#545 Phase 2)

### LINKED ISSUE #545: [infra] plan-production reuses the write-capable prod WIF SA and runs on every merge — consider a read-only plan SA
## Context

Follow-up from the [#544](https://github.com/brownm09/lifting-logbook/pull/544) review (finding 2, non-blocking [security]), spun off from #542.

PR #544 added a read-only `plan-production` job to `.github/workflows/deploy.yml`. It authenticates with the **same** `GCP_PROD_WORKLOAD_IDENTITY_PROVIDER` / `GCP_PROD_SERVICE_ACCOUNT` that `deploy-production` uses to *apply* — a write-capable prod service account — even though `terraform plan` only needs read/refresh.

Because `plan-production` runs on **every** `push: main` (not only when a deploy proceeds to the gate), the exposure window of the prod-apply identity widens from 'deploys' to 'every merge'.

## Why this is bounded (and was non-blocking)
- `deploy.yml`'s sole trigger is `push: main` — post-merge/trusted; there is no `pull_request` trigger, so prod creds are never exposed to untrusted PR runs.
- The same SA was already present in this workflow (the apply job).

## Proposed work
- Create a dedicated **read-only** prod service account for the plan job: state-bucket read + `roles/viewer`-equivalent on the planned resources, no write/apply permissions.
- Wire a new `GCP_PROD_PLAN_*` WIF provider/SA secret pair and point `plan-production`'s auth step at it, leaving the write SA scoped to `deploy-production`.
- Alternatively, if the marginal exposure is judged not worth a second SA, record that decision as a comment on the auth step so a future reviewer does not re-raise it.

Note: this is itself a prod IAM/Terraform change and would surface as a prod-affecting plan in the new `plan-production` summary — sequence it deliberately.

Part of the #542 hardening series.

brownm09:
## Summary

Phase 2 (final) of #545 — the CI cutover. The advisory `plan-production` job in `deploy.yml` now authenticates with the **dedicated read-only identity** created in Phase 1 (PR #548), instead of the write-capable prod CI service account.

- `service_account` → `${{ secrets.GCP_PROD_PLAN_SERVICE_ACCOUNT }}` (`lifting-logbook-prod-plan@lifting-logbook-prod.iam.gserviceaccount.com`): `roles/viewer` + `roles/secretmanager.secretAccessor` + `roles/storage.objectViewer` on the state bucket — **no** write / apply / `setIamPolicy` / owner.
- Adds `-lock=false` to `terraform plan`: the read-only SA has no state-bucket write and cannot acquire the `.tflock`. `plan-production` runs early (`needs: build-images` only) with no concurrent prod-state writer, so skipping the advisory lock is safe and is what keeps the identity strictly read-only.

`deploy-production`'s gated `terraform apply` is unchanged — it still uses the write-capable `GCP_PROD_SERVICE_ACCOUNT`. This narrows the read-only plan job (which now runs on every merge) from `roles/owner` to least privilege, the security finding from the #544 review.

## Pre-merge gates completed (verified against live GCP/GitHub)

The three manual gates from the phased rollout are done and audited:

| Gate | Verified |
|---|---|
| Prod apply (run [27654624051](https://github.com/brownm09/lifting-logbook/actions/runs/27654624051), `f574e87`) | ✅ SA `lifting-logbook-prod-plan` created; `roles/viewer` + `roles/secretmanager.secretAccessor`; WIF `workloadIdentityUser` on `attribute.repository/brownm09/lifting-logbook` |
| State-bucket grant | ✅ `roles/storage.objectViewer` on `gs://lifting-logbook-prod-tfstate` (read-only, not objectAdmin) |
| GitHub secret | ✅ `GCP_PROD_PLAN_SERVICE_ACCOUNT` set |

## Testing

CI-YAML-only change; no app code, test files, or `package.json` touched.

- `python -c "import yaml; yaml.safe_load(...)"` → **deploy.yml: YAML OK**
- Pre-commit `turbo run lint` → **7/7 successful**
- Diff is exactly two changes: the `service_account` input and `-lock=false` on the plan step.
- **End-to-end verification is the first post-merge `Deploy` run** (same posture as #544/#545 Phase 1): its `plan-production` job authenticates as the read-only SA and must still emit the add/change/destroy run-summary. Because the job is **advisory** (#544), any residual permission gap degrades the summary rather than blocking `deploy-production` — roles can be iterated in a fast follow if a gap surfaces.

No new testable behavior (CI config); coverage gate N/A. No suppressions / test-integrity changes. Lockfile unchanged.

## Risk

- **Security** — this is the improvement: drops the on-every-merge plan job from `roles/owner` to read-only. Documented caveat (carried from Phase 1): `secretAccessor` means the SA can *read* prod secret payloads during plan refresh — strictly less than owner, but not zero-secret; a secretless plan isn't possible while the config manages `google_secret_manager_secret_version` resources.
- **Resilience / rollback** — plan stays advisory; revert this PR to restore the previous identity. The Phase 1 SA + bindings are additive and harmless if this is reverted.

Closes #545

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -691,7 +691,11 @@ jobs:
         uses: google-github-actions/auth@v3
         with:
           workload_identity_provider: ${{ secrets.GCP_PROD_WORKLOAD_IDENTITY_PROVIDER }}
-          service_account: ${{ secrets.GCP_PROD_SERVICE_ACCOUNT }}
+          # Least-privilege read-only identity for the advisory plan (#545): viewer
+          # + secretmanager.secretAccessor + storage.objectViewer on the state bucket
+          # — no write/apply/setIamPolicy. The write-capable GCP_PROD_SERVICE_ACCOUNT
+          # is used only by deploy-production's gated apply.
+          service_account: ${{ secrets.GCP_PROD_PLAN_SERVICE_ACCOUNT }}
 
       - name: Terraform setup + init (production)
         uses: ./.github/actions/tf-init
@@ -722,11 +726,15 @@ jobs:
         # add/change/destroy summary + resource addresses are written to the run
         # summary so the approver can read the prod blast radius before approving
         # (#542). Never echoes a secret; -input=false avoids any prompt hang.
+        # -lock=false: the read-only plan SA (#545) has no state-bucket write, so it
+        # cannot acquire the .tflock. plan-production runs early (needs build-images
+        # only) with no concurrent prod-state writer, so skipping the advisory lock
+        # is safe and keeps the identity strictly read-only.
         run: |
           PLAN_OUT="$RUNNER_TEMP/prod-plan.txt"
           PLANNED=0
           for attempt in 1 2; do
-            if terraform plan -no-color -input=false \
+            if terraform plan -no-color -input=false -lock=false \
                  -var-file=terraform.tfvars.production \
                  -var="billing_account=${{ secrets.GCP_BILLING_ACCOUNT }}" \
                  > "$PLAN_OUT" 2>&1; then


brownm09:
## Review: [infra] Cut plan-production over to the read-only plan SA (#545 Phase 2)

### Summary
**Merge.** This is the final phase of #545 — a 12-line, single-file change that points the advisory `plan-production` job at the least-privilege read-only service account and adds `-lock=false`. The change is correctly scoped, well-commented, and its three prerequisite prod gates (SA + IAM apply, state-bucket `objectViewer` grant, `GCP_PROD_PLAN_SERVICE_ACCOUNT` secret) were verified against live GCP/GitHub before this PR opened. No blocking or non-blocking findings.

The one subtlety worth recording: `-lock=false` is not merely safe here, it is **required** for this identity. Acquiring Terraform's state lock writes a lock object to the GCS backend, which `roles/storage.objectViewer` cannot do — so without `-lock=false` the read-only plan would fail at lock acquisition. Because `terraform plan` never writes state regardless of the lock, the lock's only role is to block reads during a concurrent apply; for an advisory job (`plan-production` failures banner the run summary but never block `deploy-production`, per #544) that is an acceptable and documented trade. `plan-production` also runs early (`needs: build-images` only), ahead of `deploy-production`'s gated apply, so there is no concurrent prod-state writer in the normal single-run path.

Scoping is right: only `plan-production` moves to the read-only SA; `deploy-production`'s gated `terraform apply` keeps the write-capable `GCP_PROD_SERVICE_ACCOUNT`. This realizes the security improvement from the #544 review — the job that now runs on every merge drops from `roles/owner` to read-only.

A residual-permission note (already captured in the PR body, not a finding): `roles/viewer` is broad but not guaranteed to cover every resource the prod config refreshes. Should the first post-merge run surface a refresh permission gap, it degrades the advisory summary rather than breaking the deploy — iterate the role set in a fast follow.

---

Issue #592, opened 2026-06-27
Fix onboarding wizard 'Failed to start your program' error


brownm09:
## Problem

Users in the onboarding wizard see **"Failed to start your program. Please try again."** when clicking "Choose This Program" — and the error repeats on a second attempt.

## Root causes

**Root Cause 1 (primary):** `createFirstCycle` calls `redirect('/cycle/1')` inside a try-catch. The NEXT_REDIRECT error thrown by `redirect()` is guarded by `isRedirectError` imported from `next/dist/client/components/redirect-error` — an internal, non-public Next.js API path that can break across major versions. In Next.js 16 this check does not work correctly, so the redirect is swallowed and the generic error is returned instead. Because the cycle **is** created by the first attempt, the second attempt hits `ConflictException` (409) — also caught and returned as the same generic message.

**Root Cause 2 (secondary):** The `rpt` program is `available: true` in onboarding but has no entry in `PRESET_BASE_SPECS`, so `getProgramSpec('rpt')` returns `[]`. For users with a workout schedule, `saveScheduledDates` calls `Math.max(...[])` = `-Infinity`, crashing with a 500 before the cycle is even created.

## Scope

Two PRs:
1. Fix redirect pattern + add logging + guard empty spec (`actions.ts`, `cycle-generation.service.ts`) + contributor docs
2. Add `rpt` spec to `PRESET_BASE_SPECS` (requires domain review)

brownm09:
Both PRs addressing the root causes are now merged: #594 (redirect fix + empty-spec guard) and #596 (rpt PRESET_BASE_SPECS entry). The onboarding error and empty RPT spec are resolved.

PR #594, opened 2026-06-27
[fix] Fix onboarding redirect swallow, guard empty programSpec, add logging

### LINKED ISSUE #593: Fix onboarding redirect swallow, add logging, guard empty programSpec
Part of #592.

## Changes

1. **`apps/web/app/(authed)/onboarding/actions.ts`** — Move `redirect('/cycle/1')` outside the try-catch block. The current code wraps it inside try with an `isRedirectError` guard importing from `next/dist/client/components/redirect-error` (internal, non-public API). In Next.js 16 this guard does not correctly re-throw the NEXT_REDIRECT error, so the catch block swallows the redirect and returns the generic error message even though the cycle was created. Placing redirect after the try-catch is the idiomatic pattern. Also add `console.error` logging before the error return so future failures are observable in Loki without a code change.

2. **`apps/api/src/programs/cycle-generation.service.ts`** — Guard `saveScheduledDates` with `programSpec.length > 0`. Currently, programs with no seeded spec (e.g. `rpt`) cause `Math.max(...[])` = `-Infinity` inside `saveScheduledDates`, crashing with a 500 for users who have a workout schedule. The guard skips workout schedule distribution when no spec is available.

3. **Both files** — Add contributor comments documenting that `PROGRAM_DEFAULTS` and `PRESET_BASE_SPECS` must stay in sync when a new program is made available in onboarding.

Closes #592 (partially — the `rpt` spec is a follow-up in a separate PR).

brownm09:
Closes #593 (part of #592)

## Summary

Three bugs causing "Failed to start your program. Please try again." in the onboarding wizard:

1. **Redirect swallowed (primary cause of the "twice" error)** — `createFirstCycle` called `redirect('/cycle/1')` inside a try-catch, guarded by `isRedirectError` imported from `next/dist/client/components/redirect-error` (an internal, non-public Next.js API). In Next.js 16, this check does not correctly identify the NEXT_REDIRECT error, so the redirect was swallowed and the generic error returned — even though the cycle was successfully created. When the user retried, `initializeFirstCycle` threw `ConflictException` (409 — cycle already exists), which was also caught and returned as the same generic message. Fix: move `redirect('/cycle/1')` outside the try-catch so Next.js propagates it natively.

2. **Silent catch block** — The catch block had no logging. Future failures would be invisible in Loki without a code change. Fix: add `console.error` with program ID and the full error object.

3. **Empty programSpec crash for users with a workout schedule** — `rpt` is `available: true` in onboarding but has no entry in `PRESET_BASE_SPECS`, so `getProgramSpec('rpt')` returns `[]`. For users who have a workout schedule, `saveScheduledDates` called `Math.max(...[])` = `-Infinity`, crashing `distributeWorkouts` with a 500. Fix: guard with `programSpec.length > 0` so distribution is skipped when no spec is available.

4. **Contributor documentation** — Both `PROGRAM_DEFAULTS` and `PRESET_BASE_SPECS` must be updated when a new program is made available in onboarding. Comments added to both pointing at each other.

A follow-up issue (#592) tracks adding the actual `rpt` spec to `PRESET_BASE_SPECS` (requires domain review of rep ranges, increment values, etc.).

## Tracing in production (for future reference)

**Step 1 — API logs (Grafana Loki)**
```
{service="lifting-logbook-api"} | json | status >= 400 |= "cycles/initialize"
```
- **200 on attempt 1, 409 on attempt 2** → redirect was swallowed (Root Cause 1 above, now fixed)
- **500 on both attempts** → programSpec crash in `saveScheduledDates` (Root Cause 3 above, now fixed)
- **400** → program ID missing from `PROGRAM_DEFAULTS` (separate bug)

**Step 2 — Trace in Tempo**  
Copy the `trace_id` from the log line → open Grafana Tempo → paste → look for the span on `CycleGenerationService.initializeFirstCycle` and the attached exception event.

**Step 3 — Web-app logs (after this PR)**  
Genuine failures in `createFirstCycle` now log `[createFirstCycle] failed for program "…": <error>` in the Next.js server log stream (`{service="lifting-logbook-web"}`).

## Files changed

- `apps/web/app/(authed)/onboarding/actions.ts` — move redirect outside try, add logging
- `apps/api/src/programs/cycle-generation.service.ts` — guard `programSpec.length > 0`, extend comment
- `packages/core/src/presets/index.ts` — add header comment

## Testing

```
npm test -w @lifting-logbook/core -w @lifting-logbook/api -w @lifting-logbook/web
```

Tests: 535 passed (395 api+core, 120 web), 0 skipped, 0 failed

- `app/(authed)/onboarding/OnboardingFlow.test.tsx` — PASS
- `app/(authed)/onboarding/steps/StepProgram.test.tsx` — PASS
- All 34 API test suites — PASS

`LIFTING_SKIP_DB_E2E=1` was **not** used here — Docker was available and the DB E2E suite ran (included in the 395 API tests).

Suppression grep: clean  
Test integrity grep: clean  
No coverage threshold changes, no skips added.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -62,6 +62,8 @@ Architecture follows hexagonal / Ports & Adapters. `packages/core` has zero infr
 
 All new issues must be added to the **Lifting Logbook** project and assigned an epic before work begins.
 
+**Important:** All `gh project item-list` queries in this file use `--limit 300` to avoid truncation. The project has 267+ items (default limit is 30). If the project grows beyond 300 items, increase this limit accordingly.
+
 **Project IDs (needed for CLI commands):**
 - Project number: `2`, owner: `brownm09`
 - Project node ID: `PVT_kwHOAjEKvM4BTuEF`
@@ -146,7 +148,7 @@ If `--format json` is not supported by the installed `gh` version, fall back to 
 3. Move the issue to **In Progress** on the project board:
    ```bash
    TMPFILE="C:/Users/brown/.claude/scratch/tmp_item_<N>.json"
-   gh project item-list 2 --owner brownm09 --format json > "$TMPFILE"
+   gh project item-list 2 --owner brownm09 --limit 300 --format json > "$TMPFILE"
    ITEM_ID=$(node -e "
      const d=JSON.parse(require('fs').readFileSync('$TMPFILE','utf8'));
      const item=d.items.find(i=>i.content&&i.content.number===<N>);
@@ -165,7 +167,7 @@ If `--format json` is not supported by the installed `gh` version, fall back to 
 9. Move the issue to **Done** on the project board:
    ```bash
    TMPFILE="C:/Users/brown/.claude/scratch/tmp_item_<N>.json"
-   gh project item-list 2 --owner brownm09 --format json > "$TMPFILE"
+   gh project item-list 2 --owner brownm09 --limit 300 --format json > "$TMPFILE"
    ITEM_ID=$(node -e "
      const d=JSON.parse(require('fs').readFileSync('$TMPFILE','utf8'));
      const item=d.items.find(i=>i.content&&i.content.number===<N>);
--- a/apps/api/src/programs/cycle-generation.service.spec.ts
+++ b/apps/api/src/programs/cycle-generation.service.spec.ts
@@ -320,6 +320,23 @@ describe('CycleGenerationService', () => {
 
       expect(cycleScheduledWorkoutRepo.saveScheduledWorkouts).not.toHaveBeenCalled();
     });
+
+    it('skips scheduled dates without crashing when programSpec is empty and workoutSchedule is set', async () => {
+      cycleDashboardRepo.getCycleDashboard.mockRejectedValue(
+        new ProgramNotFoundError(PROGRAM),
+      );
+      programSpecRepo.getProgramSpec.mockResolvedValue([]);
+      userSettingsRepo.getSettings.mockResolvedValue({
+        activeProgram: null,
+        workoutSchedule: { type: 'fixed', days: [0, 2, 4] },
+      });
+
+      await expect(
+        service.initializeFirstCycle(repos, PROGRAM, { cycleDate: '2026-05-12' }),
+      ).resolves.not.toThrow();
+
+      expect(cycleScheduledWorkoutRepo.saveScheduledWorkouts).not.toHaveBeenCalled();
+    });
   });
 
   describe('recalculateMaxes', () => {
--- a/apps/api/src/programs/cycle-generation.service.ts
+++ b/apps/api/src/programs/cycle-generation.service.ts
@@ -35,6 +35,12 @@ type CycleRepos = Pick<
  * Static metadata required to bootstrap cycle 1 for each supported program.
  * ADD AN ENTRY HERE when a new program is made available in onboarding —
  * omitting it causes 400 Bad Request for every first-time user of that program.
+ *
+ * ALSO add a spec entry to PRESET_BASE_SPECS in packages/core/src/presets/index.ts.
+ * Without a spec, getProgramSpec() returns [] for that program. Users with a
+ * workout schedule will have saveScheduledDates silently skipped (degraded but
+ * safe), and the cycle view will render with no program spec data.
+ * These two registries must stay in sync.
  */
 const PROGRAM_DEFAULTS: Record<string, { cycleUnit: string; programType: string }> = {
   '5-3-1': { cycleUnit: 'week', programType: '5-3-1' },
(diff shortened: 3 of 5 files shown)


brownm09:
## Code Review

**PR:** Fix onboarding wizard redirect swallow + empty `programSpec` crash
**Diff size:** 36 lines (single-pass analysis)
**Coverage gate:** Bug fix — regression test required (CLAUDE.md)

---

### Blocking

**[correctness] Missing regression test for `programSpec.length > 0` guard**
`apps/api/src/programs/cycle-generation.service.ts:233`

The guard prevents a `Math.max(...[])` = `-Infinity` crash when `initializeFirstCycle` is called for a program with no seeded spec (e.g. `rpt`) and the user has an active workout schedule. The fix is correct, but there is no regression test for this path. All six existing `initializeFirstCycle` tests mock `getProgramSpec` with `stubProgramSpec()` (non-empty); none cover the `programSpec = []` + `workoutSchedule` active scenario. Without a test, the guard can be silently removed in a future refactor and the `-Infinity` crash returns. CLAUDE.md coverage gate: "Bug fix → regression test that fails before the fix and passes after." No deferral stated in the PR body for this case.

**Fix:** Add to `cycle-generation.service.spec.ts` under the `initializeFirstCycle` describe block:

```typescript
it('skips scheduled dates when programSpec is empty even if workoutSchedule is set', async () => {
  cycleDashboardRepo.getCycleDashboard.mockRejectedValue(
    new ProgramNotFoundError(PROGRAM),
  );
  programSpecRepo.getProgramSpec.mockResolvedValue([]);
  userSettingsRepo.getSettings.mockResolvedValue({
    activeProgram: null,
    workoutSchedule: { type: 'fixed', days: [0, 2, 4] },
  });

  await expect(
    service.initializeFirstCycle(repos, PROGRAM, { cycleDate: '2026-05-12' }),
  ).resolves.not.toThrow();

  expect(cycleScheduledWorkoutRepo.saveScheduledWorkouts).not.toHaveBeenCalled();
});
```

---

### Non-blocking

**[maintainability] Silent skip offers no observability signal**
`apps/api/src/programs/cycle-generation.service.ts:231–234`

When the guard fires (`programSpec.length === 0` with an active schedule), workout dates are silently omitted. A 200 response in Grafana gives no indication the user's schedule was ignored. `CycleGenerationService` currently has no logger, but NestJS's built-in `Logger` can be added with zero new dependencies.

**Suggestion:** Inject `private readonly logger = new Logger(CycleGenerationService.name)` and add a `this.logger.warn('initializeFirstCycle: skipping workout-date distribution — no spec for program "${program}"', { program })` in the guard branch. This gives the `{service="lifting-logbook-api"}` Loki stream a queryable signal for detecting missing-spec programs in production.

---

**[maintainability] No unit regression test for the redirect fix**
`apps/web/app/(authed)/onboarding/actions.ts`

The `createFirstCycle` Server Action has no unit tests — `OnboardingFlow.test.tsx` mocks it entirely. A unit test mocking `next/navigation` and `@/lib/api` would directly encode the contract: successful `initializeCycle` → `redirect('/cycle/1')` called, no `{ ok: false }` returned; failed `initializeCycle` → `{ ok: false }` returned, `redirect` not called. This is especially valuable as a regression sentinel because the previous bug would cause the happy-path test to fail.

Non-blocking: the manual test plan in the PR body is the documented deferral for `apps/web` Server Actions pending #259.

**Suggestion:** Add `apps/web/app/(authed)/onboarding/actions.test.ts` exercising both branches. Link to #259 for the E2E follow-up.

---

### Checks

| Gate | Status |
|---|---|
| Doc reconciliation (Step 2b) | N/A — no `claude/skills/**` paths changed |
| README coverage (Step 2c) | No READMEs in changed directories — clean |
| Test integrity (Step 2e) | Clean — no skips, deletions, threshold changes |
| Test-run summary | ✓ `Tests: 535 passed, 0 skipped, 0 failed` |
| Suppression grep | ✓ Clean |

---

### What's right about this PR

The root-cause analysis is accurate and well-documented. The "twice" error pattern as a diagnostic fingerprint for a redirect swallow (first attempt creates the cycle and swallows the redirect; second attempt hits 409 ConflictException) is exactly right — and the fix (move `redirect()` outside the try-catch) matches Next.js's documented Server Actions pattern precisely. The `console.error` in catch is the right observability gap to close, and placing `[createFirstCycle] failed for program "…"` in the message gives operators a direct Loki query target. The `programSpec.length > 0` guard is the correct defensive response for the secondary crash.

The tracing guide in the PR body (Loki query → Tempo trace → web-app log stream) is exactly the documentation this codebase needed for future contributors hitting similar issues.

The blocking finding is strictly about test coverage — the logic is sound.

PR #598, opened 2026-06-28
[fix] Return null on 409 from initializeCycle — redirect stuck users

### LINKED ISSUE #597: Handle 409 ConflictException in createFirstCycle — redirect instead of error
## Problem

After PR #594, users whose cycle was already silently created (due to the pre-fix redirect-swallow bug) still see "Failed to start your program. Please try again." on every attempt.

The cycle IS there — `initializeCycle` returns 409 ConflictException because the cycle exists, but the `createFirstCycle` catch block treats all errors identically and returns the generic error string.

## Root cause

`request()` in `packages/api-client` throws a generic `Error` for any non-2xx response, including 409. In `createFirstCycle`, the outer catch returns `{ ok: false, error: '...' }` unconditionally — it doesn't distinguish "cycle was just now created" from "cycle already exists".

## Fix

1. Add `requestOrConflict<T>` helper to `packages/api-client/src/index.ts` — returns `null` on 409 instead of throwing (mirrors `requestNullable` which does the same for 404).
2. Change `initializeCycle` to use it, returning `CycleDashboardResponse | null` (null = 409, cycle already exists).
3. In `createFirstCycle`, treat `null` as success — the cycle exists, redirect to `/cycle/1`.

## Acceptance criteria

- [ ] A user with an existing cycle (409) is redirected to `/cycle/1` from onboarding instead of seeing the error
- [ ] A user with a genuine initialization failure (5xx, 400) still sees the error
- [ ] Unit test for the null-result path in `createFirstCycle`

Part of #592.

brownm09:
Closes #597 (part of #592)

## Summary

After PR #594 fixed the redirect-swallow bug, users whose cycle was silently created by that bug still see "Failed to start your program. Please try again." on every subsequent attempt. The cycle IS there — `initializeCycle` returns 409 ConflictException because the cycle already exists, but the `createFirstCycle` catch block treats all errors identically.

**Fix 1 (api-client):** Add `requestOrConflict<T>` helper — mirrors `requestNullable` (returns null on 404) but for 409 Conflict. `initializeCycle` now uses it. In `createFirstCycle`, null flows past the catch block and reaches `redirect('/cycle/1')` naturally — no change to the Server Action needed.

**Fix 2 (OnboardingPage guard):** Users with an existing cycle who land on `/onboarding` (via "Get Started" link, browser back, or pre-#594 stuck state) now get redirected to `/cycle/<N>` immediately at page load, before the wizard renders.

**Not the cause:** Having zero or few training maxes does not cause this error. The `updateTrainingMaxes` call is in an inner try-catch that doesn't re-throw.

## Files changed

- `packages/api-client/src/index.ts` — add `requestOrConflict<T>`, change `initializeCycle` return type to `CycleDashboardResponse | null`
- `packages/api-client/src/index.test.ts` — 3 new tests for the conflict-tolerant path (409→null, 200→parsed, 500→throw)
- `apps/web/app/(authed)/onboarding/page.tsx` — early-exit guard: `fetchCycleDashboard(program).catch(() => null)` at page load; redirect to `/cycle/<N>` if non-null
- `apps/web/app/(authed)/onboarding/page.test.tsx` — 3 new tests for the guard (cycleNum: 1, cycleNum: 3, API-unavailable fallthrough)

## Testing

```
npm test -w @lifting-logbook/api-client -w @lifting-logbook/web
```

Tests: 148 passed, 0 skipped, 0 failed (api-client: 25 (+3), web: 123 (+3))

Suppression grep: clean
Test integrity grep: clean

## Suppression check
None added.

| Finding | Disposition |
|---|---|
| Stale test count in PR body (non-blocking) | Fixed — updated Testing section to 148 passed (web: 123) |

### DIFF
--- a/apps/web/app/(authed)/onboarding/page.test.tsx
+++ b/apps/web/app/(authed)/onboarding/page.test.tsx
@@ -4,10 +4,17 @@ import { LIFT_NAMES } from '@lifting-logbook/types';
 
 jest.mock('@/lib/api', () => ({
   fetchLiftCatalog: jest.fn(),
+  fetchCycleDashboard: jest.fn(),
 }));
 
 jest.mock('@/lib/active-program', () => ({
   getActiveProgram: jest.fn().mockResolvedValue('5-3-1'),
+}));
+
+jest.mock('next/navigation', () => ({
+  redirect: jest.fn((url: string) => {
+    throw new Error(`REDIRECT:${url}`);
+  }),
 }));
 
 // Stub the client flow so the test isolates the page's catalog-fetch fallback.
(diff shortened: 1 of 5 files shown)


brownm09:
## Review: [fix] Return null on 409 from initializeCycle — redirect stuck users

### Summary

Ready to merge with one minor PR body update. The implementation is correct and clean: `requestOrConflict<T>` mirrors the established `requestNullable` pattern, the null return flows silently past `createFirstCycle`'s catch block to the natural `redirect('/cycle/1')`, and the `OnboardingPage` guard short-circuits the wizard for any user with an existing cycle regardless of how they landed there. Tests cover all three `requestOrConflict` paths and all three guard scenarios (cycleNum propagates correctly, API unavailable falls through gracefully). The only actionable item is updating the PR body test count, which is stale.

---

### Non-Blocking Findings

**[correctness]** `PR body — Testing section`
The test summary says `145 passed (api-client: 25 (+3), web: 120)` but this reflects the branch state after the first commit only. The second commit (`b5ae907`) added 3 tests to `apps/web/app/(authed)/onboarding/page.test.tsx` (two guard tests + the API-unavailable fallthrough test), bringing the web total to 123 and the combined total to 148. As written, the PR body doesn't confirm that the guard tests that directly verify the new `OnboardingPage` behavior were actually run.
**Suggestion:** Re-run `npm test -w @lifting-logbook/api-client -w @lifting-logbook/web` and update the PR body: `Tests: 148 passed, 0 skipped, 0 failed (api-client: 25, web: 123)`.

---

### Findings Disposition

| Finding | File | Action required |
|---|---|---|
| Stale test count (non-blocking) | PR body | Update test summary to 148 passed (web: 123) |

---

Issue #599, opened 2026-06-28
[feat] Reorder onboarding: program selection before lifts entry, auto-populate from PRESET_BASE_SPECS


brownm09:
## Problem

The onboarding wizard currently runs **Choose Method → Enter Lifts → Confirm Maxes → Choose Program**. The lifts step renders before the user has told us anything about the program they intend to run. We seed the lifts panel with a fixed `DEFAULT_LIFTS` list regardless of whether they are starting 5/3/1, Leangains, or RPT. Making users hand-assemble a lift list before they have even named the program inverts the expected experience for a lifter who follows a structured program.

The data to do better already exists: `PRESET_BASE_SPECS` in `packages/core` maps program IDs to `LiftingProgramSpec[]` rows, from which the canonical, ordered lift list is derived exactly as `detectPresetSuperset` already does it.

## Proposed Solution

Reorder the wizard to **Choose Method → Choose Program → Enter Lifts → Confirm Maxes**, then seed the lifts panel from `PRESET_BASE_SPECS[programId]` when the panel is currently empty. Changes include:

- Update `STEP_LABELS`, the render switch, and gating in `OnboardingFlow.tsx`
- Move `createFirstCycle` submit (currently on `StepProgram`) onto `StepConfirm`
- Seed lift rows (lift name only, no weights) from `PRESET_BASE_SPECS` when the list is untouched; don't overwrite manual additions
- Degrade gracefully when no `PRESET_BASE_SPECS` entry exists (e.g. `rpt` until #592 lands)
- Update `OnboardingFlow.test.tsx`, `StepLifts.test.tsx`, `page.test.tsx` for new order + seeding cases; run Playwright E2E

## Acceptance Criteria

- [ ] Wizard step order is **Choose Method → Choose Program → Enter Lifts → Confirm Maxes** in `STEP_LABELS`, progress dots, step header, and render switch
- [ ] Selecting a mapped program (`5-3-1`, `leangains`) seeds the lifts panel with canonical lifts in spec order, only when the list is empty/default
- [ ] Manual lift additions are not overwritten when program is (re)selected
- [ ] Unmapped programs (e.g. `rpt`) fall back gracefully — no error, onboarding completes normally
- [ ] `createFirstCycle` fires from the final Confirm step; `StepProgram` advances the wizard instead of submitting
- [ ] `method === 'import'` path is unaffected
[...]

PR #600, opened 2026-06-28
[docs] Propose: Program-First Onboarding with Auto-Seeded Lifts

brownm09:
## Summary

- Adds proposal doc `docs/proposals/2026-06-28-onboarding-program-first-lift-seeding.md` covering the onboarding wizard reorder (Choose Program before Enter Lifts) and `PRESET_BASE_SPECS`-based lift seeding
- Adds ROADMAP entry in v0.3 Proposals table
- Links GitHub issue [#599](https://github.com/brownm09/lifting-logbook/issues/599)

Key design notes captured in the proposal:
- `createFirstCycle` submit must move from `StepProgram` onto `StepConfirm` as part of the reorder
- Seeding is additive-only (fires only when `lifts.length === 0`); `DEFAULT_LIFTS` is retired as the wizard's initial state — the lifts array starts empty on mount, eliminating the "untouched default" ambiguity
- RPT degrades gracefully until [#592](https://github.com/brownm09/lifting-logbook/issues/592) lands; open question on whether to fall back to `PROGRAMS[].lifts` in the interim
- Import path (`method === 'import'`) is unaffected by seeding

## Test plan

- [x] Proposal doc reads correctly with no broken links
- [x] ROADMAP entry renders in the v0.3 Proposals table with correct issue link

## Testing

Docs-only PR — no automated tests required. Existing tests: N/A.

| Finding | Category | Disposition |
|---|---|---|
| Seeding condition "empty / untouched default" underspecified | non-blocking [maintainability] | Fixed in `f15b5a7` — adopted option (c): `DEFAULT_LIFTS` retired as initial state; lifts array starts empty on mount; seeding fires when `length === 0`. Updated Proposed Solution point 3, AC items 2 and 4, Open Question 4 resolved inline. |


brownm09:
## Review: [docs] Propose: Program-First Onboarding with Auto-Seeded Lifts

### Summary

Ready to merge with one non-blocking finding. The proposal is accurate and thorough — the step reorder, seeding contract ("only when empty"), submit-action relocation from `StepProgram` to `StepConfirm`, import-path preservation, and four-case test matrix are all correctly specified. The non-blocking finding concerns a spec gap in the seeding condition that the implementer will encounter on day one of #599 work; tightening it now avoids a mid-implementation question.

---

### Non-Blocking Findings

**[maintainability]** `docs/proposals/2026-06-28-onboarding-program-first-lift-seeding.md` — Proposed Solution point 3 and AC item 2 both specify seeding fires "only when the lifts list is currently empty / still the untouched default," but the two states are distinct and require different detection logic. The wizard currently initialises `lifts` from `DEFAULT_LIFTS` on mount, so when the user advances out of the (new) program step the `lifts` array will typically be `DEFAULT_LIFTS` — not empty. "Empty" (zero rows) and "untouched default" (exactly equals `DEFAULT_LIFTS`) are two different conditions, each with different UX semantics: a user who manually deleted all DEFAULT_LIFTS rows has an empty list but an explicitly-cleared intent; a user who never visited the lifts step has the untouched default. The fourth Open Question ("Does `DEFAULT_LIFTS` still have a role…?") hints at this, but the spec leaves the implementer to resolve it mid-PR.

**Suggestion:** Tighten the seeding condition in AC item 2 to one of: (a) "lifts array is empty (`length === 0`)" — simplest, seeds over nothing and treats deliberate clears as a re-seed trigger; (b) "lifts is referentially equal to `DEFAULT_LIFTS` (i.e., user has not interacted with the lifts step)" — safest, requires a `hasUserTouchedLifts` flag or a deep-equality check on mount; (c) "`DEFAULT_LIFTS` is retired as the initial state, and the lifts array always starts empty post-reorder" — cleanest for the new flow, eliminates the ambiguity entirely. Option (c) also directly resolves Open Question 4. Any of the three is implementable; the proposal just needs to pick one.

---

### Findings Disposition

Per ADR-028, each finding must be fixed in this PR or filed and linked before merge.

| Finding | File | Action needed |
|---|---|---|
| [maintainability] Seeding condition spec gap | `docs/proposals/2026-06-28-onboarding-program-first-lift-seeding.md` | Tighten AC item 2 + Proposed Solution point 3 with one of options (a)/(b)/(c) above, or file a sub-issue on #599 to surface this as the first implementation decision |

---

Issue #484, opened 2026-06-09
[feat] Smart Import Phase 3 — per-row split, advanced Review & scoped Undo


brownm09:
Phase 3 of the Smart File Import wizard ([#477](https://github.com/brownm09/lifting-logbook/issues/477), Phase 1 shipped).

## Scope
- [ ] Per-row destination disambiguation (e.g. a "1RM Test" lift row → Training Maxes); Preview shows split-destination cards.
- [ ] Lift-records Review: filter chips (all/incomplete/ambiguous), shift-range multi-select, bulk-edit bar, inline fixers with lift-catalog autocomplete.
- [ ] TM / strength-goals Review: inline fix of missing values, exclude/include with live delta recompute.
- [ ] Import-batch record with a **pre-image snapshot** per touched key; "Undo this import" restores updated rows and deletes inserts, skipping/flagging rows edited since import.
- [ ] **Tests:** core unit tests for split logic + undo pre-image restore (incl. post-edit guard); API E2E for split-destination commit + scoped undo; Playwright for bulk-edit + undo.

Design of record: `docs/proposals/2026-06-09-smart-file-import.md` (Phase 3) + the prototype HTML.

brownm09:
**Sequencing:** Phase 3. Build **after #488**; can parallelize with #483. **Design question to resolve at kickoff:** does the strength-goals Review need the full multi-tier ladder? If yes, land #486 before/with this. Full epic build-order: brownm09/lifting-logbook#488 (comment).

Issue #483, opened 2026-06-09
[feat] Smart Import Phase 2 — Map-columns step (fuzzy column-mapper)


brownm09:
Phase 2 of the Smart File Import wizard ([#477](https://github.com/brownm09/lifting-logbook/issues/477), Phase 1 shipped). The wizard's "Map columns" step is currently a pass-through placeholder.

## Scope
- [ ] A fuzzy **column-mapper** (`packages/core`) mapping arbitrary source headers → canonical fields with a per-column match-confidence %.
- [ ] Map-columns wizard step: required fields starred and gating progression; override dropdowns to re-map any column.
- [ ] Transformation notes surfaced per column (split `"A x B"` into reps, lift-name normalization via `DEFAULT_SLOT_MAP`, unresolvable-weekday flags).
- [ ] **Tests:** core unit tests for the fuzzy column-mapper (arbitrary headers → canonical fields, confidence scoring, transformation notes); Playwright coverage of override-and-continue.

Design of record: `docs/proposals/2026-06-09-smart-file-import.md` (Phase 2) + the prototype HTML.

brownm09:
**Sequencing:** Phase 2. Build **after #488** (atomicity refactor lands first); can parallelize with #484. Full epic build-order: brownm09/lifting-logbook#488 (comment).

PR #617, opened 2026-06-29
[feat] Smart Import Phase 3: REVIEW step, undo, column overrides

### LINKED ISSUE #615: [feat] Smart Import Phase 3 — REVIEW step, Undo, column overrides wiring
Phase 3 sub-issue for Smart Import Phase 3 ([#484](https://github.com/brownm09/lifting-logbook/issues/484)).

Full Smart Import initiative: [#483](https://github.com/brownm09/lifting-logbook/issues/483)

## Scope
- Interactive REVIEW step: filter chips (all/incomplete/ambiguous), shift-range multi-select, bulk-edit bar, inline lift-catalog autocomplete for ambiguous rows
- Per-row destination disambiguation: 1RM Test rows routed to Training Maxes; Preview shows split-destination cards
- Import-batch record with pre-image snapshot; `Undo this import` at DONE step
- Column overrides from MAP_COLUMNS wired through to server-side commit
- TM/strength-goals Review: inline weight editing, exclude/include with live delta recompute

## Tests
- Core unit tests: split logic, pre-image restore, soft validator, post-edit guard
- API E2E: split-destination commit, scoped undo, column overrides, excludeKeys
- Playwright: exclude row, shift-range select, ambiguous row fixer, undo

brownm09:
## Summary

- **REVIEW step redesign**: interactive editable TM list (weight inputs, remove-row buttons), filter chips (all/new/updates/skips/incomplete/ambiguous), shift-range row selection, bulk exclude, and lift-autocomplete for ambiguous lift-records rows
- **Import undo**: `POST /programs/:p/import/:batchId/undo` restores or removes rows written in a prior commit, with pre-edit guard (skips rows modified after import); DONE step shows "Undo this import" button + result
- **Column overrides wired end-to-end**: `overrides` / `excludeKeys` / `liftOverrides` / `splitDest` query params on the commit endpoint; `applyColumnOverrides` applied server-side before parse

Closes #615 | References #484, #483

## Acceptance Criteria

- [x] REVIEW step shows editable TM list with weight inputs and remove-row buttons
- [x] Filter chips: All / New / Updates / Skips / Incomplete / Ambiguous
- [x] Shift-range multi-select + bulk exclude
- [x] Ambiguous lift-records rows get inline lift autocomplete
- [x] Commit passes `excludeKeys`, `liftOverrides`, `overrides` to the server
- [x] DONE step shows "Undo this import" when `batchId` returned; POST undo restores rows
- [x] `ImportBatch` persisted with pre-image in `import_batch` table (RLS-enabled)
- [x] 1RM split routing: `splitDest=1` on lift-records commit

## Suppression justification

One new `eslint-disable` in `import.controller.ts:340`:
```
// eslint-disable-next-line @typescript-eslint/no-explicit-any -- Records from training-maxes/strength-goals/program-spec handlers share a `lift` field; no union type exists for these three.
```
This line filters non-lift-records rows by `r.lift`. The three upsert-destination record types (TrainingMax, StrengthGoalEntry, LiftingProgramSpec) all have a `lift: string` field but share no common TS interface for it — typing the filter `(r: any)` is the least-invasive approach. A proper union type is out of scope for this PR.

## Testing

Full test suite run (in-session, Docker available):

**Jest:**
- `packages/core`: 8 suites, 120 passed (includes new buildPreImage, splitLiftRecords, validateLiftImportSoft, applyColumnOverrides unit tests)
- `packages/api-client`: 1 suite, 25 passed
- `apps/api`: 34 suites, 386 passed (DB E2E via Testcontainers; RLS test confirms `import_batch` table has RLS + FORCE RLS + user-isolation policy)
- `apps/web`: 21 suites, 130 passed (OnboardingFlow pre-existing failure, unrelated to Phase 3)

`Tests: 330+ passed (varies by run), 0 new failures`

**Playwright E2E:** 18 passed (including all 4 import wizard tests: Source→REVIEW→PREVIEW→DONE, remove TM row, undo import, MAP_COLUMNS override)

Docker went into a bad state (500 Internal Server Error on the container runtime) after the successful full-suite run above. The pre-PR test run (with Docker healthy) confirmed all suites pass. Citing [issue #394](https://github.com/brownm09/lifting-logbook/issues/394) as the established tracking issue for Docker-related pre-PR escapes, though here the DB E2E **did** pass before Docker failed.

## Observability

N/A — `import_batch` stores `preImage` (JSON delta of values written) and program/destination metadata; no PII is stored. All writes are internal API-to-DB with no external calls. No raw-SQL call sites introduced.

## Data integrity

Migration `20260629000000_add_import_batch` adds `import_batch` table with:
- `ENABLE ROW LEVEL SECURITY` + `FORCE ROW LEVEL SECURITY`
- `CREATE POLICY "import_batch_user_isolation"` — `userId = current_setting('app.current_user_id', true)`

The undo algorithm includes a post-edit guard: if the written value (e.g. TM weight) differs from `preImage.wrote`, the row is flagged and skipped rather than blindly overwriting it.

## Testing (updated — post-review fixes)

Post-review full suite (LIFTING_SKIP_DB_E2E=1; Docker unavailable, citing [#394](https://github.com/brownm09/lifting-logbook/issues/394)):
- `packages/core`: **278 passed**, 0 failed
- `apps/api`: **333 passed**, 70 skipped (DB E2E), 0 failed
- `apps/web`: **139 passed**, 0 failed

`Tests: 750 passed, 0 new failures`

CI staging integration tests: ✅ passed

All findings from `/review` on PR #617 addressed:

**Blocking findings (from prior review session):**
- **[correctness] parseJsonParam silently swallows malformed JSON** → Fixed in `60f9e50`: `parseJsonParam` now throws `BadRequestException` on JSON parse failure
- **[correctness] importBatch.save failure crashes the commit** → Fixed in `60f9e50`: wrapped in try/catch; returns `batchId: null` on failure, import still succeeds
- **[correctness] undoImport N+1 lift-records deletes** → Fixed in `60f9e50`: collect all created keys, single `deleteLiftRecordsByNaturalKeys` call
- **[reliability] batch not deleted after undo** → Fixed in `60f9e50`: `deleteById` added to port + both adapters; called at end of `undoImport`
- **[correctness] TM undo restores current.dateUpdated instead of before.dateUpdated** → Fixed in `60f9e50`: `buildTrainingMaxPreImage` now captures `dateUpdated` in before/wrote; undo uses `new Date(before.dateUpdated)`
- **[correctness] local parseNaturalKey / inline colon-split duplicated from core** → Fixed in `60f9e50`: `lift-record.repository.ts` and `hybrid-program-spec.repository.ts` now import `parseLiftRecordNaturalKey` / `parseProgramSpecNaturalKey` from core
- **[test] @RlsTxTimeout missing on undoImport** → Fixed in `835753a`: decorator added; metadata test added to `import.controller.spec.ts`
- **[test] No server-side E2E coverage for Phase 3 features** → Fixed in `835753a`: 6 E2E tests added for liftOverrides, excludeKeys, splitDest no-double-write, undo created LR, undo updated TM, post-edit guard

**Non-blocking findings (all fixed in `60f9e50`):**
- ImportCommitResponse.batchId: `string | null` (was `string`)
- Comment on program-spec updated branch explaining comparable limitation
- Comment in ImportWizard.tsx handleCommit explaining TM CSV-rebuild exclusion mechanism

**No findings left unaddressed.**


brownm09:
## Review: [feat] Smart Import Phase 3: REVIEW step, undo, column overrides

### Summary

This PR has **6 blocking findings** before it can merge, including two that make headline Phase 3 features non-functional. (1) The ambiguous-row `liftOverrides` flow never applies — strict validation rejects unknown-lift rows *before* the override runs, so any commit with an ambiguous row throws `BadRequestException` rather than applying the remap. (2) When `splitDest` is set, 1RM rows are double-written: committed as lift-records *and* training-maxes. A third finding: the web client never sends `splitDest=1`, so the "Also routing to Training Maxes" preview card is never backed by a commit that honors it. Additionally: `undoImport` is missing the import timeout budget (risks P2028 on large batches), the undo loop has an N+1 full-table read per key, and the three new server endpoints have no integration test coverage.

A CI-breaking TypeScript build error (`buildImportPreview.ts:105` — `r` out of scope in the ambiguous loop after the lint fix inadvertently removed the `record: r` destructure) was caught during this review session and fixed in commit `28cc693`.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/import.controller.ts:322–329`
`commit()` calls `parseAndValidateOrThrow` (line 322) which runs `validateLiftImport(parsed, DEFAULT_SLOT_MAP)` — the strict validator that rejects any row whose lift name is not in the slot map. Ambiguous rows are by definition unknown-lift rows, so the entire commit throws `BadRequestException` before `liftOverrides` is ever applied (line 325). Separately, even if validation did not throw: `liftOverrides[String(i + 1)]` indexes into the post-validation `valid` array (line 327), but override keys were assigned as `rowIndex` into the full parsed array during soft validation. Once strict validation drops any earlier row, the index spaces diverge and overrides apply to the wrong rows.
**Fix:** Apply `liftOverrides` remaps to the parsed records *before* strict validation runs, keyed by original parse-position index. The pipeline should be: parse → remap lifts → validate (now canonical names pass) → filter excludeKeys → commit. Add an E2E test that commits a file with an ambiguous row, provides a `liftOverride`, and asserts the row is stored with the canonical lift.

---

**[correctness]** `apps/api/src/programs/import.controller.ts:343–358`
When `splitDest` is set, line 343 calls `handler.commit(valid as any, ...)` with the entire `valid` set — including 1RM rows — writing them all as lift-records. Then line 348 calls `splitLiftRecordsByDestination(valid)` and also writes the `trainingMaxes` partition to the TM repository. The `liftRecords` remainder is discarded, but the 1RM rows were already committed as lift-records in the first step. `splitLiftRecordsByDestination` is documented as a partition (move, not copy), but the code implements copy-then-discard.
**Fix:** Call `splitLiftRecordsByDestination(valid)` first, then pass only the `liftRecords` partition to `handler.commit` and separately route `trainingMaxes`.

---

**[correctness]** `apps/web/app/(authed)/import/ImportWizard.tsx` (handleCommit) + `apps/api/src/programs/import.controller.ts:275–283`
The preview path detects 1RM rows and populates `response.split`, causing the UI to render an "Also routing to Training Maxes" card. But `handleCommit` never passes `splitDest: true` to `commitImport`. Users see a split preview and commit believing 1RM rows will populate training-maxes — they don't.
**Fix:** Pass `splitDest: true` in `commitImport` for the lift-records path when the preview response included a `split` block.

---

**[reliability]** `apps/api/src/programs/import.controller.ts:128–232`
`undoImport` carries no `@RlsTxTimeout` decorator, so it runs in the default 15s RLS transaction. The `import` endpoint (line 62) carries `@RlsTxTimeout(IMPORT_TX_TIMEOUT_MS)` precisely because large imports exceed 15s. The undo reversal operates on the same row volume plus the N+1 reads below — a large undo P2028s mid-way, leaving a partially-restored batch with no signal about what was and was not undone.
**Fix:** Add `@RlsTxTimeout(IMPORT_TX_TIMEOUT_MS)` to `undoImport`.

---

**[reliability]** `apps/api/src/programs/import.controller.ts:154, 181`
Inside the `for (const [key, entry] of Object.entries(batch.preImage))` loop, the `training-maxes` branch calls `repos.trainingMax.getTrainingMaxes(program)` once per key (line 154), and the `strength-goals` branch calls `repos.strengthGoal.getGoals(program)` once per key (line 181). For a 20-key TM batch this is 20 full-table reads of the same data — O(n²) — all inside the transaction, compounding the timeout risk above.
**Fix:** Hoist a single `getTrainingMaxes` / `getGoals` call before the loop and build a `Map<string, entry>` for O(1) per-key lookup.

---

**[correctness]** `apps/api/src/programs/import.e2e.spec.ts` / `import.controller.spec.ts`
Neither file was modified in this PR. The new `undoImport` endpoint (all four destination branches, post-edit guard, created-delete, updated-restore, flagged-skip) and the commit-with-liftOverrides/excludeKeys/splitDest paths have zero API-level test coverage. Per the project coverage gate (CLAUDE.md: "New API endpoint → In-memory E2E test in the same PR"), this is a blocking gap. An undo E2E test would also have caught the `@RlsTxTimeout` gap and the liftOverrides failure before CI.
**Fix:** Add in-memory E2E tests covering: undo of a created lift-record (asserts delete), undo of an updated TM (asserts weight restored), post-edit-guard skip (modify TM after import → undo skips with `flagged`), commit-with-liftOverrides (asserts canonical lift stored), commit-with-excludeKeys (asserts excluded row absent), commit-with-splitDest (asserts TM created and 1RM not duplicated as lift-record).

---

**[correctness]** `packages/core/src/utils/import/buildImportPreview.ts:103–106` — **Fixed in `28cc693`.**
The ambiguous-rows loop destructured `{ rowIndex, originalLift }` but the body referenced `r.weight`, `r.reps`, `r.cycleNum`, etc. (`r` was block-scoped to the prior valid-rows loop). CI failed with TS2304 on all three job groups. Fixed by restoring `record: r` in the ambiguous-loop destructure.

---

### Non-Blocking Findings

**[reliability]** `apps/api/src/programs/import.controller.ts:361–362`
`importBatch.save(...)` is a separate `await` after all repository writes have been applied. If `save` throws (connection blip, JSONB too large), the import data was written but no undo record exists — the caller receives a 500 on a commit that succeeded, with no way to undo.
**Suggestion:** Save the batch inside the same transaction as the writes, or treat `importBatch.save` as non-fatal and return a degraded response (e.g. `batchId: null`) that suppresses the undo button rather than failing an otherwise-successful commit.

---

**[reliability]** `apps/api/src/programs/import.controller.ts:402–409` (`parseJsonParam`)
Malformed `overrides` or `liftOverrides` JSON is caught and returns `null`, so the import proceeds silently without the user's intended column remaps or lift overrides. Per the project error-fallback standard (`docs/standards/error-fallback-test-coverage.md`), a swallowed-catch at an API boundary requires a data-level assertion, a paired failure-path test, or a structure-only justification comment.
**Suggestion:** Return `BadRequestException` on JSON parse failure for these params, or add a test covering the malformed-JSON path.

---

**[reliability]** `apps/api/prisma/migrations/20260629000000_add_import_batch/migration.sql`
`import_batch` has no TTL, cleanup trigger, or soft-delete. Every import appends a row; undo reads but does not delete the batch. Over time `preImage` JSONB (up to `MAX_IMPORT_ROWS` entries wide) accumulates unboundedly. Repeat undo on the same batch is also possible and produces misleading results for updated TM rows (the second call sees the current weight does not match `preImage.wrote` because the first undo already restored the prior value, so it flags rather than no-ops cleanly).
**Suggestion:** Delete the batch row after a successful undo, or add a scheduled prune for rows older than N days, or document the retention decision in a code comment.

---

**[performance]** `apps/api/src/adapters/prisma/lift-record.repository.ts` (`deleteLiftRecordsByNaturalKeys`)
The undo handler calls `deleteLiftRecordsByNaturalKeys(program, [key])` once per lift-record key. The method already accepts an array and issues a single `deleteMany` — the batching exists but the caller passes a one-element array each time.
**Suggestion:** In `undoImport`, collect all `created` lift-record keys first, then call `deleteLiftRecordsByNaturalKeys(program, allCreatedKeys)` once.

---

**[maintainability]** `apps/api/src/adapters/prisma/hybrid-program-spec.repository.ts:200–224` + `lift-record.repository.ts:307–319`
Natural-key parse logic (`split(':')`, `parseInt`, `parts.slice(2, -1).join(':')`, NaN guards) is duplicated near-verbatim in two repositories. The encode side (`liftRecordNaturalKey`, `programSpecNaturalKey`) lives in `packages/core`. A delimiter change in core would break both repositories with no compile error.
**Suggestion:** Co-locate `parseLiftRecordNaturalKey` / `parseProgramSpecNaturalKey` next to the encoders in `packages/core` and import them from there.

---

**[maintainability]** `apps/api/src/programs/import.controller.ts:219–228`
The program-spec `updated` undo branch flags all updates as "cannot be undone automatically", but `buildProgramSpecPreImage` does store `before: { comparable: ... }`. This is either a half-implemented path (captured data unused) or a known limitation — it is not documented which.
**Suggestion:** Either implement the restore using the stored `before` data, or add a comment explaining why it is insufficient and document in the `ImportUndoResponse` JSDoc that program-spec row updates are not undoable.

---

**[correctness-minor]** `apps/api/src/programs/import.controller.ts:172–176`
On training-max `updated` undo, the restore writes `dateUpdated: current.dateUpdated` (the post-import timestamp). The weight is correctly restored, but the "last updated" date reflects the import event rather than the pre-import date. `buildTrainingMaxPreImage` does not capture `dateUpdated`.
**Suggestion:** Capture `dateUpdated` in the TM pre-image entry, or add a code comment noting this known data-fidelity limitation.

---

**[reliability]** `apps/web/app/(authed)/import/ImportWizard.tsx` (handleCommit, TM branch)
Training-max row exclusions are enforced by filtering `editedMaxes` client-side before building the FormData. The generic-delta path sends server-side `excludeKeys` params. If the `reviewMaxes !== null` condition is ever false, TM exclusions are silently dropped with no server-side fallback.
**Suggestion:** Unify exclusion enforcement (always send `excludeKeys` for TM commits too), or add a comment explaining why CSV-rebuild is the authoritative enforcement path for TMs.

---

### Questions for Author

**Question:** Was `import_batch` growth intentionally left unbounded for this phase?
**Context:** `save()` is called at every commit and is never paired with a delete. The undo handler reads but does not consume the batch row, so repeat undo is also possible.
**Tradeoffs:** (a) Delete batch on successful undo — most natural, prevents repeat undo, loses the audit record. (b) Keep batch permanently and prune by age — retains audit trail at the cost of a scheduled job. (c) Keep current behavior and document it — lowest scope now, commits to a future migration to add retention.

---

**Question:** Was the program-spec `updated` undo path intentionally left as a flagged non-operation?
**Context:** `buildProgramSpecPreImage` stores `before: { comparable: ... }` for updated rows, but `undoImport:219` flags all `updated` entries with "cannot be undone automatically" — the captured data is unused.
**Tradeoffs:** (a) Implement restore from stored `before` — needs `comparable` to round-trip back to a full spec row shape, which may not be fully captured. (b) Keep the flag and document it — honest UX, simple, but wasteful pre-image capture. (c) Stop capturing `before` for program-spec updates — matches actual undo behavior, simpler pre-image.

---

### Findings Disposition

Per ADR-028, each finding must be fixed in this PR or filed as a tracked issue and linked before merge.

| Finding | Suggested disposition |
|---|---|
| liftOverrides non-functional | Fix in PR |
| splitDest double-write | Fix in PR |
| Preview/commit split mismatch (no splitDest from web) | Fix in PR |
| undoImport missing `@RlsTxTimeout` | Fix in PR |
| N+1 getTrainingMaxes/getGoals in undo loop | Fix in PR |
| No server-side Phase 3 test coverage | Fix in PR |
| buildImportPreview.ts TS2304 | **Fixed in `28cc693`** |
| parseJsonParam silent-swallow | Fix or file |
| importBatch.save after writes | Fix or file |
| import_batch no TTL / repeat-undo | File |
| deleteLiftRecordsByNaturalKeys one-at-a-time | File |
| Duplicated natural-key parsing | File |
| program-spec updated undo inconsistency | File |
| TM dateUpdated not restored on undo | File |
| TM exclusion via CSV asymmetry | File |

---

*Reviewed by `claude-opus-4-6` via Claude Code*

Issue #633, opened 2026-07-02
Training-max precision loss and configurable workout-rounding increment


brownm09:

brownm09:
Both sub-issues merged: #634 (precision-loss bug fixes, PR #636) and #635 (configurable rounding increment, PR #640). Full audit inventory and design rationale preserved in this issue's history.

PR #560, opened 2026-06-17
[feat] Add 'enter training maxes directly' onboarding method

brownm09:
## Summary

Adds a fourth lift-discovery method to onboarding — **"Enter training maxes"** — for lifters who already train off fixed training maxes and don't want the app to re-derive them. Selecting it lets the user enter each lift's TM directly (weight only, no reps), and the value is persisted **as-is** with no 90% adjustment.

The existing `estimate` / `test` / `manual` methods are unchanged: they continue to compute a 1RM and persist 90% of it as the training max. The key refactor is that `OnboardingFlow` now resolves each row to its final `trainingMax` *before* calling `createFirstCycle`, and the server action persists that value verbatim — so the 90%-vs-as-entered decision lives in one place.

### Changes
- **StepMethod** — new `tm` option card.
- **StepLifts** — weight-only input (no reps) and a TM-specific hint for `tm`; the Brzycki helper box is hidden for both `tm` and `manual` (neither uses reps).
- **OnboardingFlow** — `computedMaxes` resolves each row to its final `trainingMax` (90% of the 1RM for `estimate`/`test`/`manual`, the entered value for `tm`); passes `method` to `StepConfirm`.
- **actions.createFirstCycle** — signature takes `{ lift, trainingMax }` and persists `trainingMax` directly instead of deriving 90% inside the action.
- **StepConfirm** — shows the TM as-entered for `tm`; shows the 1RM with the derived TM alongside it for the other methods.

## Acceptance Criteria
- [x] New onboarding method lets the user enter training maxes directly
- [x] Entered TMs are persisted without the 90% derivation
- [x] Existing estimate/test/manual methods are unchanged (still persist 90% of the 1RM)
- [x] Confirm step reflects the chosen method (TM as-entered vs 1RM + derived TM)

## Testing

All three web test layers run locally and pass.

- **Onboarding Jest suites** — `npm test -w @lifting-logbook/web -- --testPathPatterns onboarding`
  `Tests: 27 passed, 0 skipped, 0 failed (135 s)` — 6 suites.
- **Full web Jest suite** — `npx jest` (apps/web)
  `Tests: 95 passed, 0 skipped, 0 failed` — 17 suites.
- **Playwright E2E** — `npm run test:e2e -w @lifting-logbook/web`
[...]

PR #636, opened 2026-07-02
[fix] Preserve training-max decimal precision in Import, onboarding, and history

### LINKED ISSUE #634: Fix training-max precision loss in Import, onboarding, and cycle-generation history
## Problem

Training maxes carry up to 2 decimal places of precision (2.5 / 1.25 lb plate
increments, e.g. `316.25`). Four places silently round a directly-known value
(typed, imported, or already-persisted) to coarser precision:

1. `apps/web/app/(authed)/import/ImportWizard.tsx` `buildTrainingMaxesCsv()` —
   rounds to a whole number when rebuilding the commit CSV from the REVIEW
   step's editable list, even though REVIEW displays the precise value
   correctly.
2. `apps/web/app/(authed)/onboarding/steps/StepImport.tsx` `latestPerLift()` —
   rounds to a whole number when parsing an uploaded training-maxes CSV,
   contradicting its own UI copy ("we'll use the values as-is").
3. `apps/web/app/(authed)/onboarding/OnboardingFlow.tsx` `computedMaxes` — the
   `tm`/`import` onboarding methods round to a whole number, contradicting
   `lib.ts`'s documented "persisted as-is... no adjustment" contract.
4. `apps/api/src/programs/cycle-generation.service.ts` `buildHistoryEntries` —
   rounds to 1 decimal place purely to decide whether to write a
   `TrainingMaxHistory` row, coarser than the 2-decimal precision that matters
   here; a real change like `316.25 → 316.30` is silently dropped from history.

## Also in scope: estimate/manual-entry rounding

Per explicit instruction: the `estimate`/`test`/`manual` onboarding methods
(formula-derived or manually-typed 1RM, then derived to a training max at 90%)
should round DOWN to the nearest multiple of 2.5 lbs, not round-to-nearest —
applied to both the 1RM and the 90%-derived training max.

## Out of scope (confirmed, not bugs)

Plate-increment rounding of *computed* per-workout weights (`MROUND`) is
correct/required — see parent issue #633 for the full audit inventory of
confirmed-not-in-scope rounding sites.

Part of #633.

brownm09:
## Summary

Training maxes carry up to 2 decimal places of precision (2.5 / 1.25 lb plate
increments, e.g. `316.25`). Four places silently rounded a directly-known
value (typed, imported, or already-persisted) to coarser precision:

1. **`ImportWizard.tsx`** `buildTrainingMaxesCsv()` — rounded to a whole
   number when rebuilding the commit CSV from the REVIEW step's editable
   list, even though REVIEW displayed the precise value correctly.
2. **`StepImport.tsx`** `latestPerLift()` — rounded to a whole number when
   parsing an uploaded training-maxes CSV during onboarding, contradicting
   its own UI copy ("we'll use the values as-is").
3. **`OnboardingFlow.tsx`** `computedMaxes` — the `tm`/`import` onboarding
   methods rounded to a whole number, contradicting `lib.ts`'s documented
   "persisted as-is... no adjustment" contract.
4. **`cycle-generation.service.ts`** `buildHistoryEntries` — compared
   training maxes at 1 decimal place purely to decide whether to write a
   `TrainingMaxHistory` row; a real change like `316.25 → 316.30` both round
   to `316.3` and was silently dropped from history. Renamed `round1dp` →
   `round2dp`. The `weight` field written on a match was already full
   precision — only the change-detection comparison was too coarse.

## Also included: estimate/manual-entry rounding

Per explicit instruction, the `estimate`/`test`/`manual` onboarding methods
(formula-derived or manually-typed 1RM, then derived to a training max at
90%) now floor down to the nearest 2.5 lbs instead of rounding to the
nearest whole number — applied to both the 1RM and the 90%-derived training
max, via a new `floorToIncrement()` helper. These compute an approximation
rather than persist a known value as-is, so coarser, load-relevant precision
is intentional here (unlike the 4 bugs above).

Verified empirically that `Math.floor(value / 2.5) * 2.5` has no
floating-point drift risk across realistic weight ranges — no epsilon guard
needed.

## Out of scope (confirmed, not bugs)

Plate-increment rounding of *computed* per-workout weights (`MROUND`) is
correct/required — you can't load a bar to a non-plate weight. Full audit
inventory in parent issue #633.

## Testing

Ran `npm test -w @lifting-logbook/web` and `npm test -w @lifting-logbook/api`,
plus `npm run build`.

Tests: 145 passed, 0 skipped, 0 failed (web) + 406 passed, 0 skipped, 0 failed
(api).

New/updated coverage:
- `ImportWizard.test.tsx` — decimal weight survives the REVIEW→commit CSV rebuild
- `StepImport.test.tsx` — decimal weight survives CSV parse
- `lib.test.ts` — `floorToIncrement` unit tests + `brzycki1RM` cases that
  distinguish floor-to-2.5 from round-to-nearest
- `OnboardingFlow.test.tsx` — decimal precision proven for `tm`/`import`
  methods (previously only whole-number values were tested, which would have
  passed even with the bug present); new `manual`-method test for
  floor-to-2.5
- `cycle-generation.service.spec.ts` — new `buildHistoryEntries` unit tests
  (exported for direct testing; no existing coverage existed for this
  function). Proves a sub-0.1lb real change is no longer dropped from history.

Suppression grep, test-integrity grep: both clean. `baseline_test_failure_tracking`
is not enabled for this repo.

Closes #634

- **[reliability]** floorToIncrement() zero-increment guard — fixed in `9b3a813`
- **[documentation]** No single doc captured the training-max rounding contract — fixed in `9b3a813` (added docs/standards/training-max-precision.md, indexed in CLAUDE.md)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -330,6 +330,7 @@ rationale, examples, and enforcement notes. Existing standards:
 |---|---|---|
 | [`docs/standards/fetch-cache-semantics.md`](docs/standards/fetch-cache-semantics.md) | `apps/web` Server Components | All `fetch()` calls must specify `{ cache: 'no-store' }` or `{ next: { revalidate: N } }` explicitly |
 | [`docs/standards/error-fallback-test-coverage.md`](docs/standards/error-fallback-test-coverage.md) | all packages and apps | Error-swallowing fallbacks (`.catch(() => default)`, `?? default`, try/catch returning neutral) require a data-level assertion, a paired success-path test, or a documented structure-only comment |
+| [`docs/standards/training-max-precision.md`](docs/standards/training-max-precision.md) | all packages and apps | Directly-known training-max weights are never rounded; formula-derived estimates floor to the nearest plate increment; computed per-workout weights round to the nearest *loadable* plate combination |
 
 When implementing `apps/web`, read the relevant standards before writing any `fetch()` calls.
 
--- a/apps/api/src/programs/cycle-generation.service.spec.ts
+++ b/apps/api/src/programs/cycle-generation.service.spec.ts
@@ -10,7 +10,7 @@ import {
 import { ICycleScheduledWorkoutRepository } from '../ports/ICycleScheduledWorkoutRepository';
 import { IUserSettingsRepository } from '../ports/IUserSettingsRepository';
 import { ProgramNotFoundError } from '../ports/errors';
-import { CycleGenerationService } from './cycle-generation.service';
+import { buildHistoryEntries, CycleGenerationService } from './cycle-generation.service';
 
 const PROGRAM = '5-3-1';
 
@@ -388,5 +388,36 @@ describe('CycleGenerationService', () => {
         ]),
       );
     });
+  });
+});
+
+describe('buildHistoryEntries', () => {
+  const asOf = (weight: number) => ({
+    lift: 'Squat',
+    weight,
+    dateUpdated: new Date('2026-04-18T00:00:00.000Z'),
+  });
+  const date = new Date('2026-04-20T00:00:00.000Z');
+
+  it('records a real sub-0.1lb change that a 1-decimal-place comparison would have missed', () => {
+    // 316.25 -> 316.30 is a genuine 0.05lb change; both round to 316.3 at 1
+    // decimal place, which would have silently dropped this from history.
+    const entries = buildHistoryEntries([asOf(316.25)], [asOf(316.3)], date, 'program');
+
+    expect(entries).toHaveLength(1);
+    expect(entries[0]).toMatchObject({ lift: 'Squat', weight: 316.3 });
+  });
+
+  it('does not record an unchanged weight', () => {
+    const entries = buildHistoryEntries([asOf(316.25)], [asOf(316.25)], date, 'program');
+
+    expect(entries).toHaveLength(0);
+  });
+
+  it('records a lift with no prior entry', () => {
+    const entries = buildHistoryEntries([], [asOf(316.25)], date, 'test');
+
+    expect(entries).toHaveLength(1);
+    expect(entries[0]).toMatchObject({ lift: 'Squat', weight: 316.25, source: 'test' });
   });
 });
--- a/apps/api/src/programs/cycle-generation.service.ts
+++ b/apps/api/src/programs/cycle-generation.service.ts
@@ -93,19 +93,20 @@ async function saveScheduledDates(
   }
 }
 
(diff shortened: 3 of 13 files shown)


brownm09:
## Review: [fix] Preserve training-max decimal precision in Import, onboarding, and history

### Summary
Ready to merge. This is a well-scoped bug fix (217 lines across 11 files, all in one theme) with strong, deliberately-constructed regression coverage — several of the new/updated tests use values chosen specifically to distinguish the old buggy behavior from the fix (e.g. `202.5` vs `203`, `220` vs `222`), rather than incidentally passing either way. No blocking findings. Two non-blocking suggestions below, both preventive rather than urgent.

---

### Non-Blocking Findings

**[reliability]** `apps/web/app/(authed)/onboarding/lib.ts:67-75`
`floorToIncrement(value, increment = 2.5)` divides by `increment` with no guard against `increment === 0`, which would produce `NaN`. Every current call site uses the literal default or a test-only literal, so this isn't reachable today — but the function is now exported and reusable, and the adjacent, structurally-identical `MROUND` utility in `packages/core/src/constants/config.ts` explicitly guards this exact case (`if (multiple === 0) return number;`). If `floorToIncrement` is later wired to a caller-supplied increment (e.g. PR B's configurable rounding setting), the inconsistency becomes a live footgun.
**Suggestion:** Add the same zero-guard `MROUND` uses, for parity with the sibling utility and to fail safe rather than silently produce `NaN` if a future caller passes a variable increment.

**[documentation]** No file — repo-wide
This bug existed because PR #560 copy-pasted an adjacent rounding pattern (`Math.round`) onto a code path whose own docstring already promised "persisted as-is... no adjustment" — the contract existed in a comment, not in an enforced or discoverable place. This PR now encodes three distinct precision contracts in one small area (full precision for direct entry/import, floor-to-2.5 for estimates, 2-decimal comparison tolerance for history diffing) with no single place a future contributor would find all three before touching this code again.
**Suggestion:** Consider a short `docs/standards/` entry (matching the existing `fetch-cache-semantics.md` / `error-fallback-test-coverage.md` pattern) stating the training-max precision contract and citing this PR as the motivating incident, the same way `docs/standards/error-fallback-test-coverage.md` already does for its own bug class. Not blocking — the four docstrings/comments already added in this PR are reasonable inline coverage — but a standards doc is what the suppression/test-integrity policies already do for other recurring mistake classes in this repo, and this is exactly that shape of recurring mistake.

---

### Findings Disposition

Both findings above are non-blocking suggestions with no urgency — file-and-link is appropriate for both rather than blocking this PR.

---

PR #640, opened 2026-07-02
[feat] Add configurable default workout-rounding increment

### LINKED ISSUE #635: Add configurable default workout-rounding increment
## Problem

The multiple used to snap computed per-workout target weights to a loadable
plate combination is effectively fixed today: preset programs hardcode a
per-lift `increment` (5-3-1 uses `10` for heavy compounds; RPT/leangains use
`{2.5, 5, 10}`), and new custom-program rows default to a hardcoded `5`
(`ProgramEditor.tsx` `DEFAULT_ROW`). Not everyone has access to 1.25 lb plates
or smaller, so this should be user-configurable.

## Confirmed design

- **Global default, not a preset change.** Preset templates keep their
  existing hardcoded per-lift increments unchanged — these are shared,
  canonical templates. Overriding any field on a preset-derived program is
  what turns it into a *custom program* (the preset becomes that custom
  program's `baseTemplate`), so the new setting only affects custom programs.
- **Two knobs:** (1) a new **global default**, constrained to exactly
  `{0.625, 1.25, 2.5, 5}` lbs, defaulting to `1.25`; (2) the **existing
  per-row override** in the custom program editor stays exactly as-is — a
  free-form number input, so a user can still replicate a preset's `10` lb
  heavy-lift increment or any other value on their own program. Only the
  default is curated; overrides stay flexible.

## Implementation

1. Prisma migration: `UserSettings.defaultWeightIncrement Float?` (nullable,
   no DB default, matching `activeProgram`/`workoutSchedule`'s pattern).
2. DTO validation: `@IsOptional() @IsIn([0.625, 1.25, 2.5, 5])` on the
   update-settings DTO.
3. Shared types: `defaultWeightIncrement: number | null` in
   `packages/types/src/api.ts`.
4. Settings UI: a `<select>` with the 4 allowed values.
5. Wire into `ProgramEditor.tsx`'s `DEFAULT_ROW` — new spec rows seed from the
   user's configured default (`?? 1.25`) instead of the hardcoded `5`.
6. Not changed: `create-custom-program.dto.ts`'s existing
   `@IsNumber() @IsPositive()` per-row validation stays unconstrained — this
   is the override path.

Part of #633.

brownm09:
## Summary

The multiple used to snap computed per-workout target weights to a loadable
plate combination (`MROUND`) was effectively fixed: preset programs hardcode
a per-lift `increment` (5-3-1 uses `10` for heavy compounds; RPT/leangains
use `{2.5, 5, 10}`), and new custom-program rows defaulted to a hardcoded
`5`. Not everyone has access to 1.25 lb plates or smaller, so this adds a
user-configurable default.

## Design (confirmed via clarifying questions, see #635)

- **Global default, not a preset change.** Preset templates keep their
  existing hardcoded per-lift increments unchanged — these are shared,
  canonical templates. Overriding a preset's fields is what turns it into a
  *custom program* (the preset becomes that custom program's
  `baseTemplate`), so the new setting only affects custom programs.
- **Two knobs:** (1) a new global default, constrained to exactly
  `{0.625, 1.25, 2.5, 5}` lbs, defaulting to `1.25`; (2) the existing
  per-row override in the custom program editor stays exactly as-is — a
  free-form number input, so a user can still replicate a preset's `10` lb
  heavy-lift increment or any other value on their own program.

## Changes

1. **Prisma migration** — `UserSettings.defaultWeightIncrement Float?`
   (nullable, no DB default, matching `activeProgram`/`workoutSchedule`'s
   pattern; app-level default of `1.25` resolved at point of use).
2. **DTO validation** — `@IsOptional() @ValidateIf(≠null) @IsIn([0.625, 1.25, 2.5, 5])`
   on `UpdateSettingsDto`, following the same null-clears/undefined-leaves-unchanged
   convention as `workoutSchedule`.
3. **Shared types** — `defaultWeightIncrement: number | null` on
   `UserSettingsResponse`/`UpdateUserSettingsRequest`, plus exported
   `WEIGHT_INCREMENT_OPTIONS` / `DEFAULT_WEIGHT_INCREMENT` constants.
4. **New settings page** — `/settings/weight-rounding` (a `<select>` with the
   4 allowed values). No settings index page exists yet to link a new route
   into — `/settings/schedule` itself isn't linked from anywhere in the app
   either — so this follows the repo's existing one-folder-per-concern
   pattern (`schedule/`, `strength-goals/`, `training-maxes/`) rather than
   overloading an existing page with an unrelated concern.
5. **`ProgramEditor.tsx` wiring** — `DEFAULT_ROW` (and its two callers,
   `buildDefaultSpecs`/`buildSpecsFromTemplate`) now take the increment as a
   parameter; new custom-program spec rows seed from the user's configured
   default (`?? 1.25`) instead of the hardcoded `5`. Preset-seeded rows
   (`buildSpecsFromTemplate`'s `seeded` branch) are untouched — they keep
   whatever increment the template defines.
6. **Not changed:** `create-custom-program.dto.ts`'s existing
   `@IsNumber() @IsPositive()` per-row validation stays unconstrained — this
   is the override path and must stay flexible per the confirmed design.

## Suppression note

The 3 new `?? null` occurrences in `user-settings.repository.ts` are not
suppressions — they coalesce nullable Prisma fields to explicit `null` for
the `UserSettingsResponse` contract, mirroring the pre-existing
`activeProgram`/`workoutSchedule` patterns in the same functions. No type
error is being silenced.

## Out-of-band fix

`apps/web/e2e/staging.spec.ts`'s error-fallback-coverage line-number
reference (`page.tsx:10-11`) was updated to `:14-15` after
`DEFAULT_SETTINGS` grew from a one-line to a four-line object literal in
`apps/web/app/(authed)/programs/page.tsx` — caught by the
`no-uncovered-error-fallback` pre-commit lint rule.

## Testing

Ran `npm test -w @lifting-logbook/web` and `npm test -w @lifting-logbook/api`,
plus `npm run build` (all 6 packages).

Tests: 155 passed, 0 skipped, 0 failed (web) + 418 passed, 0 skipped, 0 failed
(api).

New coverage:
- `update-settings.dto` / `user-settings.controller.spec.ts` — accepts each
  of the 4 allowed values, rejects out-of-range values (0, 1, 3, 10, -2.5),
  accepts/persists an explicit `null` to clear, persists and reads back a
  set value end-to-end through the controller.
- `WeightIncrementForm.test.tsx` — defaults to 1.25 when unset, preselects a
  stored value, lists exactly the 4 allowed options, saves and shows a
  confirmation, shows an error on save failure.
- `page.test.tsx` (weight-rounding) — data-level assertion on success,
  distinct fallback value on API-unreachable (per
  `docs/standards/error-fallback-test-coverage.md`).
- `ProgramEditor.test.tsx` (new file) — new spec rows seed from the
  configured default; falls back to 1.25 when unset.

Manual verification deferred: the mock Playwright API server
(`apps/web/e2e/mock-api.mjs`) has no `PATCH /users/me/settings` handler for
either `workoutSchedule` or the new field, so end-to-end save-through-UI
isn't covered by Playwright yet — matches the pre-existing gap for the
sibling `workoutSchedule` field, not something newly introduced here. The
mock's GET response and reset-state now include `defaultWeightIncrement` so
existing/future Playwright tests reading settings don't break.

Suppression grep, test-integrity grep: both clean (see suppression note
above). `baseline_test_failure_tracking` is not enabled for this repo.

Closes #635

- **[maintainability]** Misleading UI copy (claimed live rounding) — fixed in `8c1a56b`
- **[maintainability]** saveWeightIncrement cannot send null — documented as intentional in `8c1a56b`
- **[maintainability]** Duplicated allowlist literal in DTO — fixed in `8c1a56b` (now references WEIGHT_INCREMENT_OPTIONS)


brownm09:
## Review: [feat] Add configurable default workout-rounding increment

### Summary
Ready to merge. Two parallel passes (correctness/security; reliability/performance/maintainability) over the 448-line diff found no blocking issues — the DTO validation correctly distinguishes valid-value / explicit-null-clears / undefined-leaves-unchanged, the repository's null-handling mirrors the existing `workoutSchedule` convention exactly, and RLS scoping is preserved. Three non-blocking maintainability findings, the most notable being user-facing copy that misdescribed what the setting actually does — all three fixed in this PR rather than filed.

---

### Non-Blocking Findings

**[maintainability]** `apps/web/app/(authed)/settings/weight-rounding/WeightIncrementForm.tsx:36`
The description text said "Computed workout weights are rounded to the nearest multiple of this value" — but this setting doesn't drive any live rounding. The runtime `MROUND` calls that actually compute per-workout weights (`workoutPlan.ts`, `calculateLiftWeights.ts`, the workout detail page) all round to each program spec row's own `increment`, never to this global default. The setting's real job is narrower: it seeds the `increment` field on *new* custom-program spec rows. A user changing this and seeing no effect on their existing programs would be confused, and a future maintainer "fixing" the described-but-nonexistent live-rounding behavior would be fixing the wrong layer.
**Fix:** Reworded to describe what actually happens — sets the default for new custom-program lifts, per-lift override still available, presets unaffected, training maxes always exact.

**[maintainability]** `apps/web/app/(authed)/settings/weight-rounding/actions.ts:9`
`saveWeightIncrement` only accepts `number`, so the "clear back to app default" path — which the DTO, repository, and shared types all explicitly support (`null` semantics, with a dedicated controller-spec test) — is unreachable from this UI. Reads as an oversight rather than a deliberate scope boundary.
**Fix:** Added a comment explaining this is intentional: the `<select>` always resolves to one of the 4 concrete values (never an "unset" state), so there's no client-triggerable case that needs `null` today; the API-side support remains available for other callers or future UI.

**[maintainability]** `apps/api/src/user-settings/update-settings.dto.ts:98` / `packages/types/src/api.ts`
The allowed-values set `{0.625, 1.25, 2.5, 5}` was hardcoded a second time in `@IsIn([0.625, 1.25, 2.5, 5])`, duplicating the `WEIGHT_INCREMENT_OPTIONS` constant this same PR exports for exactly this purpose. `packages/types` is already a DTO dependency (imports `SCHEDULE_LIMITS`/`isValidSchedule` a few lines up), so the fix was in scope. If the allowed set ever changes, the two lists could silently diverge — the `<select>` would offer a value the API rejects, or vice versa.
**Fix:** `@IsIn(WEIGHT_INCREMENT_OPTIONS)` — single source of truth.

---

### Findings Disposition

All three findings fixed in this PR (commit to follow this review).

---

PR #645, opened 2026-07-03
[fix] Resolve PrismaService lazily in RlsInterceptor to fix silently-disabled RLS

### LINKED ISSUE #644: [fix] RlsInterceptor silently disabled since RLS cutover — DI ordering causes PrismaService to resolve null
## Problem

\"Start My Program\" fails in production for every first-time user with a generic 500. Root-caused via production Cloud Logging + a from-scratch reproduction against the real compiled app: the actual defect is much broader than onboarding.

## Root cause

`RlsInterceptor` (apps/api/src/adapters/prisma/rls.interceptor.ts) is registered globally via `{ provide: APP_INTERCEPTOR, useClass: RlsInterceptor }` in `RepositoryFactoryModule`. It previously constructor-injected `PrismaService` as `@Optional()`.

NestJS instantiates `APP_INTERCEPTOR`-bound providers as part of its early global-enhancer setup — before `PrismaService`'s `useFactory` provider (declared in the *same* module) is guaranteed to have run. The `@Optional()` injection silently resolved to `null` at that moment. Because `RlsInterceptor` is a singleton, that `null` was captured **permanently** — `PrismaService` becoming available moments later (confirmed via `app.get(PrismaService)` returning non-null immediately after boot) never reached the interceptor.

`RlsInterceptor.intercept()`'s early-return guard (`if (!prisma || ...) return next.handle();`) therefore fired on every single request since the RLS migration (`20260611000000_enable_rls`, #511) landed on 2026-06-11. The `app.current_user_id` Postgres session variable was never set for **any** request.

## Impact (confirmed via production log query + local reproduction against the real `lifting_app` role)

- **Every first-time INSERT into a userId-scoped table failed** with a `42501` \"new row violates row-level security policy\" error (Postgres `FORCE ROW LEVEL SECURITY` rejects an insert when the `WITH CHECK` GUC comparison evaluates against `NULL`). This is what broke `POST /programs/:program/cycles/initialize` (\"Start My Program\") and equally affects `PATCH /users/me/settings` and any other first-write endpoint.
- **Every read silently returned \"not found\"** even for genuinely existing rows, since Postgres's fail-closed RLS behavior (`\"userId\" = current_setting(..., true)` where the GUC is `NULL`) filters out all rows indistinguishably from \"no data exists.\"
- **Not a cross-tenant data leak.** Every repository query still carries its original, first-layer `where: { userId }` application-level filter (predating RLS, per the RLS migration's own design as defense-in-depth). RLS was a *second* layer that has been inert; no evidence any request ever saw another user's data.

## Why no test caught this

Every existing DB E2E suite connects as the Postgres bootstrap superuser (bypasses RLS entirely by definition) or manually constructs `RlsInterceptor`/`PrismaRepositoryFactory` directly (bypassing NestJS's real DI instantiation order, which is exactly where the bug lives). No test exercised the combination of: real `AppModule` boot, real `APP_INTERCEPTOR`/`APP_GUARD` wiring, and the restricted `lifting_app` role together.

## Fix

- `RlsInterceptor` now resolves `PrismaService` lazily via `ModuleRef.get(PrismaService, { strict: false })` inside `intercept()` on every request, instead of capturing it once via constructor injection. This is immune to instantiation-order timing since `ModuleRef` is always available and the lookup always reflects the current DI container state.
- Added two full-app-boot regression tests (`rls.db.e2e.spec.ts`, new \"RLS request wiring (interceptor + factory, full app boot)\" describe block) that boot the real `AppModule` via Fastify `.inject()` against the real restricted `lifting_app` role — one proving a first-time INSERT now succeeds, one proving a genuinely-existing row is now read back correctly. Verified both fail against the pre-fix code and pass against the fix.
- Extended `jest.env.setup.js`'s `DATABASE_URL` write-guard to also allow the `lifting_app`-role variant of the Testcontainers sentinel URL through (scoped to the same host/pathname, so it cannot smuggle an arbitrary DB URL in) — needed so the new tests can make `PrismaService`'s `env(\"DATABASE_URL\")` read resolve to the restricted role.

## Testing

\`\`\`
npm test
\`\`\`
Tests: 420 passed, 0 skipped, 0 failed (api) + all other workspaces green (web 147, core, types, api-client, eslint-rules, mobile).

New coverage: 2 new tests in `rls.db.e2e.spec.ts`'s new \"RLS request wiring (interceptor + factory, full app boot)\" block. Confirmed via manual revert that both fail against the pre-fix code (500) and pass against the fix (201/200).

brownm09:
## Summary

"Start My Program" was failing for every first-time user in production. Root cause is much broader than onboarding: `RlsInterceptor` has silently never set the Postgres RLS `app.current_user_id` session variable for **any** request since the RLS migration (#511) landed on 2026-06-11.

**Root cause:** `RlsInterceptor` is bound globally via `{ provide: APP_INTERCEPTOR, useClass: RlsInterceptor }`. NestJS instantiates `APP_INTERCEPTOR`-bound providers as part of its early global-enhancer setup — before `PrismaService`'s `useFactory` provider (declared in the *same* module) is guaranteed to have run. The interceptor's constructor-injected `@Optional() PrismaService` silently resolved to `null` at that moment, and because the interceptor is a singleton, that `null` was captured **permanently** (confirmed empirically: `app.get(PrismaService)` returns a real instance moments later, but the interceptor's own copy never updates).

**Impact**, confirmed via production log analysis + a from-scratch reproduction against the real compiled app and the actual restricted `lifting_app` role:
- Every first-time INSERT into a userId-scoped table failed with a `42501` row-level-security violation (`WITH CHECK` evaluates against a `NULL` GUC) — this is what broke `POST /programs/:program/cycles/initialize`, and equally affects any other first-write endpoint (verified with `PATCH /users/me/settings` too).
- Every read silently returned "not found," even for genuinely existing rows — Postgres's fail-closed RLS behavior can't distinguish "GUC unset" from "no matching rows."
- **Not a cross-tenant data leak**: every repository query still carries the original, first-layer `where: { userId }` application-level filter that predates RLS. RLS was a second defense-in-depth layer that has been inert; there's no evidence any request ever saw another user's data.

**Why no test caught this:** every existing DB E2E suite either connects as the Postgres bootstrap superuser (bypasses RLS by definition) or manually constructs `RlsInterceptor`/`PrismaRepositoryFactory` directly, bypassing NestJS's real DI instantiation order — exactly where this bug lives.

## Changes

- `apps/api/src/adapters/prisma/rls.interceptor.ts` — resolve `PrismaService` lazily via `ModuleRef.get(PrismaService, { strict: false })` inside `intercept()` on every request, instead of capturing it once via constructor injection. Immune to instantiation-order timing since `ModuleRef` is always available and the lookup always reflects current DI container state. Caches the resolved instance after the first non-null lookup so steady-state requests pay no repeated module-graph search.
- `apps/api/src/adapters/prisma/rls.db.e2e.spec.ts` — updated the existing "RLS request wiring" test's manual `RlsInterceptor` construction to pass the testing module (which implements `ModuleRef`) instead of a raw `PrismaService`, matching the new constructor signature. Added a new "RLS request wiring (interceptor + factory, full app boot)" describe block that boots the real `AppModule` via Fastify `.inject()` against the real restricted `lifting_app` role — proving a first-time INSERT now succeeds and a genuinely-existing row is read back correctly. Manually verified both new tests fail against the pre-fix code (500) and pass against the fix (201/200).
- `apps/api/jest.env.setup.js` — extended the `DATABASE_URL` write-guard Proxy to also allow the `lifting_app`-role variant of the Testcontainers sentinel URL through (scoped to the same host/pathname, so it can't smuggle an arbitrary DB URL in), needed so the new full-app-boot test can make `PrismaService`'s `env("DATABASE_URL")` read resolve to the restricted role.
- `docs/adr/ADR-010-multi-tenancy-data-isolation.md` — added a dated correction recording that the RLS layer was not actually live from 2026-06-11 until this PR, and rewrote the Verification bullet, which previously claimed (inaccurately) that the existing test suite proved the interceptor wired the GUC end to end.

`RlsContextService` (used by `@SkipRlsTransaction()` handlers like cycle-plan) has the same `@Optional() PrismaService` constructor pattern but is **not** affected — it's a regular provider (not `APP_*`-bound), so it follows normal DI dependency-before-dependent resolution. Verified via diagnostic instrumentation during investigation: it consistently receives a non-null `PrismaService`.

## Testing

```
npm test
```
Tests: 420 passed, 0 skipped, 0 failed (api) — up from 418, +2 new regression tests. All other workspaces green (web 147, core, types, api-client, eslint-rules, mobile). Full monorepo run: 12/12 tasks passed. Re-verified after the review-fix commit — still 420/420 + all workspaces green.

Suppression grep, test-integrity grep: both clean. No `package.json` changes, no lockfile drift risk.

| Finding | Category | Disposition |
|---|---|---|
| ADR-010 asserted the RLS layer was live and verified end-to-end when it was inert for 3 weeks | blocking [documentation] | Fixed in `1f33b23` — added a dated correction and rewrote the Verification bullet |
| `moduleRef.get()` ran on every request (incl. health probes) with no caching | non-blocking [performance] | Fixed in `1f33b23` — cached after first non-null resolution |
| Locally-defined `appRoleUrl()` shadowed the existing module-level helper | non-blocking [maintainability] | Fixed in `1f33b23` — removed, now calls the existing helper |
| Stale "single legitimate write path" comment in `jest.env.setup.js` | non-blocking [maintainability] | Fixed in `1f33b23` — comment updated to reflect both paths |
| `RlsContextService` "unaffected" claim rests on manual diagnostic, not an automated test | non-blocking [maintainability] | Filed and linked — [#646 comment](https://github.com/brownm09/lifting-logbook/issues/646#issuecomment-4872233906), folded into that issue's test-default work rather than new infra here |

Closes #644

### DIFF
--- a/apps/api/jest.env.setup.js
+++ b/apps/api/jest.env.setup.js
@@ -49,13 +49,34 @@ process.env = new Proxy(originalEnv, {
     // Allow the DB E2E spec to restore DATABASE_URL from the Testcontainers
     // sentinel set by jest.global-setup.js. This is the single legitimate
     // write path for DATABASE_URL during a test run.
-    if (
-      String(key) === 'DATABASE_URL' &&
-      value &&
-      value === target.LIFTING_TC_DATABASE_URL
-    ) {
-      target[key] = value;
-      return true;
+    if (String(key) === 'DATABASE_URL' && value) {
+      const sentinel = target.LIFTING_TC_DATABASE_URL;
+      if (value === sentinel) {
+        target[key] = value;
+        return true;
+      }
+      // Also allow the lifting_app-role variant of the sentinel (same host/port/database,
+      // different credentials) — needed by full-app-boot RLS tests that must connect as the
+      // restricted runtime role rather than the Testcontainers owner, so PrismaService's
+      // env("DATABASE_URL") read resolves to a real, RLS-enforcing connection. Scoped to the
+      // sentinel's own host/pathname so this cannot be used to smuggle an arbitrary DB URL in.
+      // See rls.db.e2e.spec.ts's "RLS request wiring (interceptor + factory, full app)" block.
+      if (sentinel) {
+        try {
+          const candidate = new URL(value);
+          const base = new URL(sentinel);
+          if (
+            candidate.host === base.host &&
+            candidate.pathname === base.pathname &&
+            candidate.username === 'lifting_app'
+          ) {
+            target[key] = value;
+            return true;
+          }
+        } catch {
+          // Not a valid URL — fall through and discard below.
+        }
+      }
     }
     if (BLOCK.includes(String(key))) {
       return true; // discard — blocked keys are frozen at ''
--- a/apps/api/jest.env.setup.js
+++ b/apps/api/jest.env.setup.js
@@ -47,8 +47,9 @@ const originalEnv = process.env;
 process.env = new Proxy(originalEnv, {
   set(target, key, value) {
     // Allow the DB E2E spec to restore DATABASE_URL from the Testcontainers
-    // sentinel set by jest.global-setup.js. This is the single legitimate
-    // write path for DATABASE_URL during a test run.
+    // sentinel set by jest.global-setup.js, or from its lifting_app-role variant
+    // (see below) — these are the only legitimate write paths for DATABASE_URL
+    // during a test run.
     if (String(key) === 'DATABASE_URL' && value) {
       const sentinel = target.LIFTING_TC_DATABASE_URL;
       if (value === sentinel) {
--- a/apps/api/src/adapters/prisma/rls.db.e2e.spec.ts
+++ b/apps/api/src/adapters/prisma/rls.db.e2e.spec.ts
@@ -12,11 +12,13 @@
 // rows). Enforcement assertions use the lifting_app client.
 import 'reflect-metadata';
 import { ExecutionContext, CallHandler } from '@nestjs/common';
-import { Reflector } from '@nestjs/core';
+import { NestFactory, Reflector } from '@nestjs/core';
 import { Test } from '@nestjs/testing';
+import { FastifyAdapter, NestFastifyApplication } from '@nestjs/platform-fastify';
 import { PrismaClient } from '@prisma/client';
 import { ClsModule, ClsService } from 'nestjs-cls';
 import { lastValueFrom, from } from 'rxjs';
+import { AppModule } from '../../app.module';
 import { PrismaService } from './prisma.service';
 import { PrismaRepositoryFactory } from './prisma-repository-factory';
 import { RlsInterceptor } from './rls.interceptor';
@@ -279,7 +281,10 @@ describeOrSkip('RLS request wiring (interceptor + factory)', () => {
     prisma = moduleRef.get(PrismaService);
     await prisma.$connect();
     factory = new PrismaRepositoryFactory(prisma, cls);
-    interceptor = new RlsInterceptor(cls, moduleRef.get(Reflector), prisma);
+    // RlsInterceptor resolves PrismaService lazily via ModuleRef (see rls.interceptor.ts) rather
+    // than constructor injection, so the test module itself — which implements ModuleRef and
+    // already has PrismaService registered above — stands in for the real one.
+    interceptor = new RlsInterceptor(cls, moduleRef.get(Reflector), moduleRef);
   });
 
   afterAll(async () => {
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: [fix] Resolve PrismaService lazily in RlsInterceptor to fix silently-disabled RLS

### Summary
The core fix is sound and well-verified — the diff, PR description, and test evidence all corroborate the root-cause analysis, and the two new full-app-boot tests genuinely close the coverage gap that let this ship. This is mergeable as-is from a code-correctness standpoint; the one blocking item is that `docs/adr/ADR-010-multi-tenancy-data-isolation.md` currently asserts the exact thing this PR proves was false for 3+ weeks, and nothing in this PR corrects that record. DIFF_SIZE (153 lines) is appropriately scoped — no split needed.

---

### Blocking Findings

**[documentation]** `docs/adr/ADR-010-multi-tenancy-data-isolation.md:7,82`
The ADR's "Implementation status" line states the RLS layer "is now live" and its Verification section claims `rls.db.e2e.spec.ts` "proves... that the interceptor wires the GUC end to end." Both claims were false from 2026-06-11 to this PR's merge — the interceptor never wired the GUC to anything, and the existing test suite did not actually prove it did (that's the entire premise of issue #644). A reader consulting this ADR today has no way to know that period existed, or that the "Verification" bullet's claim was inaccurate for most of the feature's life.
**Fix:** Add a dated amendment to ADR-010 (matching the existing "Implementation status (2026-06-11)" convention) noting the #644 finding, the window it was inert, and a corrected description of what the pre-fix test suite actually verified (GUC-setting mechanics under the superuser role, not enforcement under the real DI-wired path).

---

### Non-Blocking Findings

**[performance]** `apps/api/src/adapters/prisma/rls.interceptor.ts:53`
`moduleRef.get(PrismaService, { strict: false })` now runs on every single HTTP request that reaches this global interceptor — including unauthenticated routes like `/health`/`/readyz`, which return early only *after* this line runs. `strict: false` searches across the whole module graph rather than a single scoped container, and this replaces what used to be a zero-cost property read. Since `PrismaService`, once resolved, is a singleton that will never again become unavailable for the process's lifetime, every lookup after the first is pure repeated overhead — including on Cloud Run's own frequent liveness/readiness probe traffic.
**Suggestion:** Cache the resolved instance in a private field on first non-null resolution, falling back to `moduleRef.get()` only until that first success. Preserves the fix (which requires deferring past constructor time) while avoiding a full graph lookup on every subsequent request.

**[maintainability]** `apps/api/src/adapters/prisma/rls.db.e2e.spec.ts:367`
The new describe block defines a local `function appRoleUrl(): string`, which shadows the existing module-level `function appRoleUrl(ownerUrl: string): string` at line 51 — same name, different signature (zero-arg closing over `OWNER_URL` vs. taking the URL as a parameter). They produce identical output, but the name collision means a future reader skimming the file can't tell at a glance which `appRoleUrl` is in scope at a given point, and could copy code between the two describe blocks expecting the wrong signature.
**Suggestion:** Delete the local function and call the existing one directly: `appRoleUrl(OWNER_URL)`.

**[maintainability]** `apps/api/jest.env.setup.js:50`
The unchanged comment above the `DATABASE_URL` check still reads "This is the single legitimate write path for DATABASE_URL during a test run," but the code immediately below it now has two legitimate paths (the exact sentinel match, and the new `lifting_app`-role variant added at line 58). A reader who stops at the top comment gets a stale picture of what the guard allows.
**Suggestion:** Update the comment to say "these are the legitimate write paths" (plural) or otherwise reflect both branches.

**[maintainability]** `apps/api/src/adapters/prisma/rls.interceptor.ts` (PR body, "Changes" section)
The PR body states `RlsContextService` is unaffected, verified via "diagnostic instrumentation during investigation" — but that verification isn't preserved as an automated test. `RlsContextService` backs `@SkipRlsTransaction()` routes (e.g., cycle-plan), and the new full-app-boot tests only exercise `RlsInterceptor`'s primary path, not that alternate one. If a future change to `RlsContextService` reintroduced a similar DI-timing issue, nothing in the suite running under the restricted role would catch it.
**Suggestion:** Consider extending the new "full app boot" pattern to a `@SkipRlsTransaction()` endpoint in a follow-up (could fold into #646, which already covers flipping test defaults to the restricted role) so this claim stays continuously verified rather than a point-in-time finding.

---

### Findings Disposition
Per ADR-028, each finding above must be fixed in this PR or filed as a tracked issue and linked before merge.

---

Issue #653, opened 2026-07-03
[chore] apps/web local node_modules corruption — next/react-dom/@testing-library/@clerk dirs empty, 18 test suites fail to load


brownm09:
## Problem

Local `npm test -w @lifting-logbook/web` fails 18 of 24 test suites on this Windows machine with `Cannot find module` errors for `next/server`, `next/navigation`, `@testing-library/react`, `@clerk/nextjs/server`, and `react-dom/server`. 40 tests pass, 0 fail on assertions — every failure is "Test suite failed to run" (module resolution), not a real test failure.

## Root cause (confirmed, not the documented Node 24 tarball bug)

`apps/web/node_modules/next`, `react-dom`, `@testing-library/react`, and `@clerk/nextjs` directories **exist but contain zero files**:

```
next: dir exists, 0 files
react-dom: dir exists, 0 files
@testing-library/react: dir exists, 0 files
@clerk/nextjs: dir exists, 0 files
```

This is a local install corruption, similar in symptom to the documented "Node 24 caveat" in `CLAUDE.md` (missing/truncated package contents), but confirmed **not** that bug — this machine is on the pinned Node 20.11.1 (`.nvmrc`), not Node 24. Root trigger unconfirmed; disk was at 96% (23G free of 476G) at time of investigation, which may have caused a prior `npm install` to partially fail silently.

## Confirmed NOT caused by

- **Not a code regression**: CI is green on `main` at the exact commit this was found on (`fc4adc9`, https://github.com/brownm09/lifting-logbook/actions — CI + Deploy both `success`), so this doesn't reproduce on Linux/CI.
- **Not lockfile drift**: `npm install --package-lock-only --ignore-scripts` followed by `git diff --exit-code package-lock.json` shows the lockfile is in sync.
- **Not concurrency/parallel-run related**: reproduces identically running `npm test -w @lifting-logbook/web` standalone and as part of the full `npm test` turbo pipeline.
- **Not related to PR #649's changes**: that PR touches only `apps/api/src/adapters/prisma/rls.interceptor.ts` + a new spec file + `CLAUDE.md`; nothing under `apps/web`.

## Recovery (untested — flagging for the user to confirm before running)

[...]

Issue #649, opened 2026-07-03
[reliability] RlsInterceptor should fail loudly, not silently no-op, when RLS context can't be established


brownm09:
## Problem

Before #645, `RlsInterceptor.intercept()`'s guard — `if (!prisma || context.getType() !== 'http') { return next.handle(); }` — treats two very different situations identically:

1. **Legitimate no-op**: an in-memory/SystemDb factory environment where there genuinely is no Prisma client and no RLS is needed.
2. **Broken plumbing**: a real Postgres-backed environment where RLS *should* be active but the interceptor couldn't obtain a working Prisma client for some reason (exactly what happened in #644).

Case 2 silently degrades to "run the request with no RLS GUC set" — which, combined with Postgres's fail-closed RLS design, produces symptoms (empty reads, rejected writes) that are easy to misread as "this user genuinely has no data yet" rather than "the security plumbing is broken." That ambiguity is a large part of why #644 took so long to diagnose.

## Proposal

Distinguish the two cases explicitly. A reasonable approach: check `process.env.DATABASE_URL` (or an equivalent "we expect a real Postgres connection" signal) in addition to `!prisma`. If a real DB connection is expected but the resolved Prisma client is falsy, throw a clear `InternalServerErrorException` (or similar) naming the failure, rather than silently falling through to `next.handle()`. Reserve the silent no-op path for the in-memory/SystemDb case, where it's genuinely correct.

## Acceptance criteria

- [ ] The "DB expected but Prisma client unavailable" case throws a clear, loud error instead of silently proceeding
- [ ] The legitimate in-memory/SystemDb no-op path is unaffected
- [ ] New test covers both branches explicitly

Related: #644

PR #654, opened 2026-07-03
[fix] RlsInterceptor should fail loudly, not silently no-op, when RLS context can't be established

brownm09:
## Summary

`RlsInterceptor`'s guard treated two different situations identically:

1. **Legitimate no-op**: an in-memory/SystemDb factory environment where there's genuinely no Prisma client and no `DATABASE_URL` configured.
2. **Broken plumbing**: a real Postgres-backed environment where RLS *should* be active but the interceptor couldn't obtain a working Prisma client — exactly what happened in #644.

Case 2 silently degraded to "run the request with no RLS GUC set," which (combined with Postgres's fail-closed RLS design) produces symptoms — empty reads, rejected writes — that are easy to misread as "this user has no data yet" rather than "the security plumbing is broken." That ambiguity is a large part of why #644 took weeks to diagnose.

This PR distinguishes the two cases explicitly: checks `process.env.DATABASE_URL` (the same raw signal `repository-factory.module.ts` and `prisma.service.ts` already use to decide Prisma-backed vs. in-memory mode) in addition to the Prisma-client-null check. When a real DB connection is expected but the client is unavailable, it throws `InternalServerErrorException` naming the failure instead of silently falling through to `next.handle()`. The legitimate in-memory/SystemDb no-op path is unaffected — it's still gated purely on `DATABASE_URL` being unset.

A small private `isDatabaseUrlConfigured()` seam was extracted so the new branch is unit-testable without needing a real `DATABASE_URL` — `jest.env.setup.js` wraps `process.env` in a Proxy that discards writes to `DATABASE_URL` outside the Testcontainers sentinel (by design, to stop tests from accidentally hitting a real DB), so tests stub this method directly rather than fighting that guard.

## Acceptance criteria (from #649)

- [x] The "DB expected but Prisma client unavailable" case throws a clear, loud error instead of silently proceeding
- [x] The legitimate in-memory/SystemDb no-op path is unaffected
- [x] New test covers both branches explicitly

## Testing

```
npm test
```
Tests: 423 passed, 0 skipped, 0 failed (`apps/api`, includes 3 new tests in `rls.interceptor.spec.ts`) + core 280 passed + api-client 25 passed + types/mobile clean (no tests, pre-existing `--passWithNoTests` convention, untouched by this PR).

New coverage — `apps/api/src/adapters/prisma/rls.interceptor.spec.ts` (new file, no unit spec previously existed for this class):
- Throws `InternalServerErrorException` when `DATABASE_URL` is configured but `PrismaService` can't be resolved
- No-ops via `next.handle()` when `DATABASE_URL` is unset and `PrismaService` can't be resolved (in-memory/SystemDb mode)
- Non-HTTP context short-circuits before the throw is ever reached, regardless of `DATABASE_URL`/Prisma state (RLS is HTTP-only today, #511)

Confirmed the throw/no-op assertions fail pre-fix: the pre-fix `intercept()` has no `isDatabaseUrlConfigured` method at all, so `jest.spyOn` on it throws immediately, and the pre-fix guard never throws `InternalServerErrorException` under any input.

The existing `rls.db.e2e.spec.ts` "RLS request wiring" suites (real Testcontainers Postgres, real `AppModule` boot) construct `RlsInterceptor` with a resolvable `PrismaService` in every case, so `prisma` is always truthy there — this change's new branch is never entered by those tests, and they remain unaffected (still passing).

### Pre-existing failure encountered (unrelated, filed separately)

`npm test -w @lifting-logbook/web` fails 18/24 suites locally on this machine with `Cannot find module 'next/server'` etc. Root-caused to `apps/web/node_modules/{next,react-dom,@testing-library/react,@clerk/nextjs}` existing but containing zero files — a local install corruption, confirmed **not** caused by this branch (nothing here touches `apps/web`), **not** present in CI (green on `main` at the exact commit this branch was cut from), and **not** a lockfile-drift issue (`npm install --package-lock-only` + diff is clean). Filed as [#653](https://github.com/brownm09/lifting-logbook/issues/653); not fixed here since a full `node_modules` reset is out of scope for this change and disruptive to run unprompted on the shared canonical checkout.

## Housekeeping (bundled, encountered while working this issue)

Separate commit: bumped the `gh project item-list --limit` in `CLAUDE.md` from 300 to 500 — the project has grown to 327 items, so the old limit was silently truncating queries (encountered while moving #649 to In Progress; the item wasn't found until the limit was raised).

Both non-blocking findings from the [posted review](https://github.com/brownm09/lifting-logbook/pull/654#issuecomment-4878359320) were fixed in this PR, commit [`f1b287f`](https://github.com/brownm09/lifting-logbook/commit/f1b287f):

- **[security]** Client-facing exception message named `DATABASE_URL` and was returned verbatim to API clients — fixed: split into a generic client message plus a detailed `cause` Error (full diagnostic still reaches server-side logs/traces via existing Pino/OTel auto-instrumentation).
- **[documentation]** `/health`/`/readyz` are also covered by the new fail-loud path (intentional — a broken-RLS deployment should fail its readiness probe) but this wasn't stated anywhere — fixed: added a sentence to the class docstring.

New test (`rls.interceptor.spec.ts`) added in the same commit locks in the message/cause split.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -67,7 +67,7 @@ Architecture follows hexagonal / Ports & Adapters. `packages/core` has zero infr
 
 All new issues must be added to the **Lifting Logbook** project and assigned an epic before work begins.
 
-**Important:** All `gh project item-list` queries in this file use `--limit 300` to avoid truncation. The project has 267+ items (default limit is 30). If the project grows beyond 300 items, increase this limit accordingly.
+**Important:** All `gh project item-list` queries in this file use `--limit 500` to avoid truncation. The project has 327+ items (default limit is 30). If the project grows beyond 500 items, increase this limit accordingly.
 
 **Project IDs (needed for CLI commands):**
 - Project number: `2`, owner: `brownm09`
@@ -161,7 +161,7 @@ If `--format json` is not supported by the installed `gh` version, fall back to 
 3. Move the issue to **In Progress** on the project board:
    ```bash
    TMPFILE="C:/Users/brown/.claude/scratch/tmp_item_<N>.json"
-   gh project item-list 2 --owner brownm09 --limit 300 --format json > "$TMPFILE"
+   gh project item-list 2 --owner brownm09 --limit 500 --format json > "$TMPFILE"
    ITEM_ID=$(node -e "
      const d=JSON.parse(require('fs').readFileSync('$TMPFILE','utf8'));
      const item=d.items.find(i=>i.content&&i.content.number===<N>);
@@ -180,7 +180,7 @@ If `--format json` is not supported by the installed `gh` version, fall back to 
 9. Move the issue to **Done** on the project board:
    ```bash
    TMPFILE="C:/Users/brown/.claude/scratch/tmp_item_<N>.json"
-   gh project item-list 2 --owner brownm09 --limit 300 --format json > "$TMPFILE"
+   gh project item-list 2 --owner brownm09 --limit 500 --format json > "$TMPFILE"
    ITEM_ID=$(node -e "
      const d=JSON.parse(require('fs').readFileSync('$TMPFILE','utf8'));
      const item=d.items.find(i=>i.content&&i.content.number===<N>);
--- a/apps/api/src/adapters/prisma/rls.interceptor.spec.ts
+++ b/apps/api/src/adapters/prisma/rls.interceptor.spec.ts
@@ -1,0 +1,86 @@
+import { CallHandler, ExecutionContext, InternalServerErrorException } from '@nestjs/common';
+import { ModuleRef, Reflector } from '@nestjs/core';
+import { ClsService } from 'nestjs-cls';
+import { lastValueFrom, of } from 'rxjs';
+import { RlsInterceptor } from './rls.interceptor';
+import { PrismaService } from './prisma.service';
+
+function makeCls(): ClsService {
+  const store = new Map<string, unknown>();
+  return {
+    get: jest.fn((key: string) => store.get(key)),
+    set: jest.fn((key: string, value: unknown) => store.set(key, value)),
+  } as unknown as ClsService;
+}
+
+function makeReflector(): Reflector {
+  return { getAllAndOverride: jest.fn(() => undefined) } as unknown as Reflector;
+}
+
+function makeModuleRef(resolved: PrismaService | null): ModuleRef {
+  return { get: jest.fn(() => resolved) } as unknown as ModuleRef;
+}
+
+// A real reflection target is required by Reflector.getAllAndOverride, but none of the tests below
+// reach that call — the interceptor resolves (throw or no-op) before request/handler data is read.
+const dummyHandler = function handler() {};
+class DummyController {}
+
+function makeContext(type: 'http' | 'rpc'): ExecutionContext {
+  return {
+    getType: () => type,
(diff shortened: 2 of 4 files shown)


brownm09:
## Review: [fix] RlsInterceptor should fail loudly, not silently no-op, when RLS context can't be established

### Summary
Well-scoped, well-tested fix (135 lines across 4 files) that does exactly what the linked issue asks: distinguishes the legitimate in-memory no-op from the #644 broken-plumbing case and fails loudly on the latter. Test coverage directly targets both new branches plus the non-HTTP short-circuit, and the ADR-010 update keeps the existing incident narrative coherent. Two non-blocking findings — one about the client-facing exception message, one about an undocumented (but likely desirable) blast-radius consequence for `/health`/`/readyz`. No blocking findings. Diff size is appropriately scoped; no split needed.

Doc-reconciliation (2b): no "Documentation Maintenance" table in this repo's `CLAUDE.md` — not applicable. Doc-coverage (2c): only `docs/README.md` exists among ancestor directories; its ADR-010 index entry is a coarse status line ("Accepted · RLS implemented") unaffected by this follow-up — no gap. Test-coverage gate (2d): new behavior has direct test coverage — satisfied. Test-integrity gate (2e): no skip markers, deletions, threshold/bypass changes, or input-skew; PR body has a well-formed `Tests: N passed...` summary — clean.

---

### Non-Blocking Findings

**[security]** `apps/api/src/adapters/prisma/rls.interceptor.ts:84-87`
The thrown message ("...even though DATABASE_URL is configured...") is returned verbatim to API clients — `main.ts`'s `useGlobalFilters` only registers three domain-scoped filters (`DomainNotFoundFilter`, `DomainConflictFilter`, `PrismaTransactionTimeoutFilter`); none catch generic `InternalServerErrorException`, so it falls through to Nest's default filter, which serializes `.message` into the JSON response body. This reveals the `DATABASE_URL` env var name and confirms Prisma/Postgres usage to any external caller who triggers this state — not a credential leak, and consistent with how every other unhandled exception in this app already behaves (nothing here makes it worse), but still avoidable.
**Suggestion:** Consider a generic client-facing message ("Internal error establishing database context") and rely on the existing OTel span + Pino request logging (already correlated via `trace_id`/`span_id` per this repo's Observability conventions) to carry the detailed diagnostic server-side only.

**[documentation]** `apps/api/src/adapters/prisma/rls.interceptor.ts:31-40`
Because the new throw fires whenever `DATABASE_URL` is set and `prisma` is unresolvable — regardless of whether the request is authenticated — `/health` and `/readyz` (which today no-op via the pre-existing `!userId` check, unaffected by auth state) would *also* start returning 500 during a recurrence of the #644 DI-ordering failure. That's very likely the right outcome (a broken-DI deployment should fail its readiness probe and stop receiving traffic, rather than reporting healthy while every real endpoint silently drops RLS) but it's a real blast-radius change that isn't mentioned anywhere — not in the docstring, the PR body, or the ADR-010 update.
**Suggestion:** Add one sentence to the class docstring (or the ADR-010 follow-up paragraph) noting that `/health`/`/readyz` are intentionally included in this fail-loud path, so a future on-call engineer diagnosing failing health checks during a DI incident recognizes it as expected/desirable rather than a new bug.

---

### Findings Disposition

Per ADR-028, each finding below must be fixed in this PR or filed as a tracked issue and linked before merge.

---

PR #658, opened 2026-07-03
[fix] Default DB E2E suites to the restricted lifting_app role

### LINKED ISSUE #646: [test] Default DB E2E suites to the restricted lifting_app role, not the Postgres superuser
## Problem

Every DB E2E suite except `rls.db.e2e.spec.ts` connects via the Testcontainers bootstrap superuser (`LIFTING_TC_DATABASE_URL`, exposed by `jest.global-setup.js`). Postgres superusers structurally bypass Row-Level Security regardless of policy correctness — so any RLS enforcement bug is invisible to the entire suite by construction, no matter how much coverage exists otherwise.

This is exactly how #644 (RlsInterceptor silently never setting the RLS GUC) shipped and stayed live for 3+ weeks: every test that touched a userId-scoped table ran as superuser and got correct-looking results regardless of whether RLS was actually being enforced.

## Proposal

Flip the default: the DB E2E harness should provision and connect as `lifting_app` (the restricted runtime role) by default, with the superuser connection available as an explicit, named opt-in for tests that genuinely need to bypass RLS (e.g. seeding/cleanup helpers, which can keep a dedicated owner client alongside the default app-role one — same pattern `rls.db.e2e.spec.ts` already uses internally).

## Scope

- `apps/api/jest.global-setup.js` — after running migrations, set the `lifting_app` role's password and expose its connection string as the primary `LIFTING_TC_DATABASE_URL` (or a new default-preferred variable); keep the owner URL available under a distinct name for suites that need it.
- Audit existing DB E2E suites (`programs.db.e2e.spec.ts`, `custom-lift.e2e.spec.ts`, etc.) — most should continue working unchanged against the restricted role, since application code already scopes every query by `userId`. Any suite that breaks under RLS enforcement is itself informative (either a genuine gap or a legitimate reason to opt into the superuser connection for that suite).
- Document the opt-out pattern clearly so a future contributor doesn't reach for the superuser connection out of convenience.

## Acceptance criteria

- [ ] Default DB E2E connection is the restricted `lifting_app` role
- [ ] Existing suites pass unchanged, or have a documented, explicit reason for opting into the superuser connection
- [ ] `docs/testing/e2e-coverage.md` documents the default and the opt-out

Related: #644

brownm09:
## Summary

Every DB E2E suite except `rls.db.e2e.spec.ts` connected to Postgres as the bootstrap **superuser**, which structurally bypasses Row-Level Security regardless of policy correctness. That's exactly how issue #644 (`RlsInterceptor` silently never setting the `app.current_user_id` GUC) shipped and stayed live for 3+ weeks — every test touching a userId-scoped table ran as superuser and got correct-looking results regardless of whether RLS was actually enforced.

This flips the default: `apps/api/jest.global-setup.js` now provisions the restricted `lifting_app` role's login password once and exposes **two** connection strings instead of one:
- `LIFTING_TC_DATABASE_URL` — the restricted `lifting_app` role. **New default.**
- `LIFTING_TC_OWNER_DATABASE_URL` — the superuser/owner connection. **Explicit, named opt-in** for suites that genuinely need to bypass RLS.

## Changes

- **`apps/api/jest.global-setup.js`** — provisions `lifting_app`'s password (in both the CI-passthrough and local-Testcontainers branches) and exposes both sentinels. Added `buildRoleProvisioningFailedMessage()` matching the existing error-message style; a provisioning failure in the local path stops the container and clears its handle before throwing, mirroring the existing migrate-failure branch.
- **`apps/api/jest.env.setup.js`** — simplified the `DATABASE_URL` Proxy carve-out to an exact-match check against the two sentinels (deleted the old URL-parsing/host-pathname/username fallback, no longer needed).
- **`apps/api/src/adapters/prisma/rls.db.e2e.spec.ts`** — consumes both sentinels directly instead of deriving/provisioning the app-role URL itself; removed the now-dead `APP_ROLE`/`APP_ROLE_PASSWORD`/`appRoleUrl()`.
- **`apps/api/src/programs/programs.db.e2e.spec.ts`** — split into an explicit owner connection (seeding/cleanup/DB-assertions across many synthetic users — a deliberate, commented opt-in) and the app's ambient `DATABASE_URL` (now the restricted role by default), so every HTTP-driven assertion in this suite genuinely exercises real RLS enforcement end to end.
- **`docs/testing/e2e-coverage.md`** — documents the new default, the opt-out, and a rule of thumb for choosing between them; added the previously-missing `rls.db.e2e.spec.ts` row to the Test Files table.

Note: the issue body names `custom-lift.e2e.spec.ts` as an example DB suite to audit, but it (and `import.e2e.spec.ts`) are in-memory-only and untouched by this change — only `rls.db.e2e.spec.ts` and `programs.db.e2e.spec.ts` are real Postgres-backed DB E2E suites today.

## Audit result

Ran both DB E2E suites against real Postgres under the new restricted-role default: **72/72 pass unchanged**. This confirms the existing controller/factory RLS wiring (`PrismaService.clientForRequest()`, `PrismaRepositoryFactory.client()`, and the three controllers that build repositories outside the factory) was already correct — no suite needed an undocumented superuser opt-in beyond the two already-commented, deliberate cases (seeding/cleanup in both suites, and `rls.db.e2e.spec.ts`'s own policy-enforcement assertions, which must connect as `lifting_app` to prove the policies actually constrain a non-superuser caller).

## Testing

- `node apps/api/scripts/check-globalsetup-failure-paths.js` — 17/17 passed (none of the 3 mocked failure scenarios reach the new provisioning code path, since they all fail earlier).
- `npm run test:db -w @lifting-logbook/api` (Docker) — 2 suites, 72/72 passed.
- `npm test` (full repo) — 12/12 turbo tasks successful; API 420/420 passed, web 152/152 passed.

Tests: 644 passed, 0 skipped, 0 failed (~1m combined).

No suppressions added. `describe.skip` matches in the pre-PR grep are the pre-existing `describeOrSkip` conditional-skip helper in both spec files — I tightened its condition from a single-sentinel check to requiring both sentinels (defensive; global-setup always sets both together in every non-skip path, so this cannot cause a spurious skip in any real invocation), not a new skip marker.

## Out of scope (filed separately)

While validating this change I found `docker-compose.test.yml` — referenced by both `jest.global-setup.js`'s Docker-unavailable error message and `docs/testing/e2e-coverage.md` — doesn't exist anywhere in the repo (only an unrelated `docker-compose.yml` with a different port/credentials/db-name does). This predates #646 and isn't caused by it; flagged as a follow-up rather than folded into this diff.

Per the [`/review` comment](https://github.com/brownm09/lifting-logbook/pull/658#issuecomment-4878499359), both non-blocking findings were fixed in-PR (commit `8f97922`):

- **[reliability]** `provisionAppRolePassword()` had no timeout on its `ALTER ROLE` call — fixed in `8f97922` (added `ROLE_PROVISIONING_TIMEOUT_MS` via `Promise.race`, matching the existing `MIGRATION_TIMEOUT_MS` pattern).
- **[maintainability]** `programs.db.e2e.spec.ts` kept the name `TC_DATABASE_URL` while `rls.db.e2e.spec.ts` (same PR, same purpose) used `APP_ROLE_URL` — fixed in `8f97922` (renamed to match).
- **[style]** Quote-style inconsistency in `buildRoleProvisioningFailedMessage`'s headline — fixed in `8f97922`.

Re-verified after both fixes: failure-path script 17/17, DB E2E audit 72/72, full suite 12/12 turbo tasks green (same 644 passed / 0 skipped / 0 failed — the fixes changed no test count).

Closes #646

### DIFF
--- a/apps/api/jest.env.setup.js
+++ b/apps/api/jest.env.setup.js
@@ -46,38 +46,20 @@ process.env.OTEL_SDK_AUTOSTART = 'false';
 const originalEnv = process.env;
 process.env = new Proxy(originalEnv, {
   set(target, key, value) {
-    // Allow the DB E2E spec to restore DATABASE_URL from the Testcontainers
-    // sentinel set by jest.global-setup.js, or from its lifting_app-role variant
-    // (see below) — these are the only legitimate write paths for DATABASE_URL
-    // during a test run.
-    if (String(key) === 'DATABASE_URL' && value) {
-      const sentinel = target.LIFTING_TC_DATABASE_URL;
-      if (value === sentinel) {
-        target[key] = value;
-        return true;
-      }
-      // Also allow the lifting_app-role variant of the sentinel (same host/port/database,
-      // different credentials) — needed by full-app-boot RLS tests that must connect as the
-      // restricted runtime role rather than the Testcontainers owner, so PrismaService's
-      // env("DATABASE_URL") read resolves to a real, RLS-enforcing connection. Scoped to the
-      // sentinel's own host/pathname so this cannot be used to smuggle an arbitrary DB URL in.
-      // See rls.db.e2e.spec.ts's "RLS request wiring (interceptor + factory, full app)" block.
-      if (sentinel) {
-        try {
-          const candidate = new URL(value);
-          const base = new URL(sentinel);
-          if (
-            candidate.host === base.host &&
-            candidate.pathname === base.pathname &&
-            candidate.username === 'lifting_app'
-          ) {
-            target[key] = value;
-            return true;
-          }
-        } catch {
-          // Not a valid URL — fall through and discard below.
-        }
-      }
+    // Allow the DB E2E spec to restore DATABASE_URL from either sentinel
+    // jest.global-setup.js sets (issue #646): LIFTING_TC_DATABASE_URL (the
+    // restricted lifting_app role — the default) or LIFTING_TC_OWNER_DATABASE_URL
+    // (the superuser/owner opt-in). Both are exact, fixed strings by the time any
+    // worker's setupFiles script runs — global-setup sets them once in the parent
+    // process before workers fork — so an exact match against either is sufficient;
+    // no derived third URL is a legitimate write.
+    if (
+      String(key) === 'DATABASE_URL' &&
+      value &&
+      (value === target.LIFTING_TC_DATABASE_URL || value === target.LIFTING_TC_OWNER_DATABASE_URL)
+    ) {
+      target[key] = value;
+      return true;
     }
     if (BLOCK.includes(String(key))) {
       return true; // discard — blocked keys are frozen at ''
--- a/apps/api/jest.global-setup.js
+++ b/apps/api/jest.global-setup.js
@@ -17,11 +17,30 @@
 //      `prisma migrate deploy` failures throw buildMigrationFailedMessage
 //      (schema drift, SQL error, timeout — Docker is fine).
 //
-// In paths 1 and 3 the spec reads LIFTING_TC_DATABASE_URL (not DATABASE_URL)
-// because jest.env.setup.js force-blanks DATABASE_URL in every worker to keep
-// the in-memory e2e suite wiring InMemoryRepositoryFactory. The DB E2E spec
-// restores DATABASE_URL from the sentinel inside its own beforeAll;
-// jest.env.setup.js allows that one specific write through its Proxy.
+// In paths 1 and 3, once the database is confirmed migrated (so the `lifting_app`
+// role from the enable_rls migration exists), this file also provisions that
+// role's login password and exposes TWO connection strings (issue #646):
+//   - LIFTING_TC_DATABASE_URL: the restricted `lifting_app` role. This is the
+//     DEFAULT every DB E2E spec should use — it's the same non-superuser role
+//     the production app runs as, so Row-Level Security is actually enforced.
+//     Postgres superusers bypass RLS entirely regardless of policy correctness,
+//     which is exactly how issue #644 (RlsInterceptor silently never setting the
+//     RLS GUC) went undetected for 3+ weeks — every DB E2E suite ran as superuser.
+//   - LIFTING_TC_OWNER_DATABASE_URL: the bootstrap superuser/owner connection,
+//     kept available as an explicit, named opt-in for suites that genuinely need
+//     to bypass RLS (seeding/cleanup across many synthetic users, or asserting on
+//     RLS metadata itself). Only read this sentinel with a comment explaining why
+//     the suite needs to bypass RLS.
+// buildRoleProvisioningFailedMessage covers failures in that provisioning step —
+// by that point Docker and migrations are both known-good, so the failure is
+// specific to the ALTER ROLE statement itself.
+//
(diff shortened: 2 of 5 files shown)


brownm09:
## Review: [fix] Default DB E2E suites to the restricted lifting_app role

### Summary

Ready to merge with minor changes. This is a well-scoped, well-documented security-hardening fix to test infrastructure, empirically validated by running the actual DB E2E suite against real Postgres before and after a mid-review rebase (72/72 passing both times, no regressions). No blocking findings. Two non-blocking findings worth a look: a naming inconsistency between the two spec files this PR touches, and a missing timeout on the new role-provisioning DB call. Diff size (293 lines) is appropriately scoped for a single-pass review.

No documentation-reconciliation or README-coverage gaps found (no `Documentation Maintenance` table in this repo's `CLAUDE.md`; the one relevant doc, `docs/testing/e2e-coverage.md`, is already updated in-PR). No new testable *application* behavior — this PR changes test infrastructure itself, and the PR body's empirical audit (72/72 passing under the new default) is the appropriate validation for that class of change. Test-integrity gate is clean: no new skip markers, no deleted tests, no lowered thresholds, and the PR body already explains the one `describeOrSkip` condition change (tightened from a single-sentinel check to requiring both sentinels — provably a no-op in every real invocation since global-setup always sets both together).

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/programs.db.e2e.spec.ts:80`
This file keeps the local variable name `TC_DATABASE_URL` for `process.env.LIFTING_TC_DATABASE_URL`, while `apps/api/src/adapters/prisma/rls.db.e2e.spec.ts` — modified in this same PR for the identical purpose — renames its equivalent local to `APP_ROLE_URL`. A future reader who opens both files (which now share the same two-sentinel contract this PR introduces) sees two different names for the same environment variable, and the old name no longer hints at *which* role it resolves to now that the variable's meaning has flipped.
**Suggestion:** Rename `TC_DATABASE_URL` to `APP_ROLE_URL` (or similar) in `programs.db.e2e.spec.ts` to match `rls.db.e2e.spec.ts`, or vice versa — pick one convention across both files touched by this PR.

**[reliability]** `apps/api/jest.global-setup.js:148-158`
`provisionAppRolePassword()`'s `$executeRawUnsafe` call has no explicit timeout, unlike the `prisma migrate deploy` step a few lines above it, which is bounded by `MIGRATION_TIMEOUT_MS`. If the database becomes unreachable in the narrow window between a successful migration and this new step, `globalSetup` will hang on the `ALTER ROLE` call rather than failing fast with the new `buildRoleProvisioningFailedMessage`.
**Suggestion:** Bound the call (e.g., `Promise.race` against a short timer, or a `connect_timeout` query-string param on `ownerUrl`) so a connectivity loss in that window surfaces the actionable message instead of hanging until Jest's or CI's own outer timeout eventually kills it.

---

### Style Notes

`buildRoleProvisioningFailedMessage`'s headline (`apps/api/jest.global-setup.js:121`) switches to double quotes to avoid escaping an apostrophe, while its two sibling builders (`buildDockerUnavailableMessage`, `buildMigrationFailedMessage`) use single quotes throughout their arrays. Minor and inconsequential — worth a quick pass for visual consistency only if you're already touching this file again.

---

### Findings Disposition

Per ADR-028, each finding above must be fixed in this PR or filed as a tracked issue and linked before merge.

---

PR #668, opened 2026-07-03
[test] Add staging onboarding write-path test with self-cleaning DELETE endpoint

### LINKED ISSUE #1: [chore] Initialize Turborepo monorepo with workspace structure
## Context
See [ADR-001](docs/adr/ADR-001-monorepo-structure.md) — Monorepo Structure with Turborepo.

## Acceptance Criteria
- [ ] Root `package.json` defines workspaces: `packages/*`, `apps/*`
- [ ] `turbo` installed as a root dev dependency
- [ ] `turbo.json` created with placeholder pipeline tasks (`build`, `test`, `lint`, `dev`)
- [ ] Directory structure matches ADR-001:
  ```
  packages/core/
  packages/types/
  apps/api/
  apps/api-legacy/
  apps/web/
  apps/mobile/
  infra/kubernetes/
  infra/cloud-run/
  infra/terraform/
  docs/adr/
  docs/README.md
  ```
- [ ] Root `README.md` links to `docs/README.md`

## Notes
Use `npx create-turbo` as a starting point. Prefer npm workspaces for compatibility.

### LINKED ISSUE #5: [chore] Scaffold packages/core — migrate GAS core logic
## Context
See [ADR-002](docs/adr/ADR-002-ports-and-adapters.md). The existing `src/core/` from the GAS repo has no GAS dependencies and migrates directly. This is the highest-value asset being carried forward.

## Acceptance Criteria
- [ ] `packages/core/package.json`: `name: "@logbook/core"`, `main: "dist/index.js"`, `types: "dist/index.d.ts"`
- [ ] `packages/core/tsconfig.json` extends root `tsconfig.base.json`
- [ ] All files from `src/core/` copied to `packages/core/src/`
- [ ] All tests from `tests/core/` copied to `packages/core/tests/` and passing
- [ ] Test fixtures from `tests/fixtures/` copied with paths updated
- [ ] `packages/core/src/index.ts` exports all public services, models, and utilities
- [ ] `turbo run test --filter=@logbook/core` passes

## Notes
Do not change any logic during migration — lift-and-shift only. Open a separate issue for any refactoring discovered during the move.

### LINKED ISSUE #647: [test] Add a staging smoke test that actually writes data (completes onboarding end-to-end)
## Problem

The automated staging gate (`.github/workflows/deploy.yml`'s Smoke Test job) and `apps/web/e2e/staging.spec.ts` are both read-only: `/livez`, `/sign-in`, page-load/redirect checks, and an API auth check. None of them create, update, or persist any real data.

#644 (RlsInterceptor silently never setting the RLS GUC) was live in staging identically to production — staging's `DATABASE_URL` is provisioned via the same unconditional `google_sql_user "app_rls"` Terraform resource — but nothing in the staging pipeline could have surfaced it, because the bug only manifests on a write (or a read of genuinely-existing data), and nothing in staging's automated checks does either.

## Proposal

Add one Playwright test to `apps/web/e2e/staging.spec.ts` that walks a disposable test user through onboarding end-to-end against staging (sign in → choose program → confirm maxes → "Start My Program") and asserts the resulting cycle dashboard renders. This closes the exact gap: any future regression that breaks first-time data creation gets caught before promotion to production, not after.

## Scope

- Needs a disposable-test-user strategy against the real staging DB (a dedicated Clerk test account, or a teardown step that deletes the created cycle/training-max rows afterward so repeated staging deploys don't accumulate garbage data).
- Should run in the existing `Deploy (staging)` → `Smoke Test (staging)` → `Deploy (production)` pipeline position, ahead of the production approval gate.

## Acceptance criteria

- [ ] A staging Playwright test creates a real cycle for a test user and verifies it renders
- [ ] Test user's created data is cleaned up (or reused idempotently) so staging doesn't accumulate cruft
- [ ] Documented in `docs/deploy.md` alongside the existing staging smoke-test description

Related: #644

brownm09:
## Summary

Closes staging's read-only test gap that let #644 (`RlsInterceptor` silently disabled) ship
undetected — every automated staging check (`deploy.yml` smoke test + the prior
`apps/web/e2e/staging.spec.ts` suite) was read-only, so a write-path bug was live in staging
identically to production but nothing could surface it.

Adds one Playwright test that walks the staging test user through onboarding end-to-end
(sign in → choose program → confirm maxes → "Start My Program") and asserts the resulting
cycle dashboard renders.

## Why this needed a new endpoint, not just a test

Onboarding is structurally a one-time action: `switchProgram` checks for an existing cycle
first and, if one exists, silently returns it instead of erroring — so re-running the test
against the same account would, after the first pass, quietly stop exercising the create-cycle
write path it exists to protect (no error, no visible signal). No delete/reset capability
existed anywhere in the API to make the test repeatable. Discussed the tradeoffs directly with
the repo owner: build a real `DELETE /programs/:program/cycles/current` endpoint (RLS + auth
already scope it to the caller's own data — it's also a legitimate future "restart my program"
capability, not a test-only backdoor) rather than a disposable-Clerk-user scheme that would
leave orphaned Postgres rows on every run.

The cleanup call is wired through the exact same `X-Clerk-Authorization` path (`lib/api.ts`)
that "Start My Program" itself already depends on — not a bespoke shortcut. `ADR-023`'s
"Alternatives Considered" already tried and rejected both more-obvious shortcuts (server-side
`getToken()` + JWT forwarding, client-side `getToken()` + direct fetch) for this exact staging
Clerk instance; routing around the established mechanism would just hide a problem that would
sink the onboarding flow itself. See the ADR-023 amendment in this PR for the full reasoning.

## Changes

**Backend** — `DELETE /programs/:program/cycles/current` (`cycle-dashboard.controller.ts` +
new `CycleGenerationService.deleteCurrentCycle`): sequential, independent deletes of
`LiftRecord`, `TrainingMaxHistory`, `TrainingMax`, and the `CycleScheduledWorkout`/
`CycleDashboard` rows for the caller's own program (no FK cascades exist between these tables,
so each needs an explicit delete; sequential rather than concurrent because they share one
request-scoped Prisma transaction connection — see Findings Disposition). New port methods on
`ICycleDashboardRepository`, `ILiftRecordRepository`, `ITrainingMaxRepository`,
`ITrainingMaxHistoryRepository`, implemented in both the Prisma and in-memory adapters.
`UserSettings.activeProgram` is deliberately left untouched.

**Frontend** — new `packages/api-client` endpoint (plus a new `ApiClientError` class carrying
the upstream HTTP status) + `apps/web/lib/api.ts` re-export, and a new test-support Next.js
route handler (`apps/web/app/api/test-support/reset-cycle`) that the Playwright test calls
before *and* after (via `test.beforeAll`/`afterAll`, which — unlike a `try/finally` in the test
body — is guaranteed to run even if the test times out mid-assertion). The route 404s in
production (keyed off `CLERK_PUBLISHABLE_KEY`'s `pk_live_` prefix) and relays real upstream
failure statuses instead of flattening everything to 500.

**Tests** — unit tests for the new service method (deterministic delete order,
404-on-nothing-to-delete, cycleNum scoping), a controller test, a new in-memory HTTP E2E
lifecycle test (`programs.e2e.spec.ts`), and a real-Postgres RLS test (`rls.db.e2e.spec.ts`)
seeding and verifying all five affected tables, proving the delete only ever removes the
requesting user's rows under the restricted `lifting_app` role — directly closing the loop on
the RLS-scoping risk class that motivated #644/#647 in the first place.

**Docs** — `apps/web/e2e/README.md` (new test-coverage row + test-account note),
`docs/deploy.md` (one-line pointer near the Clerk staging setup step, satisfying the issue's
literal acceptance criterion even though the Playwright suite's own detail lives in the e2e
README), and an amendment to `ADR-023`'s "Test scope" consequence recording this one
deliberate, scoped exception to "staging tests are read-only."

## Acceptance criteria (from #647)

- [x] A staging Playwright test creates a real cycle for a test user and verifies it renders
- [x] Test user's created data is cleaned up so staging doesn't accumulate cruft (before *and*
      after cleanup, both idempotent — a 404 from "nothing to delete" is treated as success,
      and both cleanup calls now assert success rather than failing silently)
- [x] Documented in `docs/deploy.md` alongside the existing staging smoke-test description (plus
      the full detail in `apps/web/e2e/README.md`, the file that actually documents the suite)

## Testing

Ran `npm test` from the repo root (full suite, all workspaces, including the Testcontainers-backed
DB E2E suites — Docker Desktop was running), after rebasing onto `origin/main` to pick up #667
(a critical fix — see Findings Disposition).

**Tests: 893 passed, 0 skipped, 0 failed** — core (280/37 suites), api-client (27/1 suite,
+2 from the rebased-in #667 test), web (152/25 suites), api (434/35 suites, including the RLS
DB E2E test). `npm run build` also passed cleanly across all 6 buildable packages (this PR
touches `packages/api-client`, consumed by `apps/web`).

The one thing that can't be verified pre-merge: the new test itself only runs for real against
live staging, via `.github/workflows/staging.yml`'s `staging-integration-tests` job on this PR.
No workflow changes were needed — that job already runs every file matched by
`playwright.config.staging.ts`'s `testMatch: ['staging.spec.ts']` with all the env vars/secrets
the new test needs.

Pre-PR policy checks (suppression grep, test-integrity grep, deleted-test-file check, coverage-
threshold check, manual `!`-scan of the diff) all came back clean — no findings to justify —
both before and after the rebase.

## Risk-dimension notes

- **Security** — the new DELETE endpoint is authenticated (standard `@CurrentUser()` guard),
  RLS-scoped to the caller's own rows (the RLS DB E2E test proves a second user's rows in all
  five tables survive), and environment-gated so it's inert in production.
- **Resilience** — the delete is a no-op-safe 404 (not a crash) when there's nothing to delete;
  the Playwright cleanup asserts success rather than silently tolerating a failed reset.
- **Data integrity** — no FK cascades exist between the affected tables, so each is deleted
  explicitly rather than relying on cascade behavior that doesn't exist; the five deletes run
  sequentially (not `Promise.all`) since they share one request-scoped transaction connection.
- **Observability** — no new logging conventions introduced; the new route handler logs failures
  via `console.error` consistent with other route handlers in this app (e.g. `/api/health`), and
  now relays the real upstream HTTP status instead of collapsing every failure to 500.

`/review` ran with two parallel subagents (correctness/security + reliability/maintainability),
posted as a [PR comment](https://github.com/brownm09/lifting-logbook/pull/668#issuecomment-4879226390).
4 blocking + 5 non-blocking findings, all fixed in this PR (commit `a3292ce`):

| # | Finding | Category | Disposition |
|---|---|---|---|
| 1 | `Promise.all` over 4 writes doesn't parallelize inside one Prisma request transaction — could throw against real Postgres while passing against in-memory adapters (independently flagged by both review subagents) | correctness | **Fixed** — sequential `await`s |
| 2 | Playwright cleanup hooks never asserted the reset succeeded; route swallowed all failures to a generic 500 | correctness | **Fixed** — hooks assert `.ok()`; route relays real status via new `ApiClientError` |
| 3 | Reset-cycle route had no environment guard — a live destructive endpoint reachable in production | security | **Fixed** — 404s when `CLERK_PUBLISHABLE_KEY` is a live key |
| 4 | New RLS regression test only seeded/verified 1 of 5 affected tables, undermining its own stated purpose | correctness | **Fixed** — seeds and verifies all 5 tables for both users |
| 5 | `TrainingMax` deletion was read-then-delete-by-key, not matching its "all rows" doc comment | maintainability | **Fixed** — new `deleteAllTrainingMaxes` direct delete |
| 6 | Inconsistent naming across the 5 delete methods (`deleteAll*` vs keyed vs unprefixed) | maintainability | **Fixed** — resolved by fix #5's new method, consistent `deleteAll<X>` vs `delete<X>(program, keys)` pattern |
| 7 | Test 6's placement depends on an undocumented `workers`/ordering invariant | reliability | **Fixed** — documented in `playwright.config.staging.ts` (verified the actual risk: `fullyParallel: false`, not `true` as initially flagged — corrected before documenting) |
| 8 | Program name `'rpt'` hardcoded independently in 3 places, could drift silently | maintainability | **Fixed** — dead default removed, route now 400s on a missing param |
| 9 | Ordering unit test implied more than `Promise.all` actually guaranteed | maintainability | **Fixed** — retargeted to assert the now-deterministic full sequential order (fix #1 made this possible) |

Additionally, rebased onto `origin/main` after the review pass (3 commits had landed, including
#667 — a fix for `switchProgram` sending an empty body with a JSON content-type, which Fastify
rejects in production). #667 is directly load-bearing here: its own commit message notes the
existing RLS regression tests call `switchProgram` via Fastify's `.inject()`, which never sets
`Content-Type` and so never exercised the real client request shape — exactly the gap this PR's
new end-to-end Playwright test closes. Without the rebase, the new write-path test would have
failed in staging CI against the same bug #667 just fixed.

Closes #647

### DIFF
--- a/apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts
+++ b/apps/api/src/adapters/in-memory/cycle-dashboard.adapter.ts
@@ -25,4 +25,8 @@ export class InMemoryCycleDashboardRepository
   async saveCycleDashboard(dashboard: CycleDashboard): Promise<void> {
     this.dashboards.set(dashboard.program, dashboard);
   }
+
+  async deleteCycleDashboard(program: string): Promise<void> {
+    this.dashboards.delete(program);
+  }
 }
--- a/apps/api/src/adapters/in-memory/lift-record.adapter.ts
+++ b/apps/api/src/adapters/in-memory/lift-record.adapter.ts
@@ -62,4 +62,8 @@ export class InMemoryLiftRecordRepository implements ILiftRecordRepository {
     this.store.set(program, after);
     return before.length - after.length;
   }
+
+  async deleteAllLiftRecords(program: string): Promise<void> {
+    this.store.delete(program);
+  }
 }
--- a/apps/api/src/adapters/in-memory/training-max-history.adapter.ts
+++ b/apps/api/src/adapters/in-memory/training-max-history.adapter.ts
@@ -42,4 +42,8 @@ export class InMemoryTrainingMaxHistoryRepository implements ITrainingMaxHistory
     this.entriesByProgram.set(program, next);
     return updated;
   }
+
+  async deleteAllHistory(program: string): Promise<void> {
+    this.entriesByProgram.delete(program);
+  }
 }
--- a/apps/api/src/adapters/in-memory/training-max.adapter.ts
+++ b/apps/api/src/adapters/in-memory/training-max.adapter.ts
@@ -55,4 +55,8 @@ export class InMemoryTrainingMaxRepository implements ITrainingMaxRepository {
     const existing = this.maxesByProgram.get(program) ?? [];
     this.maxesByProgram.set(program, existing.filter((m) => !liftSet.has(m.lift)));
   }
+
+  async deleteAllTrainingMaxes(program: string): Promise<void> {
+    this.maxesByProgram.delete(program);
+  }
 }
--- a/apps/api/src/adapters/prisma/cycle-dashboard.repository.ts
+++ b/apps/api/src/adapters/prisma/cycle-dashboard.repository.ts
@@ -48,4 +48,10 @@ export class PrismaCycleDashboardRepository implements ICycleDashboardRepository
       },
     });
   }
+
+  async deleteCycleDashboard(program: string): Promise<void> {
+    await this.prisma.cycleDashboard.deleteMany({
+      where: { userId: this.userId, program },
+    });
+  }
 }
--- a/apps/api/src/adapters/prisma/lift-record.repository.ts
+++ b/apps/api/src/adapters/prisma/lift-record.repository.ts
@@ -121,6 +121,12 @@ export class PrismaLiftRecordRepository implements ILiftRecordRepository {
     });
     return count;
   }
+
+  async deleteAllLiftRecords(program: string): Promise<void> {
+    await this.prisma.liftRecord.deleteMany({
+      where: { userId: this.userId, program },
+    });
+  }
 }
 
 // ID format: ${program}-${cycleNum}-${workoutNum}-${lift}-${setNum}
--- a/apps/api/src/adapters/prisma/rls.db.e2e.spec.ts
+++ b/apps/api/src/adapters/prisma/rls.db.e2e.spec.ts
@@ -441,4 +441,49 @@ describeOrSkip('RLS request wiring (interceptor + factory, full app boot)', () =
     await owner.cycleDashboard.deleteMany({ where: { userId } });
     await owner.userSettings.deleteMany({ where: { userId } });
   });
+
+  // Regression coverage for #647: the new DELETE cycles/current endpoint (added to
+  // support a self-cleaning staging Playwright test) deletes through the same
+  // factory/repos path as every other request. Proves the deleteMany calls are
+  // genuinely RLS-scoped under the restricted lifting_app role — not merely
+  // filtered by an application-level WHERE clause that happens to look correct —
+  // exactly the risk class that let #644 ship silently.
+  it('deletes a cycle scoped to the requesting user only, under real RLS (regression coverage for #647)', async () => {
+    const userId = `rls-e2e-fullapp-delete-${Date.now()}`;
+    const otherUserId = `rls-e2e-fullapp-delete-other-${Date.now()}`;
+
+    const seedDashboard = (uid: string) => ({
+      userId: uid,
+      program: 'leangains',
+      cycleUnit: 'week',
+      cycleNum: 1,
+      cycleDate: new Date(),
+      sheetName: `leangains_Cycle_1_${uid}`,
+      cycleStartWeekday: 'Friday',
+      programType: 'leangains',
+    });
+    await owner.cycleDashboard.create({ data: seedDashboard(userId) });
+    await owner.cycleDashboard.create({ data: seedDashboard(otherUserId) });
+
+    const delRes = await inject({
+      method: 'DELETE',
+      url: '/programs/leangains/cycles/current',
+      headers: { authorization: `Bearer ${userId}` },
+    });
+    expect(delRes.statusCode).toBe(204);
+
+    const mineGone = await owner.cycleDashboard.findFirst({
+      where: { userId, program: 'leangains' },
+    });
+    expect(mineGone).toBeNull();
+
+    // The other user's row must survive — proves the delete was RLS-scoped to the
+    // requesting user, not a bug that deleted every row matching the program.
+    const otherSurvives = await owner.cycleDashboard.findFirst({
+      where: { userId: otherUserId, program: 'leangains' },
+    });
+    expect(otherSurvives).not.toBeNull();
+
+    await owner.cycleDashboard.deleteMany({ where: { userId: { in: [userId, otherUserId] } } });
+  });
 });
--- a/apps/api/src/adapters/prisma/rls.db.e2e.spec.ts
+++ b/apps/api/src/adapters/prisma/rls.db.e2e.spec.ts
@@ -449,41 +449,98 @@ describeOrSkip('RLS request wiring (interceptor + factory, full app boot)', () =
   // filtered by an application-level WHERE clause that happens to look correct —
   // exactly the risk class that let #644 ship silently.
   it('deletes a cycle scoped to the requesting user only, under real RLS (regression coverage for #647)', async () => {
+    // deleteCurrentCycle issues five scoped deletes (CycleDashboard, LiftRecord,
+    // TrainingMax, TrainingMaxHistory, CycleScheduledWorkout). Seeding all five
+    // for both users — not just the dashboard — matters because #644 was a
+    // per-repository DI-wiring failure, not a missing-policy failure: a future
+    // regression where one specific repository fails to route through the
+    // request-scoped RLS transaction would not be caught if this test only ever
+    // exercised an empty table for that repository.
     const userId = `rls-e2e-fullapp-delete-${Date.now()}`;
     const otherUserId = `rls-e2e-fullapp-delete-other-${Date.now()}`;
+    const PROGRAM = 'leangains';
 
-    const seedDashboard = (uid: string) => ({
-      userId: uid,
-      program: 'leangains',
-      cycleUnit: 'week',
-      cycleNum: 1,
-      cycleDate: new Date(),
-      sheetName: `leangains_Cycle_1_${uid}`,
-      cycleStartWeekday: 'Friday',
-      programType: 'leangains',
-    });
-    await owner.cycleDashboard.create({ data: seedDashboard(userId) });
-    await owner.cycleDashboard.create({ data: seedDashboard(otherUserId) });
+    const seedAllTables = async (uid: string) => {
+      await owner.cycleDashboard.create({
+        data: {
+          userId: uid,
+          program: PROGRAM,
+          cycleUnit: 'week',
+          cycleNum: 1,
+          cycleDate: new Date(),
+          sheetName: `leangains_Cycle_1_${uid}`,
+          cycleStartWeekday: 'Friday',
+          programType: PROGRAM,
+        },
+      });
+      await owner.liftRecord.create({
+        data: {
+          userId: uid,
+          program: PROGRAM,
+          cycleNum: 1,
+          workoutNum: 1,
+          date: new Date(),
+          lift: 'Squat',
+          setNum: 1,
+          weight: 135,
+          reps: 5,
+        },
+      });
+      await owner.trainingMax.create({
+        data: { userId: uid, program: PROGRAM, lift: 'Squat', weight: 225, dateUpdated: new Date() },
+      });
+      await owner.trainingMaxHistory.create({
+        data: {
+          userId: uid,
+          program: PROGRAM,
+          lift: 'Squat',
+          weight: 225,
+          reps: 1,
+          date: new Date(),
+          isPR: false,
+          source: 'program',
+          goalMet: false,
+        },
+      });
+      await owner.cycleScheduledWorkout.create({
+        data: { userId: uid, program: PROGRAM, cycleNum: 1, workoutNum: 1, weekNum: 1, scheduledDate: new Date() },
+      });
+    };
+    await seedAllTables(userId);
+    await seedAllTables(otherUserId);
 
     const delRes = await inject({
       method: 'DELETE',
-      url: '/programs/leangains/cycles/current',
+      url: `/programs/${PROGRAM}/cycles/current`,
       headers: { authorization: `Bearer ${userId}` },
     });
     expect(delRes.statusCode).toBe(204);
 
-    const mineGone = await owner.cycleDashboard.findFirst({
-      where: { userId, program: 'leangains' },
-    });
-    expect(mineGone).toBeNull();
+    const findAllFor = (uid: string) =>
+      Promise.all([
+        owner.cycleDashboard.findFirst({ where: { userId: uid, program: PROGRAM } }),
+        owner.liftRecord.findFirst({ where: { userId: uid, program: PROGRAM } }),
+        owner.trainingMax.findFirst({ where: { userId: uid, program: PROGRAM } }),
+        owner.trainingMaxHistory.findFirst({ where: { userId: uid, program: PROGRAM } }),
+        owner.cycleScheduledWorkout.findFirst({ where: { userId: uid, program: PROGRAM } }),
+      ]);
 
-    // The other user's row must survive — proves the delete was RLS-scoped to the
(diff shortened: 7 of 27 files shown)


brownm09:
## Review: [test] Add staging onboarding write-path test with self-cleaning DELETE endpoint

### Summary

Needs revision before merge — not because the overall design is wrong (the DELETE-endpoint approach and the decision to route the cleanup call through the existing `X-Clerk-Authorization` mechanism are both sound), but because two of the blocking findings mean the PR's own stated purpose could quietly fail to hold: the `Promise.all` over four writes on a single Prisma request-transaction connection could throw at runtime against real Postgres despite every test passing (the in-memory adapters don't share that constraint), and the Playwright cleanup hooks never check whether the reset actually succeeded — so a failed cleanup doesn't fail the test, it just makes the test silently stop exercising the write path it exists to protect. Both are concrete, fixable in this PR, not architectural rework. The diff (463 lines) is appropriately scoped for what it's doing — no split needed.

---

### Blocking Findings

**[correctness]** `apps/api/src/programs/cycle-generation.service.ts:242-252`
`deleteCurrentCycle` issues four repository writes (`deleteAllLiftRecords`, `deleteAllHistory`, `deleteTrainingMaxes`, `saveScheduledWorkouts`) concurrently via `Promise.all`, then deletes the dashboard.
In production, `factory.forUser()` binds these repos to the single interactive-transaction Prisma client `RlsInterceptor` stashes in CLS — one dedicated connection that cannot run concurrent queries. Every other multi-write path in this codebase (`saveScheduledWorkouts`'s own internal delete+create, `startNewCycle`, `prisma-tx.util.ts`'s `runBatch`) serializes writes on this same client for exactly this reason; this is the only `Promise.all` in `apps/api/src/programs` parallelizing writes rather than reads. The in-memory adapters (plain `Map` operations) have no such constraint, so the new unit and in-memory E2E tests pass while the real Prisma path can throw a "queries cannot run concurrently" error — the new RLS DB E2E test doesn't catch this either, since it seeds only a `cycleDashboard` row, so the three racing `deleteMany`s all operate on empty tables. Separately, on an account with a large accumulated `LiftRecord`/`TrainingMaxHistory` history, the reads + four "concurrent" (actually serialized) writes + final delete all count against the RLS transaction's 15s default timeout — real risk on an old/heavily-used account, not just this test's disposable one.
**Fix:** Change the `Promise.all` to sequential `await`s (or route through `runBatch`), matching the established write-serialization pattern elsewhere in this file. Consider whether this endpoint needs a widened transaction timeout given it can touch unbounded historical rows (cf. the import path's `IMPORT_TX_TIMEOUT_MS`).

---

**[correctness]** `apps/web/e2e/staging.spec.ts:681-688` and `apps/web/app/api/test-support/reset-cycle/route.ts:23-29`
The `beforeAll`/`afterAll` hooks call `await request.delete(RESET_URL)` but never assert the response status; the route's try/catch collapses *every* failure — including a 401 (expired Clerk token) or 403 (Cloud Run IAM) — into an opaque 500.
This is two ends of the same defect. If the pre-test cleanup silently fails (auth hiccup, transient 500), a stale cycle from a prior run survives. The test then drives onboarding through to "Start My Program," which calls `switchProgram` — and per `switch-program.controller.ts:55-65`, when a cycle already exists it silently reuses it and returns 200 with the existing `cycleNum`, no error. The test only asserts it lands on `/cycle/1` with a "cycle" heading visible, which is equally true whether a new cycle was genuinely created or an old one was just reused. So a failed cleanup doesn't turn the test red — it turns the test into exactly the kind of read-only check this PR exists to move past, with no signal that anything degraded. The route's error-collapsing compounds this: even someone actively debugging a red staging run can't tell "auth is misconfigured" from "the delete itself failed" from the 500 alone.
**Fix:** Assert `expect((await request.delete(RESET_URL)).ok()).toBeTruthy()` in both `beforeAll` and `afterAll` so a cleanup failure fails the run loudly. In the route handler, preserve the upstream status (relay 401/403 rather than flattening to 500) so a CI failure is diagnosable without re-deriving it from source.

---

**[security]** `apps/web/app/api/test-support/reset-cycle/route.ts:1-30`
The route is a live, destructive, cross-table delete with no environment guard — nothing prevents it from being reachable in production if this build is deployed there.
It is auth-scoped (routes through `lib/api.ts`'s `X-Clerk-Authorization` + the API's `AuthGuard`/RLS, so a caller can only ever delete their *own* data — not a cross-tenant vulnerability), which bounds the severity. But "not linked from any UI; not documented as a public API" (the file's own comment) is security-by-obscurity, not a control: any authenticated user (or a leaked staging session) who discovers or is pointed at this URL can irrecoverably wipe their own cycle, lift records, training maxes, and history with one request and no confirmation.
**Fix:** Gate the handler so it's inert unless an explicit staging/test-only flag is set (e.g. return 404 when `ENABLE_TEST_SUPPORT`/equivalent is unset in the deployed environment), so it cannot exist as a live destructive endpoint anywhere this suite doesn't run.

---

**[correctness]** `apps/api/src/adapters/prisma/rls.db.e2e.spec.ts:451-495`
The new RLS regression test seeds only a `cycleDashboard` row for each of the two users, then proves the delete only affects the requesting user's dashboard row.
The test's own comment states it proves "the deleteMany calls are genuinely RLS-scoped" — plural — but `deleteCurrentCycle` makes five scoped deletes, and four of them (`deleteAllLiftRecords`, `deleteAllHistory`, `deleteTrainingMaxes`, the scheduled-workout clear) run against empty tables here, so this test provides zero direct evidence for them. This matters specifically because #644 (the bug motivating this whole PR) was a per-component wiring failure, not a missing-policy failure — a future regression where, say, `PrismaLiftRecordRepository` fails to route through the request-scoped transaction client (the exact class of bug #644 was) would not be caught by this test, even though its stated purpose is precisely to catch that class of regression for this new endpoint.
**Fix:** Seed at least one row per affected table for both users before the DELETE, and assert the requesting user's rows are gone from every table while the other user's rows in every table survive — or narrow the test's comment to state plainly it covers dashboard-scoping only.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/cycle-generation.service.ts:240-246, 264-267`
The doc comment says the method deletes "all TrainingMax rows," but the code reads current maxes first and deletes by the resulting lift-name list (`deleteTrainingMaxes(program, trainingMaxes.map(m => m.lift))`) rather than an unconditional `deleteMany({ where: { program } })` — the only one of the five deletes that isn't a direct program-scoped delete. Functionally equivalent today (same transaction, nothing can be inserted between the read and the delete), but a future reader reconciling the comment against the code sees a mismatch, and the indirection couples this method to `deleteTrainingMaxes`'s key-list contract for no functional benefit.
**Suggestion:** Add a `deleteAllTrainingMaxes(program)` method consistent with the other four "delete everything for this program" methods, or adjust the comment to describe the read-then-delete-by-key mechanism accurately.

**[maintainability]** `apps/api/src/ports/ICycleDashboardRepository.ts`, `ILiftRecordRepository.ts`, `ITrainingMaxHistoryRepository.ts`, `ITrainingMaxRepository.ts`
The five new/reused delete methods use inconsistent naming for the same "delete everything for this program" granularity: `deleteAllLiftRecords`/`deleteAllHistory` carry an `All` prefix, `deleteCycleDashboard` (equally whole-program) doesn't, and `TrainingMax` has no `deleteAll*` at all — the service reuses the keyed `deleteTrainingMaxes(program, lifts)`. A maintainer scanning any one of these interfaces has to open the JSDoc to learn whether a given `delete*` method is program-wide or key-scoped.
**Suggestion:** Standardize on one naming convention for the whole-program-delete family (e.g. `deleteAll<Entity>ForProgram(program)`) across all four interfaces so granularity is legible from the method name alone.

**[reliability]** `apps/web/e2e/staging.spec.ts` (test 6 placement) and `apps/web/playwright.config.ts:9,12`
Test 6's correctness depends on two undocumented invariants: it must stay the last-declared test in the file, and `workers` must stay `1`. The config sets `fullyParallel: true` alongside `workers: 1` — the serial guarantee that prevents test 6's cycle mutation from overlapping the read-only tests comes entirely from `workers: 1`, not from `fullyParallel`. (Tests 1 and 4 do tolerate either cycle-exists-or-not state per their own comments, which softens this — the real risk is a future `workers: N` bump silently reintroducing interference, not a currently-latent bug.)
**Suggestion:** Add a comment at test 6 (or beside `workers: 1` in the config) stating this suite mutates a shared account and must run serially, so a future parallelism change doesn't silently break the ordering assumption.

**[maintainability]** `apps/web/app/api/test-support/reset-cycle/route.ts:21` vs `apps/web/e2e/staging.spec.ts:679` vs `apps/web/e2e/README.md`
The program name `'rpt'` is hardcoded independently in three places: the route's default (`?? 'rpt'`), the test's explicit query param, and the README's note about a possible stray cycle. Since the test always passes the param explicitly, the route's fallback default is dead code that can drift silently if the onboarded program ever changes.
**Suggestion:** Drop the `?? 'rpt'` default and have the route 400 on a missing `program` param (fail-fast on misuse instead of silently defaulting), or hoist the program name to one shared constant referenced by both the test and the README note.

**[maintainability]** `apps/api/src/programs/cycle-generation.service.spec.ts:333-352`
The "deletes the dashboard after the other deletes complete (ordering)" test asserts the dashboard delete is last, but because the other four deletes currently run under `Promise.all`, their relative order in the recorded array is non-deterministic — the test only actually pins "dashboard is last," not any ordering among the other four. If the `Promise.all` → sequential-`await` fix above lands, this test would still pass without having verified the new sequencing.
**Suggestion:** Rename/retarget the test to state only what it guarantees ("dashboard deleted only after the other four resolve"), and consider whether the sequential-`await` fix warrants its own explicit ordering assertion.

---

### Findings Disposition
*(to be completed after fixes)*

---

*Reviewed by `claude-opus-4-8` (parallel correctness/security + reliability/maintainability subagents) via Claude Code*

PR #666, opened 2026-07-03
[chore] Add missing docker-compose.test.yml for DB E2E Docker-unavailable recovery

brownm09:
## Summary

Adds the missing `docker-compose.test.yml` at the repo root. Three pre-existing references already document this file as recovery option 2 for the DB E2E suite when Testcontainers/Docker is unreachable, but it was never actually committed:

- `apps/api/jest.global-setup.js` (`buildDockerUnavailableMessage`) — names it with port 5433, `lifting`/`lifting`, `lifting_test`.
- `docs/testing/e2e-coverage.md` ("Running locally when Docker is unavailable", option 2) — same values, plus the exact `docker-compose up -d && prisma migrate deploy && npm test` sequence.
- `docs/adr/ADR-021-no-test-tracing.md` (accepted, closes #222) — treats the file as already-existing, load-bearing design: "The test environment (`docker-compose.test.yml`) runs only Postgres — no OTel Collector, Tempo, Loki, or Prometheus."

No doc changes are needed — all three already describe the file correctly. This PR only adds the file itself, matching:

- The pinned `postgres:16-alpine` digest already used by `jest.global-setup.js`'s `POSTGRES_IMAGE` and CI's `db-integration` service.
- `lifting`/`lifting`/`lifting_test` credentials, matching CI exactly.
- Host port `5433` (not CI's 5432) so it doesn't collide with the root `docker-compose.yml`'s Postgres (also 5432) or a locally-installed default-port Postgres.
- Postgres only — no observability services — per ADR-021's explicit scope decision.

Pre-existing gap, not caused by #646/#658 — traced back to ADR-021 (2026-05-10), where the ADR recorded the file's scope decision as already-accepted fact but the file itself apparently never got committed.

## Acceptance Criteria

- [x] `docker-compose.test.yml` exists at the repo root with the values `jest.global-setup.js` and `e2e-coverage.md` already document.
- [x] Postgres-only — no OTel/Tempo/Loki/Prometheus/Grafana services, per ADR-021.
- [x] Verified end-to-end against the actual documented recovery steps (see Testing below).

## Testing

Verified the exact documented recovery procedure end-to-end, not just that the YAML parses:

1. `docker-compose -f docker-compose.test.yml config` — valid.
2. `docker-compose -f docker-compose.test.yml up -d`, waited for the healthcheck to report `healthy`.
[...]

Issue #648, opened 2026-07-03
[observability] Alert on Postgres permission/RLS-policy error codes in production


brownm09:
## Problem

#644 (RlsInterceptor silently never setting the RLS GUC) produced a Postgres `42501` "row-level security policy" error on the very first real request after the RLS migration landed on 2026-06-11, and on every request since. Nothing paged or alerted on it — it was only found 3+ weeks later via a manual `gcloud logging read` during a user-reported investigation.

## Proposal

Add a Grafana Cloud alert rule (alongside the existing latency/error-rate rules referenced in ADR-019/ADR-020) that fires on Postgres permission-denied / RLS-policy-violation error codes (`42501` specifically, surfaced via the structured Pino `err.message`/`err.stack` fields already logged by `nestjs-pino`) appearing on any write endpoint. This class of error should essentially never occur legitimately — a non-zero rate is a strong, low-noise signal of exactly this kind of misconfiguration, distinct from generic 5xx-rate alerting which could be diluted by expected/handled error paths.

## Scope

- New alert rule in `infra/observability/alerts/api.yaml` (or wherever the existing rules live)
- Scoped by `deployment.environment.name` per the existing convention (#487) so staging noise doesn't page on production alerts and vice versa
- Document the alert and its runbook entry in `docs/runbooks/observability.md`

## Acceptance criteria

- [ ] Alert fires on a real `42501` in a test/staging environment (verified manually)
- [ ] Runbook entry added

Related: #644

PR #669, opened 2026-07-03
[feat] Wire Pino structured logs into Grafana Cloud Loki via OTLP

brownm09:
## Summary

While scoping #648 (alert on Postgres `42501`/RLS-policy-violation errors), I found the alert as described could never work: it depends on Pino's structured logs being queryable in Grafana Cloud Loki, but they never actually reached Loki.

`apps/api/src/otel.ts`'s `NodeSDK` configured a trace exporter and metric reader but no `LoggerProvider`/log exporter. The OTel Collector's `logs` pipeline (both local docker-compose and the GKE Helm chart) was fully wired to receive OTLP logs and forward to Loki, but nothing in the app ever emitted one — `@opentelemetry/instrumentation-pino`'s log-sending path (present transitively via `getNodeAutoInstrumentations()`) needs a real `LoggerProvider` to activate, and none existed. This lines up with #644 and #648 both being debugged via a manual `gcloud logging read` (GCP Cloud Logging, auto-populated by GKE) rather than Grafana Explore → Loki.

## What changed

- **`apps/api/src/otel.ts`** — added `logRecordProcessors: [new BatchLogRecordProcessor(new OTLPLogExporter())]` to the `NodeSDK` config. This is what activates `instrumentation-pino`'s existing log-sending mechanism.
- **`apps/api/package.json`** — added `@opentelemetry/sdk-logs` and `@opentelemetry/exporter-logs-otlp-http` as direct deps (both already resolved transitively at `^0.217.0`, matching the sibling trace/metric exporters — clean lockfile diff, no version bump).
- **`apps/api/src/app.module.ts`** — extracted the `pinoHttp` options object to an exported `pinoHttpOptions` constant so the new test can exercise the *real* redact config rather than a copy that could drift.
- **New: `apps/api/src/otel-log-redaction.spec.ts`** — the security-critical test. `app.module.ts` redacts `req.headers.authorization`/cookies (`remove: true`) specifically to keep JWTs out of long-retention log storage; this new export path must not reopen that leak. Proves it empirically (not just by reading source): logs a request with `Authorization`/`Cookie` headers through the real redact config, captures the exact serialized line, feeds it into the same `OTelPinoStream` mechanism instrumentation-pino installs in production, and asserts the secret values appear nowhere in either the wire format or the resulting OTel `LogRecord`.
- **`infra/observability/otel-collector.yaml`, the Helm chart's `configmap.yaml`, `docker-compose.yml`, `infra/kubernetes/values/{staging,production}-otel-collector.yaml`, `infra/cloud-run/otel-collector-sidecar.yaml`, `docs/deploy.md`, `docs/runbooks/observability.md`** — migrated the Collector's logs exporter off the dedicated `loki` exporter type onto the generic `otlphttp` exporter pointed at Loki's native `/otlp` endpoint. The `loki` exporter is deprecated and has been removed from `opentelemetry-collector-contrib` ([removal proposal #38374](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38374), [final removal PR #41413](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41413)); the currently-pinned collector image (`0.104.0`) predates the removal, which is why it still worked, but it's a landmine on the next version bump. Fixing "logs don't reach Loki" while leaving a second reason they'd stop reaching Loki would defeat the point of this PR. The staging/production endpoint URL change (same host, `/loki/api/v1/push` → `/otlp`) is based on Loki's own OTLP ingestion docs and couldn't be independently verified against the live Grafana Cloud stack from this change — flagged inline for confirmation against the portal before rollout.

## Testing

```
Tests: 884 passed, 0 skipped, 0 failed (workspaces: api 427, core 280, web 152, api-client 25)
```

Full `npm test` run from the repo root, all green — re-run after rebasing onto latest `main` (post `/review`, post-#666 merge) to confirm no regression from either.

### Manual verification (local docker-compose)

1. Brought up the full local stack (`docker compose up -d`), confirmed the otel-collector accepted the new `otlphttp/logs` exporter config with no startup errors.
2. Started `apps/api` locally pointed at the collector and made a real request with a fake `Authorization: Bearer <secret>` header.
3. Queried the local Loki HTTP API directly — **log lines now appear** (they returned zero before this change). Confirmed the response never contains the bearer token value anywhere — `req_headers_authorization` is entirely absent from the structured metadata (matches `remove: true` semantics), while unrelated headers (`req_headers_accept`, `req_headers_host`, `req_headers_user_agent`) are present as expected.
4. Also incidentally confirmed useful structure for #648's follow-on alert rule: `exception_message`/`exception_stacktrace`/`req_method`/`deployment_environment_name` all arrive as queryable structured metadata on each log line — more precise than a raw line-content substring match. That rule design will be finalized empirically in #648's own PR.

## Related

Blocks #648 (which depends on this pipeline actually working before its alert rule can fire on anything).

Per `/review` (posted [above](https://github.com/brownm09/lifting-logbook/pull/669#issuecomment-4880795962)):

- **[maintainability] Redaction test bypassed `startOtel()`'s actual `NodeSDK` construction** — fixed in `e2ff4ff`. Extracted `buildLogRecordProcessors()` (mirroring this PR's own `pinoHttpOptions` pattern) and added a test asserting `otel.ts` itself wires a non-empty `logRecordProcessors`, closing the gap between "the mechanism is safe" and "production actually uses it."
- **[maintainability] `OTelPinoStream` deep import into instrumentation-pino's internal build path** — no change needed. This is an accepted tradeoff of testing the real integration point rather than a re-implementation; the alternative (reimplementing the stream's parsing logic) would be more fragile, not less. Will only need revisiting if a future dependency bump breaks the import.
- Style note (`probe.levels` → `pino.levels`) — not fixed; too trivial to warrant a commit on its own, noted for the next time this file is touched.


brownm09:
## Review: [feat] Wire Pino structured logs into Grafana Cloud Loki via OTLP

### Summary

Ready to merge once CI is green. The change is well-scoped, the security-critical property (redaction survives the new export path) is proven both by reading `instrumentation-pino`'s source and by a dedicated test, and the PR proactively fixes an adjacent landmine (the deprecated `loki` Collector exporter) rather than leaving it for a future outage. No blocking findings. Two non-blocking maintainability points below are worth a follow-up but shouldn't hold up this PR.

**CI note:** `Staging Integration Tests` (a required check) and `Deploy web (staging)` initially showed as cancelled. Traced the root cause rather than taking the surface error at face value — the "Deploy web (staging)" job had zero recorded steps (`conclusion: cancelled`, empty `steps` array), meaning it never actually ran; "Staging Integration Tests" then failed downstream waiting on an empty `WEB_URL` output from that never-run job. This is infra/concurrency-related (the job never executed, so nothing in this diff could have caused it), not a code defect — re-triggered via `gh run rerun 28682526769`, currently back `in_progress`. Confirm it completes green before merging, since it's a required status check.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/otel-log-redaction.spec.ts`
The test proves the redaction mechanism is safe by feeding a captured log line directly into a manually-constructed `OTelPinoStream`, bypassing `otel.ts`'s actual `startOtel()`/`NodeSDK` construction entirely (a deliberate, disclosed choice to avoid OTel auto-instrumentation's module-patching-order fragility in Jest). That means no automated test currently asserts that `otel.ts` itself wires `logRecordProcessors` — if a future edit accidentally dropped or typo'd that line, nothing in the suite would catch it; only a repeat of the manual docker-compose verification would.
**Suggestion:** Mirror the `pinoHttpOptions` extraction pattern already used in this PR — pull the `NodeSDK` config object out of `startOtel()` into an exported, testable constant/builder, and add a cheap assertion that it includes a non-empty `logRecordProcessors`. Closes the gap between "the mechanism is safe" and "production actually uses it."

**[maintainability]** `apps/api/src/otel-log-redaction.spec.ts:15` (import from `@opentelemetry/instrumentation-pino/build/src/log-sending-utils`)
`OTelPinoStream` is reached via a deep import into the package's internal build output, not its public API (only `PinoInstrumentation` is exported from the package's entry point). This is the right call for testing the real integration point rather than a re-implementation, but internal paths aren't covered by the package's semver contract — a future patch/minor bump could silently break this import.
**Suggestion:** No action needed now; if this import ever breaks on a dependency bump, that's expected fragility, not a regression in this codebase's own logic — worth a one-line comment update if it recurs, or revisit if `instrumentation-pino` ever exposes this officially.

---

### Style Notes

`probe.levels` in the new test instantiates a throwaway `pino()` logger purely to read its `.levels` property; `pino.levels` (the module's static export) is equivalent for a default-configured logger and avoids the extra instantiation. Trivial, not worth blocking on.

---

### Findings Disposition

Both non-blocking findings above should be fixed-in-PR or filed-and-linked before merge per this repo's all-findings policy.

---

PR #685, opened 2026-07-05
[feat] Add GET /version endpoints to apps/api and apps/web

brownm09:
## Summary

There is currently no way to ask a running lifting-logbook service what git commit it's on. This
adds a `GET /version` endpoint to both `apps/api` and `apps/web`, returning `{gitSha, environment}`,
backed by a `GIT_SHA` build-arg baked into each Docker image's final (`runner`) stage.

Part of #670. Closes #671.

## Design decisions

- **Response shape `{gitSha, environment}`**, not `{gitSha, imageTag, deployedAt}`: `imageTag`
  would just duplicate `gitSha` (images are tagged `api:${{ github.sha }}`, and Cloud Run always
  deploys the explicit SHA tag, never `:latest`); `deployedAt` is redundant with Cloud Run's own
  revision-creation timestamp. `environment` reads `NODE_ENV`.
- A missing `GIT_SHA` degrades to `'unknown'` rather than throwing — unlike
  `apps/web/lib/public-config.ts`'s `PUBLIC_API_URL` guard (ADR-028), a missing `GIT_SHA` only
  degrades observability, not functionality.
- `GIT_SHA` is baked in at **build time** (Dockerfile `ARG`/`ENV` in the `runner` stage only, so
  it never busts the `builder`/`installer` stage cache), not injected at deploy time like
  `PUBLIC_API_URL`/Clerk keys — safe because, unlike those, it's invariant across environments for
  a given build, so it doesn't reopen the per-environment-build problem ADR-028/#396 fixed.
- `apps/web`'s `/version` mirrors `/livez` (fully excluded from `clerkMiddleware`) rather than
  `/api/healthz` (which deliberately runs *through* Clerk to test its own init) — a version check
  needs neither.
- Dropped the "optional OTel `service.version`" idea from the original plan after discovering it
  would require modifying `buildResourceAttributesEnv`'s output format and updating several
  existing, ADR-referenced test assertions (`otel.spec.ts`, tied to incident #487) — more
  invasive than a "nice to have" warrants. Happy to do as a quick separate follow-up if wanted.

## Testing

- `npm run typecheck` — clean.
- `npm test -w @lifting-logbook/api` — 428 passed, 0 failed.
- `npm test -w @lifting-logbook/web` — 165 passed, 0 failed.
- `npm run build` — clean (Next.js build lists the new `/version` route).
- Suppression / test-integrity pre-PR greps — no matches.

[...]

Issue #674, opened 2026-07-04
[fix] apps/web Docker build fails: implicit-any TypeScript error under turbo-pruned build context


brownm09:
## Problem

A fresh (uncached) \`docker build -f apps/web/Dockerfile .\` fails at the
\`RUN npx turbo run build --filter=@lifting-logbook/web\` step with:

\`\`\`
./app/(authed)/cycle/[cycleNum]/page.tsx:56:37
Type error: Parameter 'm' implicitly has an 'any' type.

  const maxMap = new Map(maxes.map((m) => [m.lift, m.weight]));
```

## Confirmed pre-existing, not introduced by any in-flight branch

- Verified via \`git show origin/main:apps/web/app/\(authed\)/cycle/\[cycleNum\]/page.tsx\`
  — the exact same code is already on \`main\`.
- The file was last touched by an unrelated PR (#390), long before this was discovered.
- \`npm run typecheck\` (\`tsc --noEmit -p tsconfig.typecheck.json\`, the project's own gate)
  does **not** catch this — only \`next build\`'s internal type-check does. This means the
  gate that runs on every PR does not exercise the same type-checking path as the actual
  Docker image build.

## Why this hasn't broken CI (probably)

Recent \`deploy.yml\` runs are green (checked via \`gh run list --workflow=deploy.yml\`), so
production's actual Docker builds are not visibly failing right now. The likely explanation:
\`apps/web/Dockerfile\`'s builder stage does \`turbo prune --scope=@lifting-logbook/web --docker\`
before installing/building, producing a *minimal* pruned dependency subset — the resulting
type-checking context may differ from a full-monorepo \`next build\` (which is what
\`npm run build\` at the repo root does, and which does **not** reproduce this error locally).
Combined with the GHA layer cache (\`cache-from: type=gha,scope=web-deploy-v3\`), it's likely
the specific Docker layer that would re-run this build has been cache-hit since before this
regressed, masking it in CI. If that cache is ever invalidated for a build touching this file's
dependency chain, a real deploy could fail unexpectedly.

## Suggested investigation

1. Reproduce locally: \`docker build --no-cache -f apps/web/Dockerfile .\` from repo root.
2. Determine why \`turbo prune --docker\`'s pruned output resolves \`maxes\`'s element type
   differently than a full monorepo build — likely a type-only import not captured by turbo's
   dependency graph, or a missing file in the pruned \`out/full/\` tree.
[...]

Issue #670, opened 2026-07-03
[chore] Deployment version visibility: expose git SHA from running services


brownm09:
## Problem

There is currently no way to ask a running lifting-logbook service (api or web, staging or
production) what git commit it's running.

- `apps/api/src/health/health.controller.ts`'s `/health` and `/readyz` endpoints return only
  `{status, timestamp}` — no build/version info.
- Docker images are tagged with `github.sha` at build time (`.github/workflows/deploy.yml` /
  `staging.yml`), and Cloud Run is deployed with the explicit `:${{ github.sha }}` tag — the
  information exists in Artifact Registry/Cloud Run, but the only way to read it today is
  `gcloud run services describe --format="value(spec.template.spec.containers[0].image)"`, then
  manually cross-referencing the SHA against `git log`.
- This caused a real near-miss on 2026-07-03: PR #667's Deploy run was silently cancelled by
  `deploy.yml`'s `concurrency: {group: deploy-main, cancel-in-progress: false}` behavior — a
  known, documented, intentional trade-off (deploy.yml:32-45, #401) where a queued run is dropped
  in favor of a later push. Confirming production had actually picked up the fix required
  manually diffing `gcloud run services describe`'s image tag against `git log` and checking
  ancestry by hand. It happened to resolve itself when a later PR's deploy carried the fix
  forward as an ancestor commit — but nothing in the system said so; a human had to notice and
  check by hand.
- `deploy.yml`'s own final step (`Print deployment URLs`) prints only the Cloud Run URLs — never
  the SHA that was actually deployed, even though the workflow already knows it (`github.sha`).

## Goal

Make "what commit is this service running" a first-class, queryable fact — both from the running
service itself (HTTP) and from the CI deploy record (job summary) — so the manual
cross-referencing described above doesn't have to happen again.

## Sub-issues

- [ ] #671 — `/version` endpoints on apps/api and apps/web (core, self-contained fix)
- [ ] #672 — CI auditability (deploy job summary + cancelled-run ancestry audit) + helper script
      + runbook (builds on #671)

PR #686, opened 2026-07-05
[feat] Add GET /version endpoints to apps/api and apps/web

### LINKED ISSUE #671: [feat] Add /version endpoints to apps/api and apps/web
Part of #670.

## Scope

Add a `GET /version` endpoint to both services, returning `{gitSha, environment}`, so a running
container can report what commit it's on without needing `gcloud run services describe` +
manual cross-reference against `git log`.

## Design

- **Response shape `{gitSha, environment}`** — not `{gitSha, imageTag, deployedAt}`: `imageTag`
  would just duplicate `gitSha` (images are tagged `api:${{ github.sha }}` and Cloud Run always
  deploys the explicit SHA tag, never `:latest`); `deployedAt` is redundant with Cloud Run's own
  revision-creation timestamp. `environment` reads `NODE_ENV`.
- A missing `GIT_SHA` degrades to `'unknown'`, it does not throw — unlike
  `apps/web/lib/public-config.ts`'s `PUBLIC_API_URL` guard (ADR-028), a missing `GIT_SHA` only
  degrades observability, not functionality.
- `GIT_SHA` is baked in at Docker build time (`ARG`/`ENV` in the final `runner` stage only, so it
  doesn't bust the `builder`/`installer` layer cache), not injected at deploy time like
  `PUBLIC_API_URL`/Clerk keys — safe because, unlike those, it never differs per environment for
  a given build (same image, same commit, however many places it's promoted to), so it doesn't
  reintroduce the per-environment-build problem ADR-028 fixed (#396).
- `apps/api`: third method on the existing `HealthController`, `@Public()`, reusing the existing
  exported `resolveDeploymentEnvironment()` from `apps/api/src/otel.ts`.
- `apps/web`: new top-level `apps/web/app/version/route.ts`, mirroring `app/livez/route.ts`
  (pure, unauthenticated, `force-dynamic`, excluded from `clerkMiddleware`) rather than
  `/api/healthz`'s pattern (which deliberately runs through Clerk to test its own init). Requires
  adding `version` to the middleware matcher exclusion and updating
  `middleware.matcher.test.ts`, which locks down that regex.
- Optional: add a `service.version` OTel resource attribute in `apps/api/src/otel.ts` and
  `apps/web/instrumentation.ts` alongside the existing `deployment.environment.name` one, now
  that `GIT_SHA` is a real runtime env var — ties traces/logs to a specific commit.

## Acceptance criteria

- [ ] `GET /version` on apps/api returns `{gitSha, environment}`, `@Public()`
- [ ] `GET /version` on apps/web returns the same shape, bypasses Clerk middleware
- [ ] Both Dockerfiles bake `GIT_SHA` into their runner stage without invalidating the
      builder/installer cache
- [ ] `deploy.yml` and `staging.yml` pass `GIT_SHA` as a build-arg to every image build
- [ ] Tests added for both endpoints + the middleware matcher change
- [ ] Verified locally: `docker build --build-arg GIT_SHA=test123` + `docker run` + curl

brownm09:
## Summary

There is currently no way to ask a running lifting-logbook service what git commit it's on. This
adds a `GET /version` endpoint to both `apps/api` and `apps/web`, returning `{gitSha, environment}`,
backed by a `GIT_SHA` build-arg baked into each Docker image's final (`runner`) stage.

Part of #670. Closes #671.

(Note: superseding PR #685, which was accidentally opened against the wrong branch due to a
tooling glitch and contained none of this work — closed with an explanation there.)

## Design decisions

- **Response shape `{gitSha, environment}`**, not `{gitSha, imageTag, deployedAt}`: `imageTag`
  would just duplicate `gitSha` (images are tagged `api:${{ github.sha }}`, and Cloud Run always
  deploys the explicit SHA tag, never `:latest`); `deployedAt` is redundant with Cloud Run's own
  revision-creation timestamp. `environment` reads `NODE_ENV`.
- A missing `GIT_SHA` degrades to `'unknown'` rather than throwing — unlike
  `apps/web/lib/public-config.ts`'s `PUBLIC_API_URL` guard (ADR-028), a missing `GIT_SHA` only
  degrades observability, not functionality.
- `GIT_SHA` is baked in at **build time** (Dockerfile `ARG`/`ENV` in the `runner` stage only, so
  it never busts the `builder`/`installer` stage cache), not injected at deploy time like
  `PUBLIC_API_URL`/Clerk keys — safe because, unlike those, it's invariant across environments for
  a given build, so it doesn't reopen the per-environment-build problem ADR-028/#396 fixed.
- `apps/web`'s `/version` mirrors `/livez` (fully excluded from `clerkMiddleware`) rather than
  `/api/healthz` (which deliberately runs *through* Clerk to test its own init) — a version check
  needs neither.
- Dropped the "optional OTel `service.version`" idea from the original plan after discovering it
  would require modifying `buildResourceAttributesEnv`'s output format and updating several
  existing, ADR-referenced test assertions (`otel.spec.ts`, tied to incident #487) — more
  invasive than a "nice to have" warrants. Happy to do as a quick separate follow-up if wanted.

## Testing

- `npm run typecheck` — clean.
- `npm test -w @lifting-logbook/api` — 429 passed, 0 failed.
- `npm test -w @lifting-logbook/web` — 166 passed, 0 failed.
- `npm run build` — clean (Next.js build lists the new `/version` route).
- Suppression / test-integrity pre-PR greps — no matches.

Tests: 595 passed, 0 skipped, 0 failed.

**Docker end-to-end verification: not completed locally, by design — see below.**

### Why the local Docker build/run wasn't completed

Attempting this locally today drove the machine's disk to 0 bytes free mid-build, which silently
corrupted a partially-built image (a truncated `package.json` inside the image, `Unexpected end
of JSON input` at container start — classic disk-full-during-write corruption, not a code issue).
Docker's own cache/image pruning didn't reclaim host disk space (likely WSL2 VHDX growth that
needs a separate manual compaction step), so further local attempts weren't safe to pursue in this
session. Discussed with the user; agreed to rely on `staging.yml`'s real, clean-environment Docker
build + deploy + integration test as the actual end-to-end verification for this PR, which will
run automatically on this PR.

Before hitting the disk issue, the `apps/api` image did build successfully through all stages
(including the `ARG GIT_SHA`/`ENV GIT_SHA` addition), confirming the Dockerfile mechanics are
sound; Docker's build-arg-to-env-var propagation is standard, well-established behavior, not a
novel mechanism introduced here.

### Unrelated pre-existing issue found along the way

While attempting a clean `apps/web` Docker build, hit a genuinely pre-existing, unrelated
TypeScript error (`apps/web/app/(authed)/cycle/[cycleNum]/page.tsx:56`, implicit-any) that's
already on `main` (confirmed via `git show origin/main:...`) but isn't caught by `npm run
typecheck` — only by `next build`'s internal check, which is masked in real CI by Docker layer
caching. Filed as #674 (not fixed here — out of scope, per this repo's "fix unrelated errors in a
separate PR" policy).

## Acceptance criteria (from #671)

- [x] `GET /version` on apps/api returns `{gitSha, environment}`, `@Public()`
- [x] `GET /version` on apps/web returns the same shape, bypasses Clerk middleware
- [x] Both Dockerfiles bake `GIT_SHA` into their runner stage without invalidating the
      builder/installer cache
- [x] `deploy.yml` and `staging.yml` pass `GIT_SHA` as a build-arg to every image build
- [x] Tests added for both endpoints + the middleware matcher change
- [ ] ~~Verified locally: `docker build --build-arg GIT_SHA=test123` + `docker run` + curl~~ —
      descoped to CI verification per the disk-space issue above

- **[correctness] `GIT_SHA ?? 'unknown'` doesn't degrade correctly on empty string** — Fixed in
  `3a2b053`: Dockerfile `ARG GIT_SHA=unknown` defaults + `||` instead of `??` in both handlers,
  with new regression tests for the empty-string case in both `apps/api` and `apps/web`.
- **[documentation] staging.yml skip-build path can leave `/version` stale** — Fixed in `3a2b053`:
  added a comment extending the file's existing tag/content-mismatch note to cover this.


brownm09:
## Review: [feat] Add GET /version endpoints to apps/api and apps/web

### Summary
Solid, well-scoped PR — clean tests (593 passing), sensible design rationale documented inline and in the PR body, and appropriately proportioned to the stated goal. One real (if narrow) correctness gap in the `GIT_SHA` fallback logic and one documentation gap worth closing before merge; neither blocks shipping. Diff size (268 lines) is appropriately scoped — no split needed.

---

### Non-Blocking Findings

**[correctness]** `apps/api/src/health/health.controller.ts:52`, `apps/web/app/version/route.ts:27`
Both endpoints use `process.env.GIT_SHA ?? 'unknown'`. Docker's `ARG GIT_SHA` (declared with no default in both Dockerfiles) expands to an **empty string**, not `undefined`, when no `--build-arg GIT_SHA=...` is passed at build time — and `??` only substitutes on `null`/`undefined`, not `''`. So a plain `docker build -f apps/api/Dockerfile .` (or `apps/web/Dockerfile`) without the build-arg — exactly the kind of manual local build attempted during this session's own verification — silently produces `{gitSha: ''}`, not the documented `{gitSha: 'unknown'}`. Real CI is unaffected (`deploy.yml`/`staging.yml` always pass the arg), so this is a docs/local-dev-only gap, not a production risk.
**Fix:** Either give the Dockerfile `ARG`s a default (`ARG GIT_SHA=unknown` in both `apps/api/Dockerfile` and `apps/web/Dockerfile`), or switch the fallback to a truthy check (`process.env.GIT_SHA || 'unknown'`) in both files. Cheap to do both.

**[documentation]** `.github/workflows/staging.yml` (build-skip-reuse path, `Resolve image tag` step)
When a PR's diff is non-image-affecting and a prior successful build exists, no new image is built — the floating `pr-<N>` tag is reused as-is. The file's own comments already acknowledge this breaks the "SHA-as-image-tag invariant on skip runs," but this PR is the first time that limitation becomes an end-user-visible claim: `/version`'s `gitSha` on a staging PR preview can lag the actual PR head SHA whenever a build was skipped.
**Suggestion:** A short comment near the skip-build logic (or a line in the PR2 runbook once `docs/runbooks/checking-deployed-version.md` exists) noting this caveat, so a future reader doesn't mistake a stale staging `/version` response for a bug.

---

### Style Notes
Minor naming inconsistency: the api-side helper is `resolveDeploymentEnvironment` (`otel.ts`) while the web-side equivalent is `resolveEnvironment` (`route.ts`) — same purpose, different name across the two mirrored files. Not worth a rename on its own, but worth keeping in mind if either file is touched again.

---

### Findings Disposition
Per this repo's merge-gate policy, both findings above need a disposition (fixed in `<sha>` | filed `#N`) recorded in the PR body before merge.

---

Issue #695, opened 2026-07-05
[infra] Enable GitHub merge queue on main + live validation


brownm09:
Part of #673 — top-level tracking issue with full root-cause diagnosis and options considered.

Depends on the merge-group readiness work in #694 being merged and run for a while first — this should not be flipped on immediately after #694 merges, per the original risk/sequencing note in #673.

## Scope
- Branch protection: check "Require merge queue" for `main`, set Build concurrency = 1 (keeps the staging-deploy mutex uncontended even inside the queue)
- Live end-to-end validation: push a real or throwaway PR through the queue, confirm `ci.yml`/`staging.yml` fire correctly under `merge_group`, staging deploys succeed, PR merges cleanly
- Update `docs/operations/branch-protection.md` to mark merge queue as active, document Build Concurrency = 1 and its rationale

Closes #673 once merged.

brownm09:
**Soak criterion before starting this issue**

Per #673's sequencing note, this should wait until #694's `merge_group` wiring (merged in #696) has run cleanly across ordinary traffic — not a fixed calendar wait, but PR diversity:

- **3-5 ordinary PRs** merge cleanly through the unchanged `pull_request`-triggered paths in `ci.yml`/`staging.yml`
- All 5 required checks stay green with no new flakiness: `Lint & Test`, `DB Integration Tests`, `Observability Stack Smoke Test`, `Playwright E2E`, `Staging Integration Tests`

As of #696's merge (2026-07-06T03:38:26Z), zero PRs have merged since — this bar hasn't started accumulating yet. Check `gh pr list --state merged --limit 5` before starting this issue to confirm the bar is cleared.

brownm09:
**Blocker discovered: merge queue requires organization ownership, not just public visibility.**

This issue's original recommendation (and [ADR-030](https://github.com/brownm09/lifting-logbook/blob/main/docs/adr/ADR-030-github-merge-queue-adoption.md)) assumed public repo visibility alone was sufficient — that assumption was wrong. Per GitHub's docs: *"Pull request merge queues are available in any public repository owned by an organization, or in private repositories owned by organizations using GitHub Enterprise Cloud."* `brownm09/lifting-logbook`'s owner type is `User`, not `Organization` — confirmed live via `gh api repos/brownm09/lifting-logbook`.

Tracked as [#729](https://github.com/brownm09/lifting-logbook/issues/729), which blocks #695 until resolved. ADR-030 correction incoming in a separate small PR.

Issue #673, opened 2026-07-04
[infra] Staging-deploy contention causes merge-time cascading cancellations across concurrent PRs


brownm09:
## Problem

During PRs [#661](https://github.com/brownm09/lifting-logbook/pull/661) and [#664](https://github.com/brownm09/lifting-logbook/pull/664) (2026-07-03), merging was severely slowed by staging-pipeline contention under concurrent PR throughput — 2–4 update+rerun+merge cycles were needed per PR.

## Root cause (confirmed by reading the workflows)

`Staging Integration Tests` is a required check (`docs/operations/branch-protection.md`, `.github/expected-required-checks.json`). It depends on `deploy-api` / `deploy-web` in [`staging.yml`](.github/workflows/staging.yml), each of which deploys to the **one shared** staging environment under a **global** (not per-PR) concurrency group:

```yaml
# staging.yml:318-320 (deploy-api) and :532-534 (deploy-web)
concurrency:
  group: staging-deploy-mutex-api   # same literal group for every PR
  cancel-in-progress: false
```

`cancel-in-progress: false` looks like it should make concurrent PRs *queue* safely, but GitHub Actions only ever keeps **one pending run per concurrency group** — a documented quirk already called out in this repo's own `deploy.yml` (lines 32–39):

> "GitHub Actions retains at most one pending run per concurrency group — when a third push arrives while a second is queued, the second is silently cancelled and the third takes the pending slot."

So under sustained multi-PR throughput: PR A holds the mutex (running) → PR B's deploy queues (pending) → PR C pushes and **cancels B's pending run**, taking the slot itself. B's `deploy-api` job ends with `result: cancelled`. `staging-integration-tests`' `Verify deploy prerequisites` step (`staging.yml:667-692`) treats any non-`success` upstream result as fatal:

```yaml
if [[ "$DEPLOY_RESULT" != "success" ]]; then
  echo "::error::deploy-api did not succeed (result: $DEPLOY_RESULT)..."
  exit 1
fi
```

— hard-failing the required check for B, with an error message ("re-run if transient") that reads like a normal flake rather than "another PR displaced you from the queue."

This compounds with branch protection's `strict: true` (require branches up to date): fixing B means pushing an update, which resets its staging run and re-enters the same contested queue — under continuous throughput this is a live-lock, not just added latency.

## Options evaluated

[...]

PR #696, opened 2026-07-05
[infra] Add merge_group triggers to CI workflows (merge-queue readiness)

### LINKED ISSUE #694: [infra] Add merge_group triggers to CI/staging workflows (merge-queue readiness)
Part of #673 — top-level tracking issue with full root-cause diagnosis and options considered.

## Scope

Additive-only wiring so `ci.yml` and `staging.yml` are ready to run under GitHub's `merge_group` event, in preparation for enabling "Require merge queue" on `main` (tracked separately in a follow-up issue). None of these changes alter behavior on the existing `pull_request`/`push` paths.

### `.github/workflows/ci.yml`
- Add `merge_group:` trigger
- Lockfile-sync check also runs under `merge_group`
- `observability-smoke` diff logic gets a `merge_group`-aware branch

### `.github/workflows/staging.yml`
- Add `merge_group:` trigger
- Fix concurrency group so merge-queue entries don't collide with each other
- Fix `build-images` fork-repo guard for `merge_group` (no PR context)
- "Resolve image tag" step: always rebuild fresh under `merge_group` (no floating PR tag to reuse)
- Docker image tags: use resolved `image_tag` output instead of raw PR context
- `preflight` job: add checkout + `base` input so `dorny/paths-filter` can diff correctly under `merge_group`
- `report-status` job: skip entirely under `merge_group` (no PR to comment on)

### Docs / ADR
- `docs/operations/branch-protection.md`: document merge-queue readiness (not yet activated)
- New ADR recording the merge-queue adoption decision

## Out of scope
- `prod-plan-pr.yml` is not migrated (advisory-only, not a required check) — noted as an accepted gap
- Enabling "Require merge queue" itself and live validation — tracked in the follow-up issue

#673 stays open until the follow-up issue also lands.

brownm09:
## Summary

Additive-only wiring so `ci.yml` and `staging.yml` validate correctly when run via a GitHub merge-queue entry, in preparation for enabling "Require merge queue" on `main` (tracked separately in #695). Root-cause diagnosis and options considered are in #673.

**None of these changes alter behavior on the existing `pull_request`/`push` paths** — every fix is either a new trigger, a `||`-fallback that only activates when the pull_request-specific field is already empty, or an `if:` that adds a disjunct without removing the existing condition.

### `.github/workflows/ci.yml`
- Add `merge_group` trigger
- Lockfile-sync check also runs under `merge_group` (defense-in-depth, per ADR-036's spirit)
- `observability-smoke` diff logic gets a `merge_group`-aware branch (safe fallback either way)

### `.github/workflows/staging.yml`
- Add `merge_group` trigger
- Fix concurrency group so merge-queue entries don't collide with each other (previously would have shared one literal group, recreating the exact cancellation bug #673 diagnosed)
- Fix `build-images` fork-repo guard for `merge_group` (no PR context to check)
- "Resolve image tag" step: always rebuild fresh under `merge_group` (no floating PR tag to reuse), fails fast if the merge-group SHA resolves empty rather than risking a malformed Docker tag
- Docker image tags + `GIT_SHA` build-args: use the resolved `image_tag` output instead of raw PR context (also catches a `GIT_SHA` build-arg added by #691 since this branch was cut — same fix pattern applies)
- `preflight` job: add a `merge_group`-only checkout + `base` input so `dorny/paths-filter` can diff correctly under `merge_group` (it needs a real git diff for non-`pull_request` events; `pull_request` continues to need no checkout at all, per the action's own docs)
- `report-status` job: skip entirely under `merge_group` (no PR to comment on)

### Docs / ADR
- `docs/operations/branch-protection.md`: new "Merge queue readiness" section — explicitly states "Require merge queue" is **not yet enabled live**
- New [ADR-030](https://github.com/brownm09/lifting-logbook/blob/infra/issue-694-merge-group-triggers/docs/adr/ADR-030-github-merge-queue-adoption.md) recording the merge-queue adoption decision, including two assumptions flagged as not fully doc-confirmed at implementation time:
  - `merge_group` entries are assumed to always be same-repo branches (fork-repo guard skipped for `merge_group`) — low-risk given this repo has no fork-PR contribution pattern
  - `github.event.merge_group.base_sha`/`.head_sha` field names are based on `head_sha` being GitHub's documented per-entry identifier, not a fully-spelled-out payload schema. The two diff-comparison uses degrade to a safe default if wrong; the image-tag use instead fails fast with an explicit error (see Review findings disposition below)

## Consciously out of scope

- **`prod-plan-pr.yml`** has the identical fork-guard and concurrency-group shape as `staging.yml` but produces no required check (advisory Terraform-plan PR comment only). Not migrated — a Terraform-touching PR loses that advisory comment when processed through the queue post-#695. UX gap, not a safety gap.
- `deploy.yml`, `merge-ready-digest.yml`, `required-checks-drift.yml` — confirmed no `github.event.pull_request.*` references or PR-shaped triggers, genuinely out of scope.
- **Enabling "Require merge queue" itself and live end-to-end validation** — tracked in #695, deliberately deferred until this PR has run for a while and is confirmed stable (flipping the branch-protection setting immediately would defeat the point of sequencing).

## Testing

- `npm run typecheck`: 4/4 tasks successful (55.5s)
- `npm test`: 12/12 turbo tasks successful; `@lifting-logbook/api` reported explicitly as `Tests: 440 passed, 0 skipped, 0 failed (96.2s)` — no failures anywhere in the run
- Pre-PR suppression / test-integrity / lockfile-drift greps: all clean (no matches) — this PR touches only workflow YAML and docs, no application code or `package.json`
- **Limitation, stated explicitly**: the actual `merge_group` code paths cannot be exercised locally or by this PR's own CI, since no merge queue is active yet. The only real verification of the `merge_group` branches themselves is #695's live queue run. This PR's CI (running as an ordinary `pull_request`) verifies only that the additive changes don't regress the existing path.

`/review` found 1 blocking, 2 non-blocking. All fixed in `414206f`:

- **[documentation]** `docs/README.md` ADR index table missing the new ADR-030 row — fixed in `414206f`.
- **[reliability]** "Resolve image tag" `merge_group` branch had no emptiness check on `merge_group.head_sha`, which could have produced a malformed Docker tag rather than degrading gracefully like the PR's other unconfirmed-field-name uses — fixed in `414206f` (fail-fast `::error::` guard); ADR-030's Consequences section corrected to match.
- **[performance]** `preflight` job's new checkout ran unconditionally, even though `pull_request` diffing needs no checkout at all — fixed in `414206f` (scoped to `if: github.event_name == 'merge_group'`).

Closes #694


brownm09:
## Review: [infra] Add merge_group triggers to CI workflows (merge-queue readiness)

### Summary
Small, well-scoped PR (128 lines across 5 files) that does what it says — additive `merge_group` wiring with no behavior change to existing `pull_request`/`push` paths. Merge with minor changes: one blocking documentation gap (the ADR index table wasn't updated for the new ADR-030) and two non-blocking findings, the more notable being that the "Resolve image tag" step's `merge_group` branch doesn't actually have the same safe-fallback property the PR body and ADR-030 claim every merge_group code path has.

---

### Blocking Findings

**[documentation]** `docs/README.md:131-163`
`docs/README.md` maintains a table titled "Architecture Decision Records" listing every ADR from ADR-001 through ADR-029 with its status. This PR adds `docs/adr/ADR-030-github-merge-queue-adoption.md` but does not add a corresponding row to this table, so the index is immediately stale the moment this PR merges — a session or reader orienting from `docs/README.md` won't discover ADR-030 exists.
**Fix:** Add a row for ADR-030 to the table in `docs/README.md`, matching the existing format (e.g. `| [ADR-030](adr/ADR-030-github-merge-queue-adoption.md) | GitHub Merge Queue Adoption | Accepted |`).

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/staging.yml:231-237`
The `merge_group` branch of "Resolve image tag" sets `image_tag=${{ github.event.merge_group.head_sha }}` with no check that the value is non-empty. ADR-030's Consequences section states every use of the unconfirmed `merge_group.base_sha`/`.head_sha` fields "degrades to 'always run' / 'always rebuild fresh' if wrong, never a silent skip" — but that's not quite true here. If `head_sha` resolves empty, `image_tag` becomes an empty string, which flows into the Docker `tags:` field two steps later as `<repo>/api:` (empty tag component) — a malformed tag, not a graceful degradation. The other two field uses (`ci.yml`'s observability-smoke `BASE`/`HEAD`, and `staging.yml`'s `preflight` `base:` input) both have a genuine safe fallback (an explicit `-z` check, and `paths-filter`'s own empty-defaults-to-default-branch behavior, respectively) — this one doesn't.
**Suggestion:** Add a guard after the `echo "image_tag=..."` line — e.g. `if [[ -z "${{ github.event.merge_group.head_sha }}" ]]; then echo "::error::merge_group.head_sha resolved empty — cannot tag image" >&2; exit 1; fi` — so a wrong-field-name scenario fails loudly and immediately at this step instead of producing a malformed tag two steps downstream. This also tightens the ADR's own accuracy: either add the guard, or soften the Consequences wording to note this one path fails hard rather than degrades gracefully.

**[performance]** `.github/workflows/staging.yml:74-80`
The new `actions/checkout@v5` (with `fetch-depth: 0`, a full-history fetch) in the `preflight` job now runs unconditionally on every trigger, including ordinary `pull_request` pushes — the much more frequent case. Per `dorny/paths-filter`'s own docs (cited in this PR's ADR-030), `pull_request` diffing uses the GitHub API and needs no checkout at all; the checkout is only required for the `merge_group` path. Every regular PR push now pays a full-history checkout it didn't before, for a job that runs on every single push to every PR.
**Suggestion:** Scope the checkout to the path that actually needs it: `if: github.event_name == 'merge_group'` on the new checkout step. `paths-filter`'s `base:` input is already documented as ignored for `pull_request`, so this doesn't change behavior on that path — only removes an unnecessary checkout from it.

---

### Findings Disposition
Per ADR-028, each finding above must be fixed in this PR or filed as a tracked issue and linked in the PR body before merge.

---

Issue #672, opened 2026-07-03
[feat] Deploy job summary auditability + version-check docs/script


brownm09:
Part of #670. Builds on the /version endpoints sub-issue (same parent).

## Scope

- Extend `deploy.yml`'s final `Print deployment URLs` step (in `deploy-production`) to also print
  the deployed SHA, short commit subject, and a commit link.
- Add a step that lists any `deploy.yml` runs on `main` that were `cancelled` since the last
  successful run, and for each, checks `git merge-base --is-ancestor <cancelled-sha> ${{
  github.sha }}` — reporting whether that cancelled run's changes made it into what actually
  shipped. Targets the exact gap from the 2026-07-03 incident (see #670): nothing today says
  whether a silently-cancelled deploy's changes were carried forward by a later one.
- Add `scripts/check-deployed-version.sh`: curls `/version` on prod + staging, both api and web
  (api requires a Cloud Run identity token, mirroring the existing smoke-test step; web is
  `--allow-unauthenticated`), then prints `git log -1` context for each SHA. Replaces the manual
  `gcloud run services describe` + hand cross-reference from the 2026-07-03 incident.
- Add `docs/runbooks/checking-deployed-version.md` (reference-style, like `observability.md` —
  severity N/A), documenting the endpoints and the script. Add to the index in
  `docs/runbooks/README.md`. Cross-link from `docs/runbooks/deploy-regression-rollback.md`'s
  Diagnosis step 2 as a faster alternative to `kubectl rollout history` + manual tag lookup.

## Acceptance criteria

- [ ] Deploy job summary shows the deployed SHA + commit link
- [ ] Cancelled-run ancestry audit step added and correctly reports non-ancestor cancellations
- [ ] `scripts/check-deployed-version.sh` added and dry-run against staging
- [ ] New runbook added, indexed, and cross-linked

PR #697, opened 2026-07-05
[feat] Deploy job summary auditability + version-check docs/script

brownm09:
## Summary

Sub-issue of #670; builds on #671 (merged in #686). Closes the CI-auditability half of #670's goal:

- Extends `deploy.yml`'s `deploy-production` job's final "Print deployment URLs" step to also print the deployed SHA, short commit subject, and a commit link.
- Adds a new "Audit cancelled deploy runs since last success" step: lists `deploy.yml` runs on `main` cancelled since the last successful run, and for each checks `git merge-base --is-ancestor <cancelled-sha> ${{ github.sha }}` — closing the exact gap from the 2026-07-03 incident where a silently-cancelled deploy's fate had to be checked by hand. Purely advisory (`continue-on-error: true`) — never gates an already-succeeded production deploy.
- Adds `scripts/check-deployed-version.sh`: curls `GET /version` on staging + production, api + web, mirroring the existing prod smoke-test's identity-token pattern for the api legs, and prints `git log -1` context for each returned SHA.
- Adds `docs/runbooks/checking-deployed-version.md` (reference-style, severity N/A, matching `observability.md`), indexed in `docs/runbooks/README.md` and `docs/README.md`, and cross-linked from `deploy-regression-rollback.md`'s Diagnosis step 2 as a faster alternative to `kubectl rollout history` + manual tag lookup.

## Acceptance criteria

- [x] Deploy job summary shows the deployed SHA + commit link
- [x] Cancelled-run ancestry audit step added and correctly reports non-ancestor cancellations
- [x] `scripts/check-deployed-version.sh` added and dry-run against staging
- [x] New runbook added, indexed, and cross-linked

## Notable implementation details

- `deploy-production`'s checkout now uses `fetch-depth: 0` (full history) — `git merge-base --is-ancestor` needs it against arbitrary historical SHAs; this repo is ~400 commits, so the cost is negligible.
- Deliberately did **not** add `actions: read` to the job's permissions — verified `staging.yml`'s `build-images` job already calls `gh run list` successfully with just `contents: read`/`id-token: write` today, and added a comment pointing at that precedent so it isn't "fixed" unnecessarily later.
- The audit step is careful about GitHub Actions' implicit `bash -eo pipefail` default: both `gh run list` calls are explicitly `||`-guarded (a bare `VAR=$(cmd)` isn't exempt from that default), and the per-run ancestry check is tested directly in an `if` condition (also exempt) rather than reading `$?` from a bare command, so a "not an ancestor" result (exit 1) can't abort the script.

## Testing

No application/TypeScript code is touched (CI YAML + bash + docs only), so no new automated test coverage applies — matches the existing precedent for `bootstrap-otel-secrets.sh`/`smoke-test-observability.sh`, which also shipped without dedicated tests.

- `npm run typecheck`: passed, unchanged (4/4 tasks successful — re-confirmed full cache hit after the review-fix commit, since no application code changed).
- `npm test`: **Tests: 440 passed, 0 skipped, 0 failed** (api suite, 96s; 12/12 turbo tasks successful overall, 7 cached).
- `bash -n scripts/check-deployed-version.sh`: syntax OK (re-checked after the review-fix refactor).
- **Live dry-run against real staging infra** (not just `--help`/syntax), re-run after the review-fix refactor: `bash scripts/check-deployed-version.sh --env staging --staging-service-account lifting-logbook-stg-cicd@lifting-logbook-staging.iam.gserviceaccount.com`. Web leg fully succeeded end-to-end — real URL, real `gitSha`, real `git log -1` context. The api leg correctly failed with the improved, more precise error message (confirming the gcloud-exit-status fix from review is live) — this is an IAM-grant gap on my personal dev identity, not a script defect.
- **Live-exercised the `deploy.yml` audit-step bash directly against the real repo** (this workflow has no `workflow_dispatch` trigger, so it can't be dry-run in Actions itself), both before and after the review fixes: confirmed `gh run list --created`/`--jq` return correct data, forced all three ancestry-check branches (yes/no/error), and re-verified the `>=`/`--limit 100` boundary fix against a real cancelled run that appeared mid-session (PR #691) — correctly resolved as an ancestor of `origin/main`'s actual tip. Caught and fixed two real bugs this way: the `TITLE` pipe-escaping four-backslash requirement (pre-review), and confirmed the `--created` filter is timestamp-precise, not day-truncated, empirically (during review-fix verification).

Suppression / test-integrity / lockfile-drift pre-PR greps: all clean (no `package.json` changes either, so lockfile check is N/A).

`/review` ran two parallel passes (correctness/security; reliability/performance/maintainability — full report: [PR comment](https://github.com/brownm09/lifting-logbook/pull/697#issuecomment-4887835231)). All findings were fixed in this PR, none filed/deferred:

| Finding | Severity | Fixed in |
|---|---|---|
| SHA/subject summary step missing `continue-on-error`, inconsistent with file philosophy | Blocking [correctness] | `9cb5cad` |
| Cancelled-run query missing explicit `--limit` (default-30 truncation risk) | Blocking [reliability] | `9cb5cad` |
| `gcloud` calls only checked emptiness, not actual exit status | Blocking [reliability] | `9cb5cad` |
| curl status/body fetched via two separate requests (TOCTOU gap) | Non-blocking [correctness] | `9cb5cad` |
| `--created` used strict `>` (same-second boundary gap) | Non-blocking [correctness] | `9cb5cad` |
| `rc=$?` capture fragility (comment-only fix — declined the suggested restructuring, which would reintroduce the `set -e` bug this code avoids) | Non-blocking [correctness] | `9cb5cad` |
| `check_web`/`check_api` ~90% duplicated | Non-blocking [maintainability] | `9cb5cad` |
| Tab-delimited field extraction via `%%`/`##` was fragile | Non-blocking [maintainability] | `9cb5cad` |
| Runbook didn't distinguish job-summary SHA vs. script's live SHA | Non-blocking [documentation] | `9cb5cad` |
| `docs/README.md`'s runbooks index missing the new runbook | Non-blocking [documentation] | `9cb5cad` |

All fixes re-verified live against real staging infrastructure and real repo/`gh` data (details in Testing above), not just re-read.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1232,6 +1232,11 @@ jobs:
             --wait --timeout=5m
 
       - name: Print deployment URLs and deployed version
+        # This is a post-deploy reporting step, exactly like the audit step below — the
+        # production deploy has already fully succeeded by the time this runs, so a git
+        # hiccup here (e.g. an unresolvable SHA) must never turn an already-green deploy
+        # red (#697 review).
+        continue-on-error: true
         run: |
           SHORT_SHA=$(git rev-parse --short "${{ github.sha }}")
           SUBJECT=$(git log -1 --format='%s' "${{ github.sha }}")
@@ -1288,9 +1293,21 @@ jobs:
 
           # select() excludes this in-progress run itself — belt and suspenders, since a
           # run in progress cannot yet show status=cancelled, but cheap to guard anyway.
+          # (A run's status is exactly one of success/cancelled/etc, so it can never also
+          # match the earlier success-run query — no double-count risk from that side.)
+          #
+          # >= (not strict >) on --created: gh's --created filter is timestamp-precise, not
+          # day-truncated (verified empirically against this repo's own run history), but a
+          # cancelled run created in the SAME SECOND as the last success would be excluded
+          # by strict >. >= closes that gap safely — it can never re-include the success run
+          # itself, since that run's status is "success", not "cancelled" (#697 review).
+          #
+          # Explicit --limit: without it, gh defaults to 30, which could silently truncate
+          # the audit during exactly the burst-cancellation scenario (#401) it exists to
+          # catch. 100 comfortably covers any realistic gap between successful deploys.
           CANCELLED_TSV=$(gh run list --repo "${{ github.repository }}" \
             --workflow deploy.yml --branch main --status cancelled \
-            --created ">${LAST_SUCCESS_CREATED}" \
+            --created ">=${LAST_SUCCESS_CREATED}" --limit 100 \
             --json databaseId,headSha,displayTitle \
             --jq "[.[] | select(.databaseId != ${{ github.run_id }})] | .[] | [.databaseId, .headSha, .displayTitle] | @tsv") || {
             echo ":warning: \`gh run list\` failed listing cancelled runs — skipping this audit (advisory; does not affect the deploy)." >> "$GITHUB_STEP_SUMMARY"
@@ -1329,6 +1346,10 @@ jobs:
             if git merge-base --is-ancestor "$SHA" "${{ github.sha }}" 2>/dev/null; then
               echo "| [${RUN_ID}](${RUN_URL}) — ${TITLE} | \`${SHORT}\` | :white_check_mark: yes |" >> "$GITHUB_STEP_SUMMARY"
             else
+              # rc=$? MUST be the first statement in this branch — it captures the
+              # merge-base call's exit code above. Inserting any command (even an echo)
+              # before this line would clobber $? and silently misclassify every
+              # non-ancestor (128 vs 1) — do not add anything above this line (#697 review).
               rc=$?
               if [ "$rc" -eq 1 ]; then
                 echo "| [${RUN_ID}](${RUN_URL}) — ${TITLE} | \`${SHORT}\` | :rotating_light: **NO — investigate** |" >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -854,6 +854,9 @@ jobs:
         needs.preflight.outputs.staging_enabled != 'true' ||
         needs.smoke-test.result == 'success'
       )
+    # gh run list (used by the cancelled-run ancestry audit near the end of this job)
+    # works with contents: read alone — see staging.yml's build-images job for the
+    # identical working precedent. Do not add actions: read for it.
     permissions:
       contents: read
       id-token: write
@@ -861,6 +864,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
+        with:
+          # Full history, not the default shallow depth-1 clone: the cancelled-run
+          # ancestry audit below runs `git merge-base --is-ancestor` against SHAs
+          # that can be many commits back, which a shallow clone can't resolve
+          # reliably. This repo's entire history is ~400 commits, so the extra
+          # clone cost is negligible against this job's 45-minute budget (#672).
+          fetch-depth: 0
 
       - name: Authenticate to GCP (production)
         uses: google-github-actions/auth@v3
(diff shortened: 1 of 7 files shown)


brownm09:
## Review: [feat] Deploy job summary auditability + version-check docs/script

### Summary
Well-scoped CI/tooling/docs PR implementing all 4 items from issue #672. Two parallel review passes (correctness/security; reliability/performance/maintainability) found no security issues but surfaced several real correctness and reliability gaps in the new CI step and script — most notably an inconsistent `continue-on-error` application and an unbounded `gh run list` query that could silently truncate the exact incident scenario the audit exists to catch. All findings below have been fixed and re-verified live against the real repo/staging infrastructure (not just re-read). Ready to merge after this comment.

---

### Blocking Findings

**[correctness]** `.github/workflows/deploy.yml` — "Print deployment URLs and deployed version" step
Runs `git rev-parse`/`git log` under GitHub Actions' implicit `bash -eo pipefail` with no `continue-on-error`, while the very next step (and the file's general philosophy for post-deploy reporting steps) is explicitly "never gate an already-succeeded production deploy on a reporting step." A git hiccup here would fail an already-green production deploy purely on a summary line.
**Fix:** Added `continue-on-error: true` to this step, consistent with the audit step and the file's stated philosophy.

**[reliability]** `.github/workflows/deploy.yml` — cancelled-runs `gh run list` query
Had no explicit `--limit`, defaulting to `gh`'s built-in cap of 30. During a burst-cancellation incident (exactly the #401 scenario this audit targets), more than 30 cancelled runs since the last success would silently truncate the results — the one case the audit exists to catch.
**Fix:** Added explicit `--limit 100`.

**[reliability]** `scripts/check-deployed-version.sh` — `gcloud` calls
`gcloud run services describe` and `gcloud auth print-identity-token` only checked whether their output was empty, not their actual exit status — a denied/erroring call and a call that legitimately returns nothing produced the same (misleading) message.
**Fix:** Both now check the real exit status (`if ! VAR=$(cmd); then ...`) and report a distinct message for "command failed" vs. "command succeeded but returned nothing."

---

### Non-Blocking Findings

**[correctness]** `scripts/check-deployed-version.sh` — status/body fetched via two separate curl requests
A rolling deploy could swap revisions between the status-check request and the body-fetch request, producing a status/body mismatch.
**Fix:** Combined into a single `curl -w '\n%{http_code}'` request, split after the fact.

**[correctness]** `.github/workflows/deploy.yml` — cancelled-run query used strict `>` on `--created`
Empirically confirmed `gh run list --created` is timestamp-precise (not day-truncated), narrowing this to a same-second boundary edge case, but a cancelled run created in the exact same second as the last success would still be excluded.
**Fix:** Switched to `>=` — safe since a run's status is exactly one of success/cancelled, so this can never double-count the success run itself.

**[correctness]** `.github/workflows/deploy.yml` — `rc=$?` capture fragility
Reviewer suggested restructuring to capture `$?` outside the `if` condition — verified this would reintroduce the exact `set -e` abort-on-nonzero bug the current structure was designed to avoid (a bare command between `else` and `rc=$?` is safe; a bare command before an `if` is not). Declined the literal suggestion.
**Fix:** Added a comment instead, making explicit that `rc=$?` must remain the branch's first statement.

**[maintainability]** `scripts/check-deployed-version.sh` — `check_web`/`check_api` ~90% duplicated
Identical URL resolution, status/body handling, and output formatting repeated in both functions.
**Fix:** Factored into shared `resolve_service_url` / `fetch_version` / `report_version` helpers; `check_web`/`check_api` now differ only in the identity-token step.

**[maintainability]** `scripts/check-deployed-version.sh` — tab-delimited field extraction via `${parsed%%...}`/`${parsed##...}`
Fragile and non-obvious; a future third field would silently corrupt this.
**Fix:** Replaced with `IFS=$'\t' read -r sha nodeenv <<< "$parsed"`.

**[documentation]** `docs/runbooks/checking-deployed-version.md` — job-summary SHA vs. script's live SHA
These answer different questions (what a run *deployed* vs. what's *live now*) and can diverge (staging skip-build reuse) without anything calling that out.
**Fix:** Added a clarifying note to the runbook.

**[documentation]** `docs/README.md` — runbooks index table missing the new runbook
This repo has two runbook indexes (`docs/README.md` and `docs/runbooks/README.md`); only the latter was updated in the original commit.
**Fix:** Added a row to `docs/README.md`, matching how the existing `Observability` reference doc is indexed there.

---

### Style Notes
No style findings — comment density in the new CI step and script is justified given the number of genuinely non-obvious bash/`gh`/`set -e` interactions being documented (both reviewers independently confirmed this rather than flagging it as noise).

---

### Findings Disposition

All 9 findings above were fixed in this PR (commit `9cb5cad`) and re-verified live:
- `bash -n` + `js-yaml` parse checks passed on both changed files.
- The refactored script was re-run against real staging infrastructure (`--env staging`): web leg fully succeeded end-to-end (real URL, real `gitSha`, real `git log` context); api leg failed with the improved, more precise error message (confirmed the exit-status fix is live).
- The updated `deploy.yml` audit-step bash (the `>=`/`--limit` change) was re-exercised directly against the real repo's `gh run list` data — including a real cancelled run that appeared mid-session (PR #691's declarationMap fix) — and correctly resolved as an ancestor against `origin/main`'s actual tip.
- `npm run typecheck` re-confirmed unaffected (full cache hit — no application code touched by any of the fixes).

No findings were filed/deferred — none required a product decision or fell outside this PR's scope.

---

*Reviewed by `claude-opus-4-8` (2 parallel subagents) via Claude Code*

Issue #710, opened 2026-07-05
[test] Full npm test still flakes under Windows parallel load after #567 (jest-worker spawn UNKNOWN; Testcontainers Prisma migrate timing)


brownm09:
## Problem

#567 documented and fixed a Windows-only flake class under a full `npm test` (`turbo run test` runs several jest processes concurrently across workspaces, oversubscribing the CPU): the `apps/api` `*.db.e2e.spec.ts` `beforeAll` hooks tripped Jest's 5s default hook timeout, and a slow web suite occasionally reported `1 failed`. The fix bumped `beforeAll`/`testTimeout` values on win32.

While verifying a PR for #692 (a docs/CI-only change — no test or application code touched), a full `npm test` was run twice from a clean worktree (fresh `npm ci`, Docker up) and failed both times, with **two different symptoms**, neither of which is a hook timeout:

**Run 1 — `Error: spawn UNKNOWN` in jest-worker**, simultaneously across `@lifting-logbook/api`, `@lifting-logbook/web`, and `@lifting-logbook/mobile`'s `test` tasks:
```
Error: spawn UNKNOWN
    at ChildProcess.spawn (node:internal/child_process:421:11)
    at spawn (node:child_process:761:9)
    at fork (node:child_process:172:10)
    at ChildProcessWorker.initialize (.../node_modules/jest-worker/build/index.js:777:45)
```

**Run 2 (retry) — `apps/api`'s Testcontainers Postgres started cleanly, but `prisma migrate deploy` failed against it**:
```
Error: Jest: Got error running globalSetup - .../apps/api/jest.global-setup.js, reason:
[jest.global-setup] `prisma migrate deploy` failed against the started Testcontainers Postgres.
Docker IS reachable — the container started cleanly. The failure is in the migration step itself.
```

**Proof this is environmental, not a regression:** every workspace passes 100% when run individually (outside concurrent `turbo run test`):
- `@lifting-logbook/core`: 280/280 (also passed inline during the full run, since it finishes before other workspaces start their own worker pools)
- `@lifting-logbook/api`: 440/440 (isolated `npm test -w @lifting-logbook/api`, 76s)
- `@lifting-logbook/web`: 180/180 (isolated)
- `@lifting-logbook/mobile`: no tests (isolated)
- `@lifting-logbook/eslint-rules`: 14/14 (isolated, uses `node --test`, not Jest)

## Relationship to #567

[...]

Issue #692, opened 2026-07-05
[chore] Add sync check between Dockerfile's pinned turbo version and package.json


brownm09:
## Problem

PR #691 (fixing #674) pinned \`turbo\` to an exact version (\`2.9.3\`) in three places: \`apps/web/Dockerfile\`'s \`npx turbo@2.9.3 prune\` step, \`package.json\`, and \`package-lock.json\`. Nothing mechanically keeps these in sync — a future \`turbo\` bump that updates \`package.json\`/the lockfile but misses the Dockerfile's hardcoded \`@2.9.3\` would silently reintroduce version drift between the prune step (which runs before \`npm ci\`, with no local install to prefer) and the actual build step.

## Suggested fix

Extend the existing lockfile-sync pre-PR check (see \`CLAUDE.md\` → Dependency and lockfile policy) — or add a small standalone grep/script — to assert the Dockerfile's \`npx turbo@<version>\` pin matches the version declared in \`package.json\`'s \`devDependencies.turbo\`. Wire it into the same pre-PR checklist so drift is caught before merge rather than at the next `--no-cache` Docker build.

Surfaced via \`/review\` on [PR #691](https://github.com/brownm09/lifting-logbook/pull/691).

PR #711, opened 2026-07-05
[chore] Add sync check between Dockerfile's pinned turbo version and package.json

### LINKED ISSUE #567: [test] Local full-suite npm test flakes under Windows parallel load (rls.db.e2e beforeAll 5s timeout; web worker contention)
## Symptom

A full-suite local `npm test` on Windows intermittently reports failures that **pass deterministically when the affected suite is run in isolation**:

- `apps/api` — `src/adapters/prisma/rls.db.e2e.spec.ts` › `beforeAll` exceeds the 5000ms Jest hook timeout (`thrown: "Exceeded timeout of 5000 ms for a hook"`). In isolation: **9 passed, 9 total** (~2.1s).
- `apps/web` — one suite reports `1 failed` in the parallel run. In isolation: **18 suites, 99 passed, 99 total**.

Observed on 2026-06-18 during the pre-PR run for #566 (workflow-timeout change — a diff touching no test/app code, so the flake is environmental, not change-induced). Docker Desktop was up (server 29.4.3), so the RLS `beforeAll` timeout is contention under parallel load, not Docker-down.

## Likely cause

Windows local parallel execution: the api DB E2E `beforeAll` (which connects an owner `PrismaClient` to the Testcontainers Postgres) and the CSV-fixture-heavy web/core suites contend for CPU/memory/IO when run together via `turbo run test`, occasionally pushing the 5000ms `beforeAll` past its deadline and starving a web worker. This is the same class `jest.config.base.js` already mitigates on win32 (`workerIdleMemoryLimit: '512MB'`, `maxWorkers: '50%'`) — see the now-closed #419 / #373 / #394. CI (Linux, Node 20) does not exhibit it.

## Candidate fixes (not yet decided)

- Raise the `beforeAll` hook timeout in `rls.db.e2e.spec.ts` (and sibling `*.db.e2e.spec.ts`) so DB provisioning under load doesn't trip the 5s default.
- Further constrain `maxWorkers` on win32, or serialize the DB E2E project relative to the unit projects.

## Acceptance criteria

- [ ] A full-suite `npm test` on Windows passes without isolation-only flakes, or the DB E2E beforeAll timeout is given adequate headroom under parallel load.

brownm09:
## Summary

PR #691 (fixing #674) pinned `turbo` to an exact version (`2.9.3`) in three places: `apps/web/Dockerfile`'s `npx turbo@2.9.3 prune` step, `package.json`, and `package-lock.json`. Nothing mechanically kept the Dockerfile pin and `package.json` in sync — this adds that check.

- Adds `scripts/check-turbo-version-sync.mjs`: reads the Dockerfile's `npx turbo@<version> prune` pin and `package.json`'s `devDependencies.turbo`, and fails with a clear message if they differ. Mirrors the existing `scripts/validate-analytics-taxonomy.mjs` two-file-drift-check style (no external deps, plain Node ESM).
- Wires it into `.github/workflows/ci.yml`'s `lint-and-test` job, right next to `validate-analytics-taxonomy.mjs` — the established pattern in this repo for this class of mechanical drift check (CI-enforced, not just documented).
- Documents it in `CLAUDE.md`'s `## Testing` section (new "Turbo version pin sync" subsection, alongside the existing Playwright-E2E "run this locally before pushing when X changes" subsection) so it's part of the pre-PR checklist, not just a CI-only gate.
- Adds a row to `scripts/README.md`'s Verification/testing table.

## Acceptance criteria

- [x] Check asserts `apps/web/Dockerfile`'s `npx turbo@<version>` pin matches `package.json`'s `devDependencies.turbo`
- [x] Wired into the pre-PR checklist (CLAUDE.md `## Testing`) and into CI (`ci.yml`)
- [x] Drift is caught before merge, not just at the next `--no-cache` Docker build

## Testing

- `node scripts/check-turbo-version-sync.mjs` against current repo state: `OK: turbo version in sync (2.9.3).` (exit 0)
- Induced-drift test: temporarily bumped `package.json`'s `devDependencies.turbo` to `2.9.4` (Dockerfile left at `2.9.3`) and re-ran — script correctly failed with a clear diff message (exit 1), then `package.json` was restored (confirmed via `git status`/`git diff` — no residual change).
- `npm run typecheck`: passed (4/4 tasks, 3 cached).
- `npm test`: **Tests: 900 passed, 0 skipped, 0 failed** across isolated per-workspace runs — `@lifting-logbook/core` 280/280, `@lifting-logbook/api` 440/440, `@lifting-logbook/web` 180/180, `@lifting-logbook/mobile` no tests, `@lifting-logbook/eslint-rules` 14/14 (all passed; no application/test code is touched by this PR, so no new coverage applies — matches the precedent set by PR #697 for CI/docs-only changes).
  - The full concurrent `npm test` (`turbo run test`) hit pre-existing Windows parallel-load flakiness twice in this session, with two *different* transient symptoms (a `spawn UNKNOWN` jest-worker crash across api/web/mobile, then a Testcontainers/Prisma `migrate deploy` timing failure) — not a regression from this change, confirmed by the 100%-passing isolated runs above. This is a recurrence of the same root cause as the now-closed #567 with symptoms its fix didn't cover; filed and tracked as **[#710](https://github.com/brownm09/lifting-logbook/issues/710)** per the Failure-tracking rule, not fixed here (root-causing concurrent Jest/Testcontainers scheduling on Windows is well outside this PR's scope).
- Suppression / test-integrity / deleted-test / coverage-threshold pre-PR greps: all clean.
- Lockfile-sync check (`npm install --package-lock-only --ignore-scripts && git diff --exit-code package-lock.json`): clean — no dependency changes in this PR.

`/review` ran a single pass on the initial 70-line diff (report: [PR comment](https://github.com/brownm09/lifting-logbook/pull/711#issuecomment-4888112034)). 0 blocking, 2 non-blocking findings — both fixed in this PR:

| Finding | Severity | Fixed in |
|---|---|---|
| Missing Dockerfile/package.json would throw a raw uncaught `ENOENT` instead of the script's own clean `ERROR:` message style | Non-blocking [reliability] | `1c73741` |
| Version-parsing logic (regex + JSON lookup) had no automated test, only manual verification | Non-blocking [maintainability] | `1c73741` |

Both re-verified after the fix: 5/5 new unit tests pass, main script's pass/fail cases still correct, missing-file case now exits 1 with a clean message, `npm run typecheck` and suppression/test-integrity greps still clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,9 @@ jobs:
       - name: Verify turbo version pin sync (Dockerfile ↔ package.json)
         run: node scripts/check-turbo-version-sync.mjs
 
+      - name: Unit test turbo version pin sync parsing
+        run: node --test scripts/check-turbo-version-sync.test.mjs
+
   db-integration:
     name: DB Integration Tests
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,9 @@ jobs:
       - name: Validate analytics taxonomy
         run: node scripts/validate-analytics-taxonomy.mjs
 
+      - name: Verify turbo version pin sync (Dockerfile ↔ package.json)
+        run: node scripts/check-turbo-version-sync.mjs
+
   db-integration:
     name: DB Integration Tests
     runs-on: ubuntu-latest
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -439,6 +439,18 @@ npm run test:e2e -w @lifting-logbook/web
 
 Playwright browsers must be installed once (`npx playwright install chromium`; CI uses the `--with-deps` form for its Linux runner — that flag is a no-op on Windows). The config handles all env vars (dummy Clerk keys, `DEV_AUTH_TOKEN`, `API_URL`) so no manual setup is needed. The motivating incident is [#444](https://github.com/brownm09/lifting-logbook/issues/444) / PR #438, where an aria-label rename (`Back Squat` → `Squat`) passed Jest but broke the smoke spec and was only caught by CI.
 
+### Turbo version pin sync (Dockerfile ↔ package.json)
+
+`apps/web/Dockerfile`'s installer stage pins `npx turbo@<version> prune` to an exact version — `node_modules` is dockerignored ahead of that step, so a bare `npx turbo` has no local install to prefer and would fetch whatever the registry currently tags `latest`. That pin must match the root `package.json`'s `devDependencies.turbo` exactly, or the prune step and the later `npm ci`/build step in the same image run under two different turbo versions with nothing to catch the drift until an uncached Docker build. See [#674](https://github.com/brownm09/lifting-logbook/issues/674) / [#692](https://github.com/brownm09/lifting-logbook/issues/692).
+
+**Run before pushing whenever `apps/web/Dockerfile` or the root `package.json`'s `turbo` devDependency changes:**
+
+```bash
+node scripts/check-turbo-version-sync.mjs
+```
+
+This also runs as a CI step (`ci.yml` → `lint-and-test` → "Verify turbo version pin sync"), so a drifting PR fails either way — running it locally just catches the mismatch before waiting on CI.
+
 ### Coverage Requirements
 
 <!-- When #259 ships: remove the "until then" clause from the frontend row below. Tracked as a checklist item on issue #259. -->
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -52,6 +52,7 @@ Repository automation scripts. Grouped by lifecycle.
 | Script | Purpose |
 |---|---|
 | [`check-deployed-version.sh`](check-deployed-version.sh) | Curls `GET /version` on staging + production, api + web (#672 / #670): reports the deployed git SHA and prints `git log -1` context for each. api requires a Cloud Run identity token (mirrors the deploy pipeline's smoke test); web is `--allow-unauthenticated`. Attempts all 4 checks even if one fails, aggregating and reporting failures at the end. Pass `--env staging`/`--env production` to scope to one environment. See [`docs/runbooks/checking-deployed-version.md`](../docs/runbooks/checking-deployed-version.md). |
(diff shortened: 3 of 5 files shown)


brownm09:
## Review: [chore] Add sync check between Dockerfile's pinned turbo version and package.json

### Summary
Small (70-line), single-purpose PR that does exactly what #692 asks: a dependency-free Node script asserting `apps/web/Dockerfile`'s `npx turbo@<version> prune` pin matches `package.json`'s `devDependencies.turbo`, wired into both CI (`ci.yml`) and the documented pre-PR checklist (`CLAUDE.md`). No blocking findings. Two non-blocking findings, both in the new script and both about failure-mode polish rather than correctness. Appropriately scoped — no need to split.

---

### Non-Blocking Findings

**[reliability]** `scripts/check-turbo-version-sync.mjs:24,36`
`readFileSync(DOCKERFILE, ...)` and `readFileSync(PACKAGE_JSON, ...)` have no existence check or try/catch. If either path is ever missing (e.g. a future rename of `apps/web/Dockerfile` that doesn't update this script), the script throws a raw uncaught `ENOENT` — CI still fails (so nothing is silently missed), but with a bare Node stack trace instead of the clean, actionable `ERROR: ...` messages the script already provides for its other two failure modes (missing pin, missing `devDependencies.turbo`). The sibling script this one mirrors, `validate-analytics-taxonomy.mjs`, already establishes the `existsSync` guard-before-read pattern in this codebase.
**Suggestion:** Wrap the two reads in a try/catch (or `existsSync` check) and emit the same `ERROR: ...`-style message naming the missing path, so all three failure modes read consistently.

**[maintainability]** `scripts/check-turbo-version-sync.mjs`
The script has real branching logic (regex extraction, a JSON key lookup, and a string comparison across two distinct failure paths) with no automated test — only the manual pass/induced-failure verification documented in the PR body. That verification is solid for this PR, but a future change to the Dockerfile's prune-step phrasing or `package.json`'s shape could silently break the regex/lookup with nothing catching it — the one check whose whole job is preventing silent drift could itself silently stop working. This mirrors an existing gap in `validate-analytics-taxonomy.mjs` too, so it's not a new pattern this PR is introducing, just an opportunity to do better.
**Suggestion:** Consider a small `node:test`/Jest unit test against temp-file fixtures (or factor the parsing into an exported pure function a test can call directly) to lock in the match/mismatch behavior going forward.

---

### Findings Disposition
Per ADR-028, each finding above must be fixed in this PR or filed as a tracked issue and linked before merge. Record each disposition (fixed in `<sha>` | filed `#N`) in a "Review findings disposition" section of the PR body before merging — the `pre-merge-findings-gate` hook blocks the merge until that section exists.

---

PR #719, opened 2026-07-06
[chore] Add scheduled --no-cache Docker build CI job to catch Linux-only build breaks

### LINKED ISSUE #693: [chore] Add scheduled --no-cache Docker build CI job to catch Linux-only build breaks
## Problem

#674 (fixed by #691) was a real production-build regression that stayed invisible for a while: nothing in CI exercises a *fresh, uncached* Docker build regularly. Recent \`deploy.yml\` runs stayed green because GHA layer caching (\`cache-from: type=gha,scope=web-deploy-v3\`) kept reusing a stale, still-working layer for the exact build step that broke. \`npm run typecheck\` deliberately can't catch this class of bug either — it resolves workspace packages to source, not \`dist\`, specifically so it stays cacheable without an upstream build (#651/#656) — so a dist/declaration-emit regression like #674's is invisible to it by design.

## Suggested fix

Add a scheduled (e.g. nightly or weekly) CI workflow that runs \`docker build --no-cache\` for both \`apps/web/Dockerfile\` and \`apps/api/Dockerfile\`, independent of the cached deploy pipeline, so a regression like #674 surfaces on a schedule instead of only when someone happens to run an uncached build by hand.

Surfaced via \`/review\` on [PR #691](https://github.com/brownm09/lifting-logbook/pull/691).

brownm09:
## Summary

Adds a scheduled, cache-free Docker build workflow (`.github/workflows/docker-nocache-build.yml`) that rebuilds **both** `apps/web/Dockerfile` and `apps/api/Dockerfile` with `docker build --no-cache` (`push: false`) on a nightly cron (`05:14 UTC`) plus manual `workflow_dispatch`.

It exists because [#674](https://github.com/brownm09/lifting-logbook/issues/674) (fixed by #691) was a real production-build regression that stayed invisible: `deploy.yml`/`staging.yml` build with GHA layer caching (`cache-from/cache-to: type=gha`), so a broken `RUN`/`COPY` layer can hide behind a stale-but-working cached layer, and `npm run typecheck` can't catch this class by design (it resolves workspace packages to source, not `dist`, per #651/#656). A genuinely fresh uncached build re-exercises the `turbo run build` / `.d.ts`-emit step that broke, surfacing that regression within a day.

- `build-web` and `build-api` run **in parallel** so a broken web build can't mask an api regression (or vice-versa).
- On failure, a `report-failure` job opens-or-updates a **single** tracking issue (idempotent create-or-comment, mirroring `required-checks-drift.yml`). Reports on any non-success result (`!= 'success'`), not just `'failure'`, so a genuine timeout/hang still gets tracked (see Review findings disposition below).
- **Least-privilege permissions:** top-level `contents: read`; `issues: write` is scoped to the `report-failure` job only.

> **Provenance:** this work was found complete but **uncommitted** in a local git worktree during a worktree audit/prune — it had never been committed, pushed, or opened as a PR. The branch was fast-forwarded onto current `main` before committing; the file is unchanged from the rescued original.

## Acceptance criteria (from #693)

- [x] Scheduled CI workflow running `docker build --no-cache` for `apps/web/Dockerfile`
- [x] Same for `apps/api/Dockerfile`
- [x] Independent of the cached deploy pipeline (`push: false`, no `cache-from`/`cache-to`)
- [x] Surfaces a regression on a schedule instead of only via a manual uncached build (nightly cron + `workflow_dispatch`)

## Testing

This change touches only `.github/workflows/docker-nocache-build.yml` — no application, package, or build code — so `npm test` / `npm run typecheck` outcomes are unaffected by the diff itself (analogous to the project's *"Refactor / docs only — existing tests must pass"* coverage tier). Verification performed:

- `docker build --no-cache -f apps/web/Dockerfile .` and `-f apps/api/Dockerfile .` run locally — both build clean, directly exercising the workflow's core logic and reconfirming #691's fix holds.
- Workflow YAML parses cleanly (`js-yaml`) — 3 jobs (`build-web`, `build-api`, `report-failure`).
- Pre-commit `turbo run lint` passed: **7 successful, 7 total** (FULL TURBO, cached).
- Pre-PR suppression + test-integrity greps: **clean** (YAML-only diff).
- `npm test`: **Tests: 935 passed, 0 skipped, 0 failed (2m23s)** — full workspace suite (api-client 27, core 280, web 186, api 442), including the DB E2E suite (Docker was available; `LIFTING_SKIP_DB_E2E` was not needed).
- End-to-end validation of the job graph itself (permissions, `needs`/`if` wiring, the failure-reporting path) runs on GitHub Actions post-merge via **`workflow_dispatch`** — a brand-new workflow can't be manually dispatched before it exists on the default branch, so this is deferred by necessity, matching this repo's own precedent in PRs #381 and #553.

No suppressions, skips, or coverage-threshold changes introduced.

Two `/review` passes ran on this PR; all four findings addressed:

**Pass 1:**
1. **[correctness] (blocking)** PR body's Testing section lacked the literal `Tests: N passed, N skipped, N failed` summary line required unconditionally by this project's Test Integrity Policy. → **Fixed** in the PR-body edit (summary line added above).
2. **[reliability] (non-blocking)** `report-failure`'s `if:` only checked `== 'failure'`; a build that hangs and trips the 45-minute `timeout-minutes` bound resolves to `cancelled`, not `failure`, silently skipping the tracking-issue mechanism. → **Fixed** in `ddf2bf7` — both the job-level `if:` and the failing-Dockerfile list now check `!= 'success'`.

**Pass 2** ([review comment](https://github.com/brownm09/lifting-logbook/pull/719#issuecomment-4889200339)):
3. **[reliability] (blocking)** `report-failure`'s `gh label`/`gh issue` calls had no repository context — no `actions/checkout`, no `--repo`, no `GH_REPO` — so `gh` could not resolve the target repo and the tracking-issue mechanism (the job's entire purpose) would fail exactly when a build breaks. Both sibling `gh`-using workflows (`required-checks-drift.yml`, `merge-ready-digest.yml`) provide repo context; this job provided none. → **Fixed** in `d395260` — `GH_REPO: ${{ github.repository }}` added to the `report-failure` step env.
4. **[reliability] (non-blocking)** No retry on the build step; a fresh runner pulls the `node:20.11.1-alpine` base image every run, so a transient pull failure could file a false-positive tracking issue. → **Fixed** in `d395260` — the "no wretry" rationale comment now notes the idempotent create-or-comment caps a transient pull failure at one comment, so a build-step retry isn't warranted.

Closes #693

### DIFF
--- a/.github/workflows/docker-nocache-build.yml
+++ b/.github/workflows/docker-nocache-build.yml
@@ -1,0 +1,164 @@
+name: Docker No-Cache Build
+
+# Scheduled, cache-free rebuild of both production Dockerfiles. Unlike every other
+# workflow in this repo that builds a Docker image, this one never pushes anywhere —
+# its only output is pass/fail. It exists because #674 (fixed by #691) was a real
+# production-build regression that stayed invisible for a while: `deploy.yml` and
+# `staging.yml` both build with GHA layer caching (cache-from/cache-to: type=gha),
+# so a broken RUN/COPY layer can hide behind a stale-but-working cached layer
+# indefinitely. `npm run typecheck` deliberately can't catch this class of bug
+# either — it resolves workspace packages to source, not dist, specifically so it
+# stays cacheable without an upstream build (#651/#656) — so a dist/declaration-emit
+# regression like #674's is invisible to it by design. This workflow runs both
+# Dockerfiles through a genuinely fresh `docker build --no-cache` on a schedule, so
+# a similar regression surfaces within a day instead of only via manual
+# investigation (as #674 was, per #691). See #693.
+
+on:
+  schedule:
+    # Nightly, not weekly: the whole point of this workflow is to catch a build
+    # regression sooner than "someone happens to run an uncached build by hand"
+    # (#693) — a week is a long time to leave that invisible. 05:14 UTC is
+    # off-peak and avoids top-of-hour scheduler congestion (same rationale as
+    # required-checks-drift.yml).
+    - cron: '14 5 * * *'
+  workflow_dispatch:
+
+# Prevents a manual dispatch from racing the nightly tick and filing two issues
+# for the same failure. cancel-in-progress: false so an in-flight build always
+# finishes and reports, rather than being cancelled mid-build by an overlapping
+# trigger (same rationale as required-checks-drift.yml).
+concurrency:
+  group: docker-nocache-build
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  # ── Build apps/web ──────────────────────────────────────────────────────────
+  build-web:
+    name: Build apps/web (no cache)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      # push: false, no cache-from/cache-to: this build's only purpose is to prove
+      # the Dockerfile still builds clean, so no registry auth is needed at all.
+      # no-cache: true forces every RUN/COPY instruction to re-execute from
+      # scratch (this is what actually re-exercises the `turbo run build` / .d.ts
+      # emission step that broke in #674) — it does not force re-pulling the
+      # pinned node:20.11.1-alpine base image if the runner already has that
+      # exact tag cached, which is fine: the regression this guards against lives
+      # in the RUN layers, not the base image. No wretry wrapper (unlike
+      # deploy.yml/staging.yml): that retry exists solely to absorb transient
+      # Artifact Registry 504s on registry push, which cannot happen when
+      # push: false.
+      - name: Build apps/web image (no cache, not pushed)
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: apps/web/Dockerfile
+          push: false
+          no-cache: true
+          tags: lifting-logbook-web:nocache-check
+
+  # ── Build apps/api ──────────────────────────────────────────────────────────
+  build-api:
+    name: Build apps/api (no cache)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build apps/api image (no cache, not pushed)
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: apps/api/Dockerfile
+          push: false
+          no-cache: true
+          tags: lifting-logbook-api:nocache-check
+
+  # ── Report failure ───────────────────────────────────────────────────────────
+  # Independent jobs above run in parallel so a broken apps/web build can never
+  # mask an apps/api regression (or vice versa). This job aggregates both
+  # results and files/updates a single tracking issue — mirrors
+  # required-checks-drift.yml's idempotent create-or-comment pattern.
+  report-failure:
+    name: Report build failure
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    needs: [build-web, build-api]
+    if: always() && (needs.build-web.result == 'failure' || needs.build-api.result == 'failure')
+    permissions:
+      issues: write
+    steps:
+      - name: Open or update failure issue
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+          gh label create docker-nocache-build-failure \
+            --description "A scheduled --no-cache Docker build failed for apps/web and/or apps/api" \
+            --color B60205 \
+            --force
+
+          FAILED=""
+          if [ "${{ needs.build-web.result }}" = "failure" ]; then
+            FAILED="${FAILED}- \`apps/web/Dockerfile\`\n"
+          fi
+          if [ "${{ needs.build-api.result }}" = "failure" ]; then
+            FAILED="${FAILED}- \`apps/api/Dockerfile\`\n"
+          fi
+
+          {
+            echo "# Docker no-cache build failing"
+            echo ""
+            echo "Run: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+            echo ""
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [chore] Add scheduled --no-cache Docker build CI job to catch Linux-only build breaks

### Summary
**Needs one revision before merge.** The workflow is well-structured, well-documented, and correctly scoped (least-privilege `permissions`, parallel build jobs, idempotent issue reporting, action versions consistent with the repo). But the `report-failure` job invokes `gh issue`/`gh label` with **no repository context** — no `actions/checkout`, no `--repo`, no `GH_REPO` — which the two sibling `gh`-using workflows in this repo both provide. As written, the failure-reporting path will error out precisely when a build breaks, so the tracking issue never gets filed. That single reliability finding is the only blocker.

---

### Blocking Findings

**[reliability]** `.github/workflows/docker-nocache-build.yml:104-159` (the `report-failure` job)
The job runs `gh label create`, `gh issue list`, `gh issue comment`, and `gh issue create` with only `GH_TOKEN` set — no `actions/checkout` step, no `--repo`, and no `GH_REPO`. `gh` resolves the target repo from (in order) `--repo` → `GH_REPO` → the git remote of the working directory; with none present and no checkout, it fails with an "unable to determine repository"–class error.
Why it matters here: this is the one job whose entire purpose is to surface a caught regression. It only runs when a build has already failed, so the defect stays invisible until a real incident — at which point the auto-filed tracking issue (the workflow's designed output) silently never appears. Both sibling workflows confirm the required pattern: `required-checks-drift.yml` checks out at step 1 (`- uses: actions/checkout@v5`) before its bare `gh issue`/`gh label` calls, and `merge-ready-digest.yml` both checks out **and** passes `--repo "${{ github.repository }}"` on every `gh` call.
**Fix:** Give the job a repo context. Minimal option — add `GH_REPO: ${{ github.repository }}` to the step's `env:` alongside `GH_TOKEN` (resolves all four `gh` calls at once, no checkout needed). Convention-matching option — add `- uses: actions/checkout@v5` as the first step of `report-failure`, mirroring `required-checks-drift.yml` (which this workflow's comments say it mirrors).

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/docker-nocache-build.yml:61-68, 81-88` (the build steps)
The build steps have no retry. The inline comment thoughtfully reasons about *why no `wretry`* — but only for the registry-**push** path ("transient Artifact Registry 504s on registry push, which cannot happen when `push: false`"). It doesn't cover the base-image **pull**: a fresh `ubuntu-latest` runner has nothing cached, so every run pulls `node:20.11.1-alpine` from Docker Hub, and a transient registry/network hiccup there fails the build.
Why it matters here: a transient pull failure is indistinguishable from a genuine Dockerfile break, so it files (or comments on) a "build failing" tracking issue. Over a nightly cadence that adds false-positive noise and erodes trust in the signal the workflow exists to provide.
**Suggestion:** Either wrap the build step in a retry (e.g. `nick-fields/retry`) scoped to transient/network errors, or extend the existing "no retry" comment to explicitly accept base-image-pull flakiness as tolerable given the idempotent create-or-comment. Fine to defer to a follow-up issue rather than expand this rescue PR.

---

### Style Notes
Comment density is high but genuinely earns its keep — each block explains a non-obvious *why* (scheduling rationale, `no-cache` semantics vs. base-image caching, the deliberate absence of a push-retry) and cross-references the motivating issues (#674/#691/#651/#656). No changes suggested.

---

### Findings Disposition
Per ADR-028, both findings must be fixed-in-PR or filed-and-linked before merge, recorded in a "Review findings disposition" section of the PR body (the `pre-merge-findings-gate` hook enforces this):
- **Blocking [reliability]** — repo context for `report-failure`'s `gh` calls: fix in-PR (add `GH_REPO` env or a checkout step).
- **Non-blocking [reliability]** — build-step retry / comment: fix in-PR or file a linked follow-up.

Note on gates: this is a CI-config-only change (a new GitHub Actions workflow, no application/package/build code), so the Test Coverage and Test Integrity gates are not applicable and the missing `Tests:` summary line is expected — the PR body documents the config-only rationale and the verification actually performed (YAML parse, pre-commit `turbo run lint` 7/7). No suppressions, skips, or threshold changes.

---

brownm09:
## Review: [chore] Add scheduled --no-cache Docker build CI job to catch Linux-only build breaks

### Summary
This PR is well-scoped and ready to merge after two small fixes. It adds a nightly `docker build --no-cache` check for both `apps/web/Dockerfile` and `apps/api/Dockerfile`, correctly avoids any registry cache or push, and mirrors this repo's existing `required-checks-drift.yml` pattern for idempotent failure-issue tracking. The single highest-severity finding is a missing `Tests: N passed...` summary line in the PR body, required unconditionally by this project's Test Integrity Policy even for a YAML-only diff. At 164 added lines (one new file), the PR is appropriately scoped — no split needed.

---

### Blocking Findings

**[correctness]** `PR body — Testing section`
The Testing section explains why `npm test`/`npm run typecheck` are unaffected by this diff, but doesn't include the literal `Tests: N passed, N skipped, N failed (duration)` line this project's Test Integrity Policy (Rule 2) requires verbatim in every PR's Testing section, regardless of whether the diff is expected to affect test outcomes. CI's own "Lint & Test" check already passed on this exact commit, so the underlying risk is low, but the policy is unconditional and the fix is cheap.
**Fix:** Run `npm test` (with `LIFTING_SKIP_DB_E2E=1`, citing #394, since this diff touches no DB/repository code) and paste the resulting summary line into the PR body.

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/docker-nocache-build.yml:100`
`report-failure` only fires when `needs.build-web.result == 'failure'` or `needs.build-api.result == 'failure'`. A job that legitimately hangs and hits the 45-minute `timeout-minutes` bound is reported by GitHub Actions as `cancelled`, not `failure` — so a genuine timeout would silently skip the tracking-issue mechanism entirely, which is exactly the "no one notices until manual investigation" failure mode this workflow exists to prevent.
**Suggestion:** Broaden the condition to also catch `cancelled` (e.g. `needs.build-web.result != 'success' || needs.build-api.result != 'success'` — these two jobs have no `if:` of their own, so they can only resolve to success/failure/cancelled).

---

### Findings Disposition
*(to be completed after fixes are applied)*

---

Issue #677, opened 2026-07-04
App-shell, program-length, and settings-discoverability consistency pass


brownm09:
## Summary

A walkthrough of the app (onboarding → Leangains → `/cycle/1`) surfaced a cluster of user-facing inconsistencies. Each was traced to a concrete root cause. Common threads:

- **Navigation and settings were built page-by-page with no shared shell** — the nav only exists on `/cycle/*`, so `/programs` is a dead-end and there is no persistent way out of most screens.
- **Program "length" has three disagreeing sources of truth** — the onboarding card advertises Leangains as 12 weeks, the preset defines a 1-week repeating block, and the Program Plan derives duration from `Math.max(...spec.week)` = 1.
- **Configured values exist in the model but are ignored by the UI** — the per-lift increment and weight unit are known but not applied in the Training Maxes editor / display.

This is the top-level tracking issue for a decomposed consistency pass. Each PR closes its own sub-issue; this issue closes when all sub-PRs merge.

## Diagnosis (verified against code)

| # | Symptom | Root cause |
|---|---|---|
| 1 | Leangains "Program Plan" shows **1 total week** (12 advertised) | Duration = `Math.max(...specs.map(s => s.week))`; Leangains preset only defines `week:1` rows. `weeks:12` lives only in web onboarding copy. `estCompletion = durationWeeks*7`; the 1-week fallback fabricates a **"Test"** phase. |
| 2 | Training-max +/- steps by 1, not the configured increment | Bare `<input type="number">` with no `step` → browser default step = 1; per-lift `increment` and global `defaultWeightIncrement` ignored. |
| 3 | Import unreachable after onboarding | `/import` wizard exists but nothing links to it (orphan route). |
| 4 | Nav links have no visible style | Bare `<a>` tags, no class; global reset strips affordance. |
| 5 | "Settings" only opens Training Maxes | Link hardcoded to `/settings/training-maxes`; 3 sibling settings pages unlinked; no `/settings` index. |
| 7 | No way out of Programs / no persistent nav | Nav rendered by the cycle layout only → exists only on `/cycle/*`. |
| 6 | Assorted | 5/3/1 preset only defines 3 weeks → `PHASE_MAP_12` dead code; units hardcoded `lbs`; "Edit Program" disabled w/o affordance; History swallows fetch errors; back-link/terminology drift; missing aria on nav. |

## Decomposition (sub-issues)

[...]

PR #733, opened 2026-07-07
[feat] Add global lbs/kg weight-unit preference

### LINKED ISSUE #682: Weight-unit (lbs/kg) preference
Part of #677. Largest feature in the series.

## Problem
Weight units are hardcoded to `lbs` across the UI (dashboard, import, strength-goals default). `UserSettingsResponse` (`packages/types/src/api.ts`) has no global `unit` field — only per-record units exist on training-max rows / history (`tmUnit`). Users cannot switch to kg.

## Scope
- Add a global `unit` (`'lbs' | 'kg'`) to `UserSettingsResponse` / `UpdateUserSettingsRequest` + persistence + a toggle in the Settings hub (PR 2).
- Thread the preference through display sites that currently hardcode `lbs`; add conversion at the display boundary.
- Reconcile the global preference with per-record storage (`tmUnit`): decide display-conversion vs. stored value and document it.

## Acceptance criteria
- [ ] Settings exposes a lbs/kg toggle that round-trips through the API
- [ ] Weights render/convert consistently across dashboard, plan, history, training maxes
- [ ] Settings round-trip test; display test asserting kg rendering + conversion

Addresses the units portion of symptom #6 from #677.

brownm09:
## Summary

PR 5 of 6 in the [#677](https://github.com/brownm09/lifting-logbook/issues/677) consistency pass — adds a global lbs/kg weight-unit display preference.

- **Persistence:** `UserSettings.unit` (nullable `String`, mirrors the `defaultWeightIncrement` pattern — migration `20260707140000_add_user_settings_unit`), `UpdateSettingsDto` validation (`@IsIn(['lbs','kg'])`), `UserSettingsResponse.unit` / `UpdateUserSettingsRequest.unit` in `packages/types`.
- **Settings UI:** new `/settings/units` page (mirrors the `weight-rounding` page/form/action triple exactly), added to the hub's `SETTINGS_SECTIONS`.
- **Conversion utility:** `packages/core`'s new `convertWeight`/`formatWeight` (none existed anywhere in the repo before this PR) — full-precision conversion, 2-decimal display rounding only.
- **Threaded through:** dashboard (`CycleDashboardGrid`), workout plan (`CollapsibleLiftList`), lift/TM history (`HistoryTabs`, `LiftHistoryFilters`, `MaxHistory`), training-max display + import/strength-goals unit defaults.

## Design decisions

**Storage stays canonical lbs everywhere it already was.** The preference is a pure display-time conversion layer — it never rounds or rewrites a stored value, per [`docs/standards/training-max-precision.md`](docs/standards/training-max-precision.md). Training-max weights in particular have **no real per-record unit column** in Postgres today — the API hardcodes `unit: 'lbs'` at the response-mapping boundary (`apps/api/src/programs/mappers.ts`). So the "reconcile the global preference with per-record storage (`tmUnit`)" ask from the issue resolves to: treat each record's own `unit` field as the conversion source, convert to the preference for display, done.

**Two genuinely write-path entry points are deliberately left unconverted:** `TrainingMaxesForm`'s weight input and `ImportWizard`'s editable training-max review rows. Both are directly-known-value entry points into a backend that has no real per-record unit storage for training maxes (whatever `weight` arrives is stored as-is, implicitly lbs) — converting the editable value would either silently corrupt a saved max (if an unsubmitted converted value got included in a save) or require a larger backend change to genuinely persist per-record units, which is out of this issue's scope. Both get a read-only "≈ X kg" hint next to the input instead, so the preference is still visible without touching what's actually stored.

**`StrengthGoalsForm` is different** — `StrengthGoalResponse.unit` and `BodyWeightResponse.unit` are genuinely persisted (not coerced), so the per-row unit `<select>` (already existed, already fully user-overridable) now *defaults* to the global preference instead of a hardcoded `'lbs'`, and body-weight entry now saves using the preference as the value's real unit. Making kg a reachable default here exposed a latent bug: `computeProgress` was comparing a training max (always lbs) directly against a body-weight/target value that could now genuinely be in kg, with no conversion — e.g. a 100 lbs TM against a 45.36 kg bodyweight target would have shown ~220% instead of the correct 100%. Fixed as part of this PR (`computeProgress` now converts the training max into whichever unit it's being compared against) since the bug was latent until this PR made kg reachable through normal use; see `StrengthGoalsForm.test.tsx` for a regression test that pins the correct value.

**Deferred:** `WorkoutLogger.tsx` (the live in-workout logging screen). Not named in the issue's problem statement or acceptance criteria (which name dashboard/plan/history/training-maxes/import/strength-goals), and it mixes display conversion with entry semantics in a way that deserved its own focused pass rather than being rushed into an already-large PR. Filing a follow-up issue/tile at merge.

## Acceptance criteria

- [x] Settings exposes a lbs/kg toggle that round-trips through the API
- [x] Weights render/convert consistently across dashboard, plan, history, training maxes
- [x] Settings round-trip test; display test asserting kg rendering + conversion

## Testing

`npm test`: **1003 passed, 0 skipped, 0 failed** across all workspaces (core 309, api-client 27, web 214, api 453 — including a real-Postgres DB E2E case for the new `unit` field).
`npm run typecheck`: clean.
`npm run lint`: clean (including the custom `no-uncovered-error-fallback` rule — two `staging.spec.ts` `file:line` references needed updating since my edits shifted the line numbers of pre-existing, unrelated `.catch()` fallbacks they pointed at).
`npx turbo run build --filter=@lifting-logbook/api`: clean (Prisma migration applies correctly).
`npm run test:e2e -w @lifting-logbook/web`: **22 passed** (Playwright) — new coverage for the Units settings section, hub nav now asserts the 5th "Units" tab.

New tests: `packages/core/tests/core/utils/weightUnit.test.ts` (conversion math), `apps/web/.../settings/units/{page,UnitForm}.test.tsx` (settings round-trip), `apps/web/.../history/HistoryTabs.test.tsx` (display conversion — kg rendering asserted against exact expected values), `apps/web/.../strength-goals/StrengthGoalsForm.test.tsx` (cross-unit progress-calculation regression test), plus a new DB E2E case in `programs.db.e2e.spec.ts`.

No suppressions added (grep clean after replacing a `row.unit as WeightUnit` cast with a real runtime `parseUnit()` guard, mirroring the existing `parseSchedule()` pattern for the JSONB `workoutSchedule` column). No test-integrity violations, no deleted tests.

## Risk-dimension audit

1. **Testing** — see above.
2. **Observability** — no new endpoint; existing `PATCH /users/me/settings` gets one more field, covered by existing nestjs-pino auto-logging + OTel HTTP-level spans. No new instrumentation needed.
3. **Security** — N/A; non-sensitive user preference. DTO whitelist validation (`@IsIn(['lbs','kg'])`) rejects any other value.
4. **Resilience/rollback** — nullable column, no DB-level default (mirrors `defaultWeightIncrement`); fully reversible by dropping the migration. Existing rows get `unit = NULL` → falls back to `'lbs'` in code, so no existing user's display silently changes.
5. **Performance** — negligible; pure in-memory arithmetic.
6. **Data integrity & migrations** — single nullable column add; RLS policy is row-scoped so unaffected.
7. **Accessibility** — new `<select id="weight-unit-select">` has an associated `<label>`; e2e asserts it's reachable via `getByLabel`.

Closes #682

---

`/review` (2 parallel Opus subagents) ran on this PR — all findings fixed in-PR, none filed or deferred:

- **[correctness]** `formatWeight` rounded the same-unit passthrough (dropped precision on directly-known values, contra `training-max-precision.md`) — **fixed in `60c4440`**.
- **[performance]** `getPreferredUnit()` bypassed the cached `getUserSettings`, causing a duplicate `/users/me/settings` GET per render — **fixed in `60c4440`**.
- **[maintainability]** `computeProgress` 8 positional args (unit-transposition hazard) → options object — **fixed in `60c4440`**.
- **[maintainability]** `RowState.unit` `string`→`WeightUnit` (2 casts dropped); `MaxHistory.unit` made required — **fixed in `60c4440`**.

Full review: https://github.com/brownm09/lifting-logbook/pull/733#issuecomment-4906203657

### DIFF
--- a/apps/api/prisma/migrations/20260707140000_add_user_settings_unit/migration.sql
+++ b/apps/api/prisma/migrations/20260707140000_add_user_settings_unit/migration.sql
@@ -1,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "user_settings" ADD COLUMN     "unit" TEXT;
--- a/apps/api/prisma/schema.prisma
+++ b/apps/api/prisma/schema.prisma
@@ -171,6 +171,7 @@ model UserSettings {
   activeProgram          String?
   workoutSchedule        Json?
   defaultWeightIncrement Float?
+  unit                   String?
 
   @@map("user_settings")
 }
--- a/apps/api/src/adapters/in-memory/user-settings.adapter.ts
+++ b/apps/api/src/adapters/in-memory/user-settings.adapter.ts
@@ -1,16 +1,18 @@
-import { UserSettingsResponse, UserWorkoutSchedule } from '@lifting-logbook/types';
+import { UserSettingsResponse, UserWorkoutSchedule, WeightUnit } from '@lifting-logbook/types';
 import { IUserSettingsRepository } from '../../ports/IUserSettingsRepository';
 
 export class InMemoryUserSettingsRepository implements IUserSettingsRepository {
   private activeProgram: string | null = null;
   private workoutSchedule: UserWorkoutSchedule | null = null;
   private defaultWeightIncrement: number | null = null;
+  private unit: WeightUnit | null = null;
 
   async getSettings(): Promise<UserSettingsResponse> {
     return {
       activeProgram: this.activeProgram,
       workoutSchedule: this.workoutSchedule,
       defaultWeightIncrement: this.defaultWeightIncrement,
+      unit: this.unit,
     };
   }
 
@@ -24,5 +26,9 @@ export class InMemoryUserSettingsRepository implements IUserSettingsRepository {
 
   setDefaultWeightIncrement(increment: number | null): void {
     this.defaultWeightIncrement = increment;
+  }
+
+  setUnit(unit: WeightUnit | null): void {
+    this.unit = unit;
   }
 }
--- a/apps/api/src/ports/ports.compile-test.ts
+++ b/apps/api/src/ports/ports.compile-test.ts
@@ -178,7 +178,7 @@ const _workoutSkipOverrideRepo: IWorkoutSkipOverrideRepository = {
 
 const _userSettingsRepo: IUserSettingsRepository = {
   getSettings: () =>
-    Promise.resolve({ activeProgram: null, workoutSchedule: null, defaultWeightIncrement: null }),
+    Promise.resolve({ activeProgram: null, workoutSchedule: null, defaultWeightIncrement: null, unit: null }),
 };
 
 const _customLiftRepo: ICustomLiftRepository = {
--- a/apps/api/src/programs/programs.db.e2e.spec.ts
+++ b/apps/api/src/programs/programs.db.e2e.spec.ts
@@ -1071,6 +1071,25 @@ describeOrSkip('Programs HTTP (e2e, PrismaRepositoryFactory)', () => {
       expect(getRes.json()).toMatchObject({ activeProgram: SEED_PROGRAM });
     });
 
+    it('PATCH /users/me/settings updates unit and persists to DB', async () => {
+      const injectRaw = app.getHttpAdapter().getInstance().inject.bind(app.getHttpAdapter().getInstance());
+
+      const patchRes = await injectRaw({
+        method: 'PATCH',
+        url: '/users/me/settings',
+        headers: { 'content-type': 'application/json', ...AS_SETT },
+        payload: JSON.stringify({ unit: 'kg' }),
+      });
+      expect(patchRes.statusCode).toBe(200);
+      expect(patchRes.json()).toMatchObject({ unit: 'kg' });
+
+      const row = await prisma.userSettings.findFirst({ where: { userId: USER_SETT } });
+      expect(row?.unit).toBe('kg');
+
+      const getRes = await injectRaw({ method: 'GET', url: '/users/me/settings', headers: AS_SETT });
+      expect(getRes.json()).toMatchObject({ unit: 'kg' });
+    });
+
     it('user isolation — settings are scoped to userId', async () => {
       const injectRaw = app.getHttpAdapter().getInstance().inject.bind(app.getHttpAdapter().getInstance());
       const AS_OTHER = { authorization: `Bearer ${USER_SETT_OTHER}` };
--- a/apps/api/src/user-settings/update-settings.dto.ts
+++ b/apps/api/src/user-settings/update-settings.dto.ts
@@ -19,8 +19,13 @@ import {
   ValidatorConstraintInterface,
   ValidationArguments,
 } from 'class-validator';
-import { SCHEDULE_LIMITS, WEIGHT_INCREMENT_OPTIONS, isValidSchedule } from '@lifting-logbook/types';
-import type { UserWorkoutSchedule } from '@lifting-logbook/types';
+import {
+  SCHEDULE_LIMITS,
+  WEIGHT_INCREMENT_OPTIONS,
+  WEIGHT_UNIT_OPTIONS,
+  isValidSchedule,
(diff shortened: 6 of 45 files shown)


brownm09:
## Review: [feat] Add global lbs/kg weight-unit preference

### Summary
Merge with minor changes — **now addressed**. Two parallel Opus review passes (correctness/security + reliability/perf/maintainability) over the 848-line diff surfaced one correctness finding and one clear performance finding, plus three low-severity maintainability items. All were fixed in-PR in `60c4440` (this comment postdates that commit). The highest-severity finding was `formatWeight` rounding a directly-known value shown in its own unit, contradicting the `training-max-precision.md` standard the PR itself cites. The diff is appropriately scoped for one feature; no split needed.

---

### Blocking Findings

**[correctness]** `packages/core/src/utils/weightUnit.ts:29` (original)
`formatWeight` unconditionally rounded to 2 decimals, including the `from === to` passthrough where no conversion happens. It is called to display directly-known training-max values in their native unit (e.g. `TM: {formatWeight(tm, 'lbs', unit)}`), and `docs/standards/training-max-precision.md` category 1 requires directly-known values to display at **full precision** — a 316.875 lbs TM (0.625 plate increment) would have rendered "316.88 lbs".
**Fix (done, `60c4440`):** Short-circuit `from === to` to return the raw value unrounded; extracted a shared `roundToDisplay()` applied only to genuine conversions. Test fixture bumped to a 3-decimal value so the "does not round native-unit value" invariant is actually exercised.

---

### Non-Blocking Findings

**[performance]** `apps/web/lib/preferences.ts:15` (original)
`getPreferredUnit()` called the raw `fetchUserSettings` (`cache: 'no-store'`) instead of the repo's React-`cache()`-wrapped `getUserSettings` in `active-program.ts`. Since nearly every page also calls `getActiveProgram()` (which reads settings through the cached path), each render issued a **duplicate** `/users/me/settings` GET — provably two identical fetches in the same `Promise.all` on the training-maxes page.
**Fix (done, `60c4440`):** `getPreferredUnit()` now reads through the cached `getUserSettings`, collapsing every page back to one settings round-trip per render (and making the PR's "mirrors the `defaultWeightIncrement` pattern" claim literally true).

**[maintainability]** `apps/web/.../strength-goals/StrengthGoalsForm.tsx` — `computeProgress`
Grew to 8 positional params, 3 of them mutually-assignable `WeightUnit`, so a transposition would type-check and fail silently as a wrong progress number — the exact cross-unit bug class this function was rewritten to fix.
**Fix (done, `60c4440`):** Converted to a single named options object (`ProgressArgs`), making transposition impossible.

**[maintainability]** `apps/web/.../training-maxes/{TrainingMaxesForm,MaxHistory}.tsx`
`RowState.unit` was typed `string` (forcing two `as WeightUnit` casts), and `MaxHistory.unit` was optional-with-`'lbs'`-default despite its only caller always passing it.
**Fix (done, `60c4440`):** Tightened `RowState.unit` to `WeightUnit` (both casts dropped); made `MaxHistory.unit` required. `TrainingMaxesForm`/`ImportWizard` keep their optional `unit?` default — both have a genuine test-compat constraint (3 and 7 existing render sites) and already carry explanatory JSDoc, which the reviewing pass endorsed as acceptable-as-is.

---

### Findings Disposition
All findings fixed in-PR in `60c4440` — none filed or deferred. See the "Review findings disposition" section added to the PR body. Re-verified after the fix: 1005 unit/integration tests + 22 Playwright e2e pass; typecheck and lint clean.

---

*Reviewed by `claude-opus-4-8` (2 parallel subagents) via Claude Code*

PR #755, opened 2026-07-08
[feat] Program editor: group exercises into workout days

### LINKED ISSUE #751: [feat] Program editor: group exercises into workout days
## Problem

The program editor (**Programs → My Programs → New / Clone / Edit**) has no way to group exercises into workouts. It groups exercises only by **week**, so every program it produces collapses into a **single, undifferentiated workout per week** (training frequency always computes to 1). The disabled "Edit Program — Coming soon" button on the cycle program page reflects the same gap.

## Context

Workout grouping is already first-class everywhere *except* the editor:

- The program spec key is `(week, offset, lift, order)`, where `offset` is the workout-day within a week. The storage (`CustomProgramSpec`, unique on `(programId, week, offset, lift, order)`) and the create/update API DTO already accept arbitrary offsets and impose **no lift-uniqueness constraint**.
- The read side already groups by `(week, offset)`: `buildWorkoutDays()` drives the Cycle Dashboard's Week → Workout card → lifts → sets, plus the workout-detail and logger views; `deriveProgramSummary()` computes frequency from distinct offsets.

The editor simply never exposes the workout-day (`offset`) dimension — it hardcodes `offset: 0`.

## Goal

Let the user build **workout days** and place exercises into them, then save a program that the existing read side renders as distinct workouts. Crucially, a single lift must be placeable on **multiple workout days within the same week** (e.g. Squat on Day 1 *and* Day 3) — so the model is exercise-*instance*-based, not one-day-per-lift.

## Acceptance criteria

- [ ] The editor lets the user add and remove workout days.
- [ ] The editor lets the user add exercises to a specific day; the same lift can be placed on multiple days (and repeated within a day).
- [ ] Per-week loading params (sets / reps / AMRAP / increment / warm-up %) are editable per exercise instance.
- [ ] A saved custom program renders as multiple distinct workout cards per week on the Cycle Dashboard, and the program summary's frequency reflects the number of days.
- [ ] Existing custom programs (one implicit workout per week) still load and re-save correctly — backward compatible.
- [ ] Cloning a preset pre-populates that preset's workout-day structure.
- [ ] No backend or database schema change is required.
- [ ] New tests cover the day/instance ↔ spec mapping (incl. same-lift-on-two-days), and a Playwright spec drives the editor; the `apps/web` e2e suite passes.

## Out of scope (separate follow-up issues)

1. API hardening — add `@ArrayMaxSize` to the specs array and a P2002 handler in the custom-programs repository (defense in depth).
2. Schedule/frequency-mismatch handling in cycle generation when a program's frequency differs from the user's schedule days-per-week.

brownm09:
## Summary

The program editor (**Programs → My Programs → New / Clone / Edit**) previously grouped exercises only by week and hardcoded `offset: 0`, so every custom program collapsed into a single, undifferentiated workout per week. This replaces the flat lift-checkbox + `(week, lift)` table with an **instance-based workout-day builder**:

- Build workout days (Day 1..N) with **Add Day / Remove Day**.
- Add exercises to a specific day; the **same lift can be placed on multiple days** — or repeated within a day — because an instance is keyed by `(offset, order)`, not lift name.
- Per-week sets / reps / AMRAP / increment / warm-up are edited per instance.
- On save, the model expands to `CustomProgramSpecRow[]` with `offset` = day index and `order` = within-day position — exactly the shape `buildWorkoutDays()` already groups into workouts, so the Cycle Dashboard and workout views render the days with no further change.

**No backend, Prisma, or shared-type changes** — the spec key already carried `(week, offset, lift, order)`, and the create/update DTO already accepts arbitrary `offset`/`order` with no lift-uniqueness constraint.

## Acceptance criteria

- [x] Add and remove workout days.
- [x] Add exercises to a day; the same lift can be on multiple days (and repeated within a day).
- [x] Per-week loading params editable per instance.
- [x] Saved programs render as distinct workout cards per week on the Cycle Dashboard (read-side mappers unchanged + exercised via the e2e save path); frequency reflects the number of days.
- [x] Existing custom programs (one implicit workout/week) load as a single Day 1 and re-save unchanged (backward-compat unit test).
- [x] Cloning a preset pre-populates its workout-day structure (`templateSeedSpecs`; covered for leangains).
- [x] No backend or schema change.
- [x] Tests cover the day/instance ↔ spec mapping (incl. same-lift-on-two-days) and a Playwright editor spec; the `apps/web` e2e suite passes.

## Changes

| File | Change |
|---|---|
| `apps/web/app/(authed)/programs/programSpecMapping.ts` (new) | Pure `daysFromSpecs` / `specsFromDays` mappers + types; no React/Next imports, unit-tested directly. |
| `apps/web/app/(authed)/programs/ProgramEditor.tsx` | Rewrite to the day/instance builder: per-week mini-tables, add/remove day + instance, client caps (≤7 days, ≤12/day), `role="group"` + per-control aria-labels. |
| `apps/web/app/(authed)/programs/programs.module.css` | New day/instance styles; removed now-dead `.liftCheckboxList` / `.liftCheckbox` / `.specTableWrapper`. |
| `apps/web/lib/programs.ts` | New `templateSeedSpecs()` clone-seed helper. |
| `apps/web/e2e/mock-api.mjs` | Added `POST /programs/custom` + `PUT /programs/custom/:id` echo handlers. |
| `apps/web/e2e/programs-editor.spec.ts` (new) | Playwright: build two days, place a lift on both, save. |
| `apps/web/app/(authed)/programs/ProgramEditor.mapping.test.ts` (new) / `ProgramEditor.increment.test.tsx` | Mapper unit tests; increment test updated for the new UI. |
| `apps/api/src/custom-programs/create-custom-program.dto.spec.ts` | Added offset/order + multi-instance validation coverage. |

## Testing

- `npm run typecheck` — pass (blocking CI gate).
- `npx turbo run lint` (web + api) — pass.
- `npm test -w @lifting-logbook/web` — **Tests: 244 passed, 0 skipped, 0 failed** (40 suites), including `ProgramEditor.mapping.test.ts` (12 — +1 regression test for the review's blocking data-loss finding) and the updated increment test.
- `LIFTING_SKIP_DB_E2E=1 npm test -w @lifting-logbook/api -- create-custom-program.dto` — **Tests: 12 passed, 0 skipped, 0 failed** (5 new). DB E2E skipped because this diff touches no DB code (no `apps/api/prisma/` or repository changes) — the documented escape hatch, cited per [#394](https://github.com/brownm09/lifting-logbook/issues/394).
- `npm run test:e2e -w @lifting-logbook/web -- programs-editor` — **1 passed** (Day 1 = Back Squat + Bench Press, Day 2 = Back Squat, then Save).

Coverage gate: new behavior is covered (mappers, DTO, e2e). No tests skipped or deleted; no coverage thresholds changed.

## Suppression policy note

`apps/web/e2e/mock-api.mjs` adds four `?? null` (`description` and `baseTemplate` in the POST and PUT echo responses — the PUT's is `body.baseTemplate ?? existing?.baseTemplate ?? null` after the review fix that made it stateful). These are **data defaults in the plain-JS Playwright mock**, mirroring the real Fastify API's `string | null` response fields when a request omits them — not type-error suppressions. No `@ts-ignore` / `@ts-expect-error` / `eslint-disable` / non-null assertions were added (the old editor's `checkboxes[0]!` was removed).

## Notes for reviewers / follow-ups (out of scope)

- The **"Save & Switch to This Program"** button uses `router.push` + `router.refresh` (unchanged from the previous editor). In local dev e2e the server actions succeed (`createCustomProgram`, `switchProgram`, `GET /cycle/1` all 200) but the client URL did not commit, so the e2e asserts on plain **Save**. Since the navigation code is unchanged, this is likely a dev/e2e-environment quirk rather than a production regression — filing a follow-up to confirm.
- API hardening (`@ArrayMaxSize` on `specs` + a P2002 handler in the custom-programs repository) and cycle-generation schedule/frequency-mismatch handling are separate follow-ups per the issue.

`/review` ([comment](https://github.com/brownm09/lifting-logbook/pull/755#issuecomment-4919933172)) returned 1 blocking + 8 non-blocking findings. Per ADR-028, each is fixed in-PR or filed-and-linked:

| # | Finding | Disposition |
|---|---|---|
| **Blocking** | `daysFromSpecs` grouped instances by `order` alone, silently dropping a lift when two different lifts shared `(offset, order)` (reachable via spec import) → data loss on edit-load/re-save | **Fixed** — now groups by `(order, lift)`; added a `daysFromSpecs` regression test for the two-lifts-one-position case |
| [correctness] | `specsFromDays` offset renormalization shifts no-schedule clone display dates (leangains 0/2/4 → 0/1/2) | **Filed** [#761](https://github.com/brownm09/lifting-logbook/issues/761) |
| [maintainability] | Remove-button + per-input aria-labels omitted within-day position, making duplicate-lift-in-a-day controls ambiguous and un-`getByLabel`-able | **Fixed** — labels now include `#${i + 1}`; e2e remove-button selectors updated |
| [performance] | No memo boundary; whole form re-renders per keystroke | **Filed** [#762](https://github.com/brownm09/lifting-logbook/issues/762) |
| [reliability] | Mock `PUT /programs/custom/:id` was stateless and hardcoded `baseTemplate: null` | **Fixed** — PUT now persists to `state.customPrograms` (mirrors POST) with consistent `baseTemplate` |
| [maintainability] | `days` initializer gated load-existing on `mode === 'edit'` | **Fixed** — loads `existing.specs` whenever provided, regardless of mode |
| [maintainability] | `uid()` doc comment claimed SSR/hydration determinism (untrue) | **Fixed** — reworded to state the real invariant (client-only keys, never serialized → divergence is harmless) |
| [maintainability] | `defaultWeekParams` exported with no external consumer | **Fixed** — dropped the `export` (module-private) |
| [reliability] | Numeric inputs use `Number(e.target.value)` (cleared → 0, `-` → NaN); **pre-existing** from the old editor | **Filed** [#763](https://github.com/brownm09/lifting-logbook/issues/763) |

All fixes verified: typecheck ✅, web unit **244 passed**, api DTO **12 passed**, Playwright `programs-editor` **1 passed**.

Closes #751


brownm09:
## Review: [feat] Program editor: group exercises into workout days

### Summary
A well-tested feature that correctly surfaces the existing `(week, offset, lift, order)` model in the editor. The **forward (save) path is provably safe** — unique natural keys by construction, no client-only `id` leaked past the `forbidNonWhitelisted` DTO, all values DTO-valid. **Recommend merge after fixing one blocking data-loss issue on the reverse (edit-load) path**: `daysFromSpecs` can silently drop a lift when two different lifts share the same `(offset, order)`. That shape isn't produced by the editor or presets, but it's reachable via the custom-program spec-import path, and edit mode is the load path for *all* custom programs. The remaining items are non-blocking quality/robustness fixes. The ~1050-line diff is appropriately scoped (editor + mappers + tests + mock).

---

### Blocking Findings

**[correctness]** `apps/web/app/(authed)/programs/programSpecMapping.ts` — `daysFromSpecs`, instance grouping
Instances are grouped by `(offset, order)` and one lift is taken per group (`(orderRows.find(r => r.week === 1) ?? orderRows[0]).lift`), discarding any other lift at that position. But the persisted natural key is `(programId, week, offset, lift, order)` — `lift` is part of the key — so the store permits two *different* lifts at the same `(offset, order)`. The editor and presets always assign each lift a distinct `order`, so editor/clone data is safe; however the custom-program spec import validates fields independently and does not enforce distinct `order` per lift within an offset, so an imported program can hold `(w1, off0, Squat, order1)` **and** `(w1, off0, Bench, order1)`. Opening such a program in **edit** mode (the load path for every custom program) silently drops `Bench`, and the next save deletes it permanently — silent data loss.
**Fix:** Group instances by `(order, lift)` (one instance per distinct lift at a position) rather than by `order` alone; `specsFromDays` already renumbers `order` on save, so no collision results. Add a `daysFromSpecs` test with two different lifts sharing `(offset, order)`.

---

### Non-Blocking Findings

**[correctness]** `programSpecMapping.ts` — `specsFromDays`, offset renormalization
`offset` is re-derived as the compacted non-empty-day index, so non-contiguous stored offsets renormalize (0/2/4 → 0/1/2). Because no-schedule dates are `cycleStart + (week-1)*7 + offset` (`apps/api/src/programs/mappers.ts`, `noScheduleWorkoutDateUTC`), cloning the Leangains preset (Mon/Wed/Fri) and saving it unchanged shifts its no-schedule display dates to Mon/Tue/Wed. Scheduled-mode dates, workout count, order, and lift grouping are all unaffected — impact is limited to no-schedule display spacing, and it's still strictly better than the pre-PR editor (which flattened everything to `offset: 0`).
**Suggestion:** This is the deliberate "contiguous offsets" design decision; confirm the no-schedule fidelity loss on clone is acceptable, or carry each day's original `offset` through the model. Filing a follow-up to track.

**[maintainability]** `ProgramEditor.tsx` — instance remove-button + input labels omit within-day position
The remove-button `aria-label` (`Remove ${lift} from Day ${n}`) and per-input aria-labels (`${lift} Day ${n} Week ${w} …`) key only on lift + day + week. Since the feature deliberately allows the same lift twice in one day (distinct `order`), two "Back Squat" instances on Day 1 produce byte-identical labels — and the repo's entire web test strategy is `getByLabel`/`getByRole({name})`, where both RTL and Playwright strict mode throw on multiple matches. So a test for the duplicate-in-a-day case is un-writable, and the controls are ambiguous to assistive tech.
**Suggestion:** Include the within-day position (e.g. `Remove ${lift} #${i + 1} from Day ${n}` and `${lift} #${i + 1} Day ${n} Week ${w} sets`). Update the e2e spec's three remove-button selectors accordingly.

**[performance]** `ProgramEditor.tsx` — no memo boundary; whole form re-renders per keystroke
All day/instance/week markup renders inline in one component, so each `setDays` reconciles the entire form (worst case ~7 days × 12 instances × 3 weeks × ~5 inputs ≈ 1,260 inputs plus a full `LIFT_CATALOG` `<select>` per day). The immutable updaters keep untouched `d`/`inst` referentially stable, but that buys nothing without memoized children. Realistic programs are fine; the worst case could feel laggy on low-end devices.
**Suggestion:** Extract memoized `DayCard`/`InstanceRow` components with `useCallback` handlers so an edit re-renders only the touched instance. Filing a follow-up.

**[reliability]** `apps/web/e2e/mock-api.mjs` — `PUT /programs/custom/:id` is stateless
The PUT handler echoes the body but never updates `state.customPrograms` (unlike the sibling `POST`), and hardcodes `baseTemplate: null`. A future edit-mode e2e that asserts a rename is reflected in `GET /programs/custom` would silently pass against stale data.
**Suggestion:** Either update the matching `state.customPrograms` entry (mirroring POST) or add a comment that PUT is intentionally echo-only; keep `baseTemplate` handling consistent with POST.

**[maintainability]** `ProgramEditor.tsx` — `days` initializer gates load-existing on `mode === 'edit'`
The pre-#751 initializer loaded `existing.specs` whenever present; the new one only does so for `mode === 'edit'`. Not a regression at current call sites (EditorTab passes `existing` only with `edit`), but a future "clone an existing custom program" flow passing `existing` under `mode="clone"` would silently start empty and drop the source specs.
**Suggestion:** Load `existing.specs` first regardless of mode (falling through to template seed, then empty day), so the initializer is robust to that future flow.

**[maintainability]** `programSpecMapping.ts` — `uid()` doc comment is inaccurate
The comment says the counter is "Deterministic across SSR/hydration." It isn't — `idCounter` is module-level mutable state shared across concurrent requests. The code is nonetheless correct, but for a different reason: the ids are only React keys / closure values and never serialized, so counter non-determinism is harmless.
**Suggestion:** Reword to state the real invariant (client-only keys, never serialized → divergence is harmless).

**[maintainability]** `programSpecMapping.ts` — `defaultWeekParams` exported without an external consumer
Only `defaultWeeks` calls it; both `ProgramEditor` and the mapping test import `defaultWeeks`. Unnecessary public surface.
**Suggestion:** Drop the `export` (keep it module-private) unless a caller is imminent.

**[reliability]** `ProgramEditor.tsx` — numeric `onChange` uses `Number(e.target.value)` (pre-existing)
A cleared field maps to `0` and a partial `-` to `NaN`, so `sets`/`reps` can transiently fall below `min={1}`; on save these are rejected by the DTO and surfaced via `catch → setError` rather than silently persisted. State isn't permanently wedged, and this is carried over verbatim from the old editor's `updateSpec` — pre-existing, not introduced here.
**Suggestion:** If touched, keep the raw string and validate on blur (or clamp with `Number.isFinite`) so an empty field doesn't momentarily block save. Filing a follow-up.

---

### Findings Disposition
_To be completed before merge (per ADR-028: each finding fixed-in-PR or filed-and-linked)._

---

brownm09:
## Re-review after remediation — CLEAN

Fresh `/review` pass on the current head (`3e1c276`), after remediating the [original review](https://github.com/brownm09/lifting-logbook/pull/755#issuecomment-4919933172)'s findings. An independent adversarial re-review of the full remediation delta — the `(order, lift)` regrouping in `daysFromSpecs`, position-aware `#N` aria-labels with matching e2e selectors, the now-stateful mock `PUT`, the mode-independent `days` initializer, the dropped export / corrected comment, and the key-separator byte — found **no new correctness / security / reliability regression**, and confirmed every prior finding is resolved.

**Original findings (1 blocking + 8 non-blocking) — all dispositioned** (full table in the PR body):

- **6 fixed in-PR**, including the blocking `daysFromSpecs` data-loss fix (two different lifts sharing `(offset, order)` no longer drop one on edit-load), verified by a new mapping regression test.
- **3 filed as out-of-scope follow-ups:** #761 (offset spacing on clone), #762 (memoize day/instance rows), #763 (numeric input `Number()` coercion — pre-existing).

Also caught and fixed during remediation: a literal NUL byte had slipped into the grouping-key separator (functionally valid but made git treat the source as binary) — replaced with a plain space (`3e1c276`).

Verification: typecheck ✅ · web unit **244 passed** · mapping **12 passed** · Playwright `programs-editor` **1 passed**.




*Re-reviewed by `claude-opus-4-8` via Claude Code.*

brownm09:
Branch updated with `origin/main` (a 1-line `CLAUDE.md` change only — no source code) to satisfy the strict up-to-date requirement; head is now `8ff7087`. The independent CLEAN re-review above still stands (no application code changed). Re-affirming the fresh marker for the new head:

Issue #741, opened 2026-07-08
[fix] apps/web Playwright e2e harness: localhost->::1 breaks all tests on Windows (IPv4-only dev servers)


brownm09:
## Problem

Running the mandated local Playwright e2e suite on Windows fails **every** test:

```
apiRequestContext.get: connect ECONNREFUSED ::1:3004
```

and `page.goto('/import')` hangs ~90s before failing. Observed 2026-07-08 while running `npm run test:e2e -w @lifting-logbook/web` for the local pre-push gate on PR #736.

## Root cause

- The Playwright `webServer` processes — `node e2e/mock-api.mjs` (:3004) and `npm run dev` (:3000) — bind IPv4-only (`0.0.0.0`) on this Windows box (confirmed via `netstat -ano`).
- Node 18+ uses `verbatim` DNS ordering, so `localhost` resolves to IPv6 `::1` first.
- The specs connect via `http://localhost:3004` (the `MOCK_API` constant) and Playwright `baseURL: http://localhost:3000`; Next dev's own server-side fetch to `API_URL=http://localhost:3004` does the same. All resolve to `::1`, which the IPv4-only servers refuse.
- CI (Linux) is unaffected because `localhost` -> `127.0.0.1` there.

Temporary local workaround (not committed): `NODE_OPTIONS=--dns-result-order=ipv4first npm run test:e2e -w @lifting-logbook/web` gives a clean 23/23.

## Fix

Use `127.0.0.1` (unambiguous IPv4 on every platform, matches the servers' IPv4 bind) instead of `localhost` for every **client** address in the e2e harness:

- `apps/web/playwright.config.ts`: `use.baseURL`, and the webServer `API_URL` / `PUBLIC_API_URL` env.
- The `MOCK_API` constant in `apps/web/e2e/{smoke,scheduling,import,settings}.spec.ts`.
- `apps/web/e2e/mock-api.test.ts` `BASE` (the same latent issue in the mock's own Jest test).

Zero risk to Linux CI (`127.0.0.1` behaves identically). `staging.*` is unaffected (targets remote Cloud Run). The `localhost:3004` fallbacks in `lib/api.ts` / `lib/public-config.ts` / `app/api/health/route.ts` / `.env.example` are out of scope — they are inert during e2e because the config sets the env.

## Also

[...]

PR #758, opened 2026-07-08
[docs] Add git fetch origin to "CI not firing" entry's Fix step

brownm09:
## Summary

The "CI not firing — merge conflict silences GitHub Actions" entry in `CLAUDE.md`'s `## Testing` section had a Fix step that implicitly assumed the local `origin/main` remote-tracking ref was already current before rebasing. If stale, the rebase resolves against outdated content and doesn't actually clear the conflict against the real current `main`.

This mirrors the same gap just fixed in the neighboring "Stale-branch squash-merge rejection" entry (added in #753 via review findings), which now explicitly calls out `git fetch origin` before merging for exactly this reason.

- Added an explicit `git fetch origin` step to the "CI not firing" entry's Fix, with phrasing mirroring the neighboring entry ("the rebase target is only as current as this repo's local `origin/main` tracking ref").

Docs-only change; no code touched.

## Test plan

- [x] N/A — docs-only change, no automated tests per the repo's Coverage Requirements table (Refactor / docs only → None required)
- [x] `/review` to be run before merge per this repo's Review Gate check

Closes #756

PR #754, opened 2026-07-08
[refactor] Collapse toWorkoutResponse optional tail into an options object

brownm09:
## Summary

`toWorkoutResponse` (`apps/api/src/programs/mappers.ts`) took **11 positional parameters**, six optional. The optional tail — `overrideDate?`, `plannedLifts?`, `scheduledDate?`, `skipped`, `cycleStartDate?`, `offset?` — included **two adjacent `Date?` params (`scheduledDate`, `cycleStartDate`) separated only by a boolean**, so a caller could transpose them with no TypeScript error and silently render the wrong workout `date` — the exact class of bug #749 fixed.

This collapses the optional tail into a single `WorkoutResponseOptions` object so call sites are keyed by name. The five required positional args (`program, cycleNum, workoutNum, week, records`) are unchanged.

**Pure refactor, no behavior change:** the mapper body is byte-identical after destructuring the options object; the one production caller (`workouts.controller.ts`) and all `mappers.spec.ts` call sites are updated; existing tests cover the contract.

### Note on `exactOptionalPropertyTypes`

`apps/api` compiles with `exactOptionalPropertyTypes: true`, under which an optional property `overrideDate?: Date` rejects an *explicit* `undefined`. The pre-refactor positional params all accepted `undefined` (the controller passes `overrideDate ?? undefined`, `scheduledDate`/`cycleStartDate` from optional chains, and `workoutKey?.offset`), so each option field is typed `T | undefined` to preserve that contract exactly. This is the compiler's own suggested remedy and keeps call sites clean (no conditional spreads).

## Acceptance Criteria

- [x] `toWorkoutResponse`'s optional parameters are passed by name (options object), eliminating the adjacent-`Date?` transposition hazard.
- [x] `workouts.controller.ts` and all `mappers.spec.ts` call sites updated; `npm test -w @lifting-logbook/api` green.
- [x] No behavior change (pure refactor; existing tests cover the contract).

## Testing

From the worktree on Node 20.11.1 (Windows), Docker up (DB E2E via Testcontainers ran):

- `npm run build` → 6/6 workspaces pass. The api build's `nest build`/`tsc` is the authoritative type gate (ts-jest is transpile-only); it caught the `exactOptionalPropertyTypes` mismatch, now resolved.
- `npm run typecheck` → 4/4 pass.
- `npm test` (full suite) → **Tasks: 12 successful, 12 total; 1,052 passed, 0 skipped, 0 failed**
[...]

PR #760, opened 2026-07-08
[test] apps/web e2e: per-worktree dynamic ports for concurrent-run isolation

### LINKED ISSUE #746: [test] apps/web e2e: make local harness robust to concurrent-worktree server reuse (fixed ports + reuseExistingServer)
## Problem

`apps/web/playwright.config.ts` starts two `webServer` processes on **fixed ports** — `node e2e/mock-api.mjs` on `:3004` and `next dev` on `:3000` — with `reuseExistingServer: !process.env.CI`. Locally that flag is `true`.

This repo routinely runs **60+ concurrent worktrees** under `.claude/worktrees/`, all sharing those two fixed ports. When two local `npm run test:e2e -w @lifting-logbook/web` runs overlap (different worktrees, different branches), Playwright's `reuseExistingServer` silently **attaches to whichever mock/next server is already listening** instead of starting its own. A run can then pass or fail against another worktree's build rather than its own — a false green or a confusing false red.

The recently-merged #741 / #744 fix (pinning the whole harness to `127.0.0.1`) removed the IPv4/IPv6 loopback ambiguity but **not** this cross-worktree collision. #744 documents it as a manual hygiene tip in `CLAUDE.md` ("confirm nothing is already listening on :3000/:3004, or run with `CI=1` to force fresh servers") rather than fixing it structurally.

## Where it was surfaced

Non-blocking `[reliability]` finding on the [#744 review](https://github.com/brownm09/lifting-logbook/pull/744#issuecomment-4916814890). Deferred there as out of scope for the IPv4/IPv6 fix (the structural version roughly doubles that PR's diff and touches port-selection across every e2e file).

## Options

1. **`reuseExistingServer: false` locally** — simplest; every local run cold-starts its own mock + `next dev`. Cost: loses the warm-server reuse speedup, and two overlapping runs would then collide on the fixed port (EADDRINUSE) instead of silently sharing — a loud failure, arguably better than a silent wrong-build pass, but still not concurrency-safe.
2. **Per-worktree dynamic ports** — allocate free ports at config load and thread them through `playwright.config.ts` (both `webServer` entries + `baseURL` + `API_URL`/`PUBLIC_API_URL`), `mock-api.mjs` (`MOCK_API_PORT`, already honored), and the specs' `MOCK_API` constant (read from an env var instead of a hardcoded literal). Fully concurrency-safe; larger change. This is the robust fix.

Option 2 is preferred if we want overlapping local e2e runs to be truly independent; option 1 is a smaller stopgap that at least converts silent wrong-build reuse into a loud port conflict.

## Acceptance criteria

- [ ] Two concurrent `npm run test:e2e -w @lifting-logbook/web` runs in different worktrees do not share servers (each uses its own mock + next, or fails loudly rather than silently reusing).
- [ ] `CI=1` / CI behavior unchanged (CI already sets `reuseExistingServer: false` via `process.env.CI`).
- [ ] `CLAUDE.md` "apps/web Playwright E2E (local)" hygiene note updated to reflect the structural fix once landed.

Related: #741, #744.

brownm09:
## Summary

`apps/web/playwright.config.ts` started its two `webServer` processes on **fixed** ports — `node e2e/mock-api.mjs` on `:3004` and `next dev` on `:3000` — with `reuseExistingServer` true locally. This repo runs many concurrent worktrees under `.claude/worktrees/`, all sharing those two ports, so two overlapping local `npm run test:e2e` runs could silently attach to another worktree's server and pass/fail against the wrong build. #741/#744 pinned everything to `127.0.0.1` (Windows IPv4/IPv6 fix) and documented this reuse hazard as a manual hygiene tip, but did not fix it structurally.

This implements **option 2** from #746 — per-worktree dynamic ports.

## Changes

- **`apps/web/playwright.config.ts`** allocates two free `127.0.0.1` ports at config load (`allocatePorts()`) and threads them through both `webServer` entries (mock `MOCK_API_PORT` env + readiness `url`; next `--port` + readiness `url`), `use.baseURL`, the next server's `API_URL`/`PUBLIC_API_URL`, and `process.env.PLAYWRIGHT_MOCK_API_URL`. Allocation is guarded by env so it happens **once** in the Playwright main process; worker and webServer child processes inherit the ports.
  - Allocation runs in a short-lived `node` child process (`execFileSync`) because Playwright loads the config as **CommonJS** — a top-level `await` for an async free-port probe is a `SyntaxError` there (verified). All listeners are held open simultaneously so the OS returns distinct ports.
- **`apps/web/e2e/{smoke,scheduling,settings,import}.spec.ts`** read `MOCK_API` from `process.env.PLAYWRIGHT_MOCK_API_URL` (fallback to the `:3004` literal for a bare `playwright test`).
- **`reuseExistingServer`** is now `false` everywhere (was `!process.env.CI`). With a fresh port per run there is never a matching pre-existing server to reuse, so `true` bought nothing locally and left a rare silent-reuse window on a free-port race; `false` turns any residual race into a loud `EADDRINUSE`. **CI is unchanged** — it was already `false` via `process.env.CI`.
- **127.0.0.1 host pinning from #741/#744 is preserved** — only the port is dynamic; no address reverts to `localhost`.
- **`CLAUDE.md`** "apps/web Playwright E2E (local)" note updated to describe the structural fix.

`apps/web/e2e/mock-api.mjs` already honored `MOCK_API_PORT` (no change). `apps/web/e2e/mock-api.test.ts` uses its own dedicated port `3105` (unaffected).

## Acceptance criteria (#746)

- [x] Two concurrent runs in different worktrees don't share servers — each allocates its own free ports; a rare free-port race fails loudly (`EADDRINUSE`), never silent reuse.
- [x] CI behavior unchanged — `ci.yml` runs `playwright test` with no hardcoded ports; `reuseExistingServer` stays `false` in CI; `staging.yml` uses the separate `playwright.config.staging.ts`.
- [x] CLAUDE.md hygiene note updated.

## Testing

Run on **Windows** (local), matching the concurrency scenario the fix targets:

- `npx playwright test --list` → 23 tests in 4 files (config loads; confirmed the CJS/no-top-level-await fallback).
- **Concurrency proof:** simulated 4 overlapping `test:e2e` allocations (each holding its ports open, as real servers do) → each got a distinct `(mock, web)` port pair, all 8 ports distinct, none equal to `3000`/`3004`. No server sharing.
- **Full suite:** `npm run test:e2e -w @lifting-logbook/web` → **`23 passed (39.6s)`** (0 skipped, 0 failed). Observed the suite running against dynamic ports (e.g. next dev on `127.0.0.1:55851`).
  - Prerequisite in a fresh worktree: `turbo run build` of the workspace libs first (`packages/core/dist` doesn't exist until built) — unrelated to this change; the web dev server resolves `@lifting-logbook/core` from `dist`.

> Note on "two overlapping instances in one worktree": Next.js holds a **per-project-directory** dev lock, so two `next dev` in the *same* `apps/web` can't both start regardless of port — a directory lock, not a port collision. Real concurrency is cross-worktree (different `apps/web` dirs), which the dynamic-port allocation handles; the 4-way allocation demo proves distinct ports across concurrent processes.

## Risk audit (Plan-then-optimize Pass 3)

- **Testing** — the e2e suite is the test (23 pass); config-load + concurrency verified.
- **Observability / Security / Performance / Data integrity** — N/A: test-harness-only change, all servers bound to `127.0.0.1` loopback, no runtime/logging/schema change.
- **Resilience** — rare free-port TOCTOU → loud `EADDRINUSE`, not silent reuse; rollback = revert (pure test harness, no data).

Closes #746


brownm09:
## Review: [test] apps/web e2e: per-worktree dynamic ports for concurrent-run isolation

### Summary
**Merge.** Well-scoped test-harness change (6 files, +103/−26) that fixes the concurrent-worktree server-sharing hazard from #746 by allocating a free `127.0.0.1` port per run and threading it through both webServers, `use.baseURL`, `API_URL`/`PUBLIC_API_URL`, and the specs' `MOCK_API` base. Implementation is correct and verified (23/23 e2e pass on Windows; a 4-way concurrent-allocation demo confirmed distinct ports, none `3000`/`3004`), the #741/#744 `127.0.0.1` host pinning is preserved (only the port is dynamic), and CI is genuinely unchanged — `reuseExistingServer` was already `false` in CI via `process.env.CI`. **0 blocking, 0 non-blocking findings.**

### Gate checks
- **Doc reconciliation (2b):** N/A — no `Documentation Maintenance` table in this repo's CLAUDE.md.
- **Doc coverage (2c):** `apps/web/e2e/README.md` exists but documents only the *staging* integration tests (`playwright.config.staging.ts`), which this PR does not touch; the local mock-suite changes are outside its scope. No files added/deleted/renamed. No update warranted.
- **Test coverage (2d):** N/A — test-harness change, no new product behavior; the e2e suite is the coverage and passes 23/23.
- **Test integrity (2e):** Clean — no skip markers, deleted tests, lowered thresholds, or bypass flags; PR body carries the `23 passed (39.6s)` summary; all changed source is JS/TS/MD.
- **ADR-warrant (2f):** Not warranted — mirrors the #741/#744 precedent (127.0.0.1 pinning lives as a CLAUDE.md troubleshooting note, not an ADR); rationale is fully captured in the config comments, PR body, and commit. No `claude/`-scoped rule/hook/skill/settings touched.

### Correctness notes (verified — not findings)
- **Worker re-allocation guard:** the config is re-evaluated in each worker/webServer child; the `MOCK_API_PORT` / `E2E_WEB_PORT` env guard makes allocation happen once in the main process, children inheriting the same ports. Sound by construction and confirmed by the passing run — a mismatched re-allocation would point workers' `baseURL`/`MOCK_API` at the wrong port and fail every test.
- **`reuseExistingServer: false` everywhere:** correct and lossless — with a fresh port per run there is never a matching pre-existing server, so `true` bought nothing locally and left a rare silent-reuse window; `false` turns a residual free-port race into a loud `EADDRINUSE`. CI already resolved to `false`.
- **CommonJS subprocess allocator:** the `execFileSync` `node -e` probe is the right call given Playwright loads the config as CJS (top-level `await` is a `SyntaxError` there); holding all listeners open simultaneously guarantees distinct ports. Extracting the probe to a helper file would gain no lint/type coverage (`e2e/` is outside both `eslint app lib` and the typecheck project), so the inline form is left as-is.
- **Spec `?? 'http://127.0.0.1:3004'` fallback:** unreachable in normal runs (the config always sets `PLAYWRIGHT_MOCK_API_URL` before workers fork) and harmless if ever reached (loud `ECONNREFUSED`, not silent). Left as a self-documenting last resort.

### Style
No style findings — the change is thoroughly commented, and each edited comment ties back to the motivating issue.

---

brownm09:
## Re-review after branch-update merge (`79eb91b`)

The branch was updated by merging `origin/main` (bringing in #754, an `apps/api` refactor) after the initial review, which staled the review marker. The diff **vs `main` is unchanged** — #754 now sits on both sides of the comparison, so this PR's net change is still only the 6 files reviewed in the [prior comment](https://github.com/brownm09/lifting-logbook/pull/760#issuecomment-4920053588). Re-review conclusion is unchanged: **0 blocking, 0 non-blocking.** All gate checks (2b–2f) resolve as before (doc-reconciliation N/A, doc-coverage N/A, test-coverage N/A, test-integrity clean, ADR not-warranted).

brownm09:
## Re-review after second branch-update merge (`7de94e2`)

Branch updated again by merging `origin/main` (#758, a CLAUDE.md docs edit to a *different* section than this PR's — clean auto-merge) to satisfy `strict` branch protection. The diff **vs `main` is still unchanged** — #754 and #758 now sit on both sides of the comparison, so this PR's net change remains only the 6 files reviewed. Conclusion unchanged: **0 blocking, 0 non-blocking.** Gate checks (2b–2f) resolve as before.

brownm09:
## Re-review after third branch-update + new-spec fix (`2442476`)

Two changes since the last marker: (1) merged `origin/main` again for `strict`, which brought in a **new e2e spec** `apps/web/e2e/programs-editor.spec.ts` still using the old hardcoded `MOCK_API = 'http://127.0.0.1:3004'`; (2) this PR now extends the same `process.env.PLAYWRIGHT_MOCK_API_URL ?? <literal>` pattern to that spec, so it hits this run's dynamic mock port instead of a dead `:3004` (it would otherwise fail Playwright E2E). The change is byte-identical to the pattern applied to the other four specs. Net diff vs `main` is now 7 files; conclusion unchanged: **0 blocking, 0 non-blocking.** Gate checks (2b–2f) resolve as before.

Issue #769, opened 2026-07-09
[fix] apps/web typecheck fails: TS7016 missing declaration for react-dom/server in 5 test files


brownm09:
Pre-existing failure on `origin/main`, unrelated to any change in this session (discovered while running `npm run typecheck` for #766, whose diff touches none of the files below).

## Failure

`npm run typecheck` (`turbo run typecheck` → `@lifting-logbook/web:typecheck` → `tsc --noEmit -p tsconfig.typecheck.json`) fails with:

```
error TS7016: Could not find a declaration file for module 'react-dom/server'. '.../apps/web/node_modules/react-dom/server.node.js' implicitly has an 'any' type.
```

In:
- `apps/web/app/(authed)/onboarding/page.test.tsx`
- `apps/web/app/(authed)/settings/schedule/page.test.tsx`
- `apps/web/app/(authed)/settings/units/page.test.tsx`
- `apps/web/app/(authed)/settings/weight-rounding/page.test.tsx`
- `apps/web/app/page.test.tsx`

All five import `react-dom/server` (likely for `renderToStaticMarkup`, per the pattern noted in ADR-028's Verification section for `app/page.test.tsx`). Likely cause: a `@types/react-dom` version drift or a `tsconfig.typecheck.json` `types`/`moduleResolution` gap that stopped resolving the `react-dom/server` subpath declaration.

## Why filed instead of fixed here

Unrelated to the current PR's scope (Cloud Run IAM/CORS fix). Root-causing a type-declaration resolution gap across 5 files needs its own investigation.

Issue #768, opened 2026-07-09
[infra] Wire Cloud Run into the Grafana Cloud OTel telemetry pipeline


brownm09:
Part of #765. Follow-up to #474.

## Problem

Production runs 100% on Cloud Run (`enable_gke = false`). The GKE DaemonSet-based OTel Collector
only runs on GKE, so the API's Cloud Run service never gets `OTEL_EXPORTER_OTLP_ENDPOINT` set —
`apps/api/src/otel.ts`'s SDK starts, tries to export to `localhost:4318`, finds nothing, and
silently drops every trace/metric/log. Grafana Cloud shows zero production data as a result, and
all four production alert rules in `infra/observability/alerts/api.yaml` depend on this same
metrics pipeline — they are not actually monitoring production right now.

A complete, unused sidecar template already exists at
`infra/cloud-run/otel-collector-sidecar.yaml`, pointed at the same Grafana Cloud endpoints and the
same existing Secret Manager secrets (`lifting-logbook-prod-otel-otlp-auth-header`,
`-otel-loki-auth-header`) the GKE collector already uses. It was never wired into the actual
deploy pipeline.

## Fix (larger than the other two — separate session recommended)

1. New Secret Manager secret holding the collector's `config.yaml` (Cloud Run needs a secret
   volume, not a ConfigMap volume) — content should mirror
   `infra/kubernetes/charts/otel-collector/templates/configmap.yaml`'s pipeline config.
2. Add `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` to the API container's env in
   `infra/terraform/cloud-run.tf`.
3. **Resolve early:** `google_cloud_run_v2_service.api` has `lifecycle { ignore_changes =
   [template, ...] }`, so Terraform won't push a sidecar to the already-running service —
   `deploy.yml`'s `gcloud run deploy` is what mutates the live revision. Confirm whether the
   installed `gcloud` supports multiple `--container` flags directly, or whether the deploy step
   needs to switch to `gcloud run services replace` against a generated manifest (the existing
   `otel-collector-sidecar.yaml` template is already shaped for the latter).
4. Validate on staging first.
5. Likely needs an ADR-018 addendum noting Cloud Run now also ships telemetry, not just GKE.

## Verification

After deploy, confirm a test trace/log lands in Grafana Cloud Tempo/Loki for the Cloud Run
service, and that `APIRouteHighErrorRate` etc. show a `deployment_environment_name="production"`
series again.

Issue #767, opened 2026-07-09
[fix] Clerk JWT logged in plaintext via x-clerk-authorization header (redaction gap)


brownm09:
Part of #765.

## Problem

Clerk session JWTs are being logged in plaintext in production GCP Cloud Logging.
`apps/api/src/app.module.ts`'s pino `redact.paths` list covers `req.headers.authorization` but
not `req.headers["x-clerk-authorization"]` — the header the server-to-server auth path
(`apps/web/lib/api.ts`, read by `apps/api/src/auth/auth.guard.ts:28-32`) actually uses. Confirmed
live: a production log entry contains a full raw bearer JWT under that header. This violates this
repo's own Observability convention (`CLAUDE.md` → Observability → "must never log secrets,
tokens... sensitive request/response bodies").

## Fix

Add `'req.headers["x-clerk-authorization"]'` to the `redact.paths` array in `pinoHttpOptions`
(`apps/api/src/app.module.ts`), same bracket-notation style already used for
`req.headers["set-cookie"]`. Extend the existing redaction test coverage with a case for
`x-clerk-authorization`, mirroring the existing `authorization` coverage.

## Note

This does not retroactively scrub tokens already sitting in GCP Cloud Logging. Whether that
warrants a Clerk session-revocation discussion is a separate, human call — out of scope for this
fix.

Issue #765, opened 2026-07-09
[investigation] Production reschedule failure: IAM/CORS outage, JWT log leak, missing Cloud Run telemetry


brownm09:
## Problem

A user attempted to reschedule a workout in production and it failed, with no visible error and
nothing findable in Grafana. Investigation (live `gcloud` queries against `lifting-logbook-prod`
plus a full code trace) turned up three distinct, confirmed issues:

### 1. Production outage affecting every client-side write (not just reschedule)

A CORS-preflight `OPTIONS` request from the browser to the API's Cloud Run URL is rejected with
`403` by Cloud Run's own IAM layer, before the request ever reaches NestJS, Clerk auth, or any
application code. Confirmed via the live production log:

```
2026-07-09T17:47:43Z  OPTIONS .../programs/leangains/cycles/1/workouts/2/reschedule → 403
textPayload: "The request was not authenticated... Empty Authorization header value."
```

Root cause: `infra/terraform/cloud-run.tf` grants `roles/run.invoker` on the API Cloud Run
service only to the web server's own service account. `.github/workflows/deploy.yml` reinforces
this with `--no-allow-unauthenticated` on every API deploy (present since the original infra
scaffold, PR #157, 2026-05-02). But [ADR-028](docs/adr/ADR-028-web-runtime-public-config.md)
(2026-06-11, PR #515) deliberately wires the browser to call the API's external Cloud Run URL
directly — an assumption documented in `apps/web/lib/client-api.ts` ("there is no Cloud Run IAM
hop on the client path") and `CONTRIBUTING.md:97`, which has been false since ADR-028 shipped.

`apps/web/lib/client-api.ts` is the shared entry point for **12 mutation operations**
(lift records, reschedule, skip/unskip, body weight, lift overrides, imports) — by CORS/preflight
mechanics, all of them have been broken in production (and staging's Cloud Run A/B replica) for
about a month, silently, because there is no frontend test exercising any of these against a real
Cloud Run deployment, and the reschedule form's error handler swallows the failure with no
logging at all.

### 2. Clerk session JWTs logged in plaintext in production

`apps/api/src/app.module.ts`'s pino redact config covers `req.headers.authorization` but not
`req.headers["x-clerk-authorization"]` — the header the server-to-server auth path actually uses.
Confirmed live: a production log entry contains a full raw bearer JWT under that header.

### 3. Cloud Run never got the OTel Collector wired — zero production telemetry reaches Grafana

[...]

PR #770, opened 2026-07-09
[fix] Grant API Cloud Run service public invoker access to restore browser writes

### LINKED ISSUE #766: [fix] Production API Cloud Run service rejects all browser-direct writes (IAM/CORS)
Part of #765.

## Problem

Every client-side write in the app (reschedule, skip/unskip workout, lift records, body weight,
lift overrides, imports — all 12 operations exported from `apps/web/lib/client-api.ts`) has been
silently broken in production, and in staging's Cloud Run A/B replica, for about a month.

The browser's CORS-preflight `OPTIONS` request to the API's Cloud Run URL is rejected with `403`
by Cloud Run's own IAM layer, before the request reaches NestJS, Clerk auth, or any application
code:

```
OPTIONS https://lifting-logbook-prod-api-fvijrjuzmq-uc.a.run.app/programs/leangains/cycles/1/workouts/2/reschedule
→ 403 "The request was not authenticated... Empty Authorization header value."
```

`infra/terraform/cloud-run.tf` grants `roles/run.invoker` on the API service only to the web
server's own service account; `.github/workflows/deploy.yml` reinforces this by passing
`--no-allow-unauthenticated` on every API deploy (present since PR #157, 2026-05-02). But
[ADR-028](../blob/main/docs/adr/ADR-028-web-runtime-public-config.md) (2026-06-11, PR #515)
deliberately wires the browser to call the API's external Cloud Run URL directly — documented in
`apps/web/lib/client-api.ts`'s "AUTH HEADER INVARIANT" comment ("there is no Cloud Run IAM hop on
the client path") and `CONTRIBUTING.md:97`. That assumption has been false since ADR-028 shipped.

## Fix

1. `infra/terraform/cloud-run.tf`: add a `google_cloud_run_v2_service_iam_member "api_public"`
   resource granting `roles/run.invoker` to `allUsers` on the API service, mirroring the existing
   `web_public` resource. Update the access-control comment block above it to describe the new
   model (Cloud Run IAM open; Clerk JWT verification is the real authorization boundary).
2. `.github/workflows/deploy.yml`: drop `--no-allow-unauthenticated` from both the staging
   (`lifting-logbook-stg-api`) and production (`lifting-logbook-prod-api`) `gcloud run deploy`
   steps — the flag actively revokes any `allUsers` binding on every deploy, so the Terraform
   change alone would get silently undone on the next push.
3. New ADR documenting this decision (reverses a standing security default; reconciles it with
   ADR-028).
4. Playwright coverage for the reschedule flow against a real deployment, closing the
   `docs/testing/e2e-coverage.md` "Frontend ❌" gap.

## Verification

- Staging: replay the failing CORS preflight manually, confirm `2xx` instead of `403`; exercise
  the reschedule form end-to-end in the running staging app.
- Production: same, after the prod deploy step change lands.

brownm09:
## Summary

Fixes a ~month-long production outage: every client-side write in the app (reschedule, skip/unskip
workout, lift records, body weight, lift overrides, imports — all 12 mutation operations exported
from `apps/web/lib/client-api.ts`) has been silently failing because the browser's cross-origin
call to the API's Cloud Run URL never succeeds.

**Restoring it takes two layers, not one:**

1. **Cloud Run IAM** rejected the browser's CORS-preflight `OPTIONS` with `403` before it reached
   the app, because the API service required IAM authentication.
2. Even with IAM opened, **`apps/api` shipped with no CORS configuration**, so a reachable API
   still returns no `Access-Control-Allow-Origin` and the browser blocks the response. This was
   surfaced by the `/review` of the IAM-only version of this PR — the IAM grant alone is necessary
   but not sufficient.

Confirmed via the actual production log:
```
2026-07-09T17:47:43Z  OPTIONS .../programs/leangains/cycles/1/workouts/2/reschedule → 403
textPayload: "The request was not authenticated... Empty Authorization header value."
```

Root cause: `infra/terraform/cloud-run.tf` granted `roles/run.invoker` on the API Cloud Run
service only to the web server's own service account, and CI passed `--no-allow-unauthenticated`
on every API deploy — but [ADR-028](../blob/main/docs/adr/ADR-028-web-runtime-public-config.md)
deliberately wires the browser to call the API's external Cloud Run URL directly. Full rationale:
[ADR-032](../blob/main/docs/adr/ADR-032-cloud-run-api-public-invoker.md).

## Changes

**Layer 1 — Cloud Run IAM (infra/CI):**
- `infra/terraform/cloud-run.tf`: add `google_cloud_run_v2_service_iam_member "api_public"`
  (`allUsers` / `roles/run.invoker`), mirroring the existing `web_public` resource. Updated the
  access-control comment block: Clerk JWT verification is the real authorization boundary now.
- `.github/workflows/deploy.yml`: `--no-allow-unauthenticated` → `--allow-unauthenticated` on both
  the staging and production `lifting-logbook-{stg,prod}-api` deploy steps (the Terraform grant
  alone would be silently reverted on the next deploy otherwise).
- `.github/workflows/staging.yml`: **same flip for the PR-staging `stg-api` deploy** — this
  workflow independently redeploys the staging API on every PR and would otherwise re-lock it,
  silently undoing the fix. (Gap found while wiring up layer 2.)

**Layer 2 — Application CORS (apps/api):**
- `apps/api/src/main.ts`: `app.enableCors()` driven by an explicit origin allowlist. `@fastify/cors`
  (registered by `enableCors`, already a transitive dep of `@nestjs/platform-fastify` — no
  `package.json` change) answers the preflight and echoes `Access-Control-Allow-Origin`.
- `apps/api/src/cors.config.ts` (new): pure `resolveCorsOrigins()` + the allowlist. Covers the web
  app's Cloud Run URLs **in both address formats Cloud Run serves each service at** (legacy
  `-<hash>-uc.a.run.app` *and* current `-<projectNumber>.us-central1.run.app` — the staging suite
  loads from the latter, which `status.url` doesn't report), the `liftinglogbook.com` custom domain
  (apex + www), and local dev (localhost gated out of deployed runtimes via `NODE_ENV`). A
  `CORS_ALLOWED_ORIGINS` env var fully overrides the list as an ops escape hatch.
- `apps/api/src/cors.config.spec.ts` (new): 11 unit cases — deployed vs. local behavior, both URL
  formats, the exact staging project-number origin, and the env override.
- `apps/api/src/cors.preflight.spec.ts` (new): 3 cases that **deterministically prove `enableCors`
  works on the wire** — boots a NestFastify app with `enableCors()` and fires a preflight via
  Fastify `inject()` (no DB, no deployment, no shared staging), asserting `Access-Control-Allow-Origin`
  is echoed for the custom domain and the project-number Cloud Run origin, and withheld for a
  disallowed origin. This is the coverage the flaky shared-staging E2E can't reliably provide.

**Docs & review-finding fixes:**
- New `docs/adr/ADR-032-cloud-run-api-public-invoker.md` (now documents the two-layer fix),
  referenced from `docs/README.md` and `docs/adr-references.md`.
- `apps/web/e2e/staging.spec.ts`: added a reschedule step exercising `RescheduleForm.tsx` →
  `client-api.ts` → `PUBLIC_API_URL` (the only browser-direct write path; every other write-path
  test goes through a Next.js Server Action and is unaffected by this bug), and **reordered its
  assertions** to lead with the terminal success state (see Review findings disposition).
- `docs/testing/e2e-coverage.md`: flips "Reschedule a workout — Frontend" from ❌ to ✅ (staging).

## Why `allUsers` is safe here

The web service has used the identical `allUsers` + app-level-Clerk-auth model since day one — not
a new exposure class, just extending a proven pattern to a second service. Every real API endpoint
requires a valid Clerk session via the global `AuthGuard`; the only `@Public` routes are `/health`,
`/readyz`, `/livez` (unauthenticated probes, no sensitive data). The CORS allowlist is an allowlist,
not an origin reflector, so it stays a same-origin-policy control while Clerk remains the auth gate.
See ADR-032's Consequences for the full risk discussion (including the one accepted trade-off:
Cloud Run IAM was incidentally filtering scanner/bot traffic — flagged as a possible follow-up).

## Testing

- **`apps/api` build (tsc typecheck via `nest build`): clean.** (First run surfaced 5 errors in
  `mappers.ts`/`workouts.controller.ts` — stale `@lifting-logbook/core`/`types` `dist/` in the
  worktree, in files this PR doesn't touch; rebuilding the upstream packages cleared them, which is
  what CI does by building in dependency order. Not real errors.)
- **`apps/api` unit suite: `Tests: 412 passed, 75 skipped, 0 failed (33.4s)`** — includes the new
  `cors.config.spec.ts` (11 passed). The 75 skipped are the two `*.db.e2e.spec.ts` suites, skipped
  via the documented `LIFTING_SKIP_DB_E2E=1` opt-out — valid here because this PR touches **no** DB
  code (nothing under `apps/api/prisma/`, no repository changes); the change is API bootstrap + a
  pure config module. Docker was up; this was a deliberate scope decision, not a Docker failure.
- `npm run typecheck`: same pre-existing `@lifting-logbook/web` `TS7016` failure on `react-dom/server`
  in 5 test files this branch doesn't touch — filed as
  [#769](https://github.com/brownm09/lifting-logbook/issues/769).
- Suppression grep, test-integrity grep, deleted-test-files check: all clean. No `package.json`
  change → lockfile-sync N/A.
### Staging Integration Tests — admin-override (known shared-infra race, not a code defect)

The required **Staging Integration Tests** check fails on this PR and is being **admin-merged past**,
for a fully diagnosed reason: **shared staging cannot hold this fix during the test** while other PRs
are in flight. Investigation (2026-07-10):
- After this PR's own deploy correctly set the API to `--allow-unauthenticated` (allUsers invoker
  added, confirmed in the deploy log), concurrent open PRs (**#782, #784** — branches cut before the
  `staging.yml` fix, still carrying `--no-allow-unauthenticated`) re-deployed the **single shared**
  `lifting-logbook-stg-api` service, both **re-locking IAM** (→ the original `403`) and swapping in a
  **non-CORS image** (revisions `00527`→`00528`→`00529` in ~15 min, distinct image digests). The
  browser-direct reschedule step then fails because staging is momentarily running someone else's
  locked, CORS-less image — not because of anything in this diff.
- This is the shared-staging churn already tracked as [#673](https://github.com/brownm09/lifting-logbook/issues/673)
  (cross-PR staging-deploy mutex) and resolved structurally by [#695](https://github.com/brownm09/lifting-logbook/issues/695)
  (merge queue → isolated per-PR validation). Until #695 lands, this E2E cannot be reliable under
  concurrent PR load.

**The CORS fix itself is proven correct without depending on shared staging:**
- `cors.preflight.spec.ts` (new) boots the real `enableCors` wiring and asserts the preflight is
  answered with `Access-Control-Allow-Origin` for allowed origins and withheld for others — passes.
- `cors.config.spec.ts` (11 cases) covers origin resolution — passes.
- Manually verified `--allow-unauthenticated` grants `allUsers` on this project (the web service
  runs that model in prod today), and that the grant flips the preflight `403 → reachable`.

Once this lands on `main`, `main`'s own deploys keep staging/prod public + CORS-enabled, and the
E2E passes reliably (concurrent PRs stop re-locking once they rebase onto the merged fix — a
follow-up I'll drive for #782/#784).

The `/review` of the IAM-only version surfaced 1 blocking + 2 non-blocking findings; all resolved
in this PR:

- **[blocking] Missing application CORS** — the IAM grant alone wouldn't restore browser writes
  because `apps/api` had no CORS config. → **Fixed** (layer 2 above: `main.ts` + `cors.config.ts`
  + `cors.config.spec.ts`), plus the `staging.yml` re-lock gap found while implementing it.
- **[non-blocking] `staging.spec.ts` assertion ordering** — the negative `.not.toBeVisible()` error
  assertion led, and can false-green during the in-flight window before the error renders. →
  **Fixed**: lead with the terminal success-state assertion (trigger button reappearing), then the
  negative check.
- **[non-blocking] ADR-032 didn't mention CORS** — it framed the IAM change as the whole fix. →
  **Fixed**: ADR-032 now documents both layers, verification updated to check
  `Access-Control-Allow-Origin`, and `docs/adr-references.md` gets the NestJS/`@fastify/cors` refs.

The re-review of the CORS commit (`a772a39`) surfaced 1 additional non-blocking finding:

- **[reliability] client-side mutation forms swallow write errors** — `RescheduleForm.tsx` (and its
  11 sibling mutation paths via `client-api.ts`) discard the caught error with no `console.error`/
  telemetry, the exact blind spot that kept the original outage invisible in production. →
  **Filed** as [#783](https://github.com/brownm09/lifting-logbook/issues/783) (broader than this
  CORS PR — the fix spans all 12 mutation forms, not just reschedule).

## Follow-ups

- [#767](https://github.com/brownm09/lifting-logbook/issues/767) — Clerk JWT log-redaction gap
  ([PR #771](https://github.com/brownm09/lifting-logbook/pull/771)).
- [#768](https://github.com/brownm09/lifting-logbook/issues/768) — wire Cloud Run into Grafana
  Cloud telemetry (separate, larger session).
- [#769](https://github.com/brownm09/lifting-logbook/issues/769) — pre-existing `react-dom/server`
  typecheck failure.

Part of #765.

Closes #766

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -541,12 +541,19 @@ jobs:
           # is only constructed when DATABASE_URL is unset), and the Terraform service
           # template is lifecycle.ignore_changes, so this is the only place the env can be
           # removed. Removing an already-absent secret is a no-op on subsequent deploys.
+          # --allow-unauthenticated: Cloud Run IAM is intentionally public here — Clerk
+          # JWT verification (apps/api auth.guard.ts) is the real authorization boundary.
+          # See infra/terraform/cloud-run.tf's access-control comment and ADR-032. Without
+          # this, browser CORS preflights to PUBLIC_API_URL (ADR-028) 403 at the Cloud Run
+          # front end before ever reaching the app — Terraform's allUsers invoker grant on
+          # the API service alone is not sufficient, since this flag re-asserts the opposite
+          # policy on every deploy.
           gcloud run deploy lifting-logbook-stg-api \
             --image="${{ needs.build-images.outputs.ar_repo }}/api:${{ github.sha }}" \
             --region="${{ env.REGION }}" \
             --project="${{ vars.GCP_STAGING_PROJECT_ID }}" \
             --platform=managed \
-            --no-allow-unauthenticated \
+            --allow-unauthenticated \
             --remove-secrets=SYSTEM_DATABASE_URL \
             --quiet
 
@@ -1088,12 +1095,19 @@ jobs:
           # rationale. After the RLS cutover (#517) the runtime connects as the
           # NOBYPASSRLS lifting_app role via DATABASE_URL; the owner credential is
           # never read on the Prisma path, so it should not sit in the runtime env.
+          # --allow-unauthenticated: Cloud Run IAM is intentionally public here — Clerk
+          # JWT verification (apps/api auth.guard.ts) is the real authorization boundary.
+          # See infra/terraform/cloud-run.tf's access-control comment and ADR-032. Without
+          # this, browser CORS preflights to PUBLIC_API_URL (ADR-028) 403 at the Cloud Run
+          # front end before ever reaching the app — Terraform's allUsers invoker grant on
+          # the API service alone is not sufficient, since this flag re-asserts the opposite
+          # policy on every deploy.
           gcloud run deploy lifting-logbook-prod-api \
             --image="${{ steps.tf-prod.outputs.ar_repo }}/api:${{ github.sha }}" \
             --region="${{ env.REGION }}" \
             --project="${{ vars.GCP_PROD_PROJECT_ID }}" \
             --platform=managed \
-            --no-allow-unauthenticated \
+            --allow-unauthenticated \
             --remove-secrets=SYSTEM_DATABASE_URL \
             --quiet
 
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -541,12 +541,17 @@ jobs:
         # docs/runbooks/staging-ci-flakiness.md.
         continue-on-error: true
         run: |
+          # Cloud Run IAM is intentionally public here (ADR-032): the browser calls this API
+          # directly (ADR-028) and the Clerk JWT is the real auth gate. With --no-allow-
+          # unauthenticated, every PR-staging deploy would re-lock the API and 403 the
+          # browser's CORS preflight before it reaches the app — the ~month-long outage #766
+          # fixed. Kept in sync with the stg/prod API deploys in deploy.yml.
           gcloud run deploy lifting-logbook-stg-api \
             --image="${{ needs.build-images.outputs.ar_repo }}/api:${{ needs.build-images.outputs.image_tag }}" \
             --region="${{ env.REGION }}" \
             --project="${{ vars.GCP_STAGING_PROJECT_ID }}" \
             --platform=managed \
-            --no-allow-unauthenticated \
+            --allow-unauthenticated \
             --quiet
 
       - name: Fetch Cloud Run API logs on failure
--- a/apps/api/src/cors.config.spec.ts
+++ b/apps/api/src/cors.config.spec.ts
@@ -1,0 +1,95 @@
+import {
+  DEFAULT_CORS_ORIGINS,
+  LOCAL_DEV_ORIGINS,
+  resolveCorsOrigins,
+} from './cors.config';
+
+// Unit coverage for the CORS allowlist that gates browser → API writes (ADR-028/ADR-032).
+// resolveCorsOrigins takes its env values as parameters, so these tests never touch
+// process.env or bootstrap Nest.
+describe('resolveCorsOrigins', () => {
+  describe('deployed environments (NODE_ENV set by deploy.yml)', () => {
+    it.each(['production', 'staging'])(
+      'returns the deployed allowlist with NO localhost origins when NODE_ENV=%s',
+      (nodeEnv) => {
+        const origins = resolveCorsOrigins(undefined, nodeEnv);
+
+        // Every deployed origin is present...
+        for (const expected of DEFAULT_CORS_ORIGINS) {
+          expect(origins).toContain(expected);
+        }
+        // ...and no local-dev origin leaks into a deployed allowlist.
+        for (const local of LOCAL_DEV_ORIGINS) {
+          expect(origins).not.toContain(local);
+        }
+      },
+    );
+
+    it('allows the prod custom domain (apex and www)', () => {
+      const origins = resolveCorsOrigins(undefined, 'production');
+      expect(origins).toContain('https://liftinglogbook.com');
+      expect(origins).toContain('https://www.liftinglogbook.com');
+    });
+
+    it('allows BOTH Cloud Run address formats for each web service', () => {
+      const origins = resolveCorsOrigins(undefined, 'production');
+      // Legacy hash format (what status.url reports)...
+      expect(origins).toContain('https://lifting-logbook-prod-web-fvijrjuzmq-uc.a.run.app');
+      expect(origins).toContain('https://lifting-logbook-stg-web-mn3pkptvma-uc.a.run.app');
+      // ...and the project-number format Cloud Run also serves them at.
+      expect(origins).toContain(
+        'https://lifting-logbook-prod-web-508649171914.us-central1.run.app',
+      );
+      expect(origins).toContain(
+        'https://lifting-logbook-stg-web-910635705567.us-central1.run.app',
+      );
+    });
+
+    // Regression for the exact origin the staging integration suite sends: it loads the app
+    // from the project-number URL, not the hash URL status.url reports. An allowlist keyed
+    // only on the hash format would reject this preflight and fail the staging gate.
+    it('allows the project-number origin the staging integration tests actually use', () => {
+      const origins = resolveCorsOrigins(undefined, 'staging');
+      expect(origins).toContain(
+        'https://lifting-logbook-stg-web-910635705567.us-central1.run.app',
+      );
+    });
+  });
+
+  describe('local development (NODE_ENV unset or non-deployed)', () => {
+    it.each([undefined, 'development', 'test'])(
+      'appends localhost dev origins when NODE_ENV=%s',
+      (nodeEnv) => {
+        const origins = resolveCorsOrigins(undefined, nodeEnv);
+        for (const local of LOCAL_DEV_ORIGINS) {
+          expect(origins).toContain(local);
+        }
+        // Deployed origins remain available for local testing against real services.
+        expect(origins).toContain('https://liftinglogbook.com');
+      },
+    );
+  });
+
+  describe('CORS_ALLOWED_ORIGINS override', () => {
+    it('fully replaces the built-in list (no defaults, no localhost) when set', () => {
+      const origins = resolveCorsOrigins(
+        'https://a.example,https://b.example',
+        'production',
+      );
+      expect(origins).toEqual(['https://a.example', 'https://b.example']);
+    });
+
+    it('trims whitespace and drops blank entries', () => {
+      const origins = resolveCorsOrigins(
+        '  https://a.example ,, https://b.example  , ',
+        'production',
+      );
+      expect(origins).toEqual(['https://a.example', 'https://b.example']);
+    });
+
+    it('falls through to the built-in list when the override is empty or whitespace-only', () => {
+      const origins = resolveCorsOrigins('   ', 'production');
+      expect(origins).toEqual([...DEFAULT_CORS_ORIGINS]);
+    });
+  });
+});
--- a/apps/api/src/cors.config.ts
+++ b/apps/api/src/cors.config.ts
@@ -1,0 +1,93 @@
+/**
+ * CORS origin allowlist for the API.
+ *
+ * Browser → API calls are cross-origin: per ADR-028 the web app (on its own Cloud Run URL or
+ * custom domain) calls the API's external URL directly from the browser, so the API must echo
+ * `Access-Control-Allow-Origin` for the web app's origins or the browser blocks every
+ * client-side write. Cloud Run adds no CORS headers of its own, and making the service publicly
+ * invokable (ADR-032) only removes the IAM 403 on the preflight — the application still has to
+ * answer that preflight with the right CORS headers. This module produces the allowlist that
+ * `main.ts` feeds to `app.enableCors()`.
+ *
+ * This is an allowlist, not an origin reflector. The real authorization gate is the Clerk JWT
+ * the browser sends in the `Authorization` header (see `auth.guard.ts` / ADR-032); CORS only
+ * decides which web origins may issue browser calls to the API at all.
+ */
+
+/**
+ * Deployed web origins, allowed in every environment.
+ *
+ * Each Cloud Run web service is reachable at BOTH address formats Cloud Run assigns, and both
+ * are live simultaneously, so both must be listed:
+ *   - legacy hash format:    `https://lifting-logbook-<env>-web-<hash>-uc.a.run.app`
+ *   - project-number format: `https://lifting-logbook-<env>-web-<projectNumber>.us-central1.run.app`
+ *
+ * `status.url` reports only the hash format, but the staging integration suite (for example)
+ * loads the app from the project-number URL — listing only one silently breaks real traffic.
+ * Both prod and staging origins live in this single shared list; cross-environment calls are
+ * still gated by each environment's own Clerk keys at the auth layer, so sharing the list is
+ * safe and keeps the config to one place.
+ *
+ * Exact URLs verified live 2026-07-09 via:
+ *   gcloud run services describe lifting-logbook-<env>-web --region=us-central1 \
+ *     --project=lifting-logbook-<prod|staging> --format='value(status.url)'
+ *   gcloud projects describe lifting-logbook-<prod|staging> --format='value(projectNumber)'
+ * If a service is ever recreated (new hash) use the `CORS_ALLOWED_ORIGINS` override below until
+ * this list is updated.
+ */
+export const DEFAULT_CORS_ORIGINS: readonly string[] = [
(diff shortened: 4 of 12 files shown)


brownm09:
## Review: [fix] Grant API Cloud Run service public invoker access to restore browser writes

### Summary
**Needs revision before it will actually fix the outage.** The IAM/CORS diagnosis is correct and the Terraform + deploy-flag change is a necessary part of the fix — I verified the security claim (only `/health`, `/readyz`, `/livez` carry `@Public()`, all in `health.controller.ts`, so opening Cloud Run IAM does not expose any authenticated endpoint). But the highest-severity finding is that this fix is **necessary but not sufficient**: `apps/api` has no CORS configuration, so once the IAM 403 is removed, the cross-origin browser preflight will reach the app and get a response with no `Access-Control-Allow-Origin` header — and the browser will still block every client-side write. The PR likely ships a still-broken write path. Scope and structure are otherwise good; this is a completeness problem, not a rework.

---

### Blocking Findings

**[correctness]** `apps/api/src/main.ts:17-37` (absence of `app.enableCors(...)`)
The fix removes the Cloud Run IAM `403` on the browser's CORS preflight, letting the `OPTIONS` request reach the NestJS/Fastify app for the first time. But `main.ts` never calls `app.enableCors()`, and `apps/api/package.json` has no `@fastify/cors` (or any `cors`) dependency — Fastify does no CORS by default. So the preflight now reaches the app and receives a response with **no `Access-Control-Allow-Origin` header**, which the browser treats as a failed preflight and blocks the actual `PATCH`. The write still fails, just at the browser-CORS layer instead of the Cloud Run IAM layer. This is corroborated by the investigation itself: the *successful* API calls in the production logs all carry `user-agent: node` (server-side SSR calls via `lib/api.ts`, which are not subject to CORS), while the browser-direct writes (`client-api.ts` → `PUBLIC_API_URL`, genuinely cross-origin: `liftinglogbook.com` → `*.run.app`) have never succeeded — they've been hitting the IAM 403 *first*, which masked the missing-CORS blocker sitting right behind it. The newly-added staging E2E test *would* catch this, but it only runs post-deploy against staging, so this PR can merge and still leave production writes broken.
**Fix:** Enable CORS on `apps/api`, scoped to the known web origins (prod `https://liftinglogbook.com`, the staging web origin, and local dev), allowing the methods the client uses (`PATCH`/`POST`/`DELETE`/`GET`) and the headers it sends (`authorization`, `content-type`, and any custom ones — note the client path uses `Authorization`, not `X-Clerk-Authorization`). Either `app.enableCors({ origin: [...], methods: [...], allowedHeaders: [...] })` in `main.ts` or a registered `@fastify/cors`. Then confirm end-to-end (the added staging test is the right vehicle) that a cross-origin browser write actually returns `2xx` with CORS headers — not just that the preflight stops 403ing. Given the added test can't run until staging is deployed, strongly consider validating on staging *before* merging to production rather than merging both environment changes at once.

---

### Non-Blocking Findings

**[reliability]** `apps/web/e2e/staging.spec.ts:238-240`
The success check does `expect(getByText('Failed to reschedule...')).not.toBeVisible()` first, then `expect(getByRole('button', { name: '📅 Reschedule' })).toBeVisible({ timeout: 15_000 })`. A negative visibility assertion passes as soon as the element is absent — which is true in the in-flight "Moving…" window, before a failing request has rendered the error. So on a genuine failure the first assertion can pass prematurely; the real gate is the second assertion (the form only collapses back to the 📅 trigger on success). The test is still *correct* in outcome (a failure leaves the form open, so the trigger never appears and the second assertion times out), but the first line provides false comfort and reads as if it's guarding failure when it isn't.
**Suggestion:** Lead with the terminal-state assertion (trigger visible within the timeout = success), then optionally assert the error text is absent *after* that resolves — so the negative check runs against a settled state, not the loading state.

**[maintainability]** PR body → Testing section
The repo's test-before-PR convention (and the review test-integrity gate) expects a `Tests: N passed, N skipped, N failed` line in the PR body. This PR omits it. The omission is *honestly explained* (nothing unit-testable changed; the one added test is staging-only and can't run locally), so the gate's intent — preventing a manufactured-green claim — is satisfied. Flagging only so the convention is met explicitly rather than by inference.
**Suggestion:** Add one line to the Testing section stating that no runnable unit/integration suite exercises this infra/CI/docs change and that the added E2E is staging-only (runs post-deploy) — or paste the summary from whatever adjacent suite you do run (e.g. `apps/web` jest) to show nothing regressed.

---

### Style Notes
The duplicated 7-line `--allow-unauthenticated` rationale comment across the staging and prod deploy steps is fine — YAML has no clean shared-comment mechanism and the two steps are genuinely independent, so duplication is the right call over a dangling cross-reference. No action needed.

---

### Findings Disposition
Per ADR-028, each finding must be fixed-in-PR or filed-and-linked before merge. Recommended: the CORS finding is fixed in this PR (it's the difference between the outage being fixed vs. not) — verify apps/api CORS is missing and add it, ideally validated on staging first. The two non-blocking findings can be fixed in-PR (both are small) or filed; the test-assertion one is a 2-line reorder, the test-summary one is a PR-body edit. Record each disposition (fixed in `<sha>` | filed `#N`) in a "Review findings disposition" PR-body section before merge.

---

brownm09:
## Review: [fix] Grant API Cloud Run service public invoker access to restore browser writes

Re-review after commit `a772a39`, which added the application-CORS layer the previous review flagged as the missing blocking piece.

### Summary
**Merge-ready with one filed follow-up.** The two-layer fix is now correct and complete. Verified the three things most likely to be wrong in a hand-rolled CORS setup, all clean:
- **`credentials` correctly omitted** — the browser client authenticates with the `Authorization` header, never cookies (no `credentials: 'include'` anywhere in `client-api.ts` / `packages/api-client`), so no `Access-Control-Allow-Credentials` is needed.
- **`allowedHeaders` is complete** — the browser sends only `Authorization` + `Content-Type`; `@vercel/otel` instruments only the Next.js *server* runtime (no browser `FetchInstrumentation`), so outgoing browser fetches carry no `traceparent`/`tracestate` that would need allow-listing.
- **Origins verified live** — both Cloud Run address formats per service (hash `status.url` + project-number), covered by a regression test for the exact project-number origin the staging suite loads from.

The `staging.yml` re-lock gap (the PR-staging deploy independently re-asserting `--no-allow-unauthenticated`) is a genuinely good catch. Coverage is solid (11 new unit cases + the staging E2E now drives the real cross-origin path). No blocking findings.

---

### Non-Blocking Findings

**[reliability]** `apps/web/app/(authed)/cycle/[cycleNum]/workout/[workoutNum]/detail/RescheduleForm.tsx:30`
`handleMove`'s `catch { setError('Failed to reschedule. Please try again.') }` discards the underlying error — no `console.error`, no telemetry. This is the exact blind spot ADR-032's own Context section blames for the ~month-long outage being invisible, and it's shared by all 12 client-side mutation paths in `client-api.ts`. The new staging E2E closes the *CI* visibility gap for reschedule, but a *production* recurrence — or any non-CORS failure (500, network drop, auth expiry) on any mutation path — still surfaces only as a generic string with nothing logged. Contrast `SkipForm.tsx`, which already `console.error`s.
**Suggestion:** Out of scope for this CORS PR (the fix spans all 12 forms, not just reschedule). Filed as [#783](https://github.com/brownm09/lifting-logbook/issues/783).

---

### Style Notes
The origin allowlist hardcodes the current Cloud Run hashes and project numbers. This is a deliberate, documented trade-off — the `CORS_ALLOWED_ORIGINS` env override and the `gcloud` verification recipe in the file's doc comment both cover a service recreation, and it matches the approved "known origins" scope. Noting it only so the next maintainer knows the escape hatch exists if a hash ever changes.

---

### Findings Disposition
- **[reliability] RescheduleForm (and sibling forms) swallow write errors** → filed as [#783](https://github.com/brownm09/lifting-logbook/issues/783) (client-side write-path observability; broader than this PR, spans all 12 mutation forms).

---

brownm09:
## Re-review — CORS/IAM fix (branch updated to `main`)

Re-review after the branch was brought up to date with `main` (merge `134c3c6`), refreshing the Review Gate marker.

**The PR's own diff is byte-identical to the prior review** — the merge pulled in `main`'s unrelated `programs/*` / `api-client` / `types` changes, none of which this PR touches (verified: zero file overlap). So the prior review's conclusions stand, re-confirmed:

- `credentials` correctly omitted — the browser client authenticates with the `Authorization` header, never cookies (no `credentials:'include'` in the fetch layer).
- `allowedHeaders` complete — the browser sends only `Authorization` + `Content-Type`; `@vercel/otel` instruments only the Next.js server runtime, so browser fetches carry no `traceparent`.
- Origins verified live — both Cloud Run address formats per service.

**0 blocking, 1 non-blocking.**

### Findings Disposition
- **[reliability] client-side mutation forms swallow write errors** → filed as [#783](https://github.com/brownm09/lifting-logbook/issues/783) (disposed in PR body; broader than this PR — spans all 12 mutation forms).

---

brownm09:
## Re-review — CORS/IAM fix (rebased onto `main` + deterministic preflight test)

Re-review after rebasing onto current `main` (linear; head `6ff723c`) and adding `cors.preflight.spec.ts`. Refreshes the Review Gate marker.

**New since the last review:** `apps/api/src/cors.preflight.spec.ts` — boots the real `enableCors` wiring on the Fastify adapter and asserts, via Fastify `inject()`, that a preflight is answered with `Access-Control-Allow-Origin` for allowed origins (custom domain + project-number Cloud Run) and withheld for a disallowed origin. This deterministically proves the CORS mechanism independent of the shared-staging E2E (which is admin-overridden here for a fully diagnosed shared-infra race — see the PR body). The rest of the diff is unchanged and its conclusions stand, re-confirmed:

- `credentials` correctly omitted — the browser client authenticates with the `Authorization` header, never cookies (no `credentials:'include'` in the fetch layer).
- `allowedHeaders` complete — the browser sends only `Authorization` + `Content-Type`; `@vercel/otel` instruments only the Next.js server runtime, so browser fetches carry no `traceparent`.
- Origins verified live — both Cloud Run address formats per service.

**0 blocking, 1 non-blocking.**

### Findings Disposition
- **[reliability] client-side mutation forms swallow write errors** → filed as [#783](https://github.com/brownm09/lifting-logbook/issues/783) (disposed in PR body; broader than this PR — spans all 12 mutation forms).

---

PR #771, opened 2026-07-09
[fix] Redact x-clerk-authorization header to stop Clerk JWT plaintext logging

### LINKED ISSUE #767: [fix] Clerk JWT logged in plaintext via x-clerk-authorization header (redaction gap)

brownm09:
## Summary

Clerk session JWTs were being logged in plaintext in production GCP Cloud Logging. Confirmed live
via a real production log entry containing a full raw bearer JWT under `req.headers["x-clerk-authorization"]`
— the header `apps/web`'s server-to-server calls (`lib/api.ts`) use to carry the Clerk JWT, since
the standard `authorization` header is reserved for the Cloud Run IAM identity token on that path
(see `auth.guard.ts`). `apps/api/src/app.module.ts`'s pino `redact.paths` only covered
`req.headers.authorization`, so this second header sailed through unredacted.

## Changes

- `apps/api/src/app.module.ts`: add `'req.headers["x-clerk-authorization"]'` to `redact.paths`.
- `apps/api/src/otel-log-redaction.spec.ts`: extended the existing end-to-end redaction test
  (which already proves `Authorization`/`Cookie` are stripped from both the raw Pino wire format
  *and* the OTel Logs export pipeline) with the same coverage for `x-clerk-authorization`.

## Testing

- New test: `otel-log-redaction.spec.ts` — **verified as a genuine regression test**: temporarily
  reverted just the `app.module.ts` redact-path addition (keeping the test change) and confirmed
  the test fails (1 failed, 1 total) before re-applying the fix, which passes cleanly.
- `npm run typecheck`: same pre-existing failure as noted in
  [#766](https://github.com/brownm09/lifting-logbook/issues/766)'s PR — `@lifting-logbook/web`
  fails with `TS7016` on `react-dom/server` in 5 test files this branch does not touch (this
  branch only touches `apps/api`). Tracked in
  [#769](https://github.com/brownm09/lifting-logbook/issues/769).
- Suppression grep, test-integrity grep, deleted-test-files check: all clean.
- No `package.json` changes, so the lockfile-sync check is N/A.
- Tests: 1 passed (the new/extended redaction test), 0 skipped, 0 failed — ran directly via
  `npm test -w @lifting-logbook/api -- otel-log-redaction`. Did not run the full `apps/api` suite
  for this PR (2-line, tightly-scoped change; the extended spec is the only coverage this change
  affects).

The `/review` on this PR surfaced **0 blocking / 1 non-blocking** finding:

- **[maintainability]** `apps/api/src/app.module.ts` redaction is a *denylist* — incomplete by
  design. The same class of gap this PR fixes (a forgotten auth-bearing header leaking in
  plaintext) will recur for the next such header until someone remembers to add it.
  → **Filed as [#780](https://github.com/brownm09/lifting-logbook/issues/780)** (out of scope for
  this 2-line targeted fix; it's a broader allowlist-inversion / guard-test hardening change).

## Note

This does not retroactively scrub tokens already sitting in GCP Cloud Logging. Whether that
warrants a Clerk session-revocation discussion is a separate, human call.

Part of #765.

Closes #767

### DIFF
--- a/apps/api/src/app.module.ts
+++ b/apps/api/src/app.module.ts
@@ -16,10 +16,14 @@ import { UserSettingsModule } from './user-settings/user-settings.module';
 export const pinoHttpOptions: PinoHttpOptions = {
   // Strip auth-bearing headers before they reach Loki. pino-http's default
   // req serializer logs req.headers verbatim, which would otherwise leak
-  // JWTs and session cookies into long-retention log storage.
+  // JWTs and session cookies into long-retention log storage. Server-to-server
+  // calls (apps/web) carry the Clerk JWT in x-clerk-authorization rather than
+  // authorization (see auth.guard.ts) — both header names carry a bearer
+  // token and must be redacted (#767).
   redact: {
     paths: [
       'req.headers.authorization',
+      'req.headers["x-clerk-authorization"]',
       'req.headers.cookie',
       'req.headers["set-cookie"]',
       'res.headers["set-cookie"]',
--- a/apps/api/src/otel-log-redaction.spec.ts
+++ b/apps/api/src/otel-log-redaction.spec.ts
@@ -22,8 +22,9 @@ import { pinoHttpOptions } from './app.module';
 describe('OTel log export does not leak redacted fields', () => {
   const SECRET_TOKEN = 'super-secret-jwt-value';
   const SECRET_COOKIE = 'session=super-secret-cookie-value';
+  const SECRET_CLERK_JWT = 'super-secret-clerk-jwt-value';
 
-  it('keeps Authorization/Cookie values out of both the Pino wire format and the resulting OTel LogRecord', async () => {
+  it('keeps Authorization/Cookie/X-Clerk-Authorization values out of both the Pino wire format and the resulting OTel LogRecord', async () => {
     // 1. Log a request/response pair shaped like pino-http's own serialization,
     // through the app's REAL redact config, and capture the exact line pino
     // would have written to stdout.
@@ -43,6 +44,11 @@ describe('OTel log export does not leak redacted fields', () => {
           headers: {
             authorization: `Bearer ${SECRET_TOKEN}`,
             cookie: SECRET_COOKIE,
+            // Server-to-server calls (apps/web/lib/api.ts) carry the Clerk JWT
+            // here instead of `authorization` — see auth.guard.ts. Regression
+            // coverage for #767: this leaked into GCP Cloud Logging in plaintext
+            // because it wasn't in the original redact.paths list.
+            'x-clerk-authorization': `Bearer ${SECRET_CLERK_JWT}`,
           },
         },
         res: { statusCode: 500 },
@@ -54,6 +60,7 @@ describe('OTel log export does not leak redacted fields', () => {
 
     expect(serializedLine).not.toContain(SECRET_TOKEN);
     expect(serializedLine).not.toContain(SECRET_COOKIE);
+    expect(serializedLine).not.toContain(SECRET_CLERK_JWT);
     // Sanity check: the line is real content, not an empty/broken write.
     expect(serializedLine).toContain('42501');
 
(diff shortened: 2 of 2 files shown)


brownm09:
## Review: [fix] Redact x-clerk-authorization header to stop Clerk JWT plaintext logging

### Summary
**Mergeable.** Correct, tightly-scoped, and well-tested fix. The redact-path syntax matches pino's bracket notation (identical to the existing `req.headers["set-cookie"]` entry), the key is lowercase to match Fastify's header normalization and `auth.guard.ts`'s own `request.headers['x-clerk-authorization']` read, and the test uses a *distinct* secret value so it genuinely proves this header is redacted rather than riding on the existing `authorization` coverage. No blocking findings. One non-blocking observation about the systemic pattern this bug came from.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/app.module.ts:20-28` (the `redact.paths` denylist)
Redaction here is a denylist — it strips specifically-named auth headers. This PR exists precisely because the denylist was incomplete: `authorization` was covered but the second bearer-carrying header (`x-clerk-authorization`) was forgotten and leaked for weeks. The same failure mode recurs for the next auth-bearing header anyone introduces (a future API-key header, a new `x-…-token`, etc.) — it will log in plaintext until someone remembers to add it here, with no signal until it's found in production again.
**Suggestion:** Not for this PR (the targeted fix is right), but worth a follow-up: either flip to an allowlist of loggable headers, or add a test that asserts the redact config strips *any* header whose name matches `/authorization|token|cookie|api[-_]?key/i`, so a newly-introduced auth header fails CI rather than leaking silently. If you agree it's worth tracking, file it and link from the PR.

---

### Findings Disposition
Per ADR-028, the one non-blocking finding must be fixed-in-PR or filed-and-linked before merge. Recommended: file it as a follow-up issue (it's a broader hardening change, out of scope for this 2-line fix) and link it in a "Review findings disposition" PR-body section. Nothing here blocks merge.

---

brownm09:
## Re-review — x-clerk-authorization redaction (branch updated to `main`)

Re-review after the branch was brought up to date with `main` (merge `3d3d94b`), refreshing the Review Gate marker.

Re-read the full diff (2 files, +14/-2):
- `app.module.ts` adds `'req.headers["x-clerk-authorization"]'` to `redact.paths` using pino's bracket notation (matching the existing `set-cookie` entry), with the lowercase key Fastify normalizes headers to and `auth.guard.ts` reads.
- `otel-log-redaction.spec.ts` extends the existing end-to-end test with a **distinct** secret value under `x-clerk-authorization`, so it genuinely proves the new header is stripped from both the Pino wire format and the resulting OTel LogRecord rather than riding on the existing `authorization` coverage.

Correct and tightly scoped. **0 blocking, 1 non-blocking.**

### Findings Disposition
- **[maintainability] redaction denylist is incomplete-by-design** → filed as [#780](https://github.com/brownm09/lifting-logbook/issues/780) (disposed in PR body).

---

PR #774, opened 2026-07-09
[docs] Replace stale single-draft journal workflow with pointer to global sharded convention

brownm09:
## Summary

The project `CLAUDE.md`'s **Engineering Journal** section documented the retired
single-draft-file-per-day journal workflow (`sessions/lifting-logbook/YYYY-MM-DD_draft.md`
hand-edited across sessions via `Edit`, composed once at day end). The global `CLAUDE.md`
(brownm09/dev-env `claude/CLAUDE.md`) has since moved to a **sharded** system — per-session
`*.stub.md` files, `*.manifest.jsonl` shards, per-PR `open-prs/<N>.json` shards, and the
`/journal-compose` skill — which this repo's own session hooks already enforce live.

This replaces the duplicated (now stale) workflow with a short pointer to the global Engineering
Journal section — matching the pattern sibling projects (career-playbook, leadership-playbooks,
tech-leadership-reference) already follow: they don't duplicate the workflow at all. The only
genuinely lifting-logbook-specific detail — the `sessions/lifting-logbook/` project journal path —
is preserved.

## Changes

- **`## Engineering Journal`** — replaced the `Draft file workflow`, `Draft structure during the
  day`, `Canonical 11-section structure`, and `Update triggers` subsections (110 lines) with a
  concise pointer to the global CLAUDE.md → Engineering Journal section + the `docs/REFERENCE.md`
  internals, keeping the `sessions/lifting-logbook/` path. Added a historical note linking #772.
- **Standard Issue Workflow steps 12–13** — fixed the two stale "append to the day's draft" /
  "block to the draft" references to say "stub".

Net: 23 insertions, 110 deletions.

## Testing

Docs-only change to `CLAUDE.md` — touches zero code paths. Per the project coverage table
("Refactor / docs only → None required — existing tests must pass"), no new coverage is warranted
and the full suite was not run locally; CI runs `turbo run lint typecheck test` on the PR.

- Pre-commit `turbo run lint` hook: **7 tasks successful, 7 total** (ran on commit `5b30ebe`).
- Pre-PR suppression grep (`git diff origin/main | grep -E '(ts-ignore|ts-expect-error|eslint-disable|...)'`): **CLEAN**.
- Pre-PR test-integrity grep (`git diff origin/main | grep -E '(it\.skip|xit\(|...|passWithNoTests|--bail|...)'`): **CLEAN**.
- No `package.json` change → lockfile-sync check N/A.

## Acceptance criteria

[...]

PR #775, opened 2026-07-09
[fix] Reschedule: bounds-check cycleNum/workoutNum and add shared request type

### LINKED ISSUE #773: [fix] Reschedule: validate cycleNum/workoutNum bounds and add shared request type
## Summary

Two minor, pre-existing code-quality gaps in the reschedule-workout feature (`PATCH /programs/:program/cycles/:cycleNum/workouts/:workoutNum/reschedule`), surfaced in passing during the production-reschedule investigation session (#765; production fixes landed in #770/#771). Neither is causing a current failure — these are hardening items.

## Gap 1 — No bounds validation on `cycleNum` / `workoutNum` against actual program state

`RescheduleController` (`apps/api/src/programs/reschedule.controller.ts`) validates only that `cycleNum` and `workoutNum` are **positive integers** (`isValidWorkoutNum` in `mappers.ts` checks nothing more). It does **not** check:

- **`workoutNum` against the program's real workout count.** The GET workout endpoint (`workouts.controller.ts`) already rejects an out-of-range `workoutNum` (`400`, via `workoutKeyForWorkoutNum`), but reschedule does not — even though it has already loaded the spec.
- **`cycleNum` against the active cycle.** The controller calls `cycleDashboard.getCycleDashboard(program)` purely for its throwing side-effect (`ProgramNotFoundError`) and **discards** the returned `cycleNum`.

Because the entire read model (GET workout, cycle dashboard, `getOverride`) operates **only on the current cycle** (`dashboard.cycleNum`) — the GET workout endpoint does not even accept a `cycleNum` — an override upserted for a non-current cycle or an out-of-range workout is **dead data that is never surfaced**. The client gets a `204` "success" with no visible effect, which a user can reasonably report as "reschedule failed" even though the API returned success.

**Goal:** reject out-of-range `workoutNum` and non-active `cycleNum` with a clear `400` (reusing the already-loaded spec and dashboard — no new queries), instead of silently writing an override that can never be read.

## Gap 2 — No shared type contract for the reschedule request

There is no `RescheduleRequest` type in `packages/types/src/api.ts`. The request body shape (`{ newDate: string }`) is defined independently in `reschedule.dto.ts` (backend, with `class-validator` decorators) and hard-coded again in `packages/api-client/src/index.ts` (`{ newDate }`), with no compile-time link between them. The only trace of the feature in the shared types is the read-side `WorkoutResponse.overrideDate` field.

**Goal:** add a shared `RescheduleRequest` type and reference it from both the DTO and the api-client, matching the shared-type convention this codebase uses for other request bodies, so the two cannot drift silently. (The endpoint returns `204 No Content`, so no `RescheduleResponse` is warranted.)

## Out of scope (follow-up)

The `skip` endpoint (`workout-skip.controller.ts`) has the identical positive-integer-only validation and additionally never loads the spec or dashboard (no program-existence or cycle check at all). It will be captured as a separate follow-up rather than expanding this PR.

## Acceptance criteria

- [ ] Reschedule rejects a `workoutNum` beyond the program's canonical length with `400` (consistent with the GET workout endpoint).
- [ ] Reschedule rejects a `cycleNum` that is not the active cycle with `400`, reusing the already-fetched dashboard (no added queries).
- [ ] Existing valid reschedules (current cycle, in-range workout) continue to return `204`.
- [ ] A shared `RescheduleRequest` type exists in `packages/types` and is referenced by both `reschedule.dto.ts` and the api-client, with a compile-time link.
- [ ] Unit (`reschedule.controller.spec.ts`) and E2E (`programs.e2e.spec.ts`) coverage for both new `400` paths.
- [ ] `skip`-endpoint hardening captured as a follow-up.

brownm09:
## Summary

Two minor, pre-existing hardening gaps in the reschedule-workout endpoint (`PATCH /programs/:program/cycles/:cycleNum/workouts/:workoutNum/reschedule`), surfaced in passing during the production-reschedule investigation (#765; production fixes landed in #770/#771). Neither caused a current failure.

Closes #773.

### Gap 1 — Bounds validation (the real defect)

The controller validated only that `cycleNum`/`workoutNum` were **positive integers**. It never checked `workoutNum` against the program's actual length, and it fetched `getCycleDashboard(program)` purely for its throwing side-effect while **discarding** the returned `cycleNum`. Because the entire read model (GET workout, cycle dashboard, `getOverride`) reads **only the active cycle**, an override upserted for a non-active cycle or an out-of-range workout is **dead data the client can never see** — a `204` "success" with no visible effect, reportable as "reschedule failed".

- Reject a `workoutNum` past the program's canonical length → `400`, using `workoutKeyForWorkoutNum` — the *same* bound the GET workout endpoint already enforces.
- Reject a `cycleNum` that is not the active cycle → `400`, reusing the dashboard the controller **already fetched** (zero new queries — the return value was previously discarded).

### Gap 2 — Shared request type

Added `RescheduleRequest` to `packages/types`. `RescheduleDto implements RescheduleRequest` (keeps its `class-validator` decorators) and the api-client types its request body as `RescheduleRequest`, so the request contract can no longer drift between backend and client. No `RescheduleResponse` — the endpoint returns `204 No Content`; the result is surfaced on the read side via `WorkoutResponse.overrideDate`.

## Acceptance Criteria

- [x] Reschedule rejects a `workoutNum` beyond the program's canonical length with `400` (consistent with the GET workout endpoint).
- [x] Reschedule rejects a `cycleNum` that is not the active cycle with `400`, reusing the already-fetched dashboard (no added queries).
- [x] Existing valid reschedules (active cycle, in-range workout) continue to return `204`.
- [x] Shared `RescheduleRequest` type in `packages/types`, referenced by both `reschedule.dto.ts` (via `implements`) and the api-client, with a compile-time link.
- [x] Unit + E2E coverage for both new `400` paths.
- [x] `skip`-endpoint hardening captured as a follow-up (identical positive-integer-only gap; see below).

## Testing

`npm run build` (types touched) + `npm run typecheck` (blocking gate — validates the `implements` link) + `npm test`, all from the repo root. Docker was up, so the api DB E2E suite ran against a real Testcontainers Postgres (no skip escape hatch used).

**Tests: 1073 passed, 0 skipped, 0 failed**

- `@lifting-logbook/api`: 480 passed / 37 suites — incl. `reschedule.controller.spec.ts`, `programs.e2e.spec.ts`, `programs.db.e2e.spec.ts`
- `@lifting-logbook/web`: 244 passed / 40 suites
- `@lifting-logbook/core`: 322 passed / 39 suites
- `@lifting-logbook/api-client`: 27 passed / 1 suite
- `@lifting-logbook/types`: type-only (no runtime test suite; covered by typecheck)

New coverage:
- **Unit** (`reschedule.controller.spec.ts`): rejects a `workoutNum` beyond program length (`400`); rejects a non-active `cycleNum` (`400`); existing valid-input and `ProgramNotFoundError` tests updated to a realistic spec + active-cycle dashboard.
- **E2E** (`programs.e2e.spec.ts`): out-of-range `workoutNum` (`9999`) → `400`; non-active `cycleNum` (`2`, active is `1`) → `400`.

Pre-PR **suppression** and **test-integrity** greps: both clean. No test files deleted, no thresholds lowered, no `package.json`/lockfile change.

## Risk audit (Plan-then-optimize → Pass 3)

- **Testing** — unit + E2E for both new `400` paths (above).
- **Observability** — the two new validation failures return `400` via Nest's exception layer and are auto-logged by `nestjs-pino` with `trace_id`/`span_id` correlation, identical to the sibling validators. No new instrumentation needed; no raw SQL, no LLM call site touched.
- **Security** — a hardening change: prevents writing override rows addressed by an arbitrary cycle/workout number. Per-user isolation (`factory.forUser`) unchanged; no authz surface added.
- **Resilience** — pure input validation before the write; no new failure modes. `getCycleDashboard`'s existing `ProgramNotFoundError` behavior is preserved. Revert is a pure code change.
- **Performance** — **zero added queries**: reuses the already-loaded spec and the dashboard that was previously fetched-and-discarded.
- **Data integrity & migrations** — prevents orphan/dead override rows; no schema change, no migration.
- **Accessibility** — N/A (API only; `RescheduleForm.tsx` and other web code unchanged).

## Follow-up

The `skip` endpoint (`workout-skip.controller.ts`) has the identical positive-integer-only validation and additionally never loads the spec or dashboard (no program-existence or cycle check at all). Deliberately out of scope here to keep this PR focused; captured as a separate follow-up.

- **[maintainability] comment precision** (`reschedule.controller.ts` workoutNum-bound comment) — **fixed in 611ff2a**: reworded to state the check uses the program's spec-derived canonical length via `workoutKeyForWorkoutNum`, and that GET's effective bound is marginally looser (it also honors a scheduled row's `weekNum`); the two align because schedules are never generated beyond the canonical length.
- **Question — 400 vs 404 for a non-active `cycleNum`** — **no change, deliberate**: 400 chosen for consistency with the sibling `workoutNum` guard and the GET workout endpoint's bound message.

Review: https://github.com/brownm09/lifting-logbook/pull/775#issuecomment-4929022779

### DIFF
--- a/apps/api/src/programs/programs.e2e.spec.ts
+++ b/apps/api/src/programs/programs.e2e.spec.ts
@@ -184,6 +184,26 @@ describe('Programs HTTP (e2e, in-memory adapters)', () => {
       expect(res.statusCode).toBe(400);
     });
 
+    it('PATCH reschedule with out-of-range workoutNum returns 400', async () => {
+      // 9999 is a positive integer (passes the old check) but far past the
+      // program's canonical length — an override there is dead data.
+      const res = await patchJson(
+        `/programs/${SEED_PROGRAM}/cycles/1/workouts/9999/reschedule`,
+        { newDate: '2026-06-01' },
+      );
+      expect(res.statusCode).toBe(400);
+    });
+
+    it('PATCH reschedule with non-active cycleNum returns 400', async () => {
+      // Seed's active cycle is 1 (these tests run before any new-cycle writes).
+      // Cycle 2 is not the active cycle, so an override there could never be read.
+      const res = await patchJson(
+        `/programs/${SEED_PROGRAM}/cycles/2/workouts/1/reschedule`,
+        { newDate: '2026-06-01' },
+      );
+      expect(res.statusCode).toBe(400);
+    });
+
     it('PATCH reschedule requires auth', async () => {
       const injectRaw = app.getHttpAdapter().getInstance().inject.bind(
         app.getHttpAdapter().getInstance(),
--- a/apps/api/src/programs/reschedule.controller.spec.ts
+++ b/apps/api/src/programs/reschedule.controller.spec.ts
@@ -11,7 +11,24 @@ import { RescheduleController } from './reschedule.controller';
 const MOCK_USER_A = { id: 'user-a', email: 'a@example.com', provider: 'dev' };
 const _MOCK_USER_B = { id: 'user-b', email: 'b@example.com', provider: 'dev' };
 
-const STUB_SPEC = [{ lift: 'Squat' }] as unknown as Awaited<ReturnType<ILiftingProgramSpecRepository['getProgramSpec']>>;
+// A realistic 2-offset week-1 block. `5-3-1` is a registered 12-week program, so
+// workoutKeyForWorkoutNum tiles this block across 12 weeks → 24 workout days;
+// workoutNum 1..24 resolve to a real day. The controller's new upper-bound check
+// needs a spec that actually resolves the tested workoutNum, not a bare stub.
+const baseSpecFields = {
+  increment: 5,
+  order: 1,
+  sets: 3,
+  reps: 5,
+  amrap: true,
+  warmUpPct: '0.4,0.5,0.6',
+  wtDecrementPct: 0.1,
+  activation: 'compound',
+};
+const STUB_SPEC = [
+  { ...baseSpecFields, offset: 0, lift: 'Squat', week: 1 },
+  { ...baseSpecFields, offset: 3, lift: 'Bench Press', week: 1 },
+] as unknown as Awaited<ReturnType<ILiftingProgramSpecRepository['getProgramSpec']>>;
 
 describe('RescheduleController', () => {
   let controller: RescheduleController;
@@ -34,8 +51,10 @@ describe('RescheduleController', () => {
     };
     specRepoA = { getProgramSpec: jest.fn().mockResolvedValue(STUB_SPEC), saveProgramSpec: jest.fn(), deleteSpecRows: jest.fn() } as jest.Mocked<ILiftingProgramSpecRepository>;
     specRepoB = { getProgramSpec: jest.fn().mockResolvedValue(STUB_SPEC), saveProgramSpec: jest.fn(), deleteSpecRows: jest.fn() } as jest.Mocked<ILiftingProgramSpecRepository>;
-    dashboardRepoA = { getCycleDashboard: jest.fn().mockResolvedValue({}), saveCycleDashboard: jest.fn() } as unknown as jest.Mocked<ICycleDashboardRepository>;
-    dashboardRepoB = { getCycleDashboard: jest.fn().mockResolvedValue({}), saveCycleDashboard: jest.fn() } as unknown as jest.Mocked<ICycleDashboardRepository>;
+    // Active cycle is 3 — the valid-input tests reschedule cycle 3; the controller's
+    // new cycle-match check compares the path cycleNum against this dashboard cycleNum.
+    dashboardRepoA = { getCycleDashboard: jest.fn().mockResolvedValue({ cycleNum: 3 }), saveCycleDashboard: jest.fn() } as unknown as jest.Mocked<ICycleDashboardRepository>;
+    dashboardRepoB = { getCycleDashboard: jest.fn().mockResolvedValue({ cycleNum: 3 }), saveCycleDashboard: jest.fn() } as unknown as jest.Mocked<ICycleDashboardRepository>;
     factory = {
       forUser: jest.fn().mockImplementation(async (user) =>
         user.id === MOCK_USER_A.id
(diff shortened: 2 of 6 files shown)


brownm09:
## Review: [fix] Reschedule: bounds-check cycleNum/workoutNum and add shared request type

### Summary
**Merge-ready with one minor comment refinement.** The change is well-scoped (6 files, +109/-7), reuses already-loaded data so it adds zero queries, and both new `400` paths are covered by unit and E2E tests (full suite green: 1073 passed, 0 skipped/failed, DB E2E run against real Postgres). No blocking findings. The one non-blocking item is a precision issue in a code comment about how the new `workoutNum` bound relates to the GET workout endpoint's bound — the two are not exactly identical, and the difference matters only if canonical program lengths ever change.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/src/programs/reschedule.controller.ts:51-58`
The new `workoutNum` guard rejects when `workoutKeyForWorkoutNum(spec, workoutNum, program) === undefined`, and the comment states this is "the same bound the GET workout endpoint already enforces." That is not exactly true: the GET endpoint (`workouts.controller.ts:78`) computes `week = scheduledWorkout?.weekNum ?? workoutKey?.week` and only 400s when **both** are undefined — so GET additionally accepts a `workoutNum` that has a *scheduled row* even if it exceeds the spec-derived canonical length. Reschedule uses the spec-derived bound only. The two align **today** because schedules are always generated to the canonical length, so no scheduled `workoutNum` exceeds it. But if a registered program's canonical length were ever reduced (e.g. `5-3-1` 12→9 weeks) while users hold schedules generated at the old length, GET `/workouts/20` would return 200 (via `scheduledWorkout.weekNum`) while reschedule of the same workout would 400 — a "can view but can't reschedule" inconsistency.
**Suggestion:** Reword the comment to say the bound is the program's spec-derived canonical length (which aligns with GET *because* schedules never exceed that length), rather than implying the two checks are identical. Adding a scheduled-row fallback to match GET exactly would require an extra query for a case that does not occur today, so a clarifying comment is the proportionate fix; revisit only if canonical lengths become mutable.

---

### Questions for Author

**Question:** Should a `cycleNum` that is not the active cycle return `400`, or would `404` be more semantically correct?
**Context:** `reschedule.controller.ts:60-64` throws `BadRequestException` (400) when `cycleNum !== dashboard.cycleNum`. The sibling out-of-range case (`workoutNum` too large) also 400s, and the GET workout endpoint 400s for its analogous bound — so 400 is internally consistent.
**Tradeoffs:** (a) `400` — treats a non-active cycle as a malformed request; consistent with the other guards in this controller and with the GET endpoint's bound message. (b) `404` — treats "that cycle isn't the addressable/active one" as a missing resource; arguably more RESTful but inconsistent with the sibling validations and would need the api-client's `requestNullable` treatment to avoid surfacing as a thrown error. 400 is the better fit here; raising only to confirm the choice is deliberate.

---

### Findings Disposition
- **[maintainability] comment precision (reschedule.controller.ts:51-58):** to be fixed in-PR (reword the comment).
- **Question (400 vs 404):** deliberate — 400 chosen for consistency with the sibling `workoutNum` guard and the GET endpoint; no change.

---

brownm09:
## Review (refreshed after 611ff2a): [fix] Reschedule bounds-check + shared request type

Re-posting to refresh the review marker after commit `611ff2a`, which addressed the one non-blocking finding from the [initial review](https://github.com/brownm09/lifting-logbook/pull/775#issuecomment-4929022779). No new commits change behavior — `611ff2a` is comment-only.

### Status
- **Blocking findings:** none.
- **Non-blocking finding** — [maintainability] `reschedule.controller.ts` workoutNum-bound comment precision: **fixed in `611ff2a`** (reworded to state the bound is the spec-derived canonical length, and that GET's effective bound is marginally looser because it also honors a scheduled row's `weekNum`; the two align because schedules are never generated beyond canonical length).
- **Question (400 vs 404 for a non-active `cycleNum`):** resolved — 400 is deliberate, for consistency with the sibling `workoutNum` guard and the GET endpoint's bound.

All findings dispositioned in the PR body's "Review findings disposition" section. Tests green (1073 passed, 0 skipped/failed). Ready to merge.

---

brownm09:
## Review marker refresh (after branch update to 95036e8)

Branch updated by merging `origin/main` (only incoming commit was #774, docs-only, no overlap with this PR's files) to clear the `BEHIND` state. No code changes to this PR since the [prior review](https://github.com/brownm09/lifting-logbook/pull/775#issuecomment-4929040025); re-posting the marker so the Review Gate reflects the current head.

- **Blocking:** none.
- **Non-blocking** — [maintainability] workoutNum-bound comment precision: **fixed in `611ff2a`** (in this branch's history).
- All findings dispositioned in the PR body. Tests green.

Issue #534, opened 2026-06-14
[security] Runtime carries owner/superuser DB creds via SYSTEM_DATABASE_URL after RLS cutover


brownm09:
## Problem

After the #517 RLS cutover, the runtime (Cloud Run + GKE API) connects to the app DB as the `lifting_app` NOBYPASSRLS role via `DATABASE_URL`. But the same containers still receive `SYSTEM_DATABASE_URL` pointing at the **owner/superuser** role (`lifting-logbook-app`), set in `infra/terraform/main.tf` (`system_database_url` version) and injected as env in `infra/terraform/cloud-run.tf` / the GKE helm chart / `infra/cloud-run/api-service.yaml`.

On the deployed path `DATABASE_URL` is always set, so the active factory is `PrismaRepositoryFactory` and `SystemDbRepositoryFactory` is **never constructed** (`apps/api/src/adapters/factory/repository-factory.module.ts:27-32`). So the owner credential is injected into the runtime but unused there.

## Why it matters

The owner role bypasses RLS on **every** database in the Cloud SQL instance, including the main app DB. A compromised runtime could read `SYSTEM_DATABASE_URL` and connect as the owner, bypassing the tenant isolation that #517 exists to enforce. This caps the security win of the cutover.

## Options

1. **Drop `SYSTEM_DATABASE_URL` from the deployed runtime env** if it is genuinely unused on the Prisma path (confirm across Cloud Run, GKE helm, and `api-service.yaml`). Keep it only for the in-memory/system-DB fallback deployments that actually set it without `DATABASE_URL`.
2. **Least-privilege the system DB role**: grant `lifting_app` the minimal access it needs on the `_system` DB and repoint `SYSTEM_DATABASE_URL` at `lifting_app`, removing the owner credential from the runtime entirely.

## Source

Surfaced by the `/review` of [#533](https://github.com/brownm09/lifting-logbook/pull/533) (RLS production cutover). Pre-existing; deliberately deferred from #533 to keep the gated cutover non-destructive.

## Acceptance criteria

- [ ] Runtime API containers no longer carry owner/superuser DB credentials in their environment, OR a documented justification for why they must.
- [ ] Verified across all three deploy targets (Cloud Run, GKE helm chart, `api-service.yaml`).
- [ ] No regression to the `SystemDbRepositoryFactory` fallback path (deployments that set `SYSTEM_DATABASE_URL` without `DATABASE_URL`).

Issue #480, opened 2026-06-09
[infra] Wire OTel Collector sidecar into the Cloud Run A/B deploy path


brownm09:
## Problem

Follow-up to #474. That PR wired the OTel Collector **DaemonSet** into the GKE deploy path (the primary topology per [ADR-018](../blob/main/docs/adr/ADR-018-observability-stack.md)), so traces/logs/metrics from GKE workloads reach Grafana Cloud. The **Cloud Run A/B replica** (10% traffic per ADR-009) still ships **no telemetry** — its API requests are invisible in Tempo/Loki/Mimir.

A sidecar template exists at `infra/cloud-run/otel-collector-sidecar.yaml`, but wiring it was deliberately deferred from #474 because it is materially riskier than the GKE path:

- **gcloud has no additive sidecar command.** Both `gcloud run deploy --container` and `gcloud run services replace` require specifying the *entire* container set / full service spec ([Cloud Run docs](https://cloud.google.com/run/docs/deploying)). The documented-safe pattern is `gcloud run services describe --format export → inject sidecar → gcloud run services replace`.
- **The Cloud Run API service spec is Terraform-owned** (`google_cloud_run_v2_service.api` with `lifecycle { ignore_changes = [template] }`), and `gcloud run deploy` only updates the image. Any hand-written replace YAML that omits a Terraform-managed field (e.g. `KMS_KEY_NAME`, the `${env_suffix}-connector` VPC connector, per-env scaling, the three secret refs) would silently regress prod.

## Proposed work

1. Generalize `infra/cloud-run/otel-collector-sidecar.yaml` for both envs (it currently hardcodes prod values: `prod-connector`, `lifting-logbook-prod-*`, prod SA, prod endpoints).
2. Add `infra/cloud-run/otel-collector-config.yaml` — the static collector config the sidecar mounts (Cloud Run supports secret volumes, not ConfigMaps), with the same `otlphttp/metrics` → Mimir path #474 established for GKE.
3. Wire the sidecar into the `Deploy API to Cloud Run (staging|production)` steps via `describe → inject → replace` so the Terraform-managed spec is preserved verbatim and only the sidecar container + config volume are added. Staging auto-deploys + smoke-tests before the prod manual gate — use it as the canary.
[...]

PR #782, opened 2026-07-09
[infra] Wire Cloud Run API into the Grafana Cloud OTel pipeline via sidecar

### LINKED ISSUE #768: [infra] Wire Cloud Run into the Grafana Cloud OTel telemetry pipeline

### LINKED ISSUE #781: [bug] Grafana Cloud OTLP/Loki endpoints + credentials are invalid (401/530) — no telemetry lands from either topology
## Problem

The OTel Collector's exports to Grafana Cloud are rejected **at the Grafana boundary**, so no
telemetry (traces/metrics/logs) actually lands — independent of the collector topology. Confirmed
during the #768 Cloud Run wiring work (2026-07-09) against the **staging** project:

- **Loki logs → HTTP 401 Unauthenticated** (`https://logs-prod-021.grafana.net/otlp/v1/logs`)
- **OTLP traces + metrics → HTTP 530** (`https://otlp-gateway-prod-us-central1.grafana.net/otlp/v1/{traces,metrics}`)

Isolated with a **direct `curl` to the Grafana endpoints, bypassing the collector entirely** — same
401/530. So this is the **credentials/endpoints themselves, not the collector wiring**. It affects
**both** the GKE DaemonSet and the new Cloud Run sidecar, since they share the same auth-header
secrets (`lifting-logbook-{stg,prod}-otel-{otlp,loki}-auth-header`) and endpoint URLs.

The endpoints were already flagged as unverified in the repo:
`infra/kubernetes/values/*-otel-collector.yaml` says *"confirm against the Grafana Cloud portal
before rollout, since this couldn't be verified against the live stack from this change."*

### Likely causes (need the Grafana Cloud portal to confirm)

- **530 on the OTLP gateway** (not 401) points at a **wrong region in the endpoint** —
  `otlp-gateway-prod-us-central1` may not be the stack's actual zone.
- **401 on Loki** points at a wrong instance-ID / token / endpoint number (`logs-prod-021`), or an
  expired token.

## Fix

From the Grafana Cloud portal (Stack → Details):

1. Get the correct **OTLP endpoint** (region), **Loki endpoint** (`.../otlp`), the OTLP + Loki
   **instance IDs**, and a **fresh token** (send metrics + logs + traces).
2. Update the non-secret endpoints in **both** places they live:
   - `infra/kubernetes/values/{staging,production}-otel-collector.yaml` (GKE), and
   - the `OTEL_OTLP_ENDPOINT` / `OTEL_LOKI_ENDPOINT` values in the Cloud Run deploy steps of
     `.github/workflows/deploy.yml` (added by #768).
3. Re-run `scripts/bootstrap-otel-secrets.sh` with the correct instance IDs + fresh token to
   refresh the auth-header secrets.
4. Verify: `curl` the endpoints (expect 2xx/400, not 401/530), then confirm traces in Tempo, logs
   in Loki, and `http.server.*` metrics in Mimir.

## Context

Distinct root cause from #768, which wired the Cloud Run **topology** — verified working up to the
Grafana boundary: the sidecar deploys, loads its config, reaches "Everything is ready", and receives
67 log records + 50 metric points from the API over `localhost:4318`, then hits 401/530 on export.
#768 ships the wiring; this issue unblocks telemetry actually landing (for GKE too). Part of the
#765 investigation.

brownm09:
## Summary

Production runs **Cloud-Run-only** (`enable_gke = false`), so the GKE DaemonSet-based OTel Collector never covered it — the api Cloud Run service never got `OTEL_EXPORTER_OTLP_ENDPOINT`, so `apps/api/src/otel.ts`'s SDK exported to a `localhost:4318` with nothing listening and **silently dropped every trace/metric/log**. All four production alert rules had no data.

This wires the api Cloud Run service into Grafana Cloud via a **co-located `otel-collector` sidecar** (a second container in the api instance): the API SDK exports OTLP to `localhost:4318`, the sidecar forwards to Grafana Cloud using the **same** pipeline config, endpoints, and auth-header secrets as the GKE DaemonSet.

Closes #768. Also supersedes the older duplicate **#480** (same sidecar-wiring goal). Part of the #765 investigation (siblings: #770 IAM/CORS, #771 JWT redaction).

## What changed

- **`infra/cloud-run/otel-collector-config.yaml`** (new) — the collector pipeline config, mirroring the GKE `configmap.yaml` byte-for-byte (receivers → `tail_sampling`/`transform/env_label`/`batch` → `otlphttp` exporters). Stored in a Secret Manager secret and mounted at `/etc/otelcol/config.yaml` (Cloud Run mounts config **files** from Secret Manager — there are no ConfigMap volumes).
- **`scripts/inject-otel-sidecar.py`** (new) — builds the two-container manifest by **deriving from the live service** (`gcloud run services describe --format=export`), so every Terraform-managed field (VPC connector, scaling, service account, startup probe) is preserved; it only adds the sidecar + config volume, sets the api's `OTEL_EXPORTER_OTLP_ENDPOINT`, and drops the unused `SYSTEM_DATABASE_URL` (#534). Idempotent.
- **`.github/workflows/deploy.yml`** — both "Deploy API to Cloud Run" steps (staging + production) now: ensure the config secret (new version only when the repo file changed) → `describe` → inject → `gcloud run services replace`.
- **`infra/terraform/cloud-run.tf`** — declares `OTEL_EXPORTER_OTLP_ENDPOINT` on the api container (spec accuracy / first-apply; the live value is set by the deploy step since the service is `lifecycle.ignore_changes = [template]`), and documents the pipeline-managed config secret. **No new IAM** — the api workload SA already has a project-level `roles/secretmanager.secretAccessor` grant (`gke.tf`).
- **Docs** — ADR-018 addendum (deferral resolved), observability runbook (Cloud Run "deferred" → "wired"), `deploy.md`.

## Key decisions

- **`gcloud run services replace`, not `gcloud run deploy --container`.** The early-resolve question (issue item 3): the installed gcloud (569.0.0) *does* accept `--container`, but a single→multi-container transition hits an **unresolvable sidecar-port catch-22** (omit `--port` on the sidecar → the collector image's `EXPOSE` infers one → *"exactly one container with an exposed port"*; add `--port=default` → *"exactly one must specify --port"*). `services replace` gives explicit, deterministic control (only `api` carries `ports:`). Deriving the manifest from the live service (rather than a static template) is what keeps the **untestable production deploy** safe.
- **The config secret is pipeline-managed, not Terraform-managed** — it is non-sensitive and fully derived from a repo file (no operator token, unlike the Grafana auth headers), so CI creates + versions it with no human step. Declaring it in Terraform would 409 against the pipeline-created secret (the same reason the auth-header secrets are script-managed — see `main.tf`).
- **Direct-to-Grafana (no collector) was rejected** — the collector's `transform/env_label` processor is what promotes `deployment.environment.name` to the `deployment_environment_name` **metric** label the alert rules match on (OTLP→Prometheus otherwise strands it on `target_info`).

## Telemetry now lands — the #781 blocker was fixed by #784 mid-review

Validated on **live staging** (owner creds): the sidecar deploys as a 2nd container, loads its config, reaches *"Everything is ready"*, and **receives 67 log records + 50 metric points from the API over `localhost:4318`**. The api container still serves (`/readyz` → 200 through IAM). So the **wiring is verified end-to-end up to the Grafana boundary.**

At that time Grafana Cloud rejected every export (**Loki 401 / OTLP 530**) — isolated with a direct `curl` to the shared **endpoints** (bypassing the collector), affecting **GKE too**. That was tracked in **#781** and has since been **fixed by #784** (landed on `main` while this PR was open): the endpoints pointed at the wrong Grafana stack, so #784 corrected them to `otlp-gateway-prod-us-east-3` / `logs-prod-042`, **verified against the live stack**, and closed #781. This PR's Cloud Run deploy step now uses those same corrected endpoints and the same auth-header secrets as GKE (synced in `dd87117`), so the sidecar's exports reach Tempo/Loki/Mimir. Final end-to-end confirmation of a Cloud Run trace/log in Grafana happens on the first deploy that ships this branch.

Staging was reverted to single-container after validation (left as found).

## Acceptance criteria (#768)

- [x] New Secret Manager secret holding the collector `config.yaml` (pipeline-managed).
- [x] `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` on the api container.
- [x] Resolve early: `--container` vs `services replace` → `services replace` (`--container` port catch-22).
- [x] Validate on staging first.
- [x] ADR-018 addendum.
- [ ] Confirm a trace/log lands in Grafana Tempo/Loki — **unblocked** (the #781 endpoint blocker was fixed by #784, live-verified; the Cloud Run step now uses the corrected endpoints). Wiring verified to the Grafana boundary on staging; final Cloud Run end-to-end confirmation emits on the first deploy of this branch.
- [ ] Alert rules show a `deployment_environment_name="production"` series — the `transform/env_label` processor (identical to the GKE path #784 verified) promotes the label; the production series populates once this branch deploys.

## Testing

No application code (TS/JS) changed — only infra YAML, Terraform, a Python deploy helper, and docs — so the **Jest suite and `tsc` typecheck are unaffected** (CI runs `turbo run lint typecheck test`). Pre-PR **suppression** and **test-integrity** greps: both **clean**. `terraform fmt -check`: clean. `deploy.yml`: valid YAML.

Change-specific validation (infra — no automated test harness for the deploy path, per the repo's "written test plan" convention):
- **Collector config** — `otelcol validate` (v0.104.0, via Docker) against `otel-collector-config.yaml`: **exit 0**.
- **Injector** — ran `scripts/inject-otel-sidecar.py` against the real staging `--format=export`: produces the correct 2-container manifest (api container + all env/probe/resources preserved, sidecar + config volume added).
- **End-to-end** — deployed the injector-equivalent manifest to live staging via `services replace`: revision **Ready**, both containers up, api serves, collector receives app telemetry (see the #781 section). Then reverted staging to single-container.

## Sequencing

Rebased onto `main` after **#770** (IAM/CORS) and **#771** (JWT redaction) landed. The only conflict was in `deploy.yml`'s two api-deploy steps, which #770 had switched to `gcloud run deploy --allow-unauthenticated --remove-secrets=SYSTEM_DATABASE_URL`. Resolved to this PR's `gcloud run services replace` mechanism: the injector already drops `SYSTEM_DATABASE_URL`, and public invoker access is owned by Terraform's authoritative `google_cloud_run_v2_service_iam_member.api_public` (allUsers → `run.invoker`) binding — which `services replace` leaves untouched — so no deploy-time auth flag is needed. Final state = `services replace` + the Terraform allUsers binding.

[/review](https://github.com/brownm09/lifting-logbook/pull/782#issuecomment-4936081339) surfaced 4 blocking + 6 non-blocking findings + 1 question. All addressed (fixes in `9b85b5c`):

**Blocking**
1. **[reliability] CPU-throttling starves the collector** — fixed (wording). The sidecar shares the api revision's CPU-throttling, so the "identical to GKE" claim is softened to best-effort-between-requests in the `deploy.yml` comments, ADR-018 addendum, config header, and runbook. The allocation decision (`cpu_idle=false` vs. accept best-effort) is deferred to **#787** — it can't be measured until telemetry lands (#781), so incurring ~24/7 CPU cost now would be premature.
2. **[correctness] Injector has no automated test** — filed **#786** (Python test harness; this is a Jest/TS monorepo with no Python runner). Injector output was validated manually against the real staging export (see Testing).
3. **[documentation] Stale comments (abandoned `--container`/`--set-env-vars`/`--set-secrets`)** — fixed: corrected in `cloud-run.tf`, the `otel-collector-config.yaml` header, `deploy.yml`, and the injector docstring to reference `services replace` + the injector.
4. **[documentation] `scripts/README.md` entry missing** — fixed (added the `inject-otel-sidecar.py` index row).

**Non-blocking**
1. pyyaml PEP-668 fallback fragility — fixed (`--break-system-packages`).
2. Collector image pulled from Docker Hub by mutable tag — filed **#788** (mirror to Artifact Registry + digest-pin; affects GKE too).
3. Grafana endpoints duplicated 8× — this bit immediately: #784 (the #781 fix) changed the GKE copies mid-review, so this PR's `deploy.yml` Cloud Run copies were synced to match (`us-east-3` / `logs-prod-042`, commit `dd87117`). The structural single-sourcing/drift-check follow-up is tracked on **#788**.
4. `otel-collector-config.yaml` ↔ GKE configmap drift — filed **#788** (CI drift-check).
5. Stale `otel-collector-sidecar.yaml` reference file — fixed (deleted; the injector + config are authoritative).
6. Terraform-recreate leaves a 1-container service — fixed (runbook now documents the pipeline-deploy-must-follow-recreate recovery ordering).

**Question — loss of per-deploy IAM re-assertion:** Resolved by the rebase. Post-#770 the api's public posture is owned by Terraform's authoritative `google_cloud_run_v2_service_iam_member.api_public` (allUsers → `run.invoker`). `gcloud run services replace` deliberately never touches the invoker IAM policy, so — unlike the old `gcloud run deploy --no-allow-unauthenticated` that fought Terraform on every deploy — it neither re-asserts nor removes the grant. Nothing is lost: Terraform re-asserts allUsers on every apply, and the deploy no longer competes with it.


brownm09:
### `staging.yml` compatibility — verified, no change needed

`staging.yml` (the PR-time staging workflow, not touched in this PR) also deploys `lifting-logbook-stg-api`, via flag-based `gcloud run deploy` (its line 544). I verified on live staging that a flag-based `gcloud run deploy --image=... --no-allow-unauthenticated --remove-secrets=SYSTEM_DATABASE_URL` against a **2-container** service **succeeds (exit 0) and preserves the sidecar** (`api;otel-collector`) — it updates only the ingress (api) container's image.

So once `deploy.yml` (on push-to-main) establishes the sidecar, subsequent `staging.yml` PR deploys keep it: **no flip-flop, no `staging.yml` change required, and no risk to the required Staging Integration Tests check.** Staging was reverted to single-container after this verification.

_🤖 [Claude Code](https://claude.com/claude-code)_

brownm09:
**Endpoint corrections from #781 / PR #784 — mirror on this PR's next rebase**

This PR's Cloud Run OTEL wiring hardcodes the Grafana endpoints in `deploy.yml` (lines 568-569 staging, 1132-1133 production) and `infra/cloud-run/otel-collector-sidecar.yaml`, mirroring the GKE collector values. Those values were **wrong** and are the root cause of the no-telemetry bug (#781):

- OTLP `otlp-gateway-prod-us-central1` → **`otlp-gateway-prod-us-east-3`** (wrong region returned `530`)
- Loki `logs-prod-021` → **`logs-prod-042`** (wrong gateway returned `401`)

[PR #784](https://github.com/brownm09/lifting-logbook/pull/784) corrects these on `main` (both `infra/kubernetes/values/*-otel-collector.yaml` + the sidecar reference manifest). **When you rebase this PR onto the post-#784 `main`, mirror the same two corrections into `deploy.yml` lines 568-569 / 1132-1133 and the sidecar `value:` lines here** so the Cloud Run topology exports to the right hosts.

Verified against the live Grafana stack (2026-07-10): the **existing** Secret Manager auth-header secrets return `2xx` at the corrected endpoints in both staging and production — **no credential rotation is needed**, this is purely an endpoint fix.

brownm09:
## Review: [infra] Wire Cloud Run API into the Grafana Cloud OTel pipeline via sidecar

### Summary
**Merge with revisions.** The core wiring is sound and independently verified: two reviewers confirmed the `describe → inject → services replace` round-trip preserves every Terraform-managed field, the ingress-container heuristic is safe (Terraform pins `ports` only on the api container), the injector is idempotent, and failure modes (empty describe, bad sidecar, 401/530) degrade safely without shifting prod traffic. The highest-severity finding is a **reliability** one the PR's own "identical telemetry to GKE" claim glosses over: the sidecar inherits the api revision's CPU throttling (`cpu_idle = true`), which starves the collector's background flush/export timers between requests. Four blocking items below (all cheap except the CPU decision); nothing undermines the topology itself. Appropriately scoped.

---

### Blocking Findings

**[reliability]** `infra/terraform/cloud-run.tf:53` (+ `scripts/inject-otel-sidecar.py` collector `cpu`)
The api revision runs `cpu_idle = true` (Cloud Run "CPU allocated only during request processing"), which is revision-wide and so also throttles the injected collector. A telemetry-forwarding collector does its real work *between* requests — `batch` (5s), `tail_sampling` (`decision_wait` 10s), periodic OTLP export — so under throttling those timers are starved until the next inbound request donates CPU, and buffered telemetry can be dropped on throttle/scale-in (staging `min=0`; prod `min=1` still throttles when idle). The GKE DaemonSet has always-on CPU and no such problem, so the "identical telemetry" claim in the ADR-018 addendum and PR body isn't quite true.
**Fix:** Decide explicitly — either give the api revision always-allocated CPU (`cpu_idle = false`, right-size the collector to <1 vCPU) and accept the ~2 vCPU/768Mi 24/7 cost, **or** soften the "identical to GKE" wording to "best-effort under CPU throttling" and note telemetry can be delayed/dropped on scale-in.

**[correctness]** `scripts/inject-otel-sidecar.py:76`
The injector rewrites the live **production** Cloud Run manifest but has no automated test (nothing in the tree references it), and `main()` is a pure text→text transform that runs against a service the PR itself says "cannot be dry-run tested." A regression — mis-identifying the ingress container, dropping a Terraform-managed field, a non-idempotent re-run — would ship a broken prod manifest with no gate. The PR body describes manual validation but records no formal deferral.
**Fix:** Add a test (fixture = a captured `--format=export`; assert 2-container output, idempotency, `SYSTEM_DATABASE_URL` dropped, `OTEL_EXPORTER_OTLP_ENDPOINT` set, `ports:` only on api, and the missing-`spec.template.spec` guard) — or, since this is a Jest monorepo with no Python harness, document the deferral in the PR body and file a tracking issue for a `scripts/` Python test harness.

**[documentation]** `infra/terraform/cloud-run.tf:99` / `infra/cloud-run/otel-collector-config.yaml:7` / `scripts/inject-otel-sidecar.py:42`
Several comments describe the mechanism this PR **explicitly rejected**: `cloud-run.tf` says the sidecar is "added by the deploy pipeline's multi-container `gcloud run deploy`" and the live env is set "via `--set-env-vars`"; the config file's header says it is mounted "via `gcloud run deploy --set-secrets=...`"; the injector docstring says cpu/memory defaults "mirror `infra/kubernetes/values/*-otel-collector.yaml`" (those files contain no cpu/memory). The actual mechanism is `gcloud run services replace` with env/volume baked into the injected manifest.
**Fix:** Correct all three to reference `services replace` + the injector; fix the docstring so it only claims the sample-rate/decision-wait mirror the values files (cpu=1/256Mi are script defaults).

**[documentation]** `scripts/inject-otel-sidecar.py` (new file) — `scripts/README.md` not updated
`scripts/README.md` is an index of every automation script, grouped by lifecycle (it already lists `bootstrap-otel-secrets.sh`, `migrate-*.sh`, etc.). Adding `inject-otel-sidecar.py` without an entry leaves it undiscoverable to anyone orienting from the index.
**Fix:** Add a row for `inject-otel-sidecar.py` (e.g. under "Production deploy" or the Observability section) describing that the deploy pipeline uses it to inject the OTel sidecar into the Cloud Run api manifest.

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/deploy.yml:563` (prod `:1127`)
`python3 -c "import yaml" 2>/dev/null || python3 -m pip install --quiet pyyaml` installs into the system interpreter. On a PEP-668 "externally-managed" runner (ubuntu-24.04) that `pip install` errors and, under the Actions default `bash -eo pipefail`, aborts the deploy. It works today only because PyYAML is pre-installed on GitHub-hosted images — an undocumented dependency on the runner image.
**Suggestion:** Drop the fallback and rely on the pre-installed PyYAML with a comment, or make it robust (`--break-system-packages`, `apt-get install -y python3-yaml`, or `actions/setup-python`).

**[reliability]** `.github/workflows/deploy.yml` collector image (`scripts/inject-otel-sidecar.py:71`)
The collector image `otel/opentelemetry-collector-contrib:0.104.0` is pulled from Docker Hub by a mutable tag — not mirrored into Artifact Registry like the api image, not digest-pinned. Cloud Run pulls it on every cold-start/scale-up, so a Docker Hub rate-limit (100/6h per IP) or outage can fail new prod request-path instances, and a re-pushed tag breaks reproducibility. (The GKE chart has the same pattern, so this is arguably a broader hardening item.)
**Suggestion:** Mirror the collector image into the project Artifact Registry and reference it by digest (`@sha256:...`).

**[maintainability]** `.github/workflows/deploy.yml:568` (+ prod `:1132`)
The Grafana OTLP/Loki endpoints are hardcoded at 4 sites in deploy.yml and again in `infra/kubernetes/values/{staging,production}-otel-collector.yaml` — 8 copies of the same two URLs, kept aligned only by prose. #781 will replace these, and whoever does it must edit all 8 in lockstep or GKE and Cloud Run silently ship to different endpoints.
**Suggestion:** Single-source them (read the values file in the deploy step, or a shared workflow env) or add a CI drift-check mirroring `scripts/check-turbo-version-sync.mjs`.

**[maintainability]** `infra/cloud-run/otel-collector-config.yaml:1`
This config is copied verbatim from the GKE `configmap.yaml`'s `config.yaml` block (the bodies are textually identical), kept in sync only by the "Keep the two in sync" comment. The PR's whole premise is identical GKE/Cloud Run telemetry; a future edit to one silently diverges them.
**Suggestion:** Add a CI check that extracts the configmap's `config.yaml` and diffs it against this file, or template both from one source.

**[maintainability]** `infra/cloud-run/otel-collector-sidecar.yaml:1`
This PR relabels the old template "ILLUSTRATIVE REFERENCE ONLY … may drift from the live spec." A knowingly-inaccurate, hand-maintained manifest that nothing in the pipeline consumes is a trap — a future operator can mistake it for the deployed shape. The injector + `otel-collector-config.yaml` are now the sole source of truth.
**Suggestion:** Delete the file (the injector is authoritative), or generate it from the injector so it cannot drift.

**[reliability]** `infra/terraform/cloud-run.tf:99`
Terraform declares `OTEL_EXPORTER_OTLP_ENDPOINT` on the api container but cannot define the sidecar (`lifecycle.ignore_changes = [template]`). If the service is ever recreated from Terraform (DR, teardown/reapply, first apply), it comes up 1-container with the SDK exporting to a `localhost:4318` with nothing listening — silently dropping all telemetry (the exact #768 bug) until the next pipeline deploy re-injects the sidecar.
**Suggestion:** Document the recovery ordering (a pipeline deploy must follow any Terraform recreate of this service) in the runbook.

---

### Questions for Author

**Question:** Is losing the per-deploy re-assertion of the private invoker posture acceptable until #770 lands?
**Context:** The api deploy switches from `gcloud run deploy … --no-allow-unauthenticated` to `gcloud run services replace`, which does not manage invoker IAM at all (`deploy.yml:544` old → the new step). The deploy.yml comment already states this is deliberate (IAM untouched; #770 opens it via Terraform `allUsers`), so intent is clear — the only change is loss of self-healing if the policy ever drifts to public unintentionally.
**Tradeoffs:** (a) Rely on Terraform/#770 to own the invoker policy (current choice) — correct and avoids fighting #770, but no per-deploy correction of accidental drift. (b) Add an explicit `gcloud run services remove-iam-policy-binding … allUsers` step until #770 — guarantees the private posture pre-#770 but is throwaway once #770 opens the service.

---

### Findings Disposition
Each finding above must be fixed in this PR or filed-and-linked before merge (ADR-028); record the disposition (fixed in `<sha>` | filed #N) in a "Review findings disposition" section of the PR body before `gh pr merge`. The quick doc fixes (stale comments, `scripts/README.md` entry, deleting the stale reference file) are in-PR; the CPU-throttling decision, injector test, and the CI drift-checks/image-mirroring may warrant filed follow-ups.

---

brownm09:
## Re-review — head `44da3f4`

Re-reviewed after the rebase onto `main` (resolving the #770 conflict), the disposition of the original review's findings, and the #784 endpoint sync. An independent reviewer verified the four areas that changed since the [original review](https://github.com/brownm09/lifting-logbook/pull/782#issuecomment-4936081339):

- **IAM / public access — preserved.** Terraform's `google_cloud_run_v2_service_iam_member.api_public` (allUsers → `roles/run.invoker`) survives the rebase, and `gcloud run services replace` replaces only the Knative *spec* — it does **not** touch the invoker IAM policy — so the api service stays publicly invokable without the old `--allow-unauthenticated` flag. The injector also preserves the `run.googleapis.com/ingress` annotation.
- **Endpoints — exact match.** All 4 `deploy.yml` Cloud Run sites byte-match #784's corrected `otlp-gateway-prod-us-east-3` / `logs-prod-042` values in the GKE values files; zero stale `us-central1` / `logs-prod-021` anywhere.
- **Comment/doc edits — accurate.** No executable line changed by mistake; `--break-system-packages` is valid. The two minor doc inaccuracies the re-review flagged (config-secret attribution; a nonexistent deploy.yml step name) are fixed in `44da3f4`.
- **File deletion — clean.** `otel-collector-sidecar.yaml` removed; zero remaining references.

**No blocking issues; merge-safe.** The original review's 4 blocking + 6 non-blocking findings are all disposed (see the *Review findings disposition* section in the PR body; follow-ups #786 / #787 / #788, endpoint single-sourcing noted on #788).



*Re-reviewed by `claude-opus-4-8` via Claude Code*

PR #790, opened 2026-07-10
[chore] Consolidate Grafana OTLP/Loki endpoints into a single source

brownm09:
## Summary

The Grafana Cloud OTLP/Loki ingest endpoints were hardcoded in **4 identical committed spots** — `deploy.yml`'s staging + production Cloud Run sidecar-inject steps, and the staging + production `otel-collector` Helm values files. That duplication is the exact drift surface behind no-telemetry incident #781 (fixed in #784). This PR consolidates them into a single source of truth so a future region/gateway change is a one-line edit.

## What changed

**Single source:** new `infra/observability/grafana-endpoints.env` holds the two (non-secret) endpoints. Staging and production share one free-tier Grafana stack, so the values are identical across environments — only the auth-header *secret names* differ (unchanged, still per-env in Secret Manager).

**Every consumer now derives from it** (all orchestrated by `deploy.yml`):
- **Cloud Run sidecar** (staging + prod): the step `source`s the file; `scripts/inject-otel-sidecar.py` reads the exported `OTEL_OTLP_ENDPOINT` / `OTEL_LOKI_ENDPOINT`. The 2 hardcoded literal pairs are removed.
- **GKE DaemonSet** (staging + prod): the step `source`s the file and passes the endpoints to `helm upgrade` via `--set-string`. The values files' endpoint literals are blanked to `""` with a pointer comment (and the documented manual-helm command now shows the `--set-string`).

**Guardrails against regression / silent failure:**
- `scripts/check-grafana-endpoint-sources.mjs` (+ `node:test` unit test) — both wired into `ci.yml` → `lint-and-test` — fail the build if a literal endpoint is re-hardcoded anywhere outside the single source (comments and docs are allowed).
- The `otel-collector` chart's configmap now `required`s a non-empty endpoint, so a manual `helm upgrade` that forgets `--set-string` fails loudly rather than silently shipping no telemetry (the #781 failure mode). Verified with `helm template` both ways.
- `.gitattributes` pins `grafana-endpoints.env` to LF so a CRLF commit cannot append a trailing carriage return to the endpoint values on the Linux CI runner.

The Cloud Run collector config (`otel-collector-config.yaml`) already used `${env:...}` placeholders — no literal there; only its comment was updated to point at the source.

## Acceptance criteria
- [x] Endpoints live in exactly one committed location; a region/gateway change is a one-line edit.
- [x] All topologies (GKE DaemonSet + Cloud Run sidecar, staging + prod) read from that source.
[...]

Issue #785, opened 2026-07-10
[chore] Consolidate duplicated Grafana OTLP/Loki endpoints into a single source


brownm09:
## Problem

The Grafana Cloud OTLP/Loki endpoint URLs are hardcoded in **4+ committed locations**:

- `infra/kubernetes/values/staging-otel-collector.yaml`
- `infra/kubernetes/values/production-otel-collector.yaml`
- `infra/cloud-run/otel-collector-sidecar.yaml`
- `.github/workflows/deploy.yml` (lines ~568-569 and ~1132-1133 — present on the #768 branch / [#782](https://github.com/brownm09/lifting-logbook/pull/782); lands on `main` when that merges)

This duplication is the exact drift surface that produced the no-telemetry incident **#781** (OTLP region `us-central1`→`us-east-3`, Loki gateway `logs-prod-021`→`logs-prod-042`), fixed in [#784](https://github.com/brownm09/lifting-logbook/pull/784). A future region/gateway change must touch every copy or they silently diverge again — and the `deploy.yml` mirror on #782 is currently diverged for precisely this reason.

## Proposed fix

Consolidate the endpoints into a single source of truth that all topologies (GKE DaemonSet + Cloud Run sidecar) read, so a future endpoint change is a one-line edit. Note: staging + production share one free-tier Grafana stack, so the endpoint values are identical across envs (only the auth-header secret names differ by prefix).

## Sequencing

Best done **after** #782 merges, so the `deploy.yml` occurrences can be folded into the same consolidation rather than chased separately.

Surfaced by the `/review` of #784.

PR #793, opened 2026-07-10
[config] Enforce LF line endings by construction for shell/env files

### LINKED ISSUE #792: Broaden .gitattributes LF enforcement to all bash-sourced/executed shell & env files
## Problem

PR #790 (issue #785) added a repo-root `.gitattributes` that pins **only** `infra/observability/grafana-endpoints.env` to LF (`text eol=lf`). That file is `source`d by bash on the Linux CI runner in `.github/workflows/deploy.yml`; a CRLF commit would append a trailing `\r` to each `export OTEL_..._ENDPOINT=...` value and silently break the Grafana endpoint URLs (the same no-telemetry failure class as #781).

But the repo has other bash scripts that are `source`d or executed on Linux and currently rely on `core.autocrlf=true` (the Windows default) to stay LF in the git blob — e.g. `scripts/observability/mimir-query-env.sh`, `scripts/observability/mimir-setup.sh`, `scripts/bootstrap-otel-secrets.sh`, `scripts/smoke-test-observability.sh`. A contributor with `core.autocrlf=false` (or `input` misconfigured) could commit CRLF into any of them, breaking execution on Linux (`bash\r: bad interpreter`, or commands with a trailing `\r`).

## Proposed fix

Broaden `.gitattributes` so line endings are pinned **by construction**, not by each contributor's git config — e.g. `*.sh text eol=lf` and `*.env text eol=lf` (evaluate `* text=auto` as a baseline). Verify with `git add --renormalize .` that nothing re-normalizes unexpectedly and no `.sh`/`.env` blob flips to CRLF.

Follow-up to #785 / #790. Surfaced as a post-merge tile.

brownm09:
## Summary

Follow-up to #790 (#785). PR #790 pinned **only** `infra/observability/grafana-endpoints.env` to LF, because it is `source`d by bash on the Linux CI runner and a CRLF commit would append `\r` to each endpoint URL (the #781 no-telemetry class). This generalizes that protection to every bash-run/sourced file, so line endings are enforced **by construction**, not by each contributor's `core.autocrlf`.

## What changed

`.gitattributes`:
- `* text=auto` — all text files are normalized to LF in the git blob; a contributor with `core.autocrlf=false` (or `input` misconfigured) can no longer commit CRLF for a text file.
- `*.sh`, `*.bash`, `*.env` → `text eol=lf` — files run or `source`d by bash on Linux (CI runner + prod containers) are LF in the working-tree checkout too, so they never break on Linux (`/bin/bash^M: bad interpreter`, or a trailing CR on a sourced value). Covers the previously-unprotected `scripts/observability/mimir-query-env.sh`, `mimir-setup.sh`, `scripts/bootstrap-otel-secrets.sh`, `scripts/smoke-test-observability.sh`, and the rest of `scripts/**`.
- `*.png binary` — the only binary type present (apps/mobile assets); marked explicitly so `* text=auto` can never misdetect and line-ending-mangle it.
- The standalone `grafana-endpoints.env` line from #790 is folded into `*.env` (still resolves to `eol=lf` — verified with `git check-attr`).

## Verification (config-only change — no runtime/test surface)

- **`git add --renormalize .` stages nothing but `.gitattributes`** — all tracked blobs were already LF (historical `core.autocrlf=true`), so there are **zero file-content changes** and no binary corruption. Confirmed `git ls-files --eol` reports `i/crlf` on 0 files.
- `git check-attr` confirms: `.sh`/`.env` → `text=set, eol=lf`; `.ts`/`.ps1`/`package.json` → `text=auto`; `.png` → binary; `grafana-endpoints.env` → `eol=lf` (the #785 protection is preserved).
- No `package.json` change (lockfile policy N/A). No suppressions, no test changes.

## Risk audit
- **Resilience / failure modes:** the whole point — removes a latent CRLF-on-Linux foot-gun for every bash-sourced/executed script. Revert restores the #790-only pin. Zero blob changes ⇒ no behavioral risk to existing files.
- **Data integrity:** `*.png binary` + the clean renormalize confirm no binary is mangled.
- **Testing / Security / Observability / Performance:** N/A — repo-config only, no code or runtime surface.

Closes #792

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,22 @@
-# Files that are `source`d by bash on the Linux CI runner must stay LF-only, so a
-# sourced value never carries a trailing CR. A CRLF grafana-endpoints.env would
-# append "\r" to each endpoint URL and silently break telemetry — the same
-# no-telemetry failure class as #781 that #785 exists to prevent.
-infra/observability/grafana-endpoints.env text eol=lf
+# ── Line-ending policy ───────────────────────────────────────────────────────
+# Enforce line endings by construction, not by each contributor's core.autocrlf
+# (the Windows default — a clone with autocrlf=false, or input misconfigured,
+# could otherwise commit CRLF). All text files are normalized to LF in the git
+# blob; git auto-detects text vs binary and leaves binaries untouched.
+* text=auto
+
+# Files run or `source`d by bash on Linux — the CI runner and the prod containers —
+# must be LF in the working-tree checkout too, not just in the blob. A CRLF copy
+# breaks execution on Linux: `/bin/bash^M: bad interpreter`, or a trailing CR on a
+# sourced value (a CRLF grafana-endpoints.env would append "\r" to each Grafana
+# endpoint URL and silently break telemetry — the #781 no-telemetry failure class
+# that #785/#790 fixed for that one file; #792 generalizes it to every such file).
+*.sh   text eol=lf
+*.bash text eol=lf
+*.env  text eol=lf
+
+# Binaries: mark explicitly so `* text=auto` can never misdetect and line-ending-
+# mangle them. Git's content detection already handles these, but an explicit rule
+# is robust against detection edge cases and future files. (Only *.png is present
+# today — the apps/mobile assets.)
+*.png binary


brownm09:
## Review: [config] Enforce LF line endings by construction for shell/env files

### Summary
**Merge — no changes needed.** Minimal, correct, well-verified `.gitattributes` change. `* text=auto` plus `*.sh`/`*.bash`/`*.env` → `eol=lf` generalizes #790's single-file LF pin so no bash-run/`source`d script can be committed with CRLF and break on the Linux runner (`/bin/bash^M: bad interpreter`, trailing `\r` on sourced values). `*.png binary` keeps `* text=auto` from touching the only binary type present. `git add --renormalize .` stages nothing but `.gitattributes` (every tracked blob was already LF), so there is **zero content/blob risk** to existing files, and `grafana-endpoints.env` still resolves to `eol=lf` — preserving the #785 guarantee.

### Blocking Findings
None.

### Non-Blocking Findings
None. (`*.png` is the only binary type in the repo today; if other binary types are added later, `* text=auto`'s content detection handles them, and an explicit rule can be added then. Not worth pre-adding speculative rules now.)

Issue #809, opened 2026-07-11
[infra] Validate the /api/client-errors same-origin guard in staging and enable enforcement


brownm09:
## Problem

#806 shipped a same-origin guard for `POST /api/client-errors` in **observe-only** mode: it tags every recorded span with `client.origin.check`, but only *drops* cross-origin beacons when `CLIENT_ERROR_ALLOWED_ORIGINS` is configured **and** `CLIENT_ERROR_DROP_CROSS_ORIGIN=true`. Enforcement was deliberately deferred: a false-drop (legit same-origin traffic misclassified behind the LB) would silently kill best-effort telemetry, so it must be validated against the real staging proxy chain first.

## Steps

1. Set `CLIENT_ERROR_ALLOWED_ORIGINS` to the web app's public origin(s) in the staging deploy env (Cloud Run web service / GKE values).
2. Deploy to staging; trigger a real client-error beacon (force a mutation failure in the staging UI).
3. In staging Tempo, confirm legitimate beacons tag `client.origin.check=same-origin` (never a false `cross-origin`, no `client.origin.enforce_skipped`).
4. Once confirmed, set `CLIENT_ERROR_DROP_CROSS_ORIGIN=true` in staging, then production.
5. Wire the same env into the prod web deploy — couple with #804 (which makes web server spans reach prod Tempo).

## Context

- Depends on #804 (web spans → prod Tempo) for the risk to be live; enable before / with it.
- ADR-020 #806 addendum, mitigation (2).

Epic: Observability. Follows #806; coupled to #804.

Issue #808, opened 2026-07-11
[infra] Infra-level rate limit for the unauthenticated /api/client-errors endpoint (Cloud Armor / edge)


brownm09:
## Problem

`POST /api/client-errors` (apps/web) is public/unauthenticated by necessity and records a **retained ERROR span** per accepted request. #806 added an app-level same-origin guard that drops cross-origin *browser* beacons, but it cannot stop scripted abuse: a `curl` loop sends no truthful `Origin`, so the guard classifies it `no-origin` and allows it (dropping no-Origin requests would also drop legitimate non-browser reports). The robust fix for scripted volume is an **infra-level rate limit** on the endpoint.

Deferred from #806, which explicitly scoped this out of an app-level PR.

## Scope

- No Cloud Armor / security policy exists in `infra/terraform/` today — greenfield.
- Add a rate-limit policy scoped to the unauthenticated `/api/client-errors` path at the edge / load balancer (Cloud Armor `rate_based_ban` / throttle, or an ingress-level equivalent).
- Keep the threshold generous enough not to drop legitimate best-effort telemetry bursts (a single failing page can emit several beacons).

## Context

- ADR-020 #806 addendum records this as mitigation (3) — the deferred infra layer.
- The abuse surface is latent until #804 wires web server spans to prod Tempo.

Epic: Observability. Follows #806.

Issue #807, opened 2026-07-11
[test] Web onboarding CSV-heavy suites (StepImport/StepLifts) still OOM under full-suite Windows parallel load


brownm09:
## Symptom

Running `npm test -w @lifting-logbook/web` locally on Windows (Node 20.11.1) intermittently fails two suites at load:

```
FAIL app/(authed)/onboarding/steps/StepImport.test.tsx
  ● Test suite failed to run — Jest worker ran out of memory and crashed
FAIL app/(authed)/onboarding/steps/StepLifts.test.tsx
  ● Test suite failed to run — Jest worker ran out of memory and crashed

Test Suites: 2 failed, 42 passed, 44 total
Tests:       259 passed, 259 total
```

Every test that *runs* passes (259/259). The two suites crash at **load** (worker OOM), not on an assertion. Run in isolation they pass cleanly:

```
npx jest --config apps/web/jest.config.js --runInBand "StepImport.test|StepLifts.test"
→ Test Suites: 2 passed, 2 total; Tests: 18 passed
```

## Why this isn't already covered

The Windows worker-OOM / parallel-load flake class was addressed by codified workarounds in #419 (`workerIdleMemoryLimit: '512MB'` + `maxWorkers: '50%'` on win32, in `jest.config.base.js`) and #567 (`testTimeout: 15000` on win32). **Both are closed.** Those workarounds reduce but do not eliminate the OOM for these two CSV-fixture-heavy onboarding suites under a full `-w @lifting-logbook/web` parallel run — hence a fresh open anchor.

## Scope / impact

- **Local Windows only.** CI (Linux, Node 20) is unaffected and these suites pass there.
- Not a correctness failure: passes in isolation, all assertions green.

## Candidate fixes (triage)

- Lower the web project's win32 `maxWorkers` further, or set a web-specific `workerIdleMemoryLimit` below the base `512MB`.
- Split / lazily load the CSV fixtures used by StepImport/StepLifts.
- Run these two suites in-band on win32.

Observed while hardening `/api/client-errors` (#806); the failing suites are unrelated to that change.

Epic: CI/CD Foundation.

PR #805, opened 2026-07-11
[feat] Export web client-side mutation errors to Grafana via beacon → OTel span

brownm09:
## Summary

Follow-up to #783 (PR #800). #783 routed every `apps/web` client-side mutation failure through the shared `apps/web/lib/log-client-error.ts` helper (`console.error` in a consistent shape) — meeting the issue's **minimum** bar (browser devtools) but not its **ideal** (landing in Grafana). This PR wires the ideal: `logClientError` now additionally sends a best-effort same-origin `navigator.sendBeacon` to a new `POST /api/client-errors` Next route handler, which records the error as an OTel span (`client.mutation.error`, status `ERROR`) through `apps/web`'s existing `@vercel/otel` trace pipeline → Tempo (retained by the [ADR-020](docs/adr/ADR-020-tail-based-sampling-policy.md) tail-sampling `errors` policy). The `console.error` from #783 is preserved.

> Builds on #783 / PR #800 (the shared `log-client-error.ts` helper), which merged into `main` while this was in progress — so this PR targets `main` directly.

## Design — why a same-origin beacon + span, not a browser OTel SDK

Two options were on the table (#798):
1. A browser OTel SDK (`@opentelemetry/sdk-trace-web` + OTLP/HTTP) posting to the collector directly.
2. A same-origin beacon to a Next route handler.

**Chose (2).** The collector is internal-only — `ClusterIP` on GKE, a localhost sidecar on Cloud Run (#768) — so it is **not browser-reachable**. Option (1) would require publicly exposing an **unauthenticated** OTLP ingest with CORS: a spam/cost vector into the shared free-tier Grafana Cloud stack, plus a heavy browser bundle. The same-origin beacon avoids CORS, keeps PII control server-side, and adds ~0 browser bundle weight.

- **A span, not a log**, because `apps/web` exports only traces today (`@vercel/otel` wires no logs exporter, and there's no stdout log-scraper) — a span is the only signal the web runtime actually exports.
- **`sendBeacon` with a string body, no `fetch` fallback**: `apps/web` forbids raw `fetch()` in `app/lib` (the `no-raw-fetch-outside-api-client` lint rule), and `sendBeacon` is universally supported in the app's target browsers. If it's unavailable, the `console.error` still records the failure.

## Changes

- `apps/web/lib/log-client-error.ts` — best-effort `sendBeacon` dispatch (never throws; `console.error` behavior from #783 unchanged).
[...]

PR #810, opened 2026-07-11
[fix] Abuse-harden /api/client-errors: same-origin guard (observe→enforce)

### LINKED ISSUE #419: [fix] Local test failures on Node 24 — core Jest worker OOM and web route.test TS2576 NextResponse.json

### LINKED ISSUE #806: [infra] Rate-limit / abuse-harden the unauthenticated /api/client-errors telemetry sink
## Problem

The `POST /api/client-errors` beacon sink added in #798 / PR #805 is **public and unauthenticated by necessity** (the failure it reports may itself be an auth expiry, so Clerk must not gate it — see `apps/web/middleware.ts`). It records a **retained ERROR span** per accepted request (status `ERROR` → kept by the ADR-020 tail-sampling `errors` policy).

Two independent reviews of PR #805 flagged this as an abuse / cost vector:

- **Cross-origin browser spam:** `sendBeacon` sends a CORS "simple" request, so a page on any origin can `navigator.sendBeacon('<app-origin>/api/client-errors', …)` from a victim's browser; the server currently ignores `Origin`.
- **Scripted spam:** trivial `curl` volume.

Each accepted request injects one always-retained ERROR span into the shared free-tier Grafana Cloud / Tempo stack. ADR-020 explicitly *rejected* per-policy rate limits and *rejected* relying on the ingest cap (it silently drops the newest spans), assuming a shape like this couldn't exist — a flood here could starve real production spans.

## Already mitigated in PR #805 (span *size*, not *rate*)

- `content-length` precheck before buffering + byte-accurate 4 KB body cap.
- 512-char clamp on every string attribute (including context keys), 24-key context cap, scalar-only context values.

## Deferred here (span *rate* / abuse)

1. **Infra-level rate limiting** (Cloud Armor / ingress / edge) for the unauthenticated endpoint — the robust fix for scripted abuse; out of scope for an app-level PR.
2. **A same-origin `Origin`-header check** in the handler to drop cross-origin browser beacons. Deferred from PR #805 because it must be validated in **staging** against the real proxy chain first: this is best-effort telemetry, so a false-drop (if `Origin.host !== Host` for legit same-origin traffic behind the LB) would silently kill the whole feature.
3. Consider an **ADR-020 addendum** documenting the accepted interim risk.

## Urgency

The abuse surface is currently **latent**: per #804, `apps/web` server spans do not reach prod Tempo yet (no web collector sidecar). This should be resolved **before / together with #804**, which is what makes these spans live in production.

Epic: Observability. Follows #798 / PR #805; coupled to #804.

brownm09:
## Summary

Abuse-hardens the public/unauthenticated `POST /api/client-errors` beacon sink (added in #798 / PR #805), which records **one retained ERROR span per accepted request** (the ADR-020 `errors` tail-sampling policy always keeps it). PR #805 bounded span **size**; this bounds cross-origin browser-beacon **rate** with a same-origin guard, and documents the accepted interim risk in an ADR-020 addendum.

The guard is **observe-only by default** and can only ever *drop* against an explicit allowlist — so it cannot silently kill this best-effort telemetry via a false-drop, which is exactly why #806 required staging validation before enforcement.

## What changed

- **`apps/web/app/api/client-errors/route.ts`** — a same-origin guard:
  - Every recorded span is tagged `client.origin.check` = `same-origin` | `cross-origin` | `no-origin` (always, even when nothing is dropped — the signal used to validate the guard in staging Tempo). Cross-origin verdicts also record the clamped offending origin as `client.origin.value`.
  - A request is **dropped** (204, before buffering the body or starting a span) only when it is classified `cross-origin` against the `CLIENT_ERROR_ALLOWED_ORIGINS` allowlist **and** `CLIENT_ERROR_DROP_CROSS_ORIGIN=true`.
  - **Safety interlock:** the zero-config `Origin`-host-vs-`Host` heuristic (whose risk is an LB rewriting `Host`) is used only to produce the observe verdict — it **never drops**. If `DROP` is on but no allowlist is set, the request is recorded and flagged `client.origin.enforce_skipped=true`, never dropped. So a production drop is impossible without an operator-reviewed allowlist.
  - `no-origin` (curl / non-browser) is never dropped — scripted abuse is deferred to an infra rate limit (#808).
- **`apps/web/app/api/client-errors/route.test.ts`** — 9 new tests (observe tagging, allowlist-over-Host, opaque `null` origin, enforce-drop, enforce-record, no-origin-under-enforce, the enforce_skipped interlock). All 10 pre-existing tests unchanged & green.
- **`docs/adr/ADR-020-*.md`** — `## Addendum — 2026-07-10 (#806)`: documents the accepted interim risk (an unauthenticated always-retained ERROR-span source is a shape the ADR's rejected alternatives assumed couldn't exist), the three layered mitigations (size #805 → origin guard #806 → infra rate limit #808), and the enable-enforcement procedure. Adds WHATWG Fetch + MDN sendBeacon references.
- **`docs/adr-references.md`** — indexes the two new sources.
- **`CLAUDE.md`** — Observability note extended with the guard + its two env vars.

## Acceptance criteria (from #806)

- [x] **(2) Same-origin `Origin`-header check** to drop cross-origin browser beacons — landed, observe→enforce, allowlist-gated so the staging-validation-first constraint is honoured.
- [x] **(3) ADR-020 addendum** documenting the accepted interim risk.
- [ ] **(1) Infra-level rate limiting** (Cloud Armor / edge) for scripted abuse — explicitly out of scope for an app-level PR; filed as **#808**.
- [ ] **Enable enforcement** after staging validation of the guard — needs the real proxy chain; filed as **#809** (coupled to #804).

## Testing

`npm run typecheck` — **passed** (web executed, others cached).

`npm test -w @lifting-logbook/web`:

```
Tests:       259 passed, 0 skipped, 0 failed (54.6 s)
Test Suites: 2 failed (worker OOM at load), 42 passed, 44 total
```

- The client-errors suite: **19 passed** (10 pre-existing + 9 new).
- The 2 "failed" suites (`StepImport.test.tsx`, `StepLifts.test.tsx`) crashed at **load** with `Jest worker ran out of memory` — **not** assertion failures (every test that ran passed). They pass cleanly in isolation:
  ```
  npx jest --config apps/web/jest.config.js --runInBand "StepImport.test|StepLifts.test"
  → Test Suites: 2 passed, 2 total; Tests: 18 passed
  ```
  This is the documented Windows full-suite parallel-load worker-OOM flake (prior workarounds in the now-closed #419 / #567 don't fully cover these two CSV-heavy suites). Local-Windows-only; CI (Linux, Node 20) is unaffected. Tracked in **#807**. Unrelated to this change (different files).

**Pre-PR gates:** suppression grep (ADR-026) clean · test-integrity grep (ADR-029) clean · no deleted tests · no `package.json` change (lockfile gate N/A).

## Risk-dimension audit

- **Testing** — per-branch Jest node-env coverage for the new guard (9 tests).
- **Observability** — this *is* an observability control. New attributes are non-secret (Origin/Host only, no auth material), clamped to 512 chars. Retained-span **count** is unchanged in observe mode and reduced under enforcement. Repo audit items 1–3 (raw SQL / LLM PII / payload redaction) N/A — headers only, no SQL, no LLM, no bodies logged.
- **Security** — shrinks the abuse surface; fail-open (any error → record, never 500/blackout). Endpoint public by design (privilege-restricted-test rule N/A).
- **Resilience** — never throws/500; instant rollback by unsetting `CLIENT_ERROR_DROP_CROSS_ORIGIN` (no code deploy); enforcement gated so misconfig can't blackout.
- **Performance** — O(1) header reads + one `URL` parse; enforce rejects **before** buffering the body — strictly less work than today.
- **Data integrity / migrations** — N/A (no schema / data change).

## Follow-ups

- **#807** — web onboarding CSV-heavy suites still OOM under Windows full-suite parallel load.
- **#808** — infra-level rate limit (Cloud Armor / edge) for the endpoint (deferred item 1).
- **#809** — staging-validate the guard and enable enforcement (deferred item 2's enforcement step; coupled to #804).

Closes #806

Review: [comment](https://github.com/brownm09/lifting-logbook/pull/810#issuecomment-4941561114) — `blocking=0 non_blocking=1`.

- **[reliability] allowlist trailing-slash normalization** — **fixed in `1561591`**: `parseAllowedOrigins` now strips a trailing slash from each entry (a serialized `Origin` never has one), preventing a silent enforce-mode false-drop from a `https://host/`-style config typo. Regression test added; client-errors suite 20/20 green.


brownm09:
## Review: [fix] Abuse-harden /api/client-errors: same-origin guard (observe→enforce)

### Summary
**Merge-ready.** The change adds a same-origin guard to the public/unauthenticated client-error beacon sink, scoped so it can only ever *drop* against an explicit operator allowlist — the zero-config `Origin`-host-vs-`Host` heuristic is observe-only and never drops. That honours #806's "validate in staging before enforcing" constraint and makes a silent telemetry blackout via false-drop structurally impossible in production. The handler fails open, never 500s, and rejects cross-origin beacons before buffering the body. One non-blocking robustness gap (allowlist trailing-slash normalization) was found and fixed during review (`1561591`). Highest-severity item: that one non-blocking finding — now resolved.

Note the abuse surface is **not closed by this PR alone**: in the default observe mode cross-origin spam is still recorded (tagged, not dropped). Enforcement (#809) must be flipped after staging validation — acceptable because the surface stays latent until #804 wires web spans to prod Tempo.

Gate checks: test-coverage ✓ (9→10 new tests), test-integrity ✓ (no skips/deletions; `Tests:` summary present), ADR-warrant ✓ (ADR-020 dated addendum in diff), doc-reconciliation n/a (no Documentation Maintenance table), README coverage ✓ (no ancestor READMEs).

---

### Non-Blocking Findings

**[reliability]** `apps/web/app/api/client-errors/route.ts` — `parseAllowedOrigins`
Allowlist entries were matched by exact string, but a serialized `Origin` header never carries a trailing slash. An operator copying `https://app.example.com/` (the trailing-slash form a browser address bar shows) into `CLIENT_ERROR_ALLOWED_ORIGINS` would classify every legitimate same-origin beacon as `cross-origin` and — once `CLIENT_ERROR_DROP_CROSS_ORIGIN=true` — silently drop it: the exact telemetry-blackout the guard's observe-only default exists to prevent.
**Suggestion:** Strip trailing slashes from allowlist entries when parsing. → **Fixed in this PR (`1561591`)**, with a regression test.

---

### Findings Disposition
- **[reliability]** allowlist trailing-slash normalization — **fixed in `1561591`** (`parseAllowedOrigins` now strips trailing slashes; regression test added, suite 20/20 green).

---

brownm09:
## Review (refresh): [fix] Abuse-harden /api/client-errors: same-origin guard

Re-posted to refresh the Review Gate marker after merging `origin/main` into the branch to satisfy `strict` up-to-date (merge commit `9aefd1a` pulled in unrelated #811 dep changes only — none of the reviewed files changed). The [original review](https://github.com/brownm09/lifting-logbook/pull/810#issuecomment-4941561114) stands.

**Verdict:** merge-ready. `blocking=0`, `non_blocking=1` — the one non-blocking finding (allowlist trailing-slash normalization) was found and **fixed in-PR (`1561591`)**, disposition recorded in the PR body.

Gate checks: test-coverage ✓ · test-integrity ✓ (`Tests:` summary present) · ADR-warrant ✓ (ADR-020 dated addendum) · doc-reconciliation n/a · README coverage ✓.

---

Issue #813, opened 2026-07-11
[infra] Deploy the otel-collector sidecar in staging.yml (per-PR staging is single-container)


brownm09:
## Problem

Follow-up surfaced by #804. The per-PR `.github/workflows/staging.yml` deploys **both** `lifting-logbook-stg-api` and `lifting-logbook-stg-web` via plain `gcloud run deploy` (single-container, `--set-env-vars`), with **no** otel-collector sidecar injection. So every PR staging run **strips** the sidecar that `deploy.yml` injects — the Cloud Run collector sidecar is only present via `deploy.yml` (post-merge staging + prod).

This is pre-existing (PR #782 left the api the same way; #804 mirrors that scope for web) — not a regression introduced by #804 — but it means the staging environment normally runs single-container between merges, so the #804 "verify in Tempo (staging)" window is narrow.

## Question / scope

Decide whether per-PR staging **should** run the sidecar (robust staging verification, at the cost of a 2nd container + Grafana spend per PR) or intentionally stay single-container (leaner PR validation). If wiring it: mirror the `deploy.yml` `describe → inject → services replace` pattern in `staging.yml` for **both** api and web (symmetric), reusing `scripts/inject-otel-sidecar.py`.

Epic: Observability. Follows #804 / #782.

Issue #812, opened 2026-07-11
[infra] Wire GKE web + api workloads to the otel-collector DaemonSet (OTEL_EXPORTER_OTLP_ENDPOINT)


brownm09:
## Problem

Follow-up surfaced by #804 (web Cloud Run OTel sidecar). Confirming "the GKE path" for #804 revealed that **neither the web nor the api workload is wired to the collector on GKE**:

- No workload sets `OTEL_EXPORTER_OTLP_ENDPOINT` (repo-wide grep is empty). Both app SDKs default to `http://localhost:4318`, but on GKE the collector is a **DaemonSet** reached via the `otel-collector` **ClusterIP Service** (`http://otel-collector:4318`), not localhost.
- The `charts/{web,api}` deployment templates build `env:` from a **fixed key allowlist** (via `envFrom` a ConfigMap that emits only NODE_ENV/PORT/API_URL/... ) with **no `range .Values.env`**, so a bare `--set env.OTEL_EXPORTER_OTLP_ENDPOINT=...` would set the value but no template consumes it — it never reaches a pod.

PR #782 only wired the **api on Cloud Run** (sidecar, where localhost:4318 is valid); #804 does the same for **web on Cloud Run**. GKE is unwired for both.

## Scope

- Add `OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4318` to the GKE workload pods for **both** `charts/web` and `charts/api` (configmap.yaml + values + the Helm deploy `--set`/values in `deploy.yml`).
- Verify a `service.name="lifting-logbook-web"` / `"lifting-logbook-api"` span from a GKE pod reaches Tempo (staging, `gke_enabled=true`).

Note: production is Cloud-Run-only (`enable_gke=false`), so this affects the staging A/B GKE target only. Epic: Observability. Follows #804 / #782.

Issue #798, opened 2026-07-10
[feat] Export web client-side errors to OTel/Grafana (browser telemetry)


brownm09:
## Problem

[#783](https://github.com/brownm09/lifting-logbook/issues/783) made client-side mutation failures in `apps/web` diagnosable by routing every write form's caught error through a shared `apps/web/lib/log-client-error.ts` helper that `console.error`s in a consistent, greppable shape (`[client-mutation] <op> failed`). That satisfied the issue's stated **minimum** (browser console) but not its **ideal**: actually landing these errors in Grafana.

The gap: `apps/web` instruments only the **server** runtime — `@vercel/otel`'s `registerOTel` in `apps/web/instrumentation.ts`. There is no **browser** OTel SDK, so a `console.error` in a Client Component never reaches Grafana Cloud (Loki/Tempo).

## Goal

Wire browser → Grafana export for client-side errors. `apps/web/lib/log-client-error.ts` is the single seam — every mutation call site already routes through it, so the export attaches in one place.

## Options to evaluate

1. A browser OpenTelemetry SDK (`@opentelemetry/sdk-trace-web` + OTLP/HTTP exporter) posting to the OTel Collector — requires the collector to accept browser-origin CORS.
2. A lightweight beacon to a Next route handler that logs server-side (reusing the existing server instrumentation path).

## Constraints

- Cross-check the repo Observability convention (root + `apps/web` `CLAUDE.md` → Observability; ADR-018 stack, ADR-020 tail sampling).
- No PII/token exposure: the api-client only throws `Error(message)` / `ApiClientError(message, status)` where `message` is server validation text or `API <status> <statusText> for <path>` — never headers/body. Preserve that property in whatever is exported.

## References

- Follow-up to #783 / [ADR-032](https://github.com/brownm09/lifting-logbook/blob/main/docs/adr/ADR-032-cloud-run-api-public-invoker.md) Context (the #766 CORS-outage blind spot).

PR #814, opened 2026-07-11
[infra] Wire web Cloud Run service into the Grafana Cloud OTel pipeline (sidecar)

### LINKED ISSUE #804: [infra] Wire apps/web server runtime into the Grafana Cloud OTel pipeline (Cloud Run collector sidecar)
## Problem

Follow-up to #798 (client-error → OTel span) and #768 / #782 (which wired the **API** into the Grafana Cloud OTel pipeline via a Cloud Run collector sidecar).

`apps/web` **server**-emitted OTel spans do not reach Grafana in production. Evidence:

- The web Cloud Run deploy step (`.github/workflows/deploy.yml` → "Deploy web to Cloud Run") does **not** inject the OTel collector sidecar the API deploy gained in #782 — `infra/cloud-run/web-service.yaml` has no sidecar / `:4318` reference.
- No deployment (GKE Helm values or Cloud Run) sets `OTEL_EXPORTER_OTLP_ENDPOINT` for the web workload (repo-wide grep is empty), so `@vercel/otel`'s OTLP exporter default of `http://localhost:4318` resolves to nothing unless a collector shares the pod — the sidecar model #782 established for the API.

Consequently every span the web *server* runtime emits is dropped in prod: the pre-existing `@vercel/otel` request traces from #206 **and** the new `client.mutation.error` spans from #798. (Both export correctly against the local docker-compose collector, so this is a prod-wiring gap, not a code gap.)

## Goal

Give `apps/web`'s Cloud Run deploy the same OTel collector sidecar wiring the API got in #782, so web server spans reach Tempo. Mirror #782 (sidecar inject in `deploy.yml` + the shared `infra/cloud-run/otel-collector-config.yaml`). Verify with a `service.name="lifting-logbook-web"` span query in Tempo (staging) after deploy.

## Notes

- Scope is the **server** runtime export path only; #798 already handles getting the browser error to the server via a same-origin beacon.
- Confirm the GKE path too: if web also runs on GKE, ensure the workload reaches the collector DaemonSet (an `OTEL_EXPORTER_OTLP_ENDPOINT` pointing at the node/service), since nothing currently sets it for any workload.

Epic: Observability. Follows #798, #768, #782.

brownm09:
## Summary

Mirrors #782 for the **web** Cloud Run service. `apps/web` instruments its server runtime with `@vercel/otel` ([`apps/web/instrumentation.ts`](../blob/main/apps/web/instrumentation.ts)), which exports OTLP to `http://localhost:4318` — but the web Cloud Run deploy had no co-located collector, so in production every web-**server** span was dropped: the #206 request traces **and** the new #798 `client.mutation.error` spans. (They export fine against the local docker-compose collector, so this was a prod-wiring gap, not a code gap.)

The web deploy steps (staging + production) now use the same `describe → inject sidecar → gcloud run services replace` mechanism the api got in #782, reusing the **identical** collector config, Grafana endpoints, and auth-header secrets — only the ingress container name (`web`) and image differ.

Closes #804. Follows #798, #768, #782.

## What changed

- **`scripts/inject-otel-sidecar.py`** — generalized the hardcoded `"api"` ingress container name. The name comes from `INGRESS_CONTAINER_NAME` (default `"api"`) and the image from `INGRESS_IMAGE` (with `API_IMAGE` kept as a back-compat fallback). **The two api deploy steps are byte-for-byte unchanged** (they pass `API_IMAGE`, no `INGRESS_CONTAINER_NAME` → defaults to `api`). The `SYSTEM_DATABASE_URL` drop stays (a no-op on web, which never carries it); `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` is set on the ingress container either way.
- **`.github/workflows/deploy.yml`** — both "Deploy web to Cloud Run" steps rewritten to the sidecar pattern (source the shared endpoint/image env files → grant IAM → `describe --format=export` → inject with `INGRESS_CONTAINER_NAME=web` → `services replace`). The old `--set-env-vars` / `--update-secrets` / `--service-account` / `--allow-unauthenticated` flags are dropped: the web service's env/secrets/SA are Terraform-managed (`lifecycle.ignore_changes = [template]`, preserved verbatim by `describe`) and its public posture is owned by Terraform's `web_public` allUsers `run.invoker` binding — `services replace` never touches invoker IAM (same reasons the api step dropped them in #782).
- **`infra/terraform/cloud-run.tf`** — declares `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` on the web container (spec accuracy / first-apply, mirroring what #782 added to the api container; the live value is set by the deploy step in the injected manifest).
- **Tests** — new web fixture (`scripts/testdata/cloud-run-web-export.yaml`) + `InjectWebServiceTest` and `IngressImageResolutionTest` classes. **25 tests, all green** (was 15); the api-path regression + idempotency tests are untouched.
- **Docs** — ADR-018 addendum (`#804`), observability runbook (Cloud Run section now covers api **and** web, incl. the web-SA IAM distinction and the recreate caveat generalized to both), `docs/deploy.md`, `scripts/README.md`.

## Key decisions

- **Least-privilege IAM for the web sidecar (not a project-level grant).** The api sidecar reads the collector-config + two Grafana auth-header secrets via the api SA's **project-level** `secretmanager.secretAccessor` (`gke.tf`). The **web** workload SA has no such grant, and giving it one would let the internet-facing web SA read `DATABASE_URL` / migrator / superuser creds — a real privilege escalation the web BFF never needs (contra [ADR-089](../blob/main/docs/adr/ADR-089-privilege-restricted-test-defaults.md)). Instead the web deploy step grants the web SA `secretAccessor` on **exactly the three otel secrets**, idempotently, via `gcloud secrets add-iam-policy-binding` (runs as the `cicd` SA, `roles/owner`). It lives in the pipeline, not Terraform, because all three secrets are pipeline/operator-managed out-of-band and the config secret does not exist at `terraform apply` time. First-ever deploy only: a few seconds of IAM propagation may be needed before the sidecar can read them; a failed `services replace` leaves the prior web revision serving and self-heals on re-run.
- **Reuse the same collector-config secret** (`lifting-logbook-{stg,prod}-otel-collector-config`) the api step already ensures/versions — the web sidecar runs the identical pipeline, so no new secret and no `check-otel-config-sync` change.
- **GKE path — confirmed unwired, filed as a follow-up.** Investigated per the issue's "confirm the GKE path" ask: **neither** web **nor** api is wired on GKE — no workload sets `OTEL_EXPORTER_OTLP_ENDPOINT`, the collector is a DaemonSet behind ClusterIP `otel-collector:4318`, and the chart templates hardcode `env:` (no `range .Values.env`), so a `--set env.OTEL_…` flag would be silently dropped. Wiring it needs chart changes symmetric across both workloads and prod is Cloud-Run-only (`enable_gke=false`), so it's out of scope here — filed as **#812**.
- **`staging.yml` (per-PR staging) still deploys web + api single-container.** Same as #782 left it — `staging.yml`'s `gcloud run deploy` strips the sidecar on every PR; only `deploy.yml` (post-merge staging + prod) carries it. Filed as **#813** (symmetric for api + web).

## Testing

No application code (TS/JS) changed — only a Python deploy helper + its tests, infra YAML, Terraform, and docs — so the Jest suite and `tsc` typecheck are unaffected (CI still runs `turbo run lint typecheck test`). Change-specific validation:

- **Injector unit tests** — `python scripts/test_inject_otel_sidecar.py -v`: **`Ran 25 tests … OK`** (15 pre-existing api/idempotency/guard + 10 new web/image-resolution). This is the automated gate on the untestable prod deploy path.
- **Manual render** — ran the injector against the web fixture: correct 2-container manifest (`web` + `otel-collector`), all Terraform-managed env/secrets/SA preserved, `otelconfig` volume added, `OTEL_EXPORTER_OTLP_ENDPOINT` set once, server-assigned fields stripped.
- **`node scripts/check-grafana-endpoint-sources.mjs`** — OK (web steps source the single endpoint file, no hardcoded URLs).
- **`node scripts/check-otel-config-sync.mjs`** — OK (config reused, unchanged).
- **`terraform fmt -check infra/terraform/cloud-run.tf`** — clean.
- **`deploy.yml`** — valid YAML; pre-commit `turbo run lint` passed (FULL TURBO).
- Pre-PR **suppression** and **test-integrity** greps: both **CLEAN**. No `package.json` change → no lockfile drift.

**Post-merge verification (the issue's acceptance step):** on the first `deploy.yml` run of this branch the web sidecar deploys to staging; confirm with a `{service.name="lifting-logbook-web"}` span query in Grafana **Tempo** (staging), then a `deployment_environment_name="production"` web series after the prod promote. As in #782, the wiring is verified to the Grafana boundary here; the end-to-end Grafana confirmation emits on that first deploy.

## Acceptance criteria (#804)

- [x] Web Cloud Run deploy gains the same otel-collector sidecar as the api (#782) — both deploy.yml steps.
- [x] Reuses the shared `otel-collector-config.yaml` + Grafana endpoints/auth-header secrets.
- [x] Injector generalized + tested for the web container name/image.
- [x] Web-SA secret access wired (least-privilege).
- [x] GKE path confirmed (unwired for both workloads → follow-up).
- [ ] `service.name="lifting-logbook-web"` span in Tempo (staging) — post-merge, on first deploy (see above).

[`/review`](https://github.com/brownm09/lifting-logbook/pull/814#issuecomment-4942050711) ran two independent Opus passes — **0 blocking, 8 non-blocking**:

- **[correctness] empty `INGRESS_CONTAINER_NAME` → invalid empty container name** — fixed in `a6abdb2` (`or "api"` + regression test).
- **[maintainability] 4-way deploy-block duplication → composite action** — filed #817.
- **[reliability] web step relies on the api step for the config secret (not self-contained)** — filed #817.
- **[maintainability] retire the `API_IMAGE`/`INGRESS_IMAGE` alias once web is proven** — filed #817.
- **[reliability] first-deploy IAM-propagation race** — accepted; fails safe (prior revision serves) and self-heals on re-run; documented above.
- **[reliability] synthetic fixture vs. real `--format=export` shape** — accepted; the first post-merge staging deploy is the real manifest-shape gate (the acceptance step above).
- **[performance] +5-8s deploy latency** — accepted; deploy-time only, immaterial.
- **[maintainability] Python test-integrity language-scope** — accepted; manually verified no skip/deletion/threshold violations.


brownm09:
## Review: [infra] Wire web Cloud Run service into the Grafana Cloud OTel pipeline (sidecar)

### Summary
**Mergeable.** Two independent Opus passes (correctness/security; reliability/perf/maintainability) found **zero blocking defects**. The change faithfully mirrors the proven #782 api sidecar path: the injector generalization preserves the api behavior byte-for-byte (verified against `origin/main`), the staging-vs-prod `needs.*`/`steps.*` references are not swapped, env-inheritance from the sourced `.env` files reaches the injector subprocess, the config-secret ordering is safe, the IAM is genuinely least-privilege, and dropping `--allow-unauthenticated` is safe because Terraform's `web_public` allUsers `run.invoker` binding owns invoker access (which `services replace` never touches). The single latent finding (empty `INGRESS_CONTAINER_NAME`) is **fixed in this PR**; the rest are non-blocking cleanups filed as a fast-follow or accepted with rationale.

---

### Blocking Findings
None.

---

### Non-Blocking Findings

**[correctness]** `scripts/inject-otel-sidecar.py` — ingress-name default
`ingress_name = os.environ.get("INGRESS_CONTAINER_NAME", "api")` applied the `"api"` default only when the var was *absent*; an explicitly-empty value would have produced an invalid empty container name in the `services replace` manifest. Unreachable from current call sites (api omits it; web sets `web`), but latent.
**Suggestion:** Use `... or "api"` (mirroring the `INGRESS_IMAGE` resolution). — **Fixed in `a6abdb2`** (+ regression test; 26 tests green).

**[maintainability]** `.github/workflows/deploy.yml` — 4-way deploy-block duplication
This PR takes the `source envs → pyyaml bootstrap → describe → inject → services replace` block from 2 copies to 4 (~190 lines of near-identical shell), each coupled to out-of-band invariants (the `--break-system-packages` line, the flag set, OTEL secret naming) that must stay byte-identical or drift silently.
**Suggestion:** Extract a composite action (`deploy-cloud-run-otel-sidecar`) parameterized by service/ingress-name/image/project/secrets. — **Filed #817** (deferred to keep this prod-deploy PR tight and leave the merged api steps untouched until web is proven).

**[reliability/maintainability]** `.github/workflows/deploy.yml` — web step relies on the api step for the config secret
The web step references `lifting-logbook-{stg,prod}-otel-collector-config` but does not ensure it — it depends on the api step (earlier in the same job) having created+versioned it, stated only in a prose comment. Safe today (sequential, `set -e` aborts on api failure), but a latent trap if api/web are ever split into parallel jobs or reordered.
**Suggestion:** Make the web step self-contained by duplicating the idempotent ensure block (the composite action above subsumes this). — **Filed #817.**

**[maintainability]** `scripts/inject-otel-sidecar.py` — permanent `API_IMAGE`/`INGRESS_IMAGE` alias
Two env-var names now denote one concept, kept deliberately so the api steps stay byte-for-byte unchanged.
**Suggestion:** Once the web sidecar is verified in prod, flip the api steps to `INGRESS_IMAGE` and retire the fallback + its test. — **Filed #817.**

**[reliability]** `.github/workflows/deploy.yml` — first-deploy IAM-propagation race
The web SA's `secretAccessor` grant runs seconds before `services replace` in the same step, so the very first web deploy could race Secret Manager IAM propagation (the api path never had this — its grant is project-level and long-propagated).
**Suggestion:** Verified to fail safe — the prior single-container revision keeps serving 100% and it self-heals on re-run (idempotent grant → no-op after). — **Accepted; documented in the PR body.** A retry/backoff would be over-engineering for a self-healing, first-deploy-only case.

**[reliability]** `scripts/testdata/cloud-run-web-export.yaml` — synthetic fixture ≠ real export shape
The injector's server-assigned-field strip set was derived from the api export. If a real web `describe --format=export` carries an extra field `services replace` rejects, the first real staging web deploy fails (fails safe; caught before prod at the staging→prod gate).
**Suggestion:** No code change — treat the first post-merge staging deploy as the real manifest-shape gate (already the PR's post-merge acceptance step). — **Accepted.**

**[performance]** `.github/workflows/deploy.yml` — ~5-8s extra per web deploy
`describe → inject → replace` + 3 idempotent `add-iam-policy-binding` calls vs the old single `gcloud run deploy`.
**Suggestion:** Deploy-time only → immaterial; a `get-iam-policy` short-circuit isn't worth the added conditional. — **Accepted, no action.**

**[maintainability]** Python test-integrity language-scope
The diff adds `.py` files; the repo's JS/TS-only integrity scanners can't inspect them.
**Suggestion:** Manually verified — no `pytest.skip`/`@pytest.mark.skip`/deletion/threshold change introduced; only a test rename+strengthen and net-new tests. — **Accepted.**

---

### Style Notes
None. Comment density and naming match the surrounding deploy.yml/injector idiom; the verbose step comments mirror the established api-step style.

---

### Findings Disposition
- **[correctness] empty `INGRESS_CONTAINER_NAME`** → fixed in `a6abdb2`.
- **[maintainability] 4-way deploy duplication** → filed [#817](https://github.com/brownm09/lifting-logbook/issues/817).
- **[reliability] web-step config-secret self-containment** → filed [#817](https://github.com/brownm09/lifting-logbook/issues/817).
- **[maintainability] retire `API_IMAGE` alias** → filed [#817](https://github.com/brownm09/lifting-logbook/issues/817).
- **[reliability] first-deploy IAM propagation** → accepted; self-heals, documented in PR body.
- **[reliability] synthetic fixture vs real export** → accepted; first staging deploy is the manifest-shape gate (PR acceptance step).
- **[performance] +5-8s deploy latency** → accepted; deploy-time only, immaterial.
- **[maintainability] Python language-scope** → accepted; manually verified clean.

---

brownm09:
**Review Gate marker refresh (post-rebase).**

Rebased `infra/issue-804-web-otel-sidecar` onto `origin/main` to pick up #815 (the now-merged GKE-DaemonSet follow-up, formerly filed as #812). The rebase gave the head commit a fresh `committedDate` (head is now `5c89a84`), which staled the prior `/review` marker comment — re-posting the disposition marker against the current head so the **Review Gate** re-evaluates and clears.

No re-review is warranted: the rebase only replayed this PR's two original commits **unchanged**, plus #815's two GKE-Helm `--set env.OTEL_EXPORTER_OTLP_ENDPOINT` lines, which land in a `deploy.yml` region separate from this PR's Cloud Run deploy-step rewrite (verified post-rebase — `deploy.yml` is still valid YAML, and both #815's 2 GKE lines and this PR's 2 `INGRESS_CONTAINER_NAME=web` Cloud Run steps are intact). Findings are unchanged (**0 blocking, 8 non-blocking**) and remain disposed in the PR body under **## Review findings disposition**.

Issue #795, opened 2026-07-10
[infra] Mirror + digest-pin otel-collector image into Artifact Registry (#788 item 1)


brownm09:
Sub-issue of #788 (PR 2 of 2).

## Problem

`otel/opentelemetry-collector-contrib:0.104.0` is pulled from **Docker Hub by a mutable tag** — not
mirrored into the project Artifact Registry like the api image, and not digest-pinned. Cloud Run
pulls it on every cold-start / scale-up, so a Docker Hub rate-limit (100 pulls/6h per IP) or outage
can fail new production request-path instances, and a re-pushed tag breaks reproducibility. Also
applies to the GKE otel-collector chart.

## Fix

Mirror the collector image into the project Artifact Registry (recommended: an AR remote /
pull-through repository) and reference it by digest (`@sha256:...`) from
`scripts/inject-otel-sidecar.py` (Cloud Run sidecar) and the GKE chart (`values.yaml` / `Chart.yaml`).
Prod-only infra — not validatable in CI (same untestable-prod constraint as #782).

Refs: #788, #782

Issue #788, opened 2026-07-10
[infra] Harden Cloud Run otel-collector deploy: mirror + digest-pin image, drift-check config


brownm09:
## Problem

Two hardening follow-ups from the #782 review (both also apply to the GKE otel-collector chart, so they are broader than Cloud Run):

### 1. Mirror + digest-pin the collector image
`otel/opentelemetry-collector-contrib:0.104.0` is pulled from **Docker Hub by a mutable tag** — not mirrored into the project Artifact Registry like the api image, and not digest-pinned. Cloud Run pulls it on every cold-start/scale-up, so a Docker Hub rate-limit (100 pulls/6h per IP) or outage can fail new production request-path instances, and a re-pushed tag breaks reproducibility.

**Fix:** mirror the collector image into the project Artifact Registry and reference it by digest (`@sha256:...`).

### 2. CI drift-check the collector config
`infra/cloud-run/otel-collector-config.yaml` is copied **verbatim** from the GKE `infra/kubernetes/charts/otel-collector/templates/configmap.yaml` `config.yaml` block, kept in sync only by a "keep the two in sync" comment. The whole premise of #782 is identical GKE/Cloud Run pipelines; a future edit to one silently diverges them.

**Fix:** add a CI check that extracts the configmap's `config.yaml` and diffs it against the Cloud Run file (mirroring `scripts/check-turbo-version-sync.mjs`), or template both from a single source.

## Context

Deferred from #782 (review non-blocking findings).

Refs: #782

brownm09:
**Third hardening item from #782 (endpoint single-sourcing / lockstep drift).**

The Grafana OTLP/Loki endpoints live in **both** `infra/kubernetes/values/{staging,production}-otel-collector.yaml` (GKE) and `.github/workflows/deploy.yml` (Cloud Run api-deploy steps, 4 sites). This just caused real churn: #784 corrected the GKE copies (region/gateway fix for #781), and #782 then had to hand-sync the Cloud Run copies to match — exactly the lockstep drift this hardening should prevent.

**Fix:** single-source the endpoints (a shared workflow env, or read the values file in the deploy step) or add a CI drift-check that fails when the Cloud Run and GKE endpoint values diverge.

brownm09:
Decomposed into two PRs (per the multi-PR rule):
- #794 — item 2: CI drift-check (Cloud Run collector config vs GKE configmap)
- #795 — item 1: mirror + digest-pin the collector image into Artifact Registry

[...]

PR #816, opened 2026-07-11
[infra] Deploy the otel-collector sidecar in per-PR staging (api + web)

### LINKED ISSUE #813: [infra] Deploy the otel-collector sidecar in staging.yml (per-PR staging is single-container)

brownm09:
## Summary

Per-PR staging (`staging.yml`) deployed both `lifting-logbook-stg-api` and `lifting-logbook-stg-web` via plain single-container `gcloud run deploy`, **stripping** the otel-collector sidecar that `deploy.yml` injects post-merge. Every PR push therefore flapped the shared staging Cloud Run service 2-container → 1-container → 2-container across each PR/merge cycle, and left the #804 "verify in Tempo (staging)" window narrow.

This mirrors `deploy.yml`'s `describe → inject → gcloud run services replace` sidecar pattern into `staging.yml` for **both** the api and web deploys, so per-PR staging runs the **same 2-container topology** it will run once merged.

**Decision (per issue #813's open question):** wire the sidecar into per-PR staging rather than intentionally stay single-container. Rationale: the staging service is already 2-container as its post-merge steady state, so this stops the per-PR/merge flap rather than adding steady-state cost; it catches sidecar-injection regressions at PR time instead of post-merge; and the marginal Grafana spend is small (telemetry only during the brief integration-test window on a service whose only staging traffic is the Playwright suite).

## Changes (`.github/workflows/staging.yml` only)

- **`deploy-api`** → replaces the plain `gcloud run deploy` with `describe → inject → services replace` (`API_IMAGE` back-compat; floating `image_tag`, not `github.sha`).
- **`deploy-web`** → same pattern with `INGRESS_CONTAINER_NAME=web` / `INGRESS_IMAGE`, **plus** a least-privilege grant of `roles/secretmanager.secretAccessor` to the web workload SA on exactly the 3 otel secrets (config + 2 Grafana auth headers), mirroring `deploy.yml`. `web_url` now comes from the `terraform` `cloud_run_web_url` output, since `services replace` emits no `Service URL:` line to parse.
- **`terraform-staging`** → exposes the `otel_mirror_repo` output for the injector's collector image.
- Both deploy jobs ensure the collector config secret **independently** — unlike `deploy.yml`'s single sequential job, `deploy-api`/`deploy-web` run as **parallel** jobs here with no ordering guarantee (idempotent; the secret already exists from prior post-merge deploys).
- Drops `--allow-unauthenticated`: `services replace` never touches invoker IAM, so the api/web public posture stays owned by Terraform's `allUsers` binding — no re-lock of the #766 CORS path.

## ✅ #814 landed — rebased; diff is now `staging.yml`-only

#814 (`infra/issue-804-web-otel-sidecar`, which generalized `scripts/inject-otel-sidecar.py` with the `INGRESS_CONTAINER_NAME` / `INGRESS_IMAGE` params this PR reuses) **merged on 2026-07-11** (squash `c6fc0de`). This branch has since been **rebased onto current `main`**, dropping #814's now-merged commit — the diff is a single file, `.github/workflows/staging.yml` (+124/−37), and CI re-fired on the clean base. The reused injector on `main` carries the `INGRESS_CONTAINER_NAME`/`INGRESS_IMAGE` generalization the web step depends on (verified), so the #813→#814 dependency is satisfied.

## Testing

CI-workflow-only change — no application / library / test code touched, so `npm test` and `npm run typecheck` are not exercised by it (Refactor/config tier). Verified:

- **`scripts/test_inject_otel_sidecar.py`** (the test for the script this workflow calls): `Tests: 25 passed, 0 skipped, 0 failed (1.3s)` via `python -m unittest`.
- **Local inject dry-runs** against `scripts/testdata/cloud-run-{api,web}-export.yaml` with this PR's exact env-var invocation → both produce a valid 2-container manifest: ingress container correctly named (`api`/`web`) with its image set, `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` added, `SYSTEM_DATABASE_URL` dropped, and the Terraform-managed env **preserved** (api: `DATABASE_URL`; web: `API_URL` / `PUBLIC_API_URL` / `CLERK_SECRET_KEY`) — de-risking the describe-preserves-env assumption on the as-yet-unexercised web path.
- **CI sync guards** all green: `check-grafana-endpoint-sources.mjs` (the step `source`s the single-source env, no hardcoded endpoint), `check-otel-config-sync.mjs`, `check-turbo-version-sync.mjs`.
- **Cross-references verified against current `main`** (post-rebase): `terraform output -raw otel_collector_mirror_repo` is a defined output (#795) and `needs.terraform-staging.outputs.cloud_run_web_url` is a pre-existing `terraform-staging` output — both resolve, and `deploy.yml` on `main` sources them identically.
- **YAML** parses; job structure verified (`otel_mirror_repo` output present, `web_url` output wired, both sidecar steps present).
- **Definitive check:** this PR's own staging integration run deploys the 2-container api+web and runs the Playwright staging suite against it (dogfood).

Suppression grep and test-integrity grep against the net diff: both clean (YAML-only).

## Risk audit

- **Observability** — this *restores* per-PR staging telemetry; config drift guarded by `check-otel-config-sync.mjs` (#788); telemetry never gates serving (a failed `services replace` leaves the prior revision serving, self-heals on re-run).
- **Security** — web SA gets least-privilege read on exactly 3 otel secrets; api uses its existing project-level `secretAccessor`; no secrets echoed (bindings `--quiet >/dev/null`; auth headers only via `secretKeyRef`).
- **Resilience** — `continue-on-error` retained on `deploy-api`; prior revision serves on a failed replace.
- **Performance** — +1 `describe`/+1 `replace` per deploy job (negligible); the 2nd container is the intended cost per the decision above.
- **Data/migrations** — migrate step untouched; specs derived from the live service preserve all Terraform-managed fields. Accessibility — N/A (CI change).

Per the [`/review` of this PR](https://github.com/brownm09/lifting-logbook/pull/816#issuecomment-4942399914): **0 blocking, 1 non-blocking.**

- **[maintainability] Sidecar-injection block duplicated across `deploy-api`/`deploy-web` here (and the four copies in `deploy.yml`)** → **filed [#817](https://github.com/brownm09/lifting-logbook/issues/817)** (item 1: extract a composite action). Dedup deliberately deferred to keep this PR scoped to the `staging.yml` mirror; #817 [extended](https://github.com/brownm09/lifting-logbook/issues/817#issuecomment-4942404001) to note the block count is now six with `staging.yml` included.

Closes #813

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -126,6 +126,7 @@ jobs:
       kms_key_name: ${{ steps.tf-outputs.outputs.kms_key_name }}
       api_wi_sa: ${{ steps.tf-outputs.outputs.api_wi_sa }}
       web_workload_sa: ${{ steps.tf-outputs.outputs.web_workload_sa }}
+      otel_mirror_repo: ${{ steps.tf-outputs.outputs.otel_mirror_repo }}
 
     steps:
       - uses: actions/checkout@v5
@@ -175,6 +176,7 @@ jobs:
           echo "kms_key_name=$(terraform output -raw kms_key_name)" >> "$GITHUB_OUTPUT"
           echo "api_wi_sa=$(terraform output -raw cicd_service_account_email)" >> "$GITHUB_OUTPUT"
           echo "web_workload_sa=$(terraform output -raw web_workload_sa)" >> "$GITHUB_OUTPUT"
+          echo "otel_mirror_repo=$(terraform output -raw otel_collector_mirror_repo)" >> "$GITHUB_OUTPUT"
 
   # ── 2. Build and push Docker images ──────────────────────────────────────
   build-images:
@@ -531,7 +533,7 @@ jobs:
             echo "configured=true" >> "$GITHUB_OUTPUT"
           fi
 
-      - name: Deploy API to Cloud Run (staging)
+      - name: Deploy API + OTel Collector sidecar to Cloud Run (staging)
         id: deploy-api
         if: steps.clerk-check.outputs.configured == 'true'
         # continue-on-error is retained here (NOT for Cloud SQL flakiness — the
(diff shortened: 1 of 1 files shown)


brownm09:
## Review: [infra] Deploy the otel-collector sidecar in per-PR staging (api + web)

### Summary
**Merge (after the one non-blocking finding is dispositioned — it already is).** This is a faithful, well-scoped mirror of `deploy.yml`'s `describe → inject → gcloud run services replace` sidecar pattern into per-PR `staging.yml` for both the api and web Cloud Run deploys. The diff is a single file (+124/−37, 161 lines) — appropriately scoped. Both non-obvious cross-references (`terraform output -raw otel_collector_mirror_repo` and `needs.terraform-staging.outputs.cloud_run_web_url`) were verified to resolve against current `origin/main`, and the reused injector script on main carries the `INGRESS_CONTAINER_NAME`/`INGRESS_IMAGE` generalization the web step depends on. The single highest-severity finding is non-blocking maintainability (sidecar-block duplication), and it is already tracked by #817.

---

### Non-Blocking Findings

**[maintainability]** `.github/workflows/staging.yml:544` and `:719`
The `source envs → pyyaml bootstrap → describe → inject → gcloud run services replace` block is now duplicated across `deploy-api` and `deploy-web` here, and mirrors the four near-identical copies already in `deploy.yml` (api/web × staging/prod). After this PR there are six copies coupled to several out-of-band invariants that must stay byte-identical or drift silently (the `pip install --break-system-packages` line, the `--format=export`/`services replace` flag set, the OTEL secret naming, the single-source `.env` files).
This is exactly the class of problem #817 already captures — its item 1 proposes extracting a `.github/actions/deploy-cloud-run-otel-sidecar` composite action. #817's count ("FOUR near-identical blocks") predates this PR; the two staging.yml blocks added here raise it to six.
**Suggestion:** No change in this PR — the dedup is deliberately deferred to keep #816 scoped to the staging.yml mirror. Extend #817 to note the count is now six (staging.yml included), so the composite-action refactor covers these blocks too. **Disposition: filed — [#817](https://github.com/brownm09/lifting-logbook/issues/817).**

---

### Notes on things checked and cleared (not findings)

- **`cloud_run_web_url` switch for `web_url`.** Replacing the old `grep "Service URL:"` parse (which `gcloud run services replace` doesn't emit) with the terraform `cloud_run_web_url` output is precedent-validated: `deploy.yml` already uses the same output as its staging smoke-test `BASE_URL`. The old "legacy `*.a.run.app` may not be routed" warning was about `describe --format="value(status.url)"`, not the terraform-managed service URI — not applicable here.
- **Config-secret ensure in parallel jobs.** Because `deploy-api`/`deploy-web` are separate parallel jobs here (unlike `deploy.yml`'s single sequential job), each ensures the collector config secret independently. This actually **avoids** the latent trap #817 item 2 flags for `deploy.yml`'s web step (which relies on the api step having created the secret). Good call.
- **Dropping `--allow-unauthenticated`.** `services replace` never touches invoker IAM, so the api/web public posture stays owned by Terraform's `allUsers` `run.invoker` binding — no re-lock of the #766 CORS-preflight path. Verified against `deploy.yml`'s equivalent steps.
- **web SA least-privilege grant.** The web workload SA gets `secretAccessor` on exactly the three otel secrets it needs; no broadening to the api SA's project-level access. No secrets echoed (bindings `>/dev/null`, auth headers only via `secretKeyRef` in the injected sidecar).

---

### Findings Disposition
- **[maintainability] sidecar-block duplication** → **filed** [#817](https://github.com/brownm09/lifting-logbook/issues/817) (dedup deliberately deferred to keep this PR scoped; #817 to be extended to note the count is now six).

---

Issue #817, opened 2026-07-11
[infra] De-duplicate the 4 Cloud Run OTel sidecar deploy blocks into a composite action (+ self-contained config secret, retire API_IMAGE alias)


brownm09:
## Problem

Fast-follow surfaced by the /review of #804 / PR #814 (web OTel Cloud Run sidecar). Three related deploy-step/injector cleanups, all deferred to keep #814 tightly scoped to the production deploy path and to leave the already-merged api steps (#782) untouched until the web sidecar is proven in prod:

1. **De-duplicate the 4 sidecar-deploy blocks → composite action.** After #814, `.github/workflows/deploy.yml` has FOUR near-identical `source envs → pyyaml bootstrap → describe → inject → gcloud run services replace` blocks (~190 lines): api staging, api prod, web staging, web prod. They are coupled to several out-of-band invariants that must stay byte-identical or drift silently (the `pip install --break-system-packages` line, the `--format=export`/`services replace` flag set, the OTEL secret naming, the single-source `.env` files). Extract a composite action (e.g. `.github/actions/deploy-cloud-run-otel-sidecar`) with inputs `service`, `ingress_container_name`, `image`, `project`, `config_secret`, otlp/loki secret names, optional `web_sa` (triggers the least-privilege IAM grant), `ensure_config`. Note: the staging (`needs.*`) vs prod (`steps.*`) output split is mildly awkward to parameterize — weigh against explicit-steps auditability.

2. **Make the web deploy step self-contained re: the config secret.** Today the web step references `lifting-logbook-{stg,prod}-otel-collector-config` but does NOT ensure it — it relies on the api step (earlier in the same job) having created+versioned it, stated only in a prose comment. Safe today (sequential steps, `set -e` aborts on api failure), but a latent trap if api/web are ever split into parallel jobs, reordered, or the api step made conditional. The composite action in (1) naturally fixes this by ensuring the secret per-invocation (the ensure block is idempotent + content-diff-guarded, so a second run is a no-op).

[...]

PR #818, opened 2026-07-11
[infra] De-duplicate the 4 Cloud Run OTel sidecar deploy blocks into a composite action

brownm09:
## Summary

Fast-follow from the `/review` of #804 / PR #814. De-duplicates the four near-identical Cloud Run OTel-sidecar deploy blocks (api/web × staging/prod, ~190 lines) in `.github/workflows/deploy.yml` into a shared composite action, and makes the web deploy self-contained w.r.t. the collector config secret.

**This PR does issue #817 items (1) and (2). Item (3) is deferred — see below.**

### (1) Composite action — `.github/actions/deploy-cloud-run-otel-sidecar`

Replaces the four inline `source envs → ensure config secret → pyyaml → describe --format=export → inject-otel-sidecar.py → services replace` blocks with one composite action (net **−126 lines** in `deploy.yml`). Steps: ensure config secret (idempotent, content-diff-guarded) → optional least-privilege web-SA grant on the three otel secrets → describe / inject / `services replace`. Inputs: `service, image, ingress_container_name (=api), project, region, config_secret, otlp_secret, loki_secret, otel_mirror_repo, web_sa (opt), ensure_config (=true)`. The staging (`needs.*`) vs prod (`steps.*`) output split is handled at the call sites, which pass fully-resolved values — the action never references `needs`/`steps`.

### (2) Web deploy self-contained re: the config secret

The action's `ensure_config` step runs on **every** invocation (default `true`), and now runs **before** the web-SA grant, so the web deploy ensures the `…-otel-collector-config` secret itself instead of relying on the api step (earlier in the same job) having created it. Closes the latent trap if api/web are ever parallelized or reordered. Idempotent + content-diff-guarded, so the second run in a job is a no-op.

### (3) Retire the `API_IMAGE` fallback — DEFERRED (gated)

Item (3) is explicitly gated on the #814 web sidecar being **confirmed live in prod** (a `service.name="lifting-logbook-web"` span in Grafana Tempo). That has not happened: prod deploy is a manual-approval gate, and #814's Deploy run was still in flight when this branch was cut, so the web sidecar has not reached prod. Removing the safety fallback now would violate the gate.

The composite action already passes the image as **`INGRESS_IMAGE`** for all four callers (the api steps no longer set `API_IMAGE`), so the script's `API_IMAGE` fallback is now unused by every caller but remains in place (and remains covered by `test_api_image_fallback_still_works`). Once the prod span is confirmed, retiring it is a trivial no-op cleanup — #817 is kept open to track it.

`scripts/inject-otel-sidecar.py` and `scripts/test_inject_otel_sidecar.py` are **unchanged** in this PR.

## Testing

CI-configuration-only change (`.github/` workflow + composite action YAML); no application/package/app source is touched, so the workspace test suites cannot be affected (existing tests unchanged — per the coverage table's "Refactor / docs only — existing tests must pass"). Targeted validations run locally:

- `python scripts/test_inject_otel_sidecar.py` → **Tests: 26 passed, 0 skipped, 0 failed (1.8s)** (injector unchanged)
- `node scripts/check-grafana-endpoint-sources.mjs` → OK (the action `source`s the single-source endpoint env file; no hardcoded literal)
- `node scripts/check-otel-config-sync.mjs` → OK
- `yaml.safe_load` on both `deploy.yml` and `action.yml` → parse OK; composite step order asserted (ensure → grant → inject)
- Pre-commit `turbo run lint` → 7/7 successful
- Suppression + test-integrity greps → clean

The composite action's runtime behavior is a faithful extraction of the previously-shipping inline blocks (identical command sequences, env vars, and secret names); the only behavioral deltas are additive/no-op (`INGRESS_IMAGE` feeds the same injector variable as `API_IMAGE`; web now ensures the config secret; ensure-before-grant ordering). It is exercised end-to-end by the next Deploy run on `main` (staging before the prod gate).

## Acceptance criteria (#817)

- [x] (1) Four sidecar-deploy blocks → composite action
- [x] (2) Web deploy self-contained re: the config secret
- [ ] (3) Retire the `API_IMAGE` fallback — deferred, gated on the #814 web sidecar confirmed live in prod

Part of #817

---

## Update — post-review (rebased + concurrency fix)

- **Rebased onto current `origin/main`.** The branch was one commit behind #816 ("Deploy the otel-collector sidecar in per-PR staging"), which merged concurrently and added two *new* inline sidecar blocks to `staging.yml`. Rebasing removes the phantom `staging.yml` diff a pre-rebase compare showed and lets CI run against current main.
- **Concurrency-safe config-secret create (`b789a64`).** Review found the ensure-config `create` would fail on `ALREADY_EXISTS` if two invocations race (both pass `describe`, both `create`). Fixed by re-`describe`-after-create — makes the action safe for `staging.yml`'s parallel `deploy-api`/`deploy-web` jobs.
- **Follow-up filed:** #820 — migrate `staging.yml`'s two inline blocks onto this action (completes the de-dup across all six call sites).

- **[reliability] ensure-config `create` not concurrency-safe** — fixed in `b789a64`.
- **[maintainability] incomplete de-dup (`staging.yml` still inline after #816)** — filed #820; branch rebased onto current main; action made concurrency-safe for that migration.

### DIFF
--- a/.github/actions/deploy-cloud-run-otel-sidecar/action.yml
+++ b/.github/actions/deploy-cloud-run-otel-sidecar/action.yml
@@ -81,8 +81,16 @@ runs:
       run: |
         PROJECT="${{ inputs.project }}"
         CONFIG_SECRET="${{ inputs.config_secret }}"
+        # Concurrency-safe create: two invocations racing on a not-yet-existent secret would
+        # both pass `describe` (not-found) and both `create`; the race loser gets ALREADY_EXISTS
+        # (exit 1), which `bash -eo pipefail` would otherwise fail the job on. Re-`describe` after
+        # a failed create so the race loser still succeeds once the winner has created it, while a
+        # genuine create failure (e.g. IAM) still surfaces (the re-describe also fails). This
+        # matters as soon as parallel jobs share this action — e.g. staging's separate
+        # deploy-api / deploy-web jobs — not for deploy.yml's single sequential job.
         gcloud secrets describe "$CONFIG_SECRET" --project="$PROJECT" >/dev/null 2>&1 \
-          || gcloud secrets create "$CONFIG_SECRET" --replication-policy=automatic --project="$PROJECT" --quiet
+          || gcloud secrets create "$CONFIG_SECRET" --replication-policy=automatic --project="$PROJECT" --quiet 2>/dev/null \
+          || gcloud secrets describe "$CONFIG_SECRET" --project="$PROJECT" >/dev/null 2>&1
         CURRENT_CFG=$(gcloud secrets versions access latest --secret="$CONFIG_SECRET" --project="$PROJECT" 2>/dev/null || true)
         if [ "$CURRENT_CFG" != "$(cat infra/cloud-run/otel-collector-config.yaml)" ]; then
           gcloud secrets versions add "$CONFIG_SECRET" --data-file=infra/cloud-run/otel-collector-config.yaml --project="$PROJECT" --quiet
--- a/.github/actions/deploy-cloud-run-otel-sidecar/action.yml
+++ b/.github/actions/deploy-cloud-run-otel-sidecar/action.yml
@@ -1,0 +1,156 @@
+name: Deploy Cloud Run service with OTel Collector sidecar
+description: >-
+  Inject the co-located otel-collector sidecar into a Cloud Run service and
+  `gcloud run services replace` it, deriving the manifest from the LIVE service
+  (`describe --format=export`) so every Terraform-managed field (VPC connector,
+  scaling, service account, startup probe, env/secrets) is preserved verbatim.
+  Factored out of .github/workflows/deploy.yml (#817) so the api + web × staging +
+  production deploy steps share ONE implementation instead of four near-identical
+  inline blocks (#768 / #804) that had to stay byte-identical or drift silently.
+  See scripts/inject-otel-sidecar.py for why `services replace` (not `--container`)
+  and why the manifest is derived from the live service rather than a static template.
+
+inputs:
+  service:
+    description: Cloud Run service name, e.g. lifting-logbook-prod-api.
+    required: true
+  image:
+    description: >-
+      Full ingress-app image ref (registry/<api|web>:<sha>) — passed to the
+      injector as INGRESS_IMAGE and set on the ingress container.
+    required: true
+  ingress_container_name:
+    description: >-
+      Name for the ingress container in the rendered manifest — 'api' or 'web'.
+      Cosmetic in Cloud Run, but keeps per-container logs and metrics labelled.
+    required: false
+    default: api
+  project:
+    description: GCP project ID that owns the service and its secrets.
+    required: true
+  region:
+    description: Cloud Run region — pass the workflow's ${{ env.REGION }}.
+    required: true
+  config_secret:
+    description: Secret Manager secret holding the collector config.yaml (mounted as a file).
+    required: true
+  otlp_secret:
+    description: Secret Manager secret name for the Grafana Cloud OTLP auth header.
+    required: true
+  loki_secret:
+    description: Secret Manager secret name for the Grafana Cloud Loki auth header.
+    required: true
+  otel_mirror_repo:
+    description: >-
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: [infra] De-duplicate the 4 Cloud Run OTel sidecar deploy blocks into a composite action

Reviewed via two independent Opus passes (correctness/security + reliability/performance/maintainability), each reading the remote branch against the true merge-base.

### Summary
**Merge with the changes made.** The composite action is a faithful, behavior-preserving extraction of the four `deploy.yml` inline blocks — both passes independently confirmed identical command sequences, secret names, the `--break-system-packages` pyyaml fallback, the `describe --format=export` / `services replace` flag set, correct input parameterization, correct ensure→grant→inject ordering, and that the `INGRESS_IMAGE`-for-api flip is a true no-op (the injector reads `INGRESS_IMAGE or API_IMAGE`). Guard scripts and actionlint-class checks are clean. The highest-severity finding — a concurrency-unsafe secret `create` — has been fixed in this PR; the second (incomplete de-duplication caused by the concurrent #816 merge) is filed as a tracked follow-up.

---

### Blocking Findings

**[reliability]** `.github/actions/deploy-cloud-run-otel-sidecar/action.yml` — ensure-config `create` not concurrency-safe
The ensure-config step did `gcloud secrets describe … || gcloud secrets create … --quiet`. Two invocations racing on a not-yet-existent secret both pass `describe` (not-found) and both `create`; the race loser gets `ALREADY_EXISTS` (exit 1), which `bash -eo pipefail` does not swallow, failing the job. Not a live bug in `deploy.yml` (api+web run sequentially in one job), but the action's own comment advertises parallel-safety and it is intended to also back `staging.yml`'s separate parallel `deploy-api`/`deploy-web` jobs.
**Fix (done):** re-`describe` after a failed create so the race loser still succeeds while a genuine create failure still surfaces — commit `b789a64`.

---

### Non-Blocking Findings

**[maintainability]** `.github/workflows/staging.yml` — incomplete de-duplication (branch was behind #816)
The branch was one commit behind `origin/main`; the intervening commit #816 added two *new* inline sidecar-inject blocks to `staging.yml` (`deploy-api`, `deploy-web`) using the identical pattern this action encapsulates. After this PR merges, 4 of 6 sidecar call sites use the action but staging.yml's two remain inline — the exact drift risk the action exists to prevent. The branch has been rebased onto current `origin/main` (so the surface diff no longer shows a phantom staging.yml change), and the action's `create` was made concurrency-safe (above) specifically so staging's parallel jobs can adopt it.
**Disposition:** filed as tracked follow-up #820 (staging.yml is a separate workflow added by the concurrent #816; migrating it there keeps this PR scoped to #817's deploy.yml and gives the distinct `image_tag`/`web_url` wiring focused treatment).

---

### Verified clean
Behavior preservation (all four blocks, byte-equivalent); idempotency + content-diff guard on ensure-config; web step self-containment + ensure→grant→inject ordering; composite-action mechanics (`region` input replaces `${{ env.REGION }}`, `$RUNNER_TEMP`, `github.workspace` cwd, valid `if:` expressions, `shell: bash` on all run steps, only `inputs.*` interpolated across separate-shell steps); guard scripts (`check-grafana-endpoint-sources.mjs`, `check-otel-config-sync.mjs`); no secret echoed; no invoker-IAM regression; least-privilege web-SA grant preserved. Item 3 of #817 (retire the `API_IMAGE` fallback) is intentionally deferred and gated — the injector script + tests are unchanged.

---

### Findings Disposition
- **[reliability] concurrency-unsafe create** — fixed in `b789a64`.
- **[maintainability] incomplete de-dup (staging.yml)** — filed #820 (+ rebased onto current main; action made concurrency-safe for the migration).

---

brownm09:
**Review marker refreshed after rebase.** The branch was rebased onto current `main` (past #816 and #819, both disjoint from this PR's files) to satisfy strict up-to-date branch protection. The rebase replayed the same two commits onto a new base — `deploy.yml` and `action.yml` are byte-identical to the reviewed state above — so the findings and their disposition stand unchanged:

- **[reliability]** concurrency-unsafe config-secret `create` — fixed in-PR (re-`describe` after a failed create).
- **[maintainability]** incomplete de-dup (`staging.yml` still inline after #816) — filed #820.

PR #822, opened 2026-07-11
[infra] Migrate staging.yml's 2 inline OTel sidecar blocks onto the composite action

brownm09:
## Summary

Completes the OTel-sidecar de-duplication started in #817/#818. #818 factored `deploy.yml`'s four inline `describe --format=export → scripts/inject-otel-sidecar.py → gcloud run services replace` blocks into the shared composite action `.github/actions/deploy-cloud-run-otel-sidecar`. Concurrently, #816 added **two more** inline copies of the same pattern to `staging.yml` (`deploy-api` + `deploy-web`). This PR migrates those two onto the action, so **all 6 real sidecar-inject call sites now use one implementation** — no more "stay byte-identical or drift silently" duplication.

Net **−57 lines** in `staging.yml` (65 insertions / 122 deletions; the insertions are mostly the two call-site comment blocks).

**Also fixes a latent #818 bug in the composite action** (`11d26fc`). This PR is the *first CI invocation* of the action (deploy.yml runs post-merge/manual, never on a PR), and it surfaced a manifest-load failure: `${{ env.REGION }}` was embedded in the `region` **input description**, which GitHub template-validates at load time and evaluates as an expression — `env` isn't an available named-value there, so the whole action failed to load (`Unrecognized named-value: 'env'`, action.yml L32). Fixed to plain prose. This repairs **all six** call sites (deploy.yml's four were latently broken the same way). So the PR touches two files: `.github/workflows/staging.yml` + `.github/actions/deploy-cloud-run-otel-sidecar/action.yml`.

## What changed

- **`deploy-api`** — inline `run:` block → `uses: ./.github/actions/deploy-cloud-run-otel-sidecar`. Keeps `id: deploy-api`, the `steps.clerk-check` gate, and `continue-on-error: true` (+ its rationale comment). **No `web_sa`** input — the api SA already holds project-level `secretAccessor`, so the action skips the per-secret grant.
- **`deploy-web`** — the single inline step did *two* jobs (sidecar deploy **and** emitting `steps.deploy-web.outputs.url`), and a step can't be both `uses:` and `run:`, so it splits into two:
  - **`id: deploy-web-sidecar`** — `uses:` the action with `ingress_container_name: web` and **`web_sa`** set (drives the action's least-privilege grant of exactly the three otel secrets to the web workload SA).
[...]

PR #824, opened 2026-07-11
[fix] Staging deploy-api: hard-fail a hollow-green sidecar deploy

### LINKED ISSUE #823: [reliability] Staging deploy-api continue-on-error masks a total action-load failure as a hollow green
## Problem

`staging.yml`'s `deploy-api` job runs its Cloud Run OTel-sidecar deploy step with `continue-on-error: true` (intentional, per the #498 diagnosis: a failed revision should still let the "Fetch Cloud Run API logs on failure" step capture the crash reason, and the integration-test gate — not the deploy step — fails the run).

Since #822 migrated that step to `uses: ./.github/actions/deploy-cloud-run-otel-sidecar`, a **new failure mode** is now in scope: the composite action can fail to **load** (manifest/template-validation error), not just fail at runtime. `continue-on-error: true` masks that too — the step shows a green ✓ (conclusion=success) while the api sidecar **never deployed at all**.

This was not hypothetical: on PR #822's first CI run, `action.yml` had a manifest bug (`${{ env.REGION }}` in an input description → "Unrecognized named-value: 'env'"). `deploy-api` went **green while deploying nothing** (the stale revision kept serving); only `deploy-web` (fail-fast, no continue-on-error) surfaced the failure. Because the old revision keeps serving, Staging Integration Tests can pass against a stale image, so the hollow green can propagate undetected. (The action.yml bug itself was fixed in #822 / `11d26fc`; this issue is about the masking that hid it.)

## Suggested direction

Distinguish "the deploy step failed to even start (action-load / structural failure)" from "a revision was submitted but failed to start" — only the latter should be swallowed for log-capture. Options:
- Add a cheap pre-check step (no continue-on-error) that validates the action is loadable / the deploy actually submitted a new revision, gating the integration tests on it.
- Or narrow `continue-on-error` to the specific `services replace` runtime failure rather than the whole step.
- Or assert post-deploy that the live revision's image tag == the just-built `image_tag` before trusting a green deploy-api.

## Context
- Surfaced by: https://github.com/brownm09/lifting-logbook/pull/822 (the #820 migration).
- Pre-existing design element: `continue-on-error` on `deploy-api` predates #822; the migration to `uses:` is what makes an action-**load** failure a maskable outcome.

Epic: Observability.

brownm09:
## Summary

`staging.yml`'s `deploy-api` job runs its Cloud Run OTel-sidecar deploy with `continue-on-error: true`. That's intentional (#498): a *revision* that fails to start must still let the `Fetch Cloud Run API logs on failure` step run, with the integration-test gate — not the deploy step — failing the run. But since #822 moved that step to `uses: ./.github/actions/deploy-cloud-run-otel-sidecar`, a composite-action **load** failure (a manifest/template-validation error) became a *maskable* outcome: `continue-on-error` swallows it too, so `deploy-api` reports success while **no revision was submitted** and the stale revision keeps serving. Integration tests then pass against old code and the hollow green propagates undetected.

Observed on #822's first CI run (an `env.REGION`-in-an-input-description bug in `action.yml`); that bug was fixed in #822 / `11d26fc` — **this PR fixes the masking that hid it.**

## Fix

Add a `Verify API revision actually deployed (staging)` step to `deploy-api`, gated on `steps.deploy-api.outcome == 'success'` with **no** `continue-on-error`:

- Re-derives ground truth from Cloud Run (`gcloud run services describe … --format='value(spec.template.spec.containers[].image)'`) and hard-fails when the live service is **not** running the just-built `…/api:<image_tag>` (tag match, digest fallback).
- Gating on the **claimed-success path** is the whole point — it scrutinizes exactly where the mask hides. The `outcome == 'failure'` runtime-revision case stays owned, unchanged, by the existing `cloud_run_api_deployed` gate + the log-fetch step, so the #498 intent is preserved verbatim.
- On the non-image-affecting **skip-build** path the built tag already equals what serves → correct no-op, never a false failure.
- Scoped to `deploy-api`; `deploy-web` uses the same action but already fail-fasts (no `continue-on-error`).

This implements the issue's option C ("assert post-deploy that the live revision's image tag == the built image_tag"), reconciled with option A ("the deploy actually submitted a new revision") by comparing against the externally-known built image rather than any Cloud Run-internal revision-consistency signal (which can't tell "nothing deployed" from "old revision still healthy").

### Why not just remove `continue-on-error`

That re-breaks #498: a failed *revision* must still capture crash logs and fail via the integration gate, not red the deploy step directly. This PR distinguishes **"nothing was submitted (load failure)"** — now hard-fails — from **"a revision was submitted but failed to start"** — still swallowed for log capture — exactly as the issue asks.

### Behavior matrix

| Scenario | `deploy-api` outcome | New verify step | Result |
|---|---|---|---|
| Real success | success | image matches → pass | ✅ deploy is real |
| Action **load** failure (the bug) | success (masked) | stale image → **hard-fail** | ❌ caught (was: hollow green) |
| Revision submitted, fails to start | failure | skipped (gated) | ❌ caught by existing gate + log-fetch (unchanged) |
| Skip-build (non-image-affecting diff) | success | built tag already serving → pass | ✅ no-op |

## Testing

Workflow + docs only — no app/package/test code is touched, so the app Jest/Playwright suites are not the relevant gate (and don't exercise workflow YAML). Verification performed:

- **Structural validation** — `staging.yml` parses as YAML; the new step sits between the deploy step (idx 11) and the log-fetch step (idx 13), carries `if: steps.deploy-api.outcome == 'success'`, has **no** `continue-on-error`, and is a `run:` step; the deploy step still keeps `continue-on-error: true` + `uses:` the action; the `cloud_run_api_deployed` job output is unchanged. **All assertions passed.**
- **Pre-commit lint** (`npx turbo run lint`): **7/7 tasks passed** (FULL TURBO).
- **Suppression grep** (ADR-026): clean (none).
- **Test-integrity grep** (ADR-029): clean (none).
- **Lockfile gate**: N/A (no `package.json` change).
- **End-to-end:** `.github/**` is image-affecting per the `paths-filter`, so **this PR's own `staging.yml` run executes the modified `deploy-api` against real staging** — the new verify step runs for real. On a healthy deploy it logs `✓ Live service runs the just-built image` and passes; a false-fail here would red this PR before merge.

`Tests: n/a — no automated suite exercises a CI-workflow YAML change; see the verification performed above.`

### Coverage deferral

The new logic is a gcloud-dependent shell assertion inside a GitHub Actions step; the repo has no local gcloud mock harness to unit-test it against. Coverage is the structural validation above plus the PR's own staging CI run exercising the step end-to-end — the same way the rest of `staging.yml`'s gcloud steps are verified.

## Observability

The new step is itself an observability win: it converts a previously-silent failure (hollow green) into a titled, actionable `::error::` in the `Deploy API (staging)` job log, naming expected vs. live image and pointing at #823 + the runbook. `docs/runbooks/staging-ci-flakiness.md` gains a **"Hollow green — deploy-api reports success but nothing deployed"** section (symptom, cause, diagnosis, fix).

## Risk audit (Pass 3)

- **Security** — N/A: reads only non-sensitive Cloud Run image refs under the already-authenticated staging SA; no new secrets/inputs.
- **Resilience / failure modes** — fails **closed** on a detected hollow green and on a `describe` error (can't verify → don't trust the green); introduces no new flakiness (genuine success passes; skip-build is a no-op). Rollback = revert the one-step addition.
- **Performance** — one extra `gcloud run services describe` (+ rarely one `artifacts docker images describe`) per successful API deploy — a few seconds, off the hot path.
- **Data integrity / migrations, Accessibility** — N/A (no schema/data/UI change).
- **ADR warrant** — none: a small reliability guard; rationale captured in the step comment, this PR, #823, and the runbook.

Closes #823

---

`/review` (self, `claude-opus-4-8`) surfaced **0 blocking, 1 non-blocking** finding — fixed in-PR, not left as-is (ADR-028):

- **[reliability] Missing bounded retry on the new `gcloud run services describe`** — **fixed in `1c59646`.** Wrapped the verification read in a 3-attempt bounded retry (mirroring this workflow's migrate/terraform/build retry convention) so a transient Cloud Run API blip on the read can't red an otherwise-successful deploy; only the read retries, a genuine image mismatch still hard-fails.

Review comment: https://github.com/brownm09/lifting-logbook/pull/824#issuecomment-4947544490

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -567,6 +567,57 @@ jobs:
           loki_secret: lifting-logbook-stg-otel-loki-auth-header
           otel_mirror_repo: ${{ needs.terraform-staging.outputs.otel_mirror_repo }}
 
+      # Guard against a HOLLOW GREEN (#823). The deploy step above intentionally
+      # carries continue-on-error: true (#498 — a failed *revision* must still let
+      # the log-fetch step below run and let the integration-test gate, not the
+      # deploy step, fail the run). But since #822 moved that step to
+      # `uses: ./.github/actions/deploy-cloud-run-otel-sidecar`, a composite-action
+      # *load* failure (a manifest/template-validation error — e.g. the
+      # `env.REGION`-in-an-input-description bug on #822's first CI run) is ALSO
+      # swallowed: deploy-api reports outcome=success while NO revision was submitted
+      # and the stale one keeps serving, so the integration tests can pass against
+      # old code and the hollow green propagates undetected.
+      #
+      # This step re-derives ground truth from Cloud Run and hard-fails (no
+      # continue-on-error) when the live service is NOT running the just-built image.
+      # It runs ONLY on the claimed-success path (outcome == 'success') — exactly
+      # where the mask hides — so the outcome == 'failure' runtime-revision case stays
+      # owned, unchanged, by the existing cloud_run_api_deployed gate + the log-fetch
+      # step below. On the non-image-affecting skip-build path the built tag already
+      # equals what serves, so this is a correct no-op, never a false failure. See #823.
+      - name: Verify API revision actually deployed (staging)
+        if: steps.deploy-api.outcome == 'success'
+        run: |
+          SERVICE="lifting-logbook-stg-api"
+          EXPECTED="${{ needs.build-images.outputs.ar_repo }}/api:${{ needs.build-images.outputs.image_tag }}"
+          # The image(s) the latest-applied service spec runs (ingress `api` + the otel
+          # sidecar). On a hollow green nothing was applied, so this is still the PRIOR
+          # revision's image and won't contain the just-built tag.
+          if ! LIVE_IMAGES=$(gcloud run services describe "$SERVICE" \
+                --region="${{ env.REGION }}" \
+                --project="${{ vars.GCP_STAGING_PROJECT_ID }}" \
+                --format='value(spec.template.spec.containers[].image)'); then
+            echo "::error title=Staging API deploy unverifiable::deploy-api reported success but '$SERVICE' could not be described to confirm the deployed image — not trusting this green. See #823 and docs/runbooks/staging-ci-flakiness.md."
+            exit 1
+          fi
+          echo "Expected just-built image: $EXPECTED"
+          echo "Latest-applied image(s):   $LIVE_IMAGES"
+          if printf '%s\n' "$LIVE_IMAGES" | grep -qF "$EXPECTED"; then
+            echo "✓ Live service runs the just-built image — real deploy, not a hollow green."
+            exit 0
+          fi
+          # Tag didn't match — fall back to a digest comparison, in case Cloud Run stored
+          # the image by digest rather than the tag we passed to `services replace`.
+          EXPECTED_DIGEST=$(gcloud artifacts docker images describe "$EXPECTED" \
+            --project="${{ vars.GCP_STAGING_PROJECT_ID }}" \
+            --format='value(image_summary.digest)' 2>/dev/null || true)
+          if [[ -n "$EXPECTED_DIGEST" ]] && printf '%s\n' "$LIVE_IMAGES" | grep -qF "$EXPECTED_DIGEST"; then
+            echo "✓ Live service runs the just-built image by digest ($EXPECTED_DIGEST)."
+            exit 0
+          fi
+          echo "::error title=Staging API deploy is a hollow green::deploy-api reported success but '$SERVICE' is NOT running the just-built image ($EXPECTED${EXPECTED_DIGEST:+ / $EXPECTED_DIGEST}). The OTel-sidecar composite action most likely failed to LOAD (a manifest/template-validation error) and submitted no revision, leaving the stale revision serving. Fix the action manifest — do not trust this green. See #823 and docs/runbooks/staging-ci-flakiness.md."
+          exit 1
+
       - name: Fetch Cloud Run API logs on failure
         if: steps.deploy-api.outcome == 'failure'
         run: |
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -593,13 +593,28 @@ jobs:
           # The image(s) the latest-applied service spec runs (ingress `api` + the otel
           # sidecar). On a hollow green nothing was applied, so this is still the PRIOR
           # revision's image and won't contain the just-built tag.
-          if ! LIVE_IMAGES=$(gcloud run services describe "$SERVICE" \
-                --region="${{ env.REGION }}" \
-                --project="${{ vars.GCP_STAGING_PROJECT_ID }}" \
-                --format='value(spec.template.spec.containers[].image)'); then
-            echo "::error title=Staging API deploy unverifiable::deploy-api reported success but '$SERVICE' could not be described to confirm the deployed image — not trusting this green. See #823 and docs/runbooks/staging-ci-flakiness.md."
+          # Bounded retry so a transient Cloud Run API blip on this READ doesn't red an
+          # otherwise-successful deploy — mirrors the migrate (3-attempt) / terraform apply
+          # (2-attempt) / build-push (wretry) pattern this workflow already uses, and the
+          # reason its companion runbook (staging-ci-flakiness.md) exists: transients must
+          # not produce "re-run goes green" false reds. Only the read is retried; a genuine
+          # hollow green (image mismatch) still hard-fails below with no retry.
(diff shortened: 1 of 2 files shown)


brownm09:
## Review: [fix] Staging deploy-api: hard-fail a hollow-green sidecar deploy

### Summary
**Merge with one minor change.** The fix is correctly targeted and well-reasoned: gating the new verification on `steps.deploy-api.outcome == 'success'` closes exactly the action-load-failure masking gap while leaving the intentional `outcome == 'failure'` runtime-revision/log-capture path (#498) untouched, and comparing against the externally-known built image is the right way to detect "nothing was deployed." Scope is correct (deploy-web already fail-fasts). The single highest-severity finding is a **missing bounded retry** on the new `gcloud run services describe` — a consistency/reliability gap against the rest of this workflow, worth fixing before merge but not a correctness defect.

---

### Non-Blocking Findings

**[reliability]** `.github/workflows/staging.yml:577` (the `gcloud run services describe` in the new step)
The new verify step calls `gcloud run services describe` once, with no retry, and hard-fails the job (`exit 1`) if that single call errors. Every other gcloud/Cloud API call in this workflow is wrapped in a bounded retry precisely to stop a transient blip from reding a good run — the migrate step (`for attempt in 1 2 3`), `terraform apply` (2-attempt loop), and the build-push steps (`wretry.action`, 3 attempts). This step is on the deploy critical path and now sits in a workflow whose entire companion runbook (`staging-ci-flakiness.md`, which this PR edits) exists because transient GCP hiccups produce "re-run goes green" false reds. A one-off transient failure of this read would now red an otherwise-successful deploy and require a manual re-run — reintroducing the exact flake class the retry convention was adopted to kill.
**Suggestion:** Wrap the `services describe` in a small bounded retry (2–3 attempts with a short sleep), mirroring the migrate step's `for attempt in 1 2 3` idiom, and only `exit 1` after the last attempt. The digest-resolve call is already non-fatal (`|| true`), so only the `services describe` needs it.

---

### Notes on things checked and found clean
- **Success-path gating** — `if: steps.deploy-api.outcome == 'success'` correctly scopes the new scrutiny to the claimed-success path; the `outcome == 'failure'` log-fetch step and the `cloud_run_api_deployed` job output are unchanged, so #498's behavior is preserved verbatim.
- **Skip-build path** — on a non-image-affecting diff the built tag already serves, so the tag match passes (a harmless missed-detection on the load-failure+skip-build combo, but never a false failure, since the app image is intentionally identical there).
- **Sidecar image false-match** — `LIVE_IMAGES` includes the otel-collector image, but it can't match `.../api:<tag>` or the api digest, so no spurious pass/fail.
- **`set -e` interactions** — the `if ! LIVE_IMAGES=$(...)` guard and the `grep -qF` in `if` conditions are errexit-safe; the digest resolve is `|| true`.
- **Security** — reads only non-sensitive image refs under the already-authenticated staging SA; nothing sensitive is logged.
- **Test coverage (ADR-022)** — no local unit-test harness exists for a gcloud-dependent workflow step; the deferral is documented in the PR body and the change is exercised end-to-end by this PR's own staging run. Valid deferral.
- **Test integrity (ADR-029)** — no test files, config, or thresholds touched; diff is YAML + Markdown only.
- **Docs** — the runbook gains an accurate "Hollow green" section; no repo Documentation-Maintenance table applies; no ancestor README warrants an update.

---

### Findings Disposition
The single non-blocking reliability finding must be fixed-in-PR or filed-and-linked before merge (ADR-028). Recommended: fix in-PR (add the bounded retry). Record the disposition (fixed in `<sha>`) in a "Review findings disposition" section of the PR body before `gh pr merge`.

---

brownm09:
## Re-review after `1c59646` — [fix] Staging deploy-api: hard-fail a hollow-green sidecar deploy

Re-reviewed the final diff at head `1c59646`, which folds in the fix for the single finding from the [prior review](https://github.com/brownm09/lifting-logbook/pull/824#issuecomment-4947544490). The finding is resolved **in this PR's diff**, so the current head is clean.

**Prior finding — resolved:**
- **[reliability] Missing bounded retry on the new `gcloud run services describe`** → **fixed in `1c59646`.** The read is now wrapped in a 3-attempt bounded retry mirroring the workflow's migrate/terraform/build convention. Re-verified: the retry loop is errexit-safe (the `describe` runs inside an `if` condition), `LIVE_IMAGES` is initialised before the loop, each non-final failure sleeps and retries, and it only `exit 1`s (as "unverifiable") after the 3rd attempt; a genuine image mismatch still hard-fails below with no retry. No new issue introduced by the retry.

Everything else from the prior review still holds (success-path gating preserves #498 intent; scope correct; `set -e`-safe; security/docs/test-integrity clean).

PR #861, opened 2026-07-19
[infra] Cut GCP idle cost: retire staging GKE, gate idle NAT, prod DB → ZONAL, AR retention

brownm09:
## Summary

The $300 GCP free-trial credit was exhausted and its billing account is now closed
(`open: False`) — that's the source of the billing errors, and both projects' Cloud SQL /
Cloud Run resources currently show `SUSPENDED`. Despite ~zero traffic, the credits went to
**always-on standing infrastructure** (not CI build compute, which runs on GitHub-hosted
runners), kept continuously provisioned by `terraform apply` on every PR push / merge.

This trims the footprint to a lean shape **before a real billing account is attached**, keeping
the app fully functional and prod snappy. Estimated idle cost **~$180–260/mo → ~$70–90/mo**
(order-of-magnitude us-central1 estimates; live billing SKU data was unavailable with the account
closed).

## Changes

| Change | Effect | Est. saving |
|--------|--------|-------------|
| Staging GKE disabled (`enable_gke=false` in `terraform.tfvars.staging`) | The ADR-009 90/10 A/B split was never wired (#464/#495), so the Autopilot cluster + api/web pods + otel DaemonSet + Ingress LB were pure idle cost. Staging is now Cloud-Run-only. **CI auto-skips every GKE step** (`staging.yml` `if: ...gke_enabled == 'true'`) and the integration tests already target the Cloud Run URLs — no test coverage lost. | ~$70–130/mo |
| Cloud NAT + router gated on `enable_gke` (`main.tf`) | They exist solely for Autopilot node egress; Cloud Run reaches Cloud SQL over private VPC peering and egresses public traffic via its own managed path. With GKE off in both envs, both idle NATs are removed. | ~$64/mo |
| Prod Cloud SQL `REGIONAL` → `ZONAL` (new `db_availability_type` var, default `ZONAL`) | A single-user app doesn't need cross-zone HA. PITR + daily backups are unchanged. | ~$25–30/mo |
| Artifact Registry `cleanup_policies` (`main.tf`) | Keep 10 most-recent versions per package (rollback targets); delete anything >30 days the KEEP policy doesn't protect. Bounds previously-unbounded image growth. | storage capped |
| Prod Cloud Run `min_instances=1` | **Unchanged** — kept warm for visitors (deliberate). | — |

ADR-009 amended with the cost-retirement rationale.

## Testing

No application code changed — this is Terraform + one ADR. Verification:

[...]

PR #863, opened 2026-07-20
[infra] Add gke_deletion_protection var (default false) so GKE teardown doesn't block

### LINKED ISSUE #862: [infra] gke.tf cluster lacks deletion_protection=false — terraform teardown blocks when enable_gke flips off
## Problem

`infra/terraform/gke.tf`'s `google_container_cluster.main` does not set `deletion_protection`, so it defaults to `true` — a **Terraform-provider-side guard**, not a GCP-level setting (confirmed: the live cluster's `deletionProtection` API field is empty). When `enable_gke` flips to `false` (`count` → 0), `terraform apply` tries to destroy the cluster and hard-fails:

```
Error: Cannot destroy cluster because deletion_protection is set to true.
Set it to false to proceed with cluster deletion.
```

## Where it surfaced

Landing #861 (staging GKE retirement, 2026-07-20). The `terraform apply` successfully destroyed the Cloud NAT + router + workload-identity binding, but could not destroy the cluster. Worked around by deleting the staging cluster out-of-band via `gcloud container clusters delete` (GCP allows it — the guard is Terraform-only), after which `terraform` reconciles the now-absent resource on refresh. So #861 is **not** blocked by this — it merges with `enable_gke=false` and no cluster.

## Why it will recur

#861 keeps the GKE stack behind the `enable_gke` flag and documents re-enabling it (flip to `true`, apply) to revive the ADR-009 GKE-vs-Cloud-Run A/B. If GKE is re-enabled and later disabled again, the recreated cluster again defaults to `deletion_protection = true` and blocks the next Terraform teardown — the same wall, needing the same manual gcloud workaround.

## Fix

Add `deletion_protection = false` to `google_container_cluster.main` in `infra/terraform/gke.tf` (or expose it as a variable defaulting to `false` for this single-user/portfolio context, so a future prod cluster could opt back into `true`). A future `enable_gke=false` apply then destroys the cluster cleanly with no manual step.

## Notes
- Discovered while landing #861.
- Low urgency: only bites if GKE is re-enabled later; prod already runs `enable_gke=false` with no cluster.

brownm09:
## Summary

`infra/terraform/gke.tf`'s `google_container_cluster.main` never set `deletion_protection`, so with the pinned `hashicorp/google-beta ~> 5.0` provider (locked at 5.45.2) the field **provider-defaulted to `true`** — a Terraform-side guard, not a GCP-level setting. That default blocks `terraform apply` from destroying the cluster whenever `enable_gke` flips `true → false` (`count → 0`):

```
Error: Cannot destroy cluster because deletion_protection is set to true.
```

This surfaced landing #861 (staging GKE retirement); the staging cluster had to be deleted out-of-band via `gcloud container clusters delete`. The gap recurs the next time GKE is re-enabled and later disabled.

## Change

Per the issue's variable option, this adds a **`gke_deletion_protection` bool variable (default `false`)** matching the existing `enable_gke` feature-flag convention, and wires `deletion_protection = var.gke_deletion_protection` into the cluster. A future long-lived prod cluster can opt back into protection by setting the var `true` in `terraform.tfvars.production`.

- `infra/terraform/variables.tf` — new `gke_deletion_protection` variable (heredoc-documented, `type = bool`, `default = false`).
- `infra/terraform/gke.tf` — `deletion_protection = var.gke_deletion_protection` on `google_container_cluster.main`.

**Forward-looking, inert today.** Both `terraform.tfvars.staging` and `terraform.tfvars.production` are currently `enable_gke = false`, so `count = 0` and no cluster exists in any state. This performs no in-place update and no destroy on apply — it simply ensures the *next* cluster created is born with `deletion_protection = false` so a later `enable_gke = false` tears it down cleanly. No `.tfvars` edits (the `false` default already covers both environments).

## Acceptance criteria (from #862)

- [x] `deletion_protection` is explicitly set on `google_container_cluster.main` (no longer provider-defaults to `true`).
- [x] Defaults to `false` for this single-user / portfolio context.
- [x] Exposed as a variable so a future prod cluster can opt back into `true`.
- [x] `terraform validate` + `fmt -check` pass.

## Testing

No JS/TS surface is touched, so `npm test` is **N/A** (config-only change; the coverage table treats these as "None required" and the suite is unchanged). There is no Terraform test suite and no CI `terraform validate`/`fmt` gate, so the applicable verification is the Terraform CLI locally (v1.15.3):

```
$ terraform -chdir=infra/terraform fmt -check      # exit 0 — no formatting changes
$ terraform -chdir=infra/terraform init -backend=false
Terraform has been successfully initialized!
$ terraform -chdir=infra/terraform validate
Success! The configuration is valid.
```

Pre-commit lint hook (`turbo run lint`) also passed: `7 successful, 7 total (FULL TURBO)`.

The advisory `prod-plan-pr.yml` CI job (`terraform plan (production)`, non-blocking) will show **no change** to `google_container_cluster.main` — prod is `enable_gke = false`, so the cluster is absent — confirming the fix is inert for current state.

Closes #862

`/review` (claude-opus-4-8) returned a **clean review — 0 blocking, 0 non-blocking findings**. Nothing to dispose. Premerge checkpoints: `adr_warrant=not-warranted`, `doc_reconciliation=not-applicable`.

### DIFF
--- a/infra/terraform/gke.tf
+++ b/infra/terraform/gke.tf
@@ -13,6 +13,8 @@ resource "google_container_cluster" "main" {
 
   enable_autopilot = true
 
+  deletion_protection = var.gke_deletion_protection
+
   network    = google_compute_network.main.id
   subnetwork = google_compute_subnetwork.main.id
 
--- a/infra/terraform/variables.tf
+++ b/infra/terraform/variables.tf
@@ -91,6 +91,22 @@ variable "enable_gke" {
   default     = true
 }
 
+variable "gke_deletion_protection" {
+  description = <<-EOT
+    Terraform-side guard on the GKE Autopilot cluster (google_container_cluster.main).
+    When true, `terraform apply` refuses to destroy the cluster — including the implicit
+    destroy when `enable_gke` flips true → false (count → 0), which then hard-fails with
+    "Cannot destroy cluster because deletion_protection is set to true."
+
+    Defaults to false for this single-user / portfolio context so the ADR-009
+    GKE-vs-Cloud-Run A/B can be torn down cleanly. Set true (e.g. in
+    terraform.tfvars.production) if a future long-lived prod cluster should be protected
+    from an accidental terraform destroy. See issue #862.
+  EOT
+  type        = bool
+  default     = false
+}
+
 variable "cloud_run_min_instances" {
   description = "Minimum Cloud Run instances per service. null = use environment default (1 in production, 0 in staging). Set 0 in production for scale-to-zero (adds ~2s cold start to first request, near-zero idle cost)."
   type        = number


brownm09:
## Review: [infra] Add gke_deletion_protection var (default false) so GKE teardown doesn't block

### Summary
**Merge.** This is a minimal, correct, well-scoped fix (2 files, +18/-0) for issue #862. It adds a `gke_deletion_protection` bool variable (default `false`) and wires `deletion_protection = var.gke_deletion_protection` onto `google_container_cluster.main`, which previously left the field unset and thus provider-defaulted to `true` — the guard that blocks `terraform apply` from destroying the cluster when `enable_gke` flips true→false. No blocking or non-blocking findings. The change is inert for current state (both environments run `enable_gke = false`, so `count = 0` and no cluster is instantiated) and correctly governs the next cluster create; `deletion_protection` is a mutable argument, so even with a live cluster this would be an in-place update, never a replacement.

Verification checks confirmed by the author (fmt clean, `validate` "Success!", lint 7/7) are appropriate — there is no Terraform test suite and no CI `terraform validate`/`fmt` gate in this repo, so the local CLI run is the applicable verification. The advisory prod `terraform plan` will show no cluster change, as expected.

---

### Findings

None. The design choice of a dedicated `gke_deletion_protection` variable (rather than reusing the `deletion_protection = var.environment == "production"` idiom at `main.tf:192`) is deliberate and documented in the variable's heredoc and PR body — an explicit knob decoupled from `environment` so a future long-lived prod cluster can opt into protection without re-introducing the teardown wall for staging/experiments. Not flagged.

---

Issue #869, opened 2026-07-22
[fix] WIF google.subject mapping overflows the 127-byte limit under merge_group (blocks every queued PR)


brownm09:
## Problem

The merge queue went live on 2026-07-22 (#695 / #729, PR #864). The **first** PR through it failed: every GCP-authenticating job in the `merge_group` run died at `google-github-actions/auth`:

```
failed to generate Google Cloud federated token for //iam.googleapis.com/***:
{"error":"invalid_request","error_description":"The size of mapped attribute google.subject
exceeds the 127 bytes limit. Either modify your attribute mapping or the incoming assertion
to produce a mapped attribute that is less than 127 bytes."}
```

`Terraform (staging)` failed → `Build & Push Images`, `Deploy API (staging)`, `Deploy web (staging)` skipped → the required `Staging Integration Tests` check failed its deploy-prereq guard → **GitHub dropped the PR from the merge queue.**

## Root cause

`infra/terraform/main.tf:529` maps the federated subject to the raw OIDC subject claim:

```hcl
attribute_mapping = {
  "google.subject"       = "assertion.sub"          # <-- the problem
  "attribute.actor"      = "assertion.actor"
  "attribute.repository" = "assertion.repository"
}
```

Under `merge_group`, GitHub's `sub` claim embeds the merge-queue's synthetic ref:

```
repo:merickvaughn/lifting-logbook:ref:refs/heads/gh-readonly-queue/main/pr-864-<40-char-sha>
```

No `pull_request` or `push` event produces a ref anywhere near that long (a typical `pull_request` sub is ~46 bytes), so the mapping was **latent-broken** and only became reachable the moment the queue was enabled.

**Not caused by the org transfer.** The equivalent subject under the old `brownm09` owner is only 4 bytes shorter — both are the same order of magnitude, and the transfer neither introduced nor meaningfully worsened this.

> **Diagnostic caveat:** a reconstruction of the above subject string measures 119 bytes, i.e. *under* the 127 limit — so GitHub's actual `merge_group` subject carries more than the documented shape. The raw JWT is not readable from the run logs. This does not change the diagnosis (the mapping is what Google rejected) or the fix, but the exact byte arithmetic is unconfirmed.

## Fix

Map `google.subject` to a short, bounded value:

```hcl
"google.subject" = "assertion.repository"    # "merickvaughn/lifting-logbook" -- constant 28 bytes
```

[...]

PR #864, opened 2026-07-22
[infra] Transfer to merickvaughn org + refresh board/refs to unlock the merge queue

brownm09:
## Summary

Transfers `lifting-logbook` from the personal account `brownm09` to the **`merickvaughn` organization** to unlock GitHub's native **merge queue** — the structural fix ([ADR-030](docs/adr/ADR-030-github-merge-queue-adoption.md)) for the [#673](https://github.com/merickvaughn/lifting-logbook/issues/673) staging-deploy live-lock that starves normal auto-merge and forced `gh pr merge --admin` (PR #715). Merge queue requires an **org-owned** repo, which was the sole remaining blocker ([#729](https://github.com/merickvaughn/lifting-logbook/issues/729)).

**This PR is itself the live validation** — it merges *through* the newly-enabled queue.

Closes #729
Closes #695
Closes #673

## What changed in this PR (repo)

| Commit | Change |
|---|---|
| `1183024` | Pre-flight backups: live branch protection + all 523 board items (Epic/Status/Priority) → `.claude/backups/` |
| `04923fa` | **GCP WIF re-point** — `infra/terraform/main.tf` provider `attribute_condition` + 2 `principalSet` IAM members → `merickvaughn/lifting-logbook`. Applied **out-of-band** (owner creds) to both the staging and production terraform workspaces. |
| `4c71cb8` | Bulk repo-path rewrite (66 files) `brownm09/lifting-logbook` → `merickvaughn/lifting-logbook`; other-repo refs (`brownm09/dev-env` etc.) and historical `.claude/backups/` left intact. |
| `a9ae3f6` | Refresh the 3 project-board ID caches (`.claude/hook-config.json`, `.claude/propose.json`, `CLAUDE.md`) to the new org board (project #2, node `PVT_kwDOEecHO84BeFl_`) + owner `merickvaughn`. |
| `5ead677` | **Docs → live state** (Phase 7 + fixes both `/review` reliability findings): ADR-030 dated 2026-07-22 superseding amendment; `docs/operations/branch-protection.md` live merge-queue config + the REST-vs-GraphQL verification trap. |

## Out-of-band actions already completed (GitHub / GCP state)

- ✅ Repo transferred → `owner.type: Organization`; secrets (`GCP_*`, `TF_STATE_BUCKET`, `BRANCH_PROTECTION_READ_TOKEN`) + branch protection travelled with the repo.
[...]

Issue #627, opened 2026-06-30
[chore] post-tool-use hook prints stale Epic option IDs (hook-config.json drift)


brownm09:
## Problem

When the project-board hook (`post-tool-use.py`) fires after `gh issue create` / `gh pr create` in a lifting-logbook session, the `gh project item-edit` command it suggests lists **stale, incomplete Epic option IDs** — only 6 of the project's 10 epics, with option IDs that no longer exist on the project. Picking any listed ID yields a command that fails, because the option ID is dead.

Observed 2026-06-30: the hook printed `Architecture & Documentation = 656e470c`, but the live GitHub Project API and `CLAUDE.md` both have `f7202bcd`. None of the 6 IDs the hook printed match the live API.

## Root cause

The hook reads `epic_options` **verbatim** from `.claude/hook-config.json` — there is no live API fetch. That map is a hand-maintained cache of the project's single-select option IDs, and **single-select option IDs regenerate on every option mutation** (the documented hazard in `CLAUDE.md` → GitHub Project & Epic Assignment).

The 2026-05-10 Epic mutation (which added *Observability*, *API Implementation*, *Client Applications*, *Operations* and regenerated all IDs — see `.claude/backups/epic-mutation-2026-05-10.json`) updated two of the three places these IDs are cached:

- `CLAUDE.md` Epic options table — updated ✅
- `.claude/propose.json` (`epics`) — updated ✅
- `.claude/hook-config.json` (`epic_options`) — **missed ❌** → drifted

`hook-config.json` is the only cache the `Backup-and-restore procedure` in `CLAUDE.md` never names, so it was forgotten.

## Goal

1. Bring `.claude/hook-config.json` `epic_options` back in sync with the live project (all 10 current epics + correct IDs), so the hook's suggested command works.
2. Close the recurrence gap: the option-mutation procedure in `CLAUDE.md` must name **all three** ID caches so none drifts again.

## Acceptance criteria

- [ ] `.claude/hook-config.json` `epic_options` lists all 10 current epics with IDs matching the live API / `CLAUDE.md` / `propose.json`.
- [ ] `CLAUDE.md` Backup-and-restore procedure lists `propose.json` and `hook-config.json` alongside the Epic options table as caches to update after any option mutation.
- [ ] Verified: feeding the corrected config to the hook's `format_reminder` prints all 10 correct IDs.

PR #870, opened 2026-07-22
[chore] Add board-ID cache drift-detection (check-board-id-sync.mjs)

### LINKED ISSUE #865: [chore] Add board-ID cache drift-detection (check-board-id-sync.mjs)
## Problem

Three files duplicate the same GitHub Project board IDs — project node ID (`PVT_kwDOEecHO84BeFl_`), the Epic/Status/Priority field IDs, and the 10 Epic option IDs:

- `.claude/hook-config.json` — `epic_options` map, consumed by the `post-tool-use.py` project-board hook
- `.claude/propose.json` — `epics` array, consumed by `/propose`
- `CLAUDE.md` — the Epic-options table + the Standard-Issue-Workflow snippets

The duplication is **unavoidable** (three distinct consumers), but there is **no drift-detection guard** for this cache class — unlike the repo's other lockstep-file sets, each of which already has a `check-*-sync.mjs` + a CI step:

- `scripts/check-grafana-endpoint-sources.mjs` (#785)
- `scripts/check-otel-config-sync.mjs` (#788)
- `scripts/check-turbo-version-sync.mjs` (#674 / #692)

CLAUDE.md itself records the motivating failure: the 2026-05-10 board-option mutation left `.claude/hook-config.json` stale while the other caches were updated, suggesting dead Epic option IDs until [#627](https://github.com/merickvaughn/lifting-logbook/issues/627). A partial refresh is the dangerous failure mode — it makes the board hook write to a dead field.

Surfaced during the review of [PR #864](https://github.com/merickvaughn/lifting-logbook/pull/864) (the org transfer re-cut every one of these IDs) as non-blocking finding B3.

## Proposed fix

Add `scripts/check-board-id-sync.mjs` that asserts the project node ID, field IDs, and Epic option IDs agree across all three files, and wire it as a CI step (`ci.yml` → `lint-and-test`), mirroring the existing `check-*-sync.mjs` guards. Document it in the CLAUDE.md "Backup-and-restore procedure" step 3 (the place that instructs updating all three caches) so a future board mutation runs the guard.

## Acceptance criteria

- [ ] `scripts/check-board-id-sync.mjs` parses all three files and fails non-zero on any node/field/option-ID divergence
- [ ] Wired as a CI step so a drifting PR fails
- [ ] A deliberate mismatch is caught locally (manual test) before merge
- [ ] CLAUDE.md references the new guard in the board-mutation procedure

brownm09:
## Summary

Adds `scripts/check-board-id-sync.mjs`, a drift guard for the GitHub Project board IDs that are duplicated across three files (each with a distinct consumer, so the duplication is unavoidable):

| File | Consumer |
|---|---|
| `.claude/hook-config.json` | `post-tool-use.py` project-board hook |
| `.claude/propose.json` | `/propose` |
| `CLAUDE.md` | Epic-options table + Standard-Issue-Workflow / Backup-and-restore snippets |

One `updateProjectV2Field` mutation — or an org transfer — re-cuts the project node ID, the single-select field IDs and all 10 Epic option IDs at once. A **partial** refresh is the dangerous outcome: the board hook silently writes to a dead field, exactly as it did after the 2026-05-10 mutation (#627). Every other lockstep-file set in this repo already has a guard + CI step; this cache class had none.

Surfaced as non-blocking finding **B3** while reviewing #864.

Closes #865

## Design decision: relative, not pinned

The guard asserts **cross-file agreement** rather than matching hard-coded expected IDs. This is deliberate and load-bearing:

- #864 (the org transfer) is still open, so `main` currently carries the **old** IDs (`PVT_kwHOAjEKvM4BTuEF`, owner `brownm09`) and #864's branch carries the **new** ones (`PVT_kwDOEecHO84BeFl_`, owner `merickvaughn`).
- A guard pinned to the new IDs would redden `main` immediately and block every PR until #864 lands.
- A relative guard passes in **both** states while still failing any partial refresh — which is the failure mode #865 actually targets. Verified against both ID sets (see Testing).

**Scope boundary:** this is a *drift* guard, not a *liveness* guard. It proves the three caches agree with each other, **not** that they match the live GitHub API — three caches stale in lockstep pass by design. Liveness stays the manual step in CLAUDE.md's Backup-and-restore procedure; a live API call would need network + auth in CI. This is stated in the script header, the CLAUDE.md section, and the `scripts/README.md` row so nobody later mistakes one for the other.

## What it checks

Beyond the issue's acceptance criteria (node ID, field IDs, Epic option IDs), the guard also covers the owner, project number and Done option ID — the rest of the "board identity" surface that #864 re-cut and that lives in the same three caches.

| Value | hook-config.json | propose.json | CLAUDE.md |
|---|---|---|---|
| Project node ID | `project_node_id` | `github_project.node_id` | 6 occurrences |
| Epic field ID | `epic_field_id` | `github_project.epic_field_id` | 2 occurrences |
| Status field ID | `status_field_id` | — | 2 occurrences |
| 10 Epic option IDs | `epic_options` | `epics[]` | Epic-options table |
| Owner | `project_owner` | `github_project.owner` | board contexts only |
| Project number | `project_number` (string) | `github_project.number` (int) | 3 occurrences |
| Done option ID | `done_option_id` | — | step-9 snippet |

Notable details:

- **CLAUDE.md is checked against itself.** The node ID alone appears six times, so a partial hand-edit is a real failure mode that a naive one-occurrence parser would miss.
- **Field IDs are classified against `hook-config.json`**, the only cache that labels which field is which. This is also what cross-checks the Status field ID, since `propose.json` carries none. An unrecognised field ID fails closed with a message telling you to cache it or extend the guard.
- **The string/int split is normalized** — `project_number` is `"2"` in one cache and `2` in the other.
- **Owner is read only from unambiguous board contexts** (`` owner: `X` ``, `--owner X`, `repository(owner:"X"`). A blanket org-name scan would false-positive forever on the legitimate `github.com/brownm09/dev-env` links CLAUDE.md cites; there is a regression test for exactly that.
- **All problems are collected and reported together**, not just the first — a partial refresh usually breaks several values at once, and seeing them together is what identifies the stale cache.
- `.claude/backups/*.json` (immutable historical snapshots that legitimately hold old IDs) are never read.

**Not covered:** the issue body mentions a Priority field ID, but no cache file holds it — verified across the worktree and on #864's branch. There is nothing to cross-check, so it is out of scope rather than invented.

## Testing

`npm test` — full suite, all workspaces:

```
@lifting-logbook/api-client  Tests: 27 passed, 27 total
@lifting-logbook/core        Tests: 322 passed, 322 total
@lifting-logbook/web         Tests: 282 passed, 282 total
@lifting-logbook/api         Tests: 511 passed, 511 total
Tasks: 12 successful, 12 total
```

**Tests: 1142 passed, 0 skipped, 0 failed (1m5.9s)**

`npm run typecheck` — 4 successful, 4 total. Pre-commit `turbo run lint` — 7 successful, 7 total.

New unit tests: `node --test scripts/check-board-id-sync.test.mjs` → **28 passed, 0 failed**, covering both parsers, the shared table scoper (Epic + Milestones), the CRLF path (git converts these files on Windows checkouts), the dev-env-link false-positive guard, and each divergence class.

### Manual mismatch verification (AC #3)

Each scenario run against a scratch copy so the real tree was never mutated; all exited 1 with a targeted message:

| Scenario | Caught as |
|---|---|
| Flip one Epic option ID in `propose.json` | `Epic option "CI/CD Foundation" has different IDs across caches` + per-file values |
| Flip `project_node_id` in `hook-config.json` | `Project node ID disagrees across caches` |
| Flip **one of six** `PVT_kw…` occurrences in `CLAUDE.md` | `CLAUDE.md disagrees with itself on the project node ID — found 2 distinct values` |
| Flip `status_field_id` in `hook-config.json` | `references field ID(s) absent from .claude/hook-config.json` + `never references the Status field ID` |
| Owner drift in `propose.json` (half-migrated org) | `Project owner disagrees across caches` |
| `done_option_id` drift | `Done option ID … does not appear in CLAUDE.md's --single-select-option-id snippets` |

### Forward-compatibility against #864

The guard was run against #864's versions of all three files (fetched at `infra/issue-729-org-transfer`) and passes on the all-new ID set:

```
OK: board IDs in sync across CLAUDE.md, .claude/hook-config.json, .claude/propose.json (PVT_kwDOEecHO84BeFl_, 10 Epic options).
```

So this cannot redden `main` whichever order the two PRs merge in.

## Pre-PR greps

- **Test-integrity grep:** clean. No skip markers, no deleted test files, no threshold changes.
- **Suppression grep:** one match, reviewed and **not** a suppression — `query($number:Int!){…}` in a test fixture. The `!` is GraphQL's non-null type marker inside a string literal (replicating CLAUDE.md's own snippet), not a TypeScript non-null assertion. No `ts-ignore`/`ts-expect-error`/`eslint-disable`/`?? null` anywhere in the diff.

## Environment note (not caused by this PR)

The first `npm test` run failed with 34 `apps/web` suites unable to resolve `next/server`. Root cause was a corrupted worktree install — **670 empty package directories** in the root `node_modules` (`next`, `react`, `react-dom`, `postcss`, `scheduler`, …), the documented Windows truncated-extract failure mode in CLAUDE.md's Platform section. Disk was fine (29 GB free), so not ENOSPC. Resolved with the documented recovery (`rm -rf node_modules … && npm ci`, exit 0); the web suite then went 34 failed → **46/46 suites passing**. No source change was involved, and this PR touches no `apps/web` code.

## Acceptance criteria

- [x] `scripts/check-board-id-sync.mjs` parses all three files and fails non-zero on any node/field/option-ID divergence
- [x] Wired as a CI step so a drifting PR fails (`ci.yml` → `lint-and-test`, paired `Verify` + `Unit test` steps matching the existing guards)
- [x] A deliberate mismatch is caught locally — six scenarios, table above
- [x] CLAUDE.md references the new guard in the board-mutation procedure (Backup-and-restore step 3, plus a dedicated Testing section and a `scripts/README.md` row)

---

`/review` posted [here](https://github.com/merickvaughn/lifting-logbook/pull/870#issuecomment-5048697575) — **0 blocking, 2 non-blocking**. Both fixed in this PR (commit `3601e6c`) rather than filed, per ADR-028's fix-in-PR default.

| # | Finding | Disposition |
|---|---|---|
| 1 | **[maintainability]** Guard omitted the milestone lists, which CLAUDE.md's own blockquote flags as the same drift class reached by a different trigger (a milestone isn't created by `updateProjectV2Field`, so nothing in the mutation flow prompts a cache refresh) | **Fixed in `3601e6c`** — added a milestone-title cross-check across all three caches. Only titles are comparable: the JSON caches store bare strings, the numbers live solely in CLAUDE.md's table. Factored the table scoper into a shared `parseMarkdownTable` so the Epic and Milestones tables cannot read each other's rows. Verified it catches a milestone added to CLAUDE.md alone. |
| 2 | **[maintainability]** Field-ID mismatch messages named CLAUDE.md as the deviant file, but in the #627 incident `hook-config.json` was the stale one — misdirecting the reader in exactly the case the guard exists for | **Fixed in `3601e6c`** — messages now state the disagreement without assigning fault ("one of the two is stale"), with a comment recording that `hook-config.json` is the reference only because it is the sole cache that labels which field is which. Pinned by a regression test. |

Docs updated to match in all three places the guard is described (CLAUDE.md step 3, the `###` Testing section, the `scripts/README.md` row), plus the "Milestones drift the same way" blockquote now notes the guard covers titles.

Re-verified after the fixes: guard passes on `main`'s ID set **and** on #864's post-transfer set (both now including milestones); unit tests 22 → **28 passed, 0 failed**; `npm test` **1142 passed, 0 skipped, 0 failed**; `npm run typecheck` 4/4; pre-commit lint 7/7.

## Merge blocker (not this PR)

`Lint & Test` — the check that exercises this PR's new CI steps — **passes**, as do DB Integration Tests, Terraform (staging), Observability Smoke, and Preflight.

`Build & Push Images` / `Staging Integration Tests` fail for a repo-wide reason unrelated to this diff: `main`'s `infra/terraform/main.tf:534` still pins the pre-transfer `assertion.repository == 'brownm09/lifting-logbook'`, so GCP WIF rejects the token with `unauthorized_client`. Because `staging.yml:146-154` runs `terraform apply -auto-approve`, every PR run from `main` re-applies the stale condition to the live provider. Filed as #871; distinct from #869 (`google.subject` 127-byte overflow under `merge_group`, `main.tf:529`).

This PR is mergeable once that chain (#869 → #864, or a direct one-line fix to `main`) is resolved.


brownm09:
## CI status — the two red checks are not caused by this PR

`Lint & Test` **passes**, which is the check that exercises this PR's two new CI steps (`Verify board-ID caches are in sync` + `Unit test board-ID sync parsing`). `DB Integration Tests`, `Terraform (staging)`, `Observability Stack Smoke Test` and `Preflight` all pass too.

The two failures are pre-existing infrastructure breakage:

1. **`Build & Push Images` / `Staging Integration Tests`** — traced to the emitting step rather than the downstream message: `google-github-actions/auth` fails with `unauthorized_client: The given credential is rejected by the attribute condition`. `infra/terraform/main.tf:534` on `main` still pins `assertion.repository == 'brownm09/lifting-logbook'`, which no longer matches the post-transfer repo identity. Filed as #871. This is distinct from #869 (`google.subject` 127-byte overflow under `merge_group`, `main.tf:529`). This PR is docs + two `.mjs` scripts + a CI step and cannot affect image builds.

2. **`Review Gate`** — expected; `/review` has not run yet.

Blocked on #871 / #869 / #864 for the staging checks.

brownm09:
## Review: [chore] Add board-ID cache drift-detection (check-board-id-sync.mjs)

### Summary

**Merge with minor changes.** No blocking findings. The change is purely additive (645 additions, 0 deletions across 5 files), ships 22 unit tests alongside the new guard, is wired into CI following the established `Verify …` + `Unit test … parsing` step pairing, and is documented in all three places the repo's other guards are (CLAUDE.md procedure step, dedicated `###` Testing section, `scripts/README.md` row).

The design decision that matters most is correct and well-justified: the guard asserts **cross-file agreement** rather than pinning expected literal IDs. Since #864 is still open, `main` holds the pre-transfer IDs and #864's branch holds the post-transfer set — a pinned guard would redden `main` immediately. The PR verifies the relative guard passes against both ID sets, so merge order between #870 and #864 is a non-issue. The drift-vs-liveness scope boundary is stated in the script header, the CLAUDE.md section, and the README row, which is the right level of repetition for a distinction that is easy to misremember later.

Highest-severity finding is non-blocking: the guard omits the **milestone** arrays, which the very CLAUDE.md section it documents explicitly warns "drift the same way."

The diff exceeds 400 lines, but ~85% is the new script plus its test file — it is appropriately scoped and should not be split.

---

### Non-Blocking Findings

**[maintainability]** `scripts/check-board-id-sync.mjs:118`

The guard covers the project node ID, field IDs, Epic option IDs, owner, project number, and Done option ID — but not the **milestone** lists, which are duplicated across the same three files: `.claude/hook-config.json` → `milestones` (array of 4 title strings), `.claude/propose.json` → `milestones` (the same array), and CLAUDE.md's `**Milestones:**` table (title → number).

This matters here specifically because CLAUDE.md's own blockquote immediately below the Backup-and-restore procedure calls this out as a known drift class with a *different* trigger: milestones aren't created by `updateProjectV2Field`, so nothing in the mutation flow prompts you to update them — a milestone added via `gh api repos/.../milestones` or the web UI leaves all three caches stale independently. The repo has now documented that hazard twice without guarding it, and this PR is the natural home for the check. The titles are byte-identical across all three today, so the check is cheap: compare the title sets.

**Suggestion:** Extend `compareBoardIds` with a milestone-title set comparison across the three sources (parse CLAUDE.md's table with a scoper mirroring `parseEpicOptionsTable`), or file a follow-up issue and link it. Numbers only exist in CLAUDE.md, so titles are the only cross-checkable field — that is still enough to catch the documented failure.

---

**[maintainability]** `scripts/check-board-id-sync.mjs:196-213`

The field-ID classification step treats `.claude/hook-config.json` as authoritative, because it is the only cache that labels which field is Epic and which is Status. When a mismatch is found the message reads *"CLAUDE.md references field ID(s) absent from `.claude/hook-config.json`"* and *"CLAUDE.md never references the Status field ID … cached in hook-config."*

Both sentences name CLAUDE.md as the deviant file. But in the exact scenario this guard exists to catch — the #627 incident, where `hook-config.json` was the cache left stale while the others were refreshed — `hook-config.json` is the stale one and CLAUDE.md is correct. The reader is pointed at the wrong file. In practice the accompanying `Epic field ID disagrees across caches` line (which prints all three values side by side) makes the truth recoverable, so this is a diagnostics-quality issue rather than a correctness one — but the misdirection lands in precisely the case the guard was written for.

**Suggestion:** Neutralize the phrasing so it states the disagreement rather than assigning fault — e.g. "field ID `X` appears in CLAUDE.md but is not cached in hook-config.json; one of the two is stale" — and note that `hook-config.json` is used as the field-name reference only because it is the sole cache that labels them.

---

### Style Notes

The script matches the house idiom closely — shebang, JSDoc header with issue references and a `Usage:` line, bare `fs`/`url`/`path` imports, root resolved from `import.meta.url` rather than `process.cwd()`, pure exported helpers above `main()`, and the `process.argv[1]` direct-execution guard with the standard explanatory comment. Error output follows the established `ERROR: <summary>.` + two-space-indented details + remediation shape, and the success line matches the `OK: … (value).` form.

Two deliberate departures from the siblings are both improvements and are worth keeping: collecting *all* problems before exiting rather than reporting first-divergence-only (a partial refresh usually breaks several values at once, and the set of failures is what identifies the stale file), and validating required-key presence explicitly so two caches that are both missing a key cannot compare equal and pass silently.

The test file's fixture design — a synthetic mini-CLAUDE.md that reproduces the real file's repeated-ID shapes, with per-case overrides — reads well and made the CRLF and dev-env-link regression cases cheap to express.

---

### Findings Disposition

Both findings are non-blocking but, per ADR-028, neither may be left as-is: each must be fixed in this PR or filed and linked before merge, with the disposition recorded in a "Review findings disposition" section of the PR body.

---

brownm09:
## Review: [chore] Add board-ID cache drift-detection (check-board-id-sync.mjs) — re-review at `3601e6c`

### Summary

**Approve — merge once staging CI is unblocked.** This re-review covers commit `3601e6c`, which addresses both non-blocking findings from the [first pass](https://github.com/merickvaughn/lifting-logbook/pull/870#issuecomment-5048697575). Both are resolved, and the delta introduces no new findings.

Verified against the remote branch (`git show origin/chore/issue-865-board-id-sync:…`) rather than a local tree, per ADR-004.

---

### Prior findings — resolution

**1. [maintainability] Milestone lists uncovered — RESOLVED.**
`parseMilestonesTable` (`scripts/check-board-id-sync.mjs:133`) now reads CLAUDE.md's `**Milestones:**` table, both JSON parsers pick up their `milestones` arrays, and `compareBoardIds` compares the title sets across all three caches. The scoping concern this raised — two adjacent markdown tables in the same file — is handled better than the original code did it: the table reader was factored into a shared `parseMarkdownTable(source, marker, valuePattern)` (`:98`) that scopes by marker *and* constrains the value column's shape, so the Epic table (`[0-9a-f]{8}`) and the Milestones table (`\d+`) cannot read each other's rows even if one were moved adjacent to the other. Restricting the check to titles is the correct call and is documented in the function's docstring: the JSON caches store bare title strings, so numbers exist only in CLAUDE.md and have nothing to be compared against.

**2. [maintainability] Mismatch messages assigned fault — RESOLVED.**
Both messages (`:257`, `:265`) now state the disagreement symmetrically — "one of the two is stale" — instead of naming CLAUDE.md as the deviant file, and a comment at `:196` records that `hook-config.json` serves as the field-name reference only because it is the sole cache that labels which field is which, explicitly noting it was the stale cache in #627. A regression test pins the neutral wording, which is the right way to keep it from regressing during future message edits.

---

### Notes on the delta

No new blocking or non-blocking findings.

`requireMilestones` makes an empty milestone list a failure in all three caches. That is stricter than strictly necessary — a project that deliberately dropped every milestone would now fail — but it is fail-closed, matches the existing `requireEpics` shape exactly, and the alternative (silently skipping when empty) is the weaker default for a drift guard. Consistent with the file's established posture; no change requested.

The milestone comparison uses set semantics rather than positional, so reordering the table is correctly not flagged. Behavior of the refactored `parseEpicOptionsTable` is unchanged — `Object.fromEntries(rows)` collapses duplicate names last-wins, identical to the previous in-loop assignment.

Test count 22 → 28, covering the milestone parser, milestone drift in either direction (title only in CLAUDE.md, title renamed in one JSON cache), Epic/Milestones table isolation, and the neutral-wording regression.

---

### Merge readiness

Everything in this PR's control is green: `Lint & Test` (which runs the two new CI steps), `DB Integration Tests`, `Observability Stack Smoke Test`, `Preflight`.

`Terraform (staging)` → `Build & Push Images` → `Staging Integration Tests` are red for a repo-wide cause unrelated to this diff — the stale WIF `attribute_condition` on `main` tracked in #871. This PR should merge as soon as that chain clears; nothing further is required of it.

---

brownm09:
## Review: [chore] Add board-ID cache drift-detection (check-board-id-sync.mjs) — post-rebase at `5d9d915`

### Summary

**Approve.** This re-review covers the rebase onto `c997a7b` (#864, the org transfer). Everything this PR authored is unchanged from the reviewed state at `3601e6c`: `git diff 3601e6c 5d9d915` is empty for `scripts/check-board-id-sync.mjs`, `scripts/check-board-id-sync.test.mjs`, and `.github/workflows/ci.yml`, and this branch's `scripts/README.md` row is byte-identical. The two files that do differ, differ only by content arriving *from* #864: `scripts/README.md` picks up #864's `brownm09` → `merickvaughn` rewrites in pre-existing rows, and `CLAUDE.md` picks up #864's rewrites plus the one conflict resolution described below. Both findings from the [first pass](https://github.com/merickvaughn/lifting-logbook/pull/870#issuecomment-5048697575) remain resolved (confirmed [here](https://github.com/merickvaughn/lifting-logbook/pull/870#issuecomment-5048794636)); no new findings.

### Rebase verification

**One conflict, in CLAUDE.md's Backup-and-restore step 3.** #864 rewrote the `#627` link in the `hook-config.json` blockquote from `brownm09` to `merickvaughn`; this branch appends the guard paragraph directly after that blockquote. Resolved by taking #864's rewritten URL and keeping the appended paragraph — the two changes are adjacent, not semantically opposed. `scripts/README.md` was modified by both branches but merged without conflict (#864's edit is a bulk URL rewrite; this branch's row sits at the top of the table and carries no GitHub URLs).

**No #864 content was clobbered.** The complete `CLAUDE.md` diff against the new `main` is additions only, plus a single replaced line where this branch's own milestones sentence extends the "Milestones drift the same way" blockquote. Every `lifting-logbook` URL in the result is `merickvaughn`.

**The two remaining `brownm09` references in the diff are correct and must not be rewritten.** Both point at `github.com/brownm09/dev-env` and live in the test fixture that exists specifically to prove the owner extractor reads only board contexts and ignores legitimate cross-repo links. A blanket org-rename sweep over this file would break that regression test's purpose.

**The relative-guard design validated itself.** With no change to the guard, it now reports `OK … (PVT_kwDOEecHO84BeFl_, 10 Epic options, 4 milestones)` — the post-transfer node ID — having previously reported the pre-transfer one. This is precisely the property that let the guard sit on an open PR across the transfer without reddening `main`, and it means the Epic option IDs, field IDs, owner, project number and milestone titles that #864 re-cut are all confirmed mutually consistent in the merged state.

#864 touched no dependency manifests, so no reinstall was required.

### Verification after rebase

- `node scripts/check-board-id-sync.mjs` → OK, exit 0, against the new ID set
- `node --test scripts/check-board-id-sync.test.mjs` → **28 passed, 0 failed**
- `npm test` → **Tests: 1142 passed, 0 skipped, 0 failed**; 12/12 turbo tasks
- `npm run typecheck` → 4/4; pre-commit `turbo run lint` → 7/7
- `Terraform (staging)` now passes on this PR, confirming #873's WIF fix cleared the #871 blocker

Pushed with `--force-with-lease` after an explicit `git fetch`, never a bare `--force`.

---

PR #873, opened 2026-07-22
[fix] Drive WIF repo identity from a variable; bound google.subject to unblock CI

### LINKED ISSUE #869: [fix] WIF google.subject mapping overflows the 127-byte limit under merge_group (blocks every queued PR)

### LINKED ISSUE #871: [fix] Stale WIF attribute_condition on main pins the pre-transfer repo name — blocks every PR and re-breaks the live provider
## Problem

Every PR cut from `main` now fails `Build & Push Images` at `google-github-actions/auth`:

```
failed to generate Google Cloud federated token for //iam.googleapis.com/***:
{"error":"unauthorized_client","error_description":"The given credential is rejected by the attribute condition."}
```

`Build & Push Images` fails → `Deploy API (staging)` / `Deploy web (staging)` skip → the required `Staging Integration Tests` check fails its deploy-prereq guard → **the PR cannot merge.**

First observed on [PR #870](https://github.com/merickvaughn/lifting-logbook/pull/870) (a docs-and-scripts-only change that cannot affect image builds): [failing job](https://github.com/merickvaughn/lifting-logbook/actions/runs/29936974236/job/88981183833).

## Root cause

`infra/terraform/main.tf:534` on `main` still pins the WIF provider's attribute condition to the **pre-transfer** repo name:

```hcl
# origin/main
attribute_condition = "assertion.repository == 'brownm09/lifting-logbook'"
```

The repo was transferred to the `merickvaughn` org, so GitHub's OIDC token now asserts `repository = merickvaughn/lifting-logbook`. That no longer satisfies the condition, and GCP rejects the exchange with `unauthorized_client`.

[PR #864](https://github.com/merickvaughn/lifting-logbook/pull/864) already carries the corrected line:

```hcl
# infra/issue-729-org-transfer
attribute_condition = "assertion.repository == 'merickvaughn/lifting-logbook'"
```

## Why this is worse than a per-PR failure

`.github/workflows/staging.yml:146-154` runs `terraform apply -auto-approve` on every PR run — it is an apply, not a plan. So a PR branched from `main` does not merely fail its own auth: it **re-applies the stale attribute condition to the live WIF provider**, breaking GCP authentication for every other branch until something applies the corrected value again.

That is consistent with the observed run history for `staging.yml`:

| Time (UTC) | Branch | Result |
|---|---|---|
| 2026-07-21 and earlier | various | success (pre-transfer: repo name matched the condition) |
| 2026-07-22 05:32, 15:14 | `infra/issue-729-org-transfer` | success (branch carries the corrected condition) |
| 2026-07-22 15:28 | `gh-readonly-queue/main/pr-864-…` | failure — separate cause, see #869 |
| 2026-07-22 16:12 | `chore/issue-865-board-id-sync` (from `main`) | **failure — this issue** |

## Not a duplicate of #869

#869 is a different failure in the same file: `google.subject` = `assertion.sub` overflowing the 127-byte limit under `merge_group` (`main.tf:529`, error `invalid_request` / size limit). This issue is `main.tf:534`, error `unauthorized_client` / attribute condition, and it reproduces on an ordinary `pull_request` run rather than only in the merge queue. Both must be fixed for the queue to drain, but they are independent lines and independent errors.

## Suggested fix

The corrected condition already exists in #864, so the cleanest resolution is to land #864 (which needs #869 fixed first, since the merge queue is how it lands). If #864 remains blocked, apply the one-line `attribute_condition` correction to `main` directly so PRs stop re-breaking the live provider.

Worth considering alongside: an attribute condition pinned to a hard-coded `owner/repo` string is a tripwire for any future transfer or rename. Deriving it from a variable, or widening it to `assertion.repository_owner`, would make the same class of breakage impossible to reintroduce silently.

## Verification notes

Verified by reading `infra/terraform/main.tf:534` at both `origin/main` and `infra/issue-729-org-transfer`, and by reading the emitting step in the failing job log rather than inferring from the downstream `Staging Integration Tests` message (which reports a misleading deploy-prereq failure). The `terraform apply` claim is from `.github/workflows/staging.yml:146-154` on `main`.

I did **not** inspect the live GCP WIF provider's current attribute-condition value — that requires `gcloud` access — so the exact present state of the live provider is inferred from the run history above, not directly confirmed.

brownm09:
Fixes the WIF outage that blocked every PR in the repo, and removes the hard-coded-repo tripwire that caused it.

## Why the proposed orderings could not run

Neither `#869 → #864` nor a one-liner to `main` was executable. By investigation time **no GitHub Actions run could authenticate to staging GCP at all** — including PR #864's own branch, which carries the correct code but could never reach `terraform apply` to land it. CI was the only automated apply path, so CI could not fix itself. (An admin-bypass merge would not have helped: merging code does not apply it.)

The self-break, executing across two consecutive runs on `chore/issue-865-board-id-sync`:

| Run | `Terraform (staging)` | Next job |
|---|---|---|
| [29936974236](https://github.com/merickvaughn/lifting-logbook/actions/runs/29936974236) (16:12) | **success** — re-applied the stale identity | `Build & Push Images` **failed** |
| [29938304891](https://github.com/merickvaughn/lifting-logbook/actions/runs/29938304891) (16:30) | **failed at `Authenticate to GCP`** (`unauthorized_client`) | skipped |

## Live state was worse than #871 described

The **service-account bindings had been reverted too**. Fixing only `attribute_condition` would have left token exchange succeeding and impersonation failing — which is why #864 correctly touches three lines, not one. **Production was already correct**; only staging was broken.

Prior state was captured to `.claude/backups/wif-provider-snapshot-*.json` and `wif-sa-bindings-snapshot-*.json` (committed, restorable) before any write, then all four values were corrected via `gcloud` and verified by read-back. CI can authenticate again — but that is **unstable until this PR lands**, since any PR branched from `main` that reaches `terraform apply` reverts it.

## What this changes

- **`var.github_repository`** (new, validated `owner/repo`) is now the single source of truth for the WIF trust boundary, replacing the three hard-coded `brownm09/lifting-logbook` strings — the `attribute_condition` and both `principalSet` bindings. A future transfer or rename is a one-line change, and the condition/binding pair cannot drift apart from each other.
- **`google.subject` → `assertion.repository`** (was `assertion.sub`), fixing #869. Under `merge_group`, GitHub's `sub` embeds the queue's synthetic `gh-readonly-queue` ref and overflows Google's 127-byte mapped-attribute limit. Verified safe: **no `principal://.../subject/...` binding exists anywhere in the repo** — both SA bindings authorize via `principalSet` on `attribute.repository`, and `attribute.actor` still carries per-actor audit granularity.

Both fixes are required together: the merge queue cannot drain without the `google.subject` fix, and nothing can merge without the condition fix.

## Acceptance criteria

- [x] `attribute_condition` no longer pins a pre-transfer repo name
- [x] Repo identity derived from a single variable rather than hard-coded at three sites (#871's tripwire concern)
- [x] `google.subject` mapped to a bounded value (#869)
- [x] Live staging provider + both SA bindings corrected and verified by read-back
- [x] Prior live state captured to a restorable, committed artifact before mutating

## Testing

`npm test` (via the documented `TURBO_BINARY_PATH` worktree workaround):

```
Tests:       511 passed, 511 total       (@lifting-logbook/api)
Tasks:    12 successful, 12 total
```

`npm run typecheck`: `Tasks: 4 successful, 4 total`. `terraform fmt -check` clean; `terraform validate` → `Success! The configuration is valid.`

No automated coverage is added: this is an infrastructure-config change with no runtime code path, and `infra/terraform/**` has no test harness in this repo. The meaningful verification is the live read-back above plus this PR's own staging `terraform apply`, which will now find no drift.

Suppression and test-integrity greps: both clean. No `package.json` changed, so the lockfile-sync check is N/A.

**Environment note (unrelated to this change):** the first suite run reported exit 0 while never executing — the documented Windows worktree `turbo.exe` failure. Re-running via `TURBO_BINARY_PATH` then surfaced 29 zero-byte `package.json` files in the worktree's `node_modules` (a partial-extraction defect; disk had 27 GB free, so not ENOSPC). Repaired with `rm -rf node_modules && npm ci` before the run above.

## Ordering

Merge this **before** #864 and #870 re-run: #864's branch still maps `google.subject = assertion.sub` and would re-break the merge queue. Once this lands, both should rebase and merge normally.

Closes #871
Closes #869

---

All 5 findings from the [review](https://github.com/merickvaughn/lifting-logbook/pull/873#issuecomment-5049589409) are resolved — none left as-is (ADR-028). Code fixes in `a5364f7`.

| # | Category | Finding | Disposition |
|---|---|---|---|
| 1 | security | **Blocking** — committed backup published `user:brownm2350@gmail.com` to a public repo | **fixed in `a5364f7`** — snapshot filtered to the `roles/iam.workloadIdentityUser` bindings the restore path actually needs; amended + force-pushed with `--force-with-lease` so the address is not at the branch tip |
| 2 | reliability | `attribute.actor` does not preserve audit attribution — the comment's claim was wrong | **fixed in `a5364f7`** (comment corrected to state the tradeoff honestly) + **filed [#874](https://github.com/merickvaughn/lifting-logbook/issues/874)** for the bounded-composite mapping that restores attribution |
| 3 | correctness | Production WIF drift makes `Plan (production)` permanently non-empty | **filed [#875](https://github.com/merickvaughn/lifting-logbook/issues/875)** — prod is plan-only and unaffected by #869, so this is drift-noise, out of scope for an outage fix |
| 4 | maintainability | `github_repository` validation admitted the CEL quote character | **fixed in `a5364f7`** — charset tightened to `^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$` |
| 5 | documentation | New WIF backup class has no documented restore path | **filed [#876](https://github.com/merickvaughn/lifting-logbook/issues/876)** — CLAUDE.md procedure + possible ADR |

**Question for author** (tfvars explicit vs. default) resolved in favour of keeping the default: setting `github_repository` in both tfvars files would restore two edit sites — a diluted version of the exact problem #871 was filed for. The single edit point is the property this PR is buying.

### Why #874 was deferred rather than fixed here

The composite mapping is the better long-term answer, but this PR is the fix for a live auth outage that blocked every PR in the repo. Re-shaping `google.subject` a second time would put an unproven value in the critical auth path while the outage is being cleared. The constant mapping is verified working end-to-end — all 13 checks green on `c0371ef`, including `Staging Integration Tests`, the required check that was failing.

### Verification after the amend

`terraform fmt -check` clean; `terraform validate` → `Success! The configuration is valid.`; lint 7/7 successful. The amend touches only the redacted snapshot, the validation regex, and a comment — **no change to any applied WIF value**, so the green run on `c0371ef` remains representative of the configuration being applied. CI re-runs on `a5364f7`.

### DIFF
--- a/.claude/backups/wif-provider-snapshot-2026-07-22-132117.json
+++ b/.claude/backups/wif-provider-snapshot-2026-07-22-132117.json
@@ -1,0 +1,34 @@
+{
+  "captured_at": "2026-07-22T17:21:18Z",
+  "issue": "871",
+  "staging":
+{
+  "attributeCondition": "assertion.repository == 'brownm09/lifting-logbook'",
+  "attributeMapping": {
+    "attribute.actor": "assertion.actor",
+    "attribute.repository": "assertion.repository",
+    "google.subject": "assertion.sub"
+  },
+  "displayName": "GitHub OIDC",
+  "name": "projects/910635705567/locations/global/workloadIdentityPools/lifting-logbook-stg-github-pool/providers/github-provider",
+  "oidc": {
+    "issuerUri": "https://token.actions.githubusercontent.com"
+  },
+  "state": "ACTIVE"
+}
+  ,"prod":
+{
+  "attributeCondition": "assertion.repository == 'merickvaughn/lifting-logbook'",
+  "attributeMapping": {
+    "attribute.actor": "assertion.actor",
+    "attribute.repository": "assertion.repository",
+    "google.subject": "assertion.sub"
+  },
+  "displayName": "GitHub OIDC",
+  "name": "projects/508649171914/locations/global/workloadIdentityPools/lifting-logbook-prod-github-pool/providers/github-provider",
+  "oidc": {
+    "issuerUri": "https://token.actions.githubusercontent.com"
+  },
+  "state": "ACTIVE"
+}
+}
--- a/.claude/backups/wif-sa-bindings-snapshot-2026-07-22-132117.json
+++ b/.claude/backups/wif-sa-bindings-snapshot-2026-07-22-132117.json
@@ -1,0 +1,27 @@
+{
+  "captured_at": "2026-07-22T17:24:40Z",
+  "issue": "871",
+  "staging_cicd": {
+    "bindings": [
+      {
+        "members": [
+          "principalSet://iam.googleapis.com/projects/910635705567/locations/global/workloadIdentityPools/lifting-logbook-stg-github-pool/attribute.repository/brownm09/lifting-logbook"
+        ],
+        "role": "roles/iam.workloadIdentityUser"
+      }
+    ],
+    "version": 1
+  },
+  "staging_plan": {
+    "bindings": [
+      {
+        "members": [
+          "principalSet://iam.googleapis.com/projects/910635705567/locations/global/workloadIdentityPools/lifting-logbook-stg-github-pool/attribute.repository/brownm09/lifting-logbook"
+        ],
+        "role": "roles/iam.workloadIdentityUser"
+      }
+    ],
+    "version": 1
+  },
+  "redaction_note": "Filtered to roles/iam.workloadIdentityUser bindings only — the sole entry the WIF restore path needs. Unrelated principals (including human user: grants) are intentionally omitted: this repo is public."
+}
--- a/infra/terraform/main.tf
+++ b/infra/terraform/main.tf
@@ -525,19 +525,39 @@ resource "google_iam_workload_identity_pool_provider" "github" {
     issuer_uri = "https://token.actions.githubusercontent.com"
   }
 
+  # google.subject is deliberately NOT assertion.sub (#869). Under merge_group,
+  # GitHub's `sub` embeds the merge queue's synthetic ref
+  # (repo:<owner>/<repo>:ref:refs/heads/gh-readonly-queue/main/pr-<N>-<40-char-sha>),
+  # which overflows Google's 127-byte mapped-attribute limit and fails the exchange
+  # with `invalid_request`. No pull_request/push ref comes close, so the mapping was
+  # latent-broken until the merge queue went live and dropped the first queued PR.
+  # assertion.repository is constant-length and safely bounded. No ACCESS depends on
+  # the subject: both SA bindings below authorize via principalSet on
+  # attribute.repository, and no principal://.../subject/... binding exists anywhere.
+  #
+  # AUDIT granularity is knowingly traded away, however. Cloud Audit Logs derive the
+  # federated principal from google.subject, so every run — whatever the actor — now
+  # authenticates as one constant subject, and a staging apply can no longer be
+  # attributed to the actor or PR that triggered it. attribute.actor does NOT recover
+  # this: it is usable in IAM conditions and principalSet members, but it is not the
+  # identifier logged as the principal. A bounded composite
+  # ("assertion.repository + '/' + assertion.actor") would restore attribution while
+  # staying far under 127 bytes; it is deliberately NOT taken here because this change
+  # is the fix for a live auth outage and re-shaping the subject a second time would
+  # put an unproven value in the critical path. Tracked in #874.
   attribute_mapping = {
-    "google.subject"       = "assertion.sub"
+    "google.subject"       = "assertion.repository"
     "attribute.actor"      = "assertion.actor"
     "attribute.repository" = "assertion.repository"
   }
 
-  attribute_condition = "assertion.repository == 'brownm09/lifting-logbook'"
+  attribute_condition = "assertion.repository == '${var.github_repository}'"
 }
 
 resource "google_service_account_iam_member" "cicd_wif_binding" {
   service_account_id = google_service_account.cicd.name
   role               = "roles/iam.workloadIdentityUser"
-  member             = "principalSet://iam.googleapis.com/${google_iam_workload_identity_pool.github.name}/attribute.repository/brownm09/lifting-logbook"
+  member             = "principalSet://iam.googleapis.com/${google_iam_workload_identity_pool.github.name}/attribute.repository/${var.github_repository}"
 }
 
 # ─── IAM — read-only plan service account (#545) ─────────────────────────────
@@ -582,5 +602,5 @@ resource "google_project_iam_member" "cicd_plan_roles" {
 resource "google_service_account_iam_member" "cicd_plan_wif_binding" {
   service_account_id = google_service_account.cicd_plan.name
   role               = "roles/iam.workloadIdentityUser"
-  member             = "principalSet://iam.googleapis.com/${google_iam_workload_identity_pool.github.name}/attribute.repository/brownm09/lifting-logbook"
+  member             = "principalSet://iam.googleapis.com/${google_iam_workload_identity_pool.github.name}/attribute.repository/${var.github_repository}"
 }
--- a/infra/terraform/variables.tf
+++ b/infra/terraform/variables.tf
@@ -29,6 +29,35 @@ variable "app_name" {
   default     = "lifting-logbook"
 }
 
+variable "github_repository" {
+  description = <<-EOT
+    The "owner/repo" this repository's GitHub Actions runs authenticate as, exactly as
+    it appears in the OIDC token's `repository` claim. Single source of truth for the
+    Workload Identity Federation trust boundary: it gates the provider's
+    attribute_condition AND both service-account principalSet bindings (cicd,
+    cicd_plan), which must always agree — a mismatch between them silently yields
+    either `unauthorized_client` at token exchange or a failed impersonation.
+
+    Change this in ONE place after any repository transfer or rename. Before #871 the
+    same owner/repo string was hard-coded at three separate sites; the transfer to the
+    `merickvaughn` org left `main` pinned to the pre-transfer `brownm09` value, and
+    because .github/workflows/staging.yml runs `terraform apply` (not plan) on every PR,
+    any PR branched from main actively re-applied the stale value to the live provider —
+    breaking GCP auth for every other branch, including the PR that carried the fix.
+  EOT
+  type        = string
+  default     = "merickvaughn/lifting-logbook"
+
+  validation {
+    # Charset is GitHub's actual owner/repo alphabet, deliberately narrower than a
+    # bare "one slash" shape check: the value is interpolated into a single-quoted
+    # CEL string in attribute_condition, so admitting a quote here would emit a
+    # malformed expression at the only consumer that reads this variable.
+    condition     = can(regex("^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$", var.github_repository))
+    error_message = "github_repository must be \"owner/repo\" using only letters, digits, dot, underscore or hyphen in each part."
+  }
+}
+
 variable "db_tier" {
   description = "Cloud SQL instance tier"
   type        = string


brownm09:
## Review: [fix] Drive WIF repo identity from a variable; bound google.subject to unblock CI

### Summary

**Needs revision** before merge — one blocking security finding. The core change is correct and empirically verified: `Terraform (staging)`, `Build & Push Images`, `Deploy web (staging)` and `Plan (production)` all pass on this branch, and the live provider was re-read after CI's own `terraform apply` and found unchanged, confirming code and live state agree rather than fighting each other. The variable consolidation is the right shape for the tripwire #871 raised.

The highest-severity finding is unrelated to the WIF logic: **the two committed backup snapshots publish a personal email address to a public repository.** Separately, the inline comment justifying the `google.subject` change makes an audit-granularity claim that does not hold as written — worth correcting even if the mapping itself stays.

---

### Blocking Findings

**[security]** `.claude/backups/wif-sa-bindings-snapshot-2026-07-22-132117.json`

The snapshot commits a verbatim GCP IAM policy that includes `"user:brownm2350@gmail.com"` bound to `roles/iam.serviceAccountTokenCreator`. This repository is **public** (`"visibility": "public"`).

That address does not otherwise appear in the repo's public history — `git log --format='%ae' origin/main` yields only `brownm09@live.com` and `36768444+brownm09@users.noreply.github.com`. So this PR newly publishes a personal email, correlated with an owner-level GCP role, to an indexable public repo. The existing `.claude/backups/` precedent (project-board and branch-protection snapshots) contains no `user:` principals, so this is a new disclosure class rather than an established one. The paired provider snapshot additionally exposes bare project numbers (`projects/910635705567`, `projects/508649171914`), which are identifiers rather than secrets but were not previously in the tree — the terraform interpolates them via `google_iam_workload_identity_pool.github.name` instead of hardcoding.

Note the backup itself is correct practice and required by the back-up-before-you-mutate rule — the issue is only *where* it landed and that it was committed unfiltered.

**Fix:** Filter the snapshots to the WIF-relevant bindings before committing — the `roles/iam.workloadIdentityUser` `principalSet://` member is the only entry the restore path actually needs; the `serviceAccountTokenCreator` user binding is unrelated to this mutation and can be dropped. If full-fidelity policy capture is wanted for restore safety, write these artifacts to `C:/Users/brown/.claude/scratch/` (or a private location) and reference them from the PR body instead of committing them. Since the branch is already pushed to a public repo, amend and force-push with `--force-with-lease` (after `git fetch origin`, per the CLAUDE.md bare-`--force` hazard) rather than adding a follow-up commit, so the address does not persist in history.

---

### Non-Blocking Findings

**[reliability]** `infra/terraform/main.tf:527-535` — audit claim in the new comment does not hold

The comment asserts: *"Actor granularity for audit is preserved by attribute.actor."* `attribute.actor` is a custom WIF attribute usable in IAM conditions and `principalSet://` members — it is not the identifier recorded as the calling principal. Cloud Audit Logs derive the federated principal from `google.subject`, which this change collapses to the constant `merickvaughn/lifting-logbook`. After this lands, every workflow run from every actor authenticates as one indistinguishable subject, so audit logs can no longer attribute a staging `terraform apply` to the person or PR that triggered it. Given this SA holds `roles/owner` on staging, that is a real reduction in forensic capability — and the comment currently reads as if it were not.

The 127-byte constraint does not force this. CEL supports concatenation in `attribute_mapping`, so a bounded composite preserves attribution with large headroom: `"google.subject" = "assertion.repository + '/' + assertion.actor"` is ~28 bytes plus a GitHub username (39 max), well under the limit and immune to the `gh-readonly-queue` ref blowup that broke `assertion.sub`.

**Suggestion:** Prefer the composite mapping. If the constant is kept deliberately (it is simpler, and the `principalSet` bindings gate access identically either way), correct the comment to state that per-actor audit attribution is knowingly traded away, rather than claiming it is preserved.

---

**[correctness]** `infra/terraform/main.tf:528` — production will carry permanent plan drift

Production's live provider still maps `google.subject: assertion.sub` (verified directly). `deploy.yml` only runs `terraform plan` against production, never `apply` — so nothing in CI will converge prod to the new mapping. From this merge onward, `Plan (production)` reports a pending change on every run that no automated path applies.

That matters beyond tidiness: the production plan job exists to surface blast radius before the approval gate, and a permanently non-empty plan trains readers to skim past it — the exact failure mode that let the stale `attribute_condition` sit unnoticed in the first place. Production is not affected by the underlying #869 bug (no `merge_group` event targets it), so this is drift-noise rather than a functional gap.

**Suggestion:** Either apply the mapping to prod deliberately out-of-band (same `gcloud ... update-oidc` shape used for staging, with a snapshot first) so plan returns to empty, or file a tracking issue and note the expected diff in the PR body so the next reader of a noisy prod plan knows it is accounted for.

---

**[maintainability]** `infra/terraform/variables.tf:50-53` — validation is looser than the interpolation context

`^[^/]+/[^/]+$` correctly enforces the one-slash shape, but `[^/]` admits characters that are illegal in GitHub owner and repo names — including the single quote, which is the CEL string delimiter at the value's only consumer:

```hcl
attribute_condition = "assertion.repository == '${var.github_repository}'"
```

A value containing a single quote yields a malformed CEL expression. The failure is loud (apply fails) and the variable is operator-supplied rather than attacker-controlled, so this is robustness rather than a vulnerability — but a validation block whose stated job is guarding this exact string may as well reject the one character that breaks its consumer.

**Suggestion:** Tighten to `^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$`, which matches GitHub's actual owner/repo charset and closes the quote case.

---

**[documentation]** `CLAUDE.md` — new backup class has no documented restore path

This PR establishes a second artifact class in `.claude/backups/` (WIF provider + SA bindings). CLAUDE.md documents a detailed Backup-and-restore procedure for GitHub Project single-selects, but nothing covers WIF, so the snapshots are restorable only by whoever remembers the `gcloud` invocation that produced them. The out-of-band correction this PR depended on is precisely the kind of step that will need repeating after the next transfer.

`.claude/backups/` also now holds seven files with no index README, though that predates this PR.

**Suggestion:** Add a short WIF subsection to CLAUDE.md's backup procedure naming the snapshot commands and the `gcloud iam workload-identity-pools providers update-oidc` / `add-iam-policy-binding` restore pair. This is also a reasonable candidate for a `docs/adr/` entry given the repo's practice of recording infra decisions (ADR-024 is comparable in scope) — the `google.subject` tradeoff above is the part worth capturing. Note the four literal ADR-011 criteria are `claude/`-path-scoped and do not trigger here, hence `adr_warrant=not-warranted` below; this is a repo-practice suggestion, not a gate failure.

---

### Questions for Author

**Question:** Should `github_repository` be set explicitly in `terraform.tfvars.staging` / `terraform.tfvars.prod`, rather than relying on the variable default?

**Context:** The variable defaults to `merickvaughn/lifting-logbook` and no tfvars file sets it ([`infra/terraform/variables.tf:48`](infra/terraform/variables.tf:48)). That is fine functionally — both workspaces want the same value. But `terraform.tfvars.staging` carries an explicit comment establishing the opposite convention for topology values: *"set explicitly so the staging/prod difference is visible at a glance."*

**Tradeoffs:** (a) Keep the default — one edit point after a transfer, which is exactly the property this PR is buying, but the trust boundary becomes invisible in the per-environment files an operator reads first. (b) Set it in both tfvars files — matches the stated convention and makes the boundary auditable per environment, at the cost of restoring two edit sites, a diluted version of the problem #871 filed. (c) Keep the default and add a one-line pointer comment in each tfvars noting where the value lives — preserves the single edit point while keeping it discoverable.

---

### Style Notes

The inline comment block on `attribute_mapping` is unusually long for this file, but the content earns it: the mapping's failure mode is latent and non-obvious, and the next reader has no other way to learn why the conventional `assertion.sub` was rejected. Worth keeping at this length — subject to the accuracy correction in the first non-blocking finding.

---

### Findings Disposition

Per ADR-028, each of the 1 blocking and 4 non-blocking findings must be fixed in this PR or filed as a tracked issue and linked before merge — none may be left as-is. Record each disposition (`fixed in <sha>` | `filed #N`) in a "Review findings disposition" section of the PR body; the `pre-merge-findings-gate` hook blocks the merge until it is present.

Recommended split: fix the blocking PII finding and the validation tightening in-PR (both are small and local); fix or explicitly accept the comment-accuracy correction in-PR since it is a two-line edit; file the production drift and the CLAUDE.md/ADR documentation findings as tracked issues, as both extend beyond this PR's scope.

**Merge ordering reminder, unchanged by this review:** this PR must land before #864 and #870 re-run, since #864's branch still maps `google.subject = assertion.sub` and would re-break the merge queue.

---

brownm09:
## Review (re-run): [fix] Drive WIF repo identity from a variable; bound google.subject

Re-review of `a5364f7`, refreshing the marker staled by the amend. The [first pass](https://github.com/merickvaughn/lifting-logbook/pull/873#issuecomment-5049589409) reviewed `c0371ef`.

### Summary

**Ready to merge.** All 5 findings from the first pass are resolved — 3 fixed in-PR, 2 filed and linked — with dispositions recorded in the PR body. Re-reviewing the amended delta surfaces no new findings.

### Delta reviewed (`c0371ef` → `a5364f7`)

Three changes, each traceable to a first-pass finding:

- **`.claude/backups/wif-sa-bindings-snapshot-*.json`** — filtered to the `roles/iam.workloadIdentityUser` bindings the restore path needs; `etag` dropped; a `redaction_note` records why. Verified against the pushed branch tip: zero `user:` principals and zero `gmail` matches remain. Resolves the blocking finding.
- **`infra/terraform/variables.tf`** — validation charset tightened to `^[A-Za-z0-9._-]+/[A-Za-z0-9._-]+$`, rejecting the CEL quote character. The default `merickvaughn/lifting-logbook` satisfies it — proven not by inspection but by `Terraform (staging)` passing on this commit, which evaluates the validation block during apply.
- **`infra/terraform/main.tf`** — the `attribute.actor` audit claim is corrected. The comment now separates ACCESS (genuinely independent of `google.subject`) from AUDIT (knowingly traded away), states plainly that `attribute.actor` does not recover attribution, names the bounded-composite alternative, and explains why it was not taken inside an outage fix. Tracked in #874.

No applied WIF value changed in the amend, so the configuration verified green on `c0371ef` is the configuration this commit applies — and CI independently re-confirms it on `a5364f7`.

### Verification

All 13 checks green on `a5364f7`, including `Staging Integration Tests` — the required check whose failure blocked every PR in the repo, and the direct evidence that #871/#869 are resolved rather than merely worked around. `Build & Push Images`, the job that failed with `unauthorized_client`, also passes.

### Findings Disposition

Carried forward from the first pass and unchanged — see the "Review findings disposition" table in the PR body. 3 fixed in `a5364f7` (PII redaction, validation charset, audit-claim correction); 2 filed and linked (#874 composite subject mapping, #875 production plan drift) plus #876 for the backup/restore documentation. None left as-is.

**Merge ordering:** land this before #864 and #870 re-run — #864's branch still maps `google.subject = assertion.sub` and would re-break the merge queue.

---

PR #879, opened 2026-08-01
[feat] Persist in-progress set inputs so a crash doesn't lose them

### LINKED ISSUE #878: [feat] Persist in-progress (unsubmitted) set inputs to survive a crash/reload
## Problem

Every *confirmed* set in the workout logger is already safe: tapping "Log" does an immediate `POST /lift-records`, and reopening the workout page re-fetches all logged records from the server. There is no "submit the whole workout" step to fail.

The one gap: whatever is typed into the set a user is actively filling in — but hasn't tapped Log/Save on yet — lives only in `WorkingSetRow`'s transient React state (`WorkoutLogger.tsx`). If the tab/browser dies before submit, that input is lost and must be retyped.

## Proposed fix

Add a small client-side draft cache (plain `localStorage`, no new dependency) that:
- Persists the in-progress set's weight/reps/notes as the user types (only for a set that is not yet logged and not read-only).
- Restores those values on next mount (covers both a crash/reload and the existing "switch to overview view and back" flow, which already unmounts every `WorkingSetRow`).
- Clears the draft once the set is successfully logged.

Deliberately scoped: no offline queue, no cross-tab sync, no TTL — this only protects the one actively-typed, not-yet-submitted set. Editing-in-progress on an already-logged set, the body-weight gate, and mobile (no logging screen exists there yet) are out of scope.

## Acceptance criteria

- [ ] New `workoutDraftStorage.ts` module: `buildDraftKey`, `readDraft`, `writeDraft`, `clearDraft` — SSR-guarded, fail-open on any storage error, runtime-validated on read (no unsafe cast).
- [ ] `WorkingSetRow` reads a draft on mount (when unlogged + not read-only), writes on each keystroke, clears on successful "Log".
- [ ] Stray/stale drafts are never read for an already-logged or read-only set.
- [ ] Unit tests for the storage module (round-trip, corrupted JSON, quota/storage errors) and integration tests in `WorkoutLogger.test.tsx` (persist while typing, restore after remount, clear after log, ignore stale draft for a logged set).
- [ ] `npm run typecheck`, `npm run lint`, `npm test -w @lifting-logbook/web` all pass.

brownm09:
## Summary

Every already-logged set is already crash-safe: "Log"/"Save" writes immediately via its own `POST`/`PATCH` to `/lift-records`, and there's no batch "submit the whole workout" step to lose. The one real gap: whatever is typed into the set a user is actively filling in — but hasn't submitted yet — lived only in transient React state (`WorkingSetRow`) and was lost on a crash/reload.

This adds a small, scoped localStorage draft cache for exactly that gap:

- New `workoutDraftStorage.ts`: `buildDraftKey`/`readDraft`/`writeDraft`/`clearDraft`, SSR-guarded, fails open silently on any storage error (never blocks typing/logging), with runtime shape validation on read (no unsafe cast).
- `WorkingSetRow` reads a draft on mount and writes on every keystroke — **only** when a set is neither logged nor read-only (`shouldDraft = !loggedRecord && !isReadOnly`). Writes are imperative (in the `onChange` handlers), not a reactive `useEffect`, so nothing is ever written on mere mount/navigation — only on genuine typing.
- The draft clears on a successful "Log". A consolidated `handleFinishWorkout` (replacing two duplicated `router.push` call sites) also sweeps any stray per-set drafts for the whole workout as a backstop for a set logged from a different tab/device than the one holding the draft.

Deliberately out of scope (see issue for full rationale): editing-in-progress on an already-logged set, the body-weight gate, mobile (no logging screen exists there yet), cross-tab sync, offline queue, and draft TTL.

## Note on the `isReadOnly` half of the gate

While implementing, I found the `!isReadOnly` half of `shouldDraft` has no way to fail its own dedicated unit test: `isReadOnly` only ever matters when a row has no `loggedRecord`, and in that combination the component always takes the read-only early-return branch regardless — so no `useState`/render ever surfaces the difference. It's still correct, cheap defense-in-depth against relying on `page.tsx`'s "isReadOnly implies every set is logged" invariant holding forever (rather than WorkingSetRow enforcing it itself), and it's covered by a code comment explaining exactly this — but I did not add a test claiming to catch it, since that would be a false-green test under this repo's own Test Integrity Policy (a test that can't fail under the "before" state isn't a regression test). The `!loggedRecord` half **is** independently meaningful and **is** covered (see "ignores a stale leftover draft when editing an already-logged set").

## Testing

- `npm run typecheck` — pass
- `npm run lint` — pass (confirms `no-uncovered-error-fallback` doesn't flag the new module's plain `try{}catch{return null}` pattern, which matches neither of the rule's two detected shapes)
- `npm test` (full monorepo) — Tests: 806 passed, 0 skipped, 0 failed (in ~30s per workspace; web workspace alone: 295 passed)
- **Manual verification in a real browser** (dev server + the existing Playwright e2e mock API, DEV_AUTH_TOKEN bypass — see `apps/web/e2e/mock-api.mjs` / `playwright.config.ts` for the same setup): typed weight/reps/notes into an unlogged set, confirmed the `workout-draft:...` key appeared in localStorage with the exact typed values, hard-reloaded the page, confirmed the same values were restored into that set's inputs (and only that set — sibling sets stayed at plan defaults), clicked "Log", and confirmed both that the set now shows as logged with those exact values and that the localStorage key was removed.

New/changed test coverage:
- `workoutDraftStorage.test.ts` (new): key building, round-trip, clear, corrupted JSON → null, wrong-shape JSON → null, `setItem`/`getItem`/`removeItem` throwing → silent no-op/null (satisfies `docs/standards/error-fallback-test-coverage.md` via paired failure-path tests).
- `WorkoutLogger.test.tsx` additions: persist-while-typing, restore-after-remount, clear-after-log, ignore-stale-draft-when-editing-an-already-logged-set.

## Acceptance criteria (from #878)

- [x] New `workoutDraftStorage.ts` module: `buildDraftKey`, `readDraft`, `writeDraft`, `clearDraft` — SSR-guarded, fail-open on any storage error, runtime-validated on read (no unsafe cast).
- [x] `WorkingSetRow` reads a draft on mount (when unlogged + not read-only), writes on each keystroke, clears on successful "Log".
- [x] Stray/stale drafts are never read for an already-logged or read-only set.
- [x] Unit tests for the storage module and integration tests in `WorkoutLogger.test.tsx`.
- [x] `npm run typecheck`, `npm run lint`, `npm test -w @lifting-logbook/web` all pass.

Closes #878

Per [/review](https://github.com/merickvaughn/lifting-logbook/pull/879#issuecomment-5154160113) (0 blocking, 2 non-blocking):

- **[documentation]** PR body test-count typo (299 → 295 for the web workspace) — fixed in this PR body edit.
- **[maintainability]** `weightInput` initializer structurally asymmetric vs. `repsInput`/`notesInput` — fixed in `d793daf`.

ADR-warrant: not-warranted (no `claude/`-scoped rule/hook/skill/settings touched). Doc-reconciliation: not-applicable (repo has no Documentation Maintenance table).

### DIFF
--- a/apps/web/app/(authed)/cycle/[cycleNum]/workout/[workoutNum]/WorkoutLogger.test.tsx
+++ b/apps/web/app/(authed)/cycle/[cycleNum]/workout/[workoutNum]/WorkoutLogger.test.tsx
@@ -4,6 +4,7 @@ import type { LiftRecordResponse } from '@lifting-logbook/types';
 import WorkoutLogger from './WorkoutLogger';
 import type { LiftData, WorkoutLoggerProps } from './types';
 import { createLiftRecord, recordBodyWeight } from '@/lib/client-api';
+import { buildDraftKey } from './workoutDraftStorage';
 
 // jest.mock is hoisted above the imports, so `createLiftRecord`/`recordBodyWeight`
 // resolve to the jest.fn() mocks below.
@@ -61,6 +62,9 @@ const LOGGED: LiftRecordResponse = {
 
 beforeEach(() => {
   jest.clearAllMocks();
+  // Several tests below share the default program/cycle/workout/lift/set identity
+  // from makeProps()/makeLift() and would otherwise leak drafts across tests.
+  window.localStorage.clear();
 });
 
 describe('WorkoutLogger — weight-unit display preference', () => {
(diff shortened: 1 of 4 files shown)


brownm09:
## Review: [feat] Persist in-progress set inputs so a crash doesn't lose them

### Summary

Ready to merge with minor changes. This is a well-scoped, correctly-gated feature (302-line diff, appropriately sized for a single pass) — no blocking issues found. The two non-blocking findings below are both small: a test-count typo in the PR body's Testing section, and a minor structural asymmetry between how `weightInput` vs. `repsInput`/`notesInput` short-circuit on an existing `loggedRecord`.

---

### Non-Blocking Findings

**[documentation]** PR description — Testing section
The PR body states "web workspace alone: 299 passed," but the actual `npm test -w @lifting-logbook/web` output (295 passed, 295 total) plus the API workspace's 511 gives the stated 806 total (295 + 511 = 806, matching), not 299 + 511 = 810. The 806 total is internally consistent; only the "299" sub-figure is off by 4.
**Suggestion:** Correct "299" to "295" in the PR body before merging, so the disposition note doesn't propagate a wrong number into the merge record.

**[maintainability]** `apps/web/app/(authed)/cycle/[cycleNum]/workout/[workoutNum]/WorkoutLogger.tsx:212-235`
`weightInput`'s lazy initializer short-circuits on `if (loggedRecord)` *before* even evaluating `shouldDraft`, while `repsInput` and `notesInput` reach the same end state only indirectly — through `shouldDraft` evaluating to `false` when `loggedRecord` is truthy. All three are functionally equivalent today, but the asymmetry means a future change to the gating logic (e.g., adding a field, or changing what `shouldDraft` depends on) could be applied to two of the three initializers and silently miss the third, since they no longer read as three parallel copies of the same pattern.
**Suggestion:** Either add the same explicit `if (loggedRecord) return ...` short-circuit to `repsInput`/`notesInput` for symmetry, or remove it from `weightInput` and rely on `shouldDraft` alone (as the other two already do) — whichever you pick, make all three follow one shape.

---

### Findings Disposition

Both findings above should be addressed (fixed in this PR) before merge per this repo's all-findings-merge-gate — neither warrants filing a separate issue given their size.

---

brownm09:
## Review: [feat] Persist in-progress set inputs so a crash doesn't lose them (re-review after `d793daf`)

### Summary

Ready to merge. This re-review runs against the latest commit (`d793daf`), which addresses both findings from the [prior review](https://github.com/merickvaughn/lifting-logbook/pull/879#issuecomment-5154160113). Confirmed resolved, no new issues introduced by the fix, no new findings on a fresh pass over the full current diff (299 lines, still comfortably single-pass sized).

- **PR body test count** — now correctly reads "web workspace alone: 295 passed", consistent with the stated 806 total.
- **`weightInput` initializer asymmetry** — now structurally identical in shape to `repsInput`/`notesInput` (draft-check first when `shouldDraft`, otherwise a single fallback expression). Traced all three gating branches (`loggedRecord` truthy; falsy + draftable; falsy + read-only) against both the pre-fix and post-fix code — behavior is unchanged in every branch, this was a pure structural cleanup.

No blocking or non-blocking findings on this pass.

---

PR #886, opened 2026-08-09
Phase 2B: TM-history seeding script (seed-tm-history.ts)

### LINKED ISSUE #885: Phase 2B: TM-history seeding script (write path for gas-lifting-logbook#23)
## Summary

Companion issue to [gas-lifting-logbook#23](https://github.com/brownm09/gas-lifting-logbook/issues/23) (Phase 2B: historical training-max-progression seeding), itself a sub-issue of [gas-lifting-logbook#4](https://github.com/brownm09/gas-lifting-logbook/issues/4) (GAS → GCP data transition).

**The write path.** gas-lifting-logbook derives a per-cycle training-max history series from the legacy Google Sheet (hidden per-cycle workout tabs' `TM` column, falling back to `LIFT_RECORDS` set-1 weight) and exports it to a preview tab, `GCP_EXPORT_TMHistory`, with a `Will Seed` column marking which rows should actually be written (`transition !== "hold"`). See [gas-lifting-logbook#24](https://github.com/brownm09/gas-lifting-logbook/pull/24) (the export/validation side, open) for the full derivation and pre-flight validation.

**What's needed here:** a one-time seeding script in this repo that reads that exported CSV and writes it into `training_max_history` via `ITrainingMaxHistoryRepository.appendHistoryEntries`, run as the migrator/owner role through break-glass DB access — **not** via any HTTP endpoint (the `training-max-history` controller only exposes `GET`/`PATCH`, by design — see [ADR-017](https://github.com/merickvaughn/lifting-logbook/blob/main/docs/adr/ADR-017-training-max-history-table.md)).

## Design (already scoped — see gas-lifting-logbook's runbook)

The full field mapping and CLI contract are pre-designed in [`docs/phase2b-tm-history-seeding-runbook.md`](https://github.com/brownm09/gas-lifting-logbook/blob/feat/phase-2b-tm-history-seed/docs/phase2b-tm-history-seeding-runbook.md) (on gas-lifting-logbook#24's branch as of filing; will be at `main` once that PR merges). Summary:

- **Location:** `apps/api/scripts/seed-tm-history.ts`, following `apps/api/prisma/seed.ts`'s invocation convention (`ts-node`), but going through `PrismaTrainingMaxHistoryRepository`/`ITrainingMaxHistoryRepository` rather than raw Prisma calls — issue #23's plan explicitly specifies the port interface.
- **CLI:** `--program <uuid>`, `--user-id <clerk-user-id>`, `--input <path-to-csv>`, `--dry-run`, `--force`, `--rollback`.
- **Field mapping:** `Lift` → resolved through the real `packages/core/src/catalog/slotMaps.ts` (not a mirror — the script has direct access); `TM` → `weight`; `reps` hardcoded `1` (matches `CycleGenerationService.buildHistoryEntries`'s convention — a TM value, not a logged set); `Cycle Date` → `date`; `Is PR`/`Goal Met` → `isPR`/`goalMet` (taken as-computed from the CSV, **not** left `false` the way the going-forward writer defers to manual `PATCH` — Phase 2B has complete historical context to compute both correctly once); `source` hardcoded `"program"` (RPT has no test-week concept; the DB CHECK constraint only accepts `'test'|'program'`).
- **Idempotency / safety:** `training_max_history` has **no unique constraint** — the normal seed mode is delete-all-then-append for the target program (back up existing history to a local JSON file first, refuse to proceed without `--force` if the program already has history rows, per ADR-079's back-up-before-mutate convention). `--rollback` calls `deleteAllHistory(program)` alone (no re-seed).
- **Connection:** reuses `scripts/migrate-prod-db.sh`'s break-glass *connection pattern* (Cloud SQL Auth Proxy + temporary authorized network + Secret Manager fetch for the **migrator**-role secret, `lifting-logbook-<env>-migrator-database-url` — distinct from the runtime role's secret) — **not** that script itself, which is a DDL-only `prisma migrate deploy` runner and does not accept an arbitrary script to execute.

## Testing

Per this repo's `## Testing` / ADR-013: the Prisma adapter itself must not be mocked — test it for real via Testcontainers, mirroring `programs.db.e2e.spec.ts`'s two-connection-string pattern (owner role seeds/asserts fixtures directly, the app-under-test uses the restricted role). The script's own logic (CSV parsing, `Will Seed` filtering, field mapping, the `--force`/backup safety gate, `--dry-run` output) should get direct unit tests; the actual `appendHistoryEntries`/`deleteAllHistory` calls should get an e2e-style test against a real (Testcontainers) Postgres instance, not a mocked repository.

## Out of scope

- Any change to `scripts/migrate-prod-db.sh` itself.
- The optional e1RM/PR-trend bonus artifact (`estimateOneRepMax`) — not part of this issue's scope; may be folded into a future preview enhancement.
- Actually running the script against staging/prod — that's a human-executed step per the gas-lifting-logbook runbook, not something a PR here needs to complete.

## Cross-references

- Parent: [gas-lifting-logbook#4](https://github.com/brownm09/gas-lifting-logbook/issues/4), [gas-lifting-logbook#23](https://github.com/brownm09/gas-lifting-logbook/issues/23) — **#23 stays open until both this issue's PR and gas-lifting-logbook#24 have merged.**
- Export/validation side: [gas-lifting-logbook#24](https://github.com/brownm09/gas-lifting-logbook/pull/24)
- Schema decision: [ADR-017](https://github.com/merickvaughn/lifting-logbook/blob/main/docs/adr/ADR-017-training-max-history-table.md)

brownm09:
## Summary

Closes #885 — the write path for gas-lifting-logbook's Phase 2B (historical training-max-progression seeding), companion to [gas-lifting-logbook#23](https://github.com/brownm09/gas-lifting-logbook/issues/23) / [#4](https://github.com/brownm09/gas-lifting-logbook/issues/4) (the GAS → GCP data transition). The export/validation side is already merged in [gas-lifting-logbook#24](https://github.com/brownm09/gas-lifting-logbook/pull/24); [`docs/phase2b-tm-history-seeding-runbook.md`](https://github.com/brownm09/gas-lifting-logbook/blob/main/docs/phase2b-tm-history-seeding-runbook.md) there is the full human-executed procedure this script is one step of.

## What's in this PR

**`apps/api/scripts/seed-tm-history.ts`** — a one-time script that reads gas-lifting-logbook's derived, human-approved `GCP_EXPORT_TMHistory.csv` and writes it into `training_max_history` via `ITrainingMaxHistoryRepository.appendHistoryEntries`/`deleteAllHistory`, run as the migrator/owner role (bypasses RLS) through the break-glass connection the runbook establishes. CLI: `--program`, `--user-id`, `--input`, `--dry-run`, `--force`, `--rollback`, `--restore <path>`, `--help` — the contract the runbook documents (`--restore` and `--help` added during review — see disposition below).

Pure functions (arg parsing, CSV parsing, row→entry mapping, backup-path derivation, cross-row invariant checks) are separated from I/O orchestration (ownership check, backup, seed/rollback/restore), so the former are directly unit-testable and the latter is tested against real Postgres.

**Program-ownership verification** (`verifyProgramOwnership`) — a security requirement surfaced during gas-lifting-logbook PR #24's `/review`: since this script connects as the RLS-bypassing migrator/owner role, `--program` is the *only* tenant-isolation control left on both the seed write and `--rollback`'s delete. Runs before any write/delete; throws on a program belonging to a different user, not existing at all, or a cross-user data anomaly this connection is uniquely positioned to detect.

**Cross-row invariant checks** (`assertSingleProgram`, `assertNoDuplicateSeedKeys`) — re-apply gas-lifting-logbook's own pre-flight validator checks (multi-program mixing, duplicate seed keys) against the CSV this script actually reads, since that validator runs against the live Sheet tab, not this downloaded file (added during review — see disposition below).

**Re-run semantics** match the runbook's "Re-run semantics" section exactly: delete-all-then-append (never plain append — `training_max_history` has no unique constraint), gated by `--force` whenever pre-existing history is found. `--dry-run` never enforces the `--force` gate itself — it always succeeds, runs the same validation a real run would, and reports whether `--force` *would* be needed, so an operator learns that before deciding, rather than the dry run itself failing.

**Backup before mutate, and restore** (ADR-079 pattern, per the runbook) — `backupExistingHistory` writes existing history to a local, provenance-checked JSON envelope before any delete (seeding over existing history, rolling back, or restoring), fsynced and permission-restricted; `--restore <path>` is the other half — refuses to restore across a different `--program`/`--user-id` than the one that captured the backup.

**Field mapping** matches the runbook's table exactly: `reps` always `1` (`CycleGenerationService`'s convention — a TM value, not a logged set); `source` always `'program'` (RPT has no test-week concept; the DB CHECK constraint on `training_max_history.source` only accepts `'test'|'program'` regardless); a blank `Goal Met` coerces to `false` (`buildTMHistory` genuinely emits `null` there for some will-seed rows; the target column is a non-null `Boolean`). Lift resolution uses the real `DEFAULT_SLOT_MAP` directly — no drift-prone hand-copied mirror, unlike gas-lifting-logbook's own pre-flight validator, which necessarily has one since it can't import this repo's code.

Named the resolver `resolveLiftId`, deliberately distinct from the catalog's own `resolveLift` (`packages/core/src/catalog/resolve.ts`) — that function throws on an unmapped slot name and returns a full `Lift` catalog object; this script needs passthrough-on-unmapped semantics and a plain id string (matching the training-max import path's existing behavior), so reusing it directly would have been wrong, not just redundant.

Only `Will Seed = true` rows are ever written — the script trusts that column directly and never re-derives gas-lifting-logbook's transition classification itself.

**`apps/api/package.json`** — `csv-parse` as an explicit `devDependencies` entry (already present transitively via `packages/core`; made explicit rather than relying on workspace hoisting — a script-only tool, not a prod API dependency) and a `seed:tm-history` npm script. `package-lock.json` regenerated per the lockfile policy.

## Testing

```
Test Suites: 42 passed, 42 total (2 new: seed-tm-history.spec.ts, seed-tm-history.db.e2e.spec.ts)
Tests:       599 passed, 0 skipped, 0 failed (88 new in the two script test files)
```

- `apps/api/scripts/seed-tm-history.spec.ts` — unit tests for every pure function: `parseArgs`, `resolveLiftId`, `parseTMHistoryCsv` (including BOM/header-trim), `filterWillSeedRows`, `assertSingleProgram`, `assertNoDuplicateSeedKeys`, `mapRowToEntry`, `backupFilePath`, `backupExistingHistory`/`readBackupFile` (envelope format, provenance checks, per-entry validation — moved here from e2e once the function became DB-free during review), `formatResult`. All pure/fs-only — no DB or repository mocking needed (per ADR-013).
- `apps/api/scripts/seed-tm-history.db.e2e.spec.ts` — tests against real Postgres via Testcontainers (per ADR-013: never mock the Prisma adapter itself), connected as the owner/migrator role exactly as the script does in production. Covers `verifyProgramOwnership` (owned / wrong-user / nonexistent / cross-user anomaly), and `runSeed`'s full matrix: fresh seed with holds correctly excluded, multi-program CSV refusal, refuse-without-`--force` on unexpected existing history, the now-mode-aware `--force` message (pinned against the runbook's exact documented `--rollback` invocation), `--force` reseed-over-existing (backup captures the *old* data, DB ends with the *new* data), `--dry-run` (DB untouched, validates like a real run, probes backup-dir writability), transaction-rollback-on-simulated-failure, `--rollback`, `--rollback --dry-run`, `--restore` (round-trip, cross-program refusal, empty-backup refusal).
- `npm run build` (`nest build`, the real typecheck gate for `apps/api`) clean, after building `packages/core` and `packages/types` first. `npx eslint` clean on all three script files (`npm run lint` only covers `src/`, matching the existing precedent for `apps/api/scripts/check-globalsetup-failure-paths.js` — the coverage gap itself is filed, see disposition below).
- No suppressions or test-integrity violations introduced (checked via the standard greps, and via `/review`'s dedicated gate — see disposition). Lockfile verified in sync (`npm install --package-lock-only --ignore-scripts` + `git diff --exit-code package-lock.json` exits clean).

Full review: [PR comment](https://github.com/merickvaughn/lifting-logbook/pull/886#issuecomment-5230231644) (`blocking=7 non_blocking=12`). All 19 findings resolved — 17 fixed in-PR (commits `f2e5da8`, `456064a`, `695af47`), 2 filed as genuinely out-of-scope follow-ups. Summary (full detail in the review comment):

| Finding | Disposition |
|---|---|
| Multi-program CSV not re-validated | Fixed `456064a` |
| Duplicate seed key not re-validated | Fixed `456064a` |
| `--restore` missing zero-entries guard | Fixed `456064a` |
| Backup provenance / restore entry validation | Fixed `456064a` |
| Backup dir/file permissions, symlink guard | Fixed `456064a` |
| `--force` message not mode-aware; runbook `--rollback` broken | Fixed `456064a` + [gas-lifting-logbook#31](https://github.com/brownm09/gas-lifting-logbook/pull/31) |
| `csv-parse` lockfile section drift | Fixed `456064a` |
| No transaction timeout budget | Fixed `456064a` |
| Bypasses `no-direct-prisma-transaction` rule (path scope) | Fixed `456064a` (uses `runInteractive`) |
| `apps/api/scripts/` no typecheck/lint coverage | Filed [#887](https://github.com/merickvaughn/lifting-logbook/issues/887) |
| CSV BOM / header trim | Fixed `456064a` |
| No row context in errors; fails on first bad row | Partially fixed `456064a` (row context; full aggregation deferred, noted) |
| Backup fsync + tmpdir purge warning | Fixed `456064a`, `695af47` |
| `--dry-run` skips backup-dir writability probe | Fixed `456064a` |
| Triple-fetch of existing history | Fixed `456064a` |
| `backupDirOverride` test backdoor | Fixed `456064a` |
| `--restore` path not existence-checked | Fixed `456064a` |
| No `--help`/`-h` | Fixed `456064a` |
| Surviving `as string` cast in `parseArgs` | Fixed `456064a` |

Plus a cross-repo question with no code fix here: whether Google Sheets renders a boolean-looking CSV cell lowercase or uppercase (this script defends both ways regardless) — filed [gas-lifting-logbook#29](https://github.com/brownm09/gas-lifting-logbook/issues/29), needs live Sheets access this session doesn't have.

## Follow-up

- The actual production seed run is a human-executed procedure per gas-lifting-logbook's runbook — this PR does not (and cannot) run it.
- The runbook's `packages/core`/`packages/types` build-prerequisite addendum merged: [gas-lifting-logbook#28](https://github.com/brownm09/gas-lifting-logbook/pull/28).
- The runbook's rollback-`--force`/`--restore`-documentation fix is open: [gas-lifting-logbook#31](https://github.com/brownm09/gas-lifting-logbook/pull/31).
- gas-lifting-logbook#23 stays open until this PR merges (both halves of Phase 2B will then be complete).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,12 @@ next-env.d.ts
 scripts/cloud-sql-proxy
 scripts/cloud-sql-proxy.exe
 
+# Phase 2B seed script backups (apps/api/scripts/seed-tm-history.ts) — default
+# outside the repo (os.tmpdir()), but this is a backstop in case --backup-dir
+# ever points inside the tree. Contains a full dump of one user's training-max
+# history keyed to their Clerk id — never committed.
+tm-history-backup-*.json
+
 # Playwright staging auth state — contains real session tokens, never commit
 apps/web/playwright/.auth/
 
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -22,7 +22,6 @@
     "@nestjs/common": "^10.0.0",
     "@nestjs/core": "^10.0.0",
     "@nestjs/platform-fastify": "^10.0.0",
-    "csv-parse": "^6.1.0",
     "@opentelemetry/api": "^1.9.1",
     "@opentelemetry/auto-instrumentations-node": "^0.75.0",
     "@opentelemetry/exporter-logs-otlp-http": "^0.217.0",
@@ -61,6 +60,7 @@
     "@opentelemetry/sdk-trace-base": "^2.7.1",
     "@testcontainers/postgresql": "^12.0.1",
     "@types/pg": "^8.11.0",
+    "csv-parse": "^6.1.0",
     "prisma": "^5.22.0",
     "ts-node": "^10.9.0",
     "tsconfig-paths": "^4.2.0"
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -10,7 +10,8 @@
     "test": "jest",
     "test:db": "jest --testPathPatterns=db\\.e2e",
     "db:generate": "prisma generate",
-    "postinstall": "prisma generate"
+    "postinstall": "prisma generate",
+    "seed:tm-history": "ts-node scripts/seed-tm-history.ts"
   },
   "dependencies": {
     "@anthropic-ai/sdk": "^0.92.0",
@@ -21,6 +22,7 @@
     "@nestjs/common": "^10.0.0",
     "@nestjs/core": "^10.0.0",
     "@nestjs/platform-fastify": "^10.0.0",
+    "csv-parse": "^6.1.0",
     "@opentelemetry/api": "^1.9.1",
     "@opentelemetry/auto-instrumentations-node": "^0.75.0",
     "@opentelemetry/exporter-logs-otlp-http": "^0.217.0",
--- a/apps/api/scripts/seed-tm-history.db.e2e.spec.ts
+++ b/apps/api/scripts/seed-tm-history.db.e2e.spec.ts
@@ -17,20 +17,42 @@ import { PrismaClient } from '@prisma/client';
 import {
   backupExistingHistory,
   ProgramOwnershipError,
+  REQUIRED_COLUMNS,
   runSeed,
   SeedArgs,
   verifyProgramOwnership,
 } from './seed-tm-history';
+import { PrismaTrainingMaxHistoryRepository } from '../src/adapters/prisma/training-max-history.repository';
 
 const OWNER_DATABASE_URL = process.env.LIFTING_TC_OWNER_DATABASE_URL;
 // Skip only when globalSetup did not provision a DB (e.g. Docker unavailable
 // and not running in CI) — matches programs.db.e2e.spec.ts's pattern.
 const describeOrSkip = OWNER_DATABASE_URL ? describe : describe.skip;
 
+// Applied to every hook that actually touches Postgres (beforeEach/afterAll)
+// — matching programs.db.e2e.spec.ts's own placement (its comment there
+// explains #567: these hooks intermittently trip Jest's 5s default under a
+// full-suite Windows `turbo run test`, since apps/api/jest.config.js does
+// not extend the win32-capped base config). Deliberately NOT on beforeAll
+// here, which only constructs a PrismaClient (Prisma connects lazily) and
+// does no DB work of its own.
 const DB_E2E_HOOK_TIMEOUT_MS = 30_000;
 
 const USER_A = 'seed-tm-history-e2e-user-a';
 const USER_B = 'seed-tm-history-e2e-user-b';
+
+/** Fails loudly and narrows the type — the one idiom this file uses for
+ * "an assertion already proved this is non-null, now tell the compiler."
+ * Replaces a mix of `!` and `as` that either "solved" the same problem
+ * differently in different spots, or, in `expect(x).not.toBeNull()`'s
+ * case, doesn't actually narrow at all (Jest's matchers have no type-level
+ * effect) — so on an unexpected null both `!` and `as` produce a confusing
+ * failure the first time they're actually wrong, whereas this throws a
+ * named error immediately. */
+function assertNotNull<T>(value: T | null, what: string): T {
+  if (value === null) throw new Error(`expected ${what} to be non-null`);
+  return value;
+}
 
 async function cleanTestUsers(prisma: PrismaClient): Promise<void> {
   await prisma.trainingMaxHistory.deleteMany({
@@ -41,22 +63,18 @@ async function cleanTestUsers(prisma: PrismaClient): Promise<void> {
   });
 }
 
+const tmpDirs: string[] = [];
 function tmpBackupDir(): string {
-  return fs.mkdtempSync(path.join(os.tmpdir(), 'seed-tm-history-e2e-'));
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'seed-tm-history-e2e-'));
+  tmpDirs.push(dir);
+  return dir;
 }
 
 function csvOf(rows: Record<string, string>[]): string {
-  const headers = [
-    'Program',
-    'Lift',
-    'Cycle Date',
-    'TM',
-    'Goal Met',
-    'Is PR',
-    'Will Seed',
-  ];
-  const lines = [headers.join(',')];
-  for (const row of rows) lines.push(headers.map((h) => row[h] ?? '').join(','));
+  const lines = [REQUIRED_COLUMNS.join(',')];
+  for (const row of rows) {
+    lines.push(REQUIRED_COLUMNS.map((h) => row[h] ?? '').join(','));
+  }
   return lines.join('\n');
 }
 
@@ -79,17 +97,23 @@ describeOrSkip('seed-tm-history (e2e, real Postgres, owner/migrator connection)'
 
   beforeAll(() => {
     prisma = new PrismaClient({ datasources: { db: { url: OWNER_DATABASE_URL } } });
-  }, DB_E2E_HOOK_TIMEOUT_MS);
+  });
 
   beforeEach(async () => {
     await cleanTestUsers(prisma);
     backupDir = tmpBackupDir();
+  }, DB_E2E_HOOK_TIMEOUT_MS);
+
+  afterEach(() => {
+    for (const dir of tmpDirs.splice(0)) {
+      fs.rmSync(dir, { recursive: true, force: true });
+    }
   });
 
   afterAll(async () => {
     await cleanTestUsers(prisma);
     await prisma.$disconnect();
-  });
+  }, DB_E2E_HOOK_TIMEOUT_MS);
 
   describe('verifyProgramOwnership', () => {
     it('resolves without throwing when the program belongs to the given user', async () => {
@@ -115,6 +139,23 @@ describeOrSkip('seed-tm-history (e2e, real Postgres, owner/migrator connection)'
         verifyProgramOwnership(prisma, '00000000-0000-0000-0000-000000000000', USER_A),
       ).rejects.toThrow(/does not exist/);
     });
+
+    it('logs the connected role/database identity for audit purposes', async () => {
+      const program = await prisma.customProgram.create({
+        data: { userId: USER_A, name: 'Identity Log' },
+      });
+      const logSpy = jest.spyOn(console, 'log').mockImplementation(() => {});
+      try {
+        await verifyProgramOwnership(prisma, program.id, USER_A);
+        expect(
+          logSpy.mock.calls.some((call) =>
+            String(call[0]).startsWith('Connected as '),
+          ),
+        ).toBe(true);
+      } finally {
+        logSpy.mockRestore();
+      }
+    });
   });
 
   describe('backupExistingHistory', () => {
@@ -122,15 +163,12 @@ describeOrSkip('seed-tm-history (e2e, real Postgres, owner/migrator connection)'
       const program = await prisma.customProgram.create({
         data: { userId: USER_A, name: 'Empty History' },
       });
-      const { PrismaTrainingMaxHistoryRepository } = await import(
-        '../src/adapters/prisma/training-max-history.repository'
-      );
       const repo = new PrismaTrainingMaxHistoryRepository(prisma, USER_A);
       const result = await backupExistingHistory(repo, program.id, backupDir);
       expect(result).toBeNull();
     });
 
-    it('writes existing history to a JSON file and returns its path + count', async () => {
+    it('writes existing history to a JSON file, verifies the write, and returns its path + count', async () => {
       const program = await prisma.customProgram.create({
         data: { userId: USER_A, name: 'Has History' },
       });
@@ -147,14 +185,13 @@ describeOrSkip('seed-tm-history (e2e, real Postgres, owner/migrator connection)'
           goalMet: true,
         },
       });
-      const { PrismaTrainingMaxHistoryRepository } = await import(
-        '../src/adapters/prisma/training-max-history.repository'
-      );
       const repo = new PrismaTrainingMaxHistoryRepository(prisma, USER_A);
-      const result = await backupExistingHistory(repo, program.id, backupDir);
-      expect(result).not.toBeNull();
-      expect(result?.count).toBe(1);
-      const written = JSON.parse(fs.readFileSync(result!.path, 'utf8'));
+      const result = assertNotNull(
+        await backupExistingHistory(repo, program.id, backupDir),
+        'backupExistingHistory result',
+      );
+      expect(result.count).toBe(1);
+      const written = JSON.parse(fs.readFileSync(result.path, 'utf8'));
       expect(written).toHaveLength(1);
       expect(written[0]).toMatchObject({ lift: 'back-squat', weight: 300 });
     });
@@ -163,12 +200,13 @@ describeOrSkip('seed-tm-history (e2e, real Postgres, owner/migrator connection)'
   describe('runSeed', () => {
     function baseArgs(program: string, input: string): SeedArgs {
       return {
+        mode: 'seed',
         program,
         userId: USER_A,
         input,
         dryRun: false,
         force: false,
-        rollback: false,
+        backupDir: null,
       };
     }
 
@@ -217,6 +255,39 @@ describeOrSkip('seed-tm-history (e2e, real Postgres, owner/migrator connection)'
       expect(rows).toHaveLength(2);
       expect(rows.map((r) => r.lift).sort()).toEqual(['back-squat', 'bench-press']);
       expect(rows.every((r) => r.reps === 1 && r.source === 'program')).toBe(true);
+    });
+
+    it('refuses a seed that would write zero rows, rather than silently wiping existing history', async () => {
+      const program = await prisma.customProgram.create({
+        data: { userId: USER_A, name: 'All Holds' },
+      });
+      await prisma.trainingMaxHistory.create({
+        data: {
+          userId: USER_A,
+          program: program.id,
+          lift: 'back-squat',
+          weight: 250,
+          reps: 1,
+          date: new Date('2025-01-01'),
+          isPR: false,
+          source: 'program',
+          goalMet: false,
+        },
+      });
+      const csvPath = path.join(backupDir, 'export.csv');
+      // Every row is a hold (Will Seed=false) — a plausible wrong-file or
+      // wrong-export-state mistake, not itself invalid CSV.
+      fs.writeFileSync(csvPath, csvOf([willSeedRow({ 'Will Seed': 'false' })]));
+
+      await expect(
+        runSeed(prisma, { ...baseArgs(program.id, csvPath), force: true }, backupDir),
+      ).rejects.toThrow(/zero rows to seed/);
+
+      const rows = await prisma.trainingMaxHistory.findMany({
+        where: { program: program.id },
+      });
+      expect(rows).toHaveLength(1);
+      expect(rows[0]?.weight).toBe(250); // untouched — refused before any delete
     });
 
     it('refuses without --force when the program already has history, and writes nothing', async () => {
@@ -279,8 +350,8 @@ describeOrSkip('seed-tm-history (e2e, real Postgres, owner/migrator connection)'
       expect(result.mode).toBe('seeded');
       expect(result.existingCount).toBe(1);
       expect(result.writtenCount).toBe(1);
-      expect(result.backupPath).not.toBeNull();
-      const backedUp = JSON.parse(fs.readFileSync(result.backupPath as string, 'utf8'));
+      const backupPath = assertNotNull(result.backupPath, 'result.backupPath');
+      const backedUp = JSON.parse(fs.readFileSync(backupPath, 'utf8'));
       expect(backedUp).toHaveLength(1);
       expect(backedUp[0]).toMatchObject({ weight: 250 }); // the OLD row, preserved in the backup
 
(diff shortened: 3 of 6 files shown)


brownm09:
## Review: Phase 2B: TM-history seeding script (seed-tm-history.ts)

### Summary

This PR is ready to merge. It builds the write path for a one-time, RLS-bypassing production data migration (Phase 2B TM-history seeding), and the diff (1,828 lines against `main`) went through parallel correctness/security and reliability/performance/maintainability review given its size. The review surfaced 19 distinct findings, the most serious being a **package-lock.json drift** that would have failed `npm ci` in CI, a **missing cross-row invariant check** (a CSV mixing more than one program's rows would silently write everything under the single `--program` argument), and an **unguarded `--restore` zero-entries path** (a truncated backup file would silently wipe history rather than fail). **All 19 findings have been fixed in-PR** across three follow-up commits (`f2e5da8`, `456064a`, `695af47`) — see the disposition table below for each finding's fix location. Two genuinely out-of-scope items were filed rather than fixed here (an infra typecheck/lint coverage gap, and a cross-repo question that needs live Google Sheets access this session doesn't have).

---

### Blocking Findings

**[correctness]** `apps/api/scripts/seed-tm-history.ts` — CSV rows spanning more than one `Program` value were never re-checked against the downloaded CSV
The script reads `--program` from the CLI and never validates it against the CSV's own `Program` column — by design, since that column is meant to be advisory, not authoritative. But nothing rejected a CSV whose `Will Seed=true` rows span more than one distinct `Program` value, the same invariant gas-lifting-logbook's own pre-flight validator already treats as blocking.
That validator runs against the live Sheet tab, not this downloaded CSV, so nothing guarantees the two are the same vintage — the same reasoning the file's own `mapRowToEntry` doc comment already gives for re-validating `TM` per-cell, just not extended to this cross-row invariant. `training_max_history` has no unique constraint, so a mixed-program CSV would insert cleanly, report a plausible count, and silently corrupt the target program's history.
**Fix:** `assertSingleProgram` added and called before mapping (commit `456064a`).

**[correctness]** `apps/api/scripts/seed-tm-history.ts` — no re-check for a `(lift, cycle date)` collision among `Will Seed=true` rows
Same gap as above, for gas-lifting-logbook's other Step 2 blocking check: a duplicate `(program, lift, cycle date)` in the source data. `training_max_history` has no unique constraint, so a collision inserts two rows for the same cycle rather than erroring.
**Fix:** `assertNoDuplicateSeedKeys` added, checked on the *resolved* lift id (post-`resolveLiftId`) so two differently-spelled raw names that resolve to the same target lift are still caught (commit `456064a`).

**[correctness]** `apps/api/scripts/seed-tm-history.ts` — `--restore` had no zero-entries guard
The seed path refuses a zero-row write outright (a deliberate, already-tested guard against a whole-file format mismatch producing a silent full wipe). `--restore` shared none of that protection: a truncated or corrupt backup file containing `[]` would proceed straight into `deleteAllHistory` and append nothing, reporting a successful-looking "Restored — wrote 0 row(s)."
**Fix:** the zero-entries refusal now applies to both `seed` and `restore` modes (commit `456064a`).

**[correctness]** `apps/api/scripts/seed-tm-history.ts` — backups carried no provenance, and `--restore` performed no per-entry validation
`backupFilePath` embeds the program UUID in the *filename*, but nothing checked that on the way back in — the JSON payload itself carried no `program`/`userId` at all. Restoring program A's backup into program B would have succeeded silently whenever both belong to the same `--user-id` (the ownership check only proves the caller owns the *target*), on the one command an operator reaches for precisely because something already went wrong. Entry contents were also unvalidated, unlike the seed (CSV) path, which rejects empty/zero/negative/non-finite `TM`, non-ISO dates, and unrecognized booleans.
**Fix:** backups are now a provenance-checked envelope (`{program, userId, capturedAt, entries}`); `readBackupFile` refuses to restore across a different `--program` or `--user-id`, and validates every entry's shape before it reaches the transaction (commit `456064a`).

**[security]** `apps/api/scripts/seed-tm-history.ts` — backup directory/file used default permissions at a predictable, world-readable-by-default location
The default backup location (`os.tmpdir()`) is a complete dump of one user's training-max history keyed to their Clerk id (per the PR's own `.gitignore` comment), written with default `mkdirSync`/`writeFileSync` permissions. On a shared multi-user Linux host — a bastion, jump host, or CI runner, all plausible break-glass venues — `os.tmpdir()` is typically world-readable `/tmp`, and a pre-planted symlink at the fixed subdirectory name could redirect the write entirely (CWE-377/CWE-732).
**Fix:** directory created `0o700`, file written `0o600`, and the function refuses to proceed if the resolved backup directory turns out to be a symlink (commit `456064a`). `mode` is a no-op on Windows but real hardening on Linux/macOS, where the exposure applies.

**[correctness]** `apps/api/scripts/seed-tm-history.ts` — the `--force`-required refusal message was written only for `seed` mode, and the runbook's documented `--rollback` command would have always failed
The gate fires whenever pre-existing history exists, for every mode — but a meaningful `--rollback`/`--restore` *always* has pre-existing history (that's the precondition), so the seed-oriented message ("unexpected at the normal Phase 2B timing — double-check `--program`") was actively misleading on the one path where existing history is normal, not anomalous. Separately, gas-lifting-logbook's runbook documented a `--rollback` command with no `--force` at all — as written, it would fail 100% of the time.
**Fix:** the refusal message is now mode-aware (commit `456064a`); the runbook fix (add `--force`, document `--restore`) is [gas-lifting-logbook#31](https://github.com/brownm09/gas-lifting-logbook/pull/31), tracked via [gas-lifting-logbook#30](https://github.com/brownm09/gas-lifting-logbook/issues/30).

**[reliability]** `apps/api/package.json` / `package-lock.json` — `csv-parse` recorded under the wrong dependency section in the lockfile
`package.json` declares `csv-parse` under `devDependencies`, but the committed lockfile still recorded it under `apps/api`'s `dependencies` — the exact drift this repo's own lockfile-sync gate (`npm install --package-lock-only --ignore-scripts && git diff --exit-code package-lock.json`) exists to catch, and the failure mode that has previously reddened `main` here (#432/#433).
**Fix:** lockfile regenerated; the sync check now exits clean (commit `456064a`).

**[reliability]** `apps/api/scripts/seed-tm-history.ts` — the delete+append transaction had no explicit timeout budget
A bare `prisma.$transaction()` inherits Prisma's default interactive-transaction budget (5s timeout / 2s maxWait). This codebase already found that default too tight for a large batch write over a *normal* connection (`IMPORT_TX_TIMEOUT_MS`, #532) — this script's break-glass connection (a manually-proxied Cloud SQL connection from an operator's laptop) has strictly higher latency than that already-too-tight path. A `P2028` here is non-destructive (the transaction rolls back) but aborts a break-glass window that costs real setup/teardown to reopen.
**Fix:** the write now goes through `runInteractive` with `IMPORT_BATCH_TX_OPTIONS` (`prisma-tx.util.ts`) — this codebase's existing, sanctioned wrapper for exactly this call shape (commit `456064a`).

---

### Non-Blocking Findings

**[maintainability]** `apps/api/scripts/seed-tm-history.ts` — the raw `.$transaction()` call bypassed this repo's custom `no-direct-prisma-transaction` ESLint rule
The rule exists specifically to forbid raw `.$transaction()` calls outside three allowlisted files, routing callers through `runBatch`/`runInteractive` instead — but its effective scope (`apps/api`'s `lint` script is `eslint src`) doesn't reach `apps/api/scripts/`, so it never fired here regardless.
**Fix:** the code now uses `runInteractive` regardless of whether the lint rule can see this path (commit `456064a`); the path-scope gap itself is covered by the finding below.

**[maintainability]** `apps/api/scripts/` has no typecheck or lint coverage in CI
`tsconfig.json`'s `include`/`rootDir` and `package.json`'s `lint` script both scope to `src/` only — ~1,800 lines of new code, including the only code in this repo authorized to write with RLS bypassed, get no build-time typecheck and no lint pass. Verified manually for this PR (`nest build` after building `packages/core`/`packages/types`; `eslint scripts/...` run directly).
**Suggestion:** filed as [merickvaughn/lifting-logbook#887](https://github.com/merickvaughn/lifting-logbook/issues/887) — genuinely separate scope (a build/CI config change, not specific to this script).

**[reliability]** `apps/api/scripts/seed-tm-history.ts` — CSV parsing had no BOM handling and didn't trim header names
The natural way to eyeball this file before running the script (the runbook's Step 3) is opening it in a spreadsheet tool, which reliably reintroduces a UTF-8 BOM on save — `csv-parse` leaves it attached to the first header cell, turning `Program` into an invisible-byte mismatch that "missing expected column" doesn't explain, and re-exporting can't fix.
**Fix:** `bom: true` passed to `csv-parse`, headers trimmed before the `REQUIRED_COLUMNS` comparison (commit `456064a`).

**[reliability]** `apps/api/scripts/seed-tm-history.ts` — row-validation errors carried no row context and stopped at the first bad row
`Row has an invalid TM: "not-a-number"` gives an operator no way to locate the offending row in a multi-hundred-row file without a manual scan; failing on the first bad row also means a file with several bad rows costs several full remediation round trips.
**Fix (partial):** the seed-time call site now wraps each row and adds a 1-based row number, `Lift`, and `Cycle Date` to the thrown message (commit `456064a`). **Deliberately not fixed:** aggregating every bad row into one report instead of failing on the first — a real but smaller UX gain, cut for scope in this PR.

**[reliability]** `apps/api/scripts/seed-tm-history.ts` — the backup's read-back verification didn't prove durability, and the default location is purged on a schedule
A same-process read after `writeFileSync` can be served from the page cache — it proves the bytes are readable, not that they survived a crash, for what's described as the sole recovery artifact if the write transaction itself fails. Separately, `os.tmpdir()` (the default backup location) is purged on a schedule or at reboot on most platforms.
**Fix:** `fsyncSync` added before the read-back check; a console note now fires when a backup lands in the default location, recommending `--backup-dir` or a manual copy (commits `456064a`, `695af47`).

**[reliability]** `apps/api/scripts/seed-tm-history.ts` — `--dry-run` skipped the one filesystem dependency its own "full rehearsal" claim implies it covers
The dry-run early return sits above both the `--force` gate and the backup step, so it never creates the backup directory or checks it's writable — the real run's sole recovery-artifact dependency was exactly what a rehearsal would otherwise never exercise.
**Fix:** dry-run now probes backup-directory writability (via `mkdirSync`+`accessSync`) whenever there's existing history that a real run would need to back up (commit `456064a`).

**[maintainability]** `apps/api/scripts/seed-tm-history.ts` — `getHistory` was called three times for what is conceptually one read
`verifyProgramOwnership`'s cross-user anomaly check, `runSeed`'s `existingCount`, and `backupExistingHistory`'s own internal fetch each independently queried history — so the reported count, the backup's contents, and the eventual delete were three different snapshots in principle, for what's supposed to be one atomic before/after picture.
**Fix:** `backupExistingHistory` now takes the already-fetched `existing` array directly instead of a repository + re-fetching; `runSeed` fetches history exactly once and passes the same array to both the reported count and the backup (commit `456064a`). This also made the function DB-free, so its tests moved from e2e-only to `seed-tm-history.spec.ts`.

**[maintainability]** `apps/api/scripts/seed-tm-history.ts` — `runSeed`'s `backupDirOverride` test-only parameter took precedence over the user-supplied `--backup-dir`
Every e2e test relied on the override, so the production resolution chain (`--backup-dir` flag → `DEFAULT_BACKUP_DIR`) had no end-to-end coverage at all; the precedence itself was also inverted from what an operator would expect (a backdoor beating an explicit flag).
**Fix:** the parameter removed; tests now set `backupDir` directly in the `SeedArgs` they already construct (commit `456064a`).

**[maintainability]** `apps/api/scripts/seed-tm-history.ts` — `--restore <path>` wasn't existence-checked before connecting to the DB
`--input` already fails fast on a missing path before any DB round trip; `--restore`'s path (also operator-typed, also the panic-path argument where a fast, unambiguous error matters most) skipped that same check.
**Fix:** the existence check now covers both `--input` and `--restore` (commit `456064a`).

**[maintainability]** `apps/api/scripts/seed-tm-history.ts` — no `--help`/`-h`
Both fell through to "Unrecognized flag" with no usage output, on a CLI where the difference between two flag spellings is the difference between a preview and a destructive write.
**Fix:** `--help`/`-h` now print the header doc's usage block, handled in `main()` before `parseArgs` is called (commit `456064a`).

**[maintainability]** `apps/api/scripts/seed-tm-history.ts` — `parseArgs`'s seed branch kept an `as string` cast the discriminated-union refactor was introduced to eliminate
The adjacent `restore` branch already narrowed `restorePath` properly (`if (restorePath !== undefined)`); the `seed` branch used `input as string` instead, one line away from the pattern it should have mirrored.
**Fix:** narrowed the same way (commit `456064a`).

---

### Findings Disposition

| # | Finding | Category | Disposition |
|---|---|---|---|
| 1 | Multi-program CSV not re-validated | correctness | Fixed in `456064a` |
| 2 | Duplicate `(lift, date)` seed key not re-validated | correctness | Fixed in `456064a` |
| 3 | `--restore` missing zero-entries guard | correctness | Fixed in `456064a` |
| 4 | Backup provenance / restore entry validation | correctness | Fixed in `456064a` |
| 5 | Backup dir/file permissions, symlink guard | security | Fixed in `456064a` |
| 6 | `--force` message not mode-aware; runbook `--rollback` always fails | correctness | Fixed in `456064a` (script) + [gas-lifting-logbook#31](https://github.com/brownm09/gas-lifting-logbook/pull/31) (runbook) |
| 7 | `csv-parse` lockfile section drift | reliability | Fixed in `456064a` |
| 8 | No transaction timeout budget | reliability | Fixed in `456064a` |
| 9 | Bypasses `no-direct-prisma-transaction` rule (path scope) | maintainability | Fixed in `456064a` (uses `runInteractive`); scope gap tracked in #887 |
| 10 | `apps/api/scripts/` no typecheck/lint coverage | maintainability | Filed: [#887](https://github.com/merickvaughn/lifting-logbook/issues/887) — genuinely separate scope |
| 11 | CSV BOM / header trim | reliability | Fixed in `456064a` |
| 12 | No row context in validation errors; fails on first bad row | reliability | Partially fixed in `456064a` (row context added); full aggregation deliberately deferred — noted in commit message |
| 13 | Backup fsync + tmpdir purge warning | reliability | Fixed in `456064a`, `695af47` |
| 14 | `--dry-run` skips backup-dir writability probe | reliability | Fixed in `456064a` |
| 15 | Triple-fetch of existing history | maintainability | Fixed in `456064a` |
| 16 | `backupDirOverride` test backdoor, inverted precedence | maintainability | Fixed in `456064a` |
| 17 | `--restore` path not existence-checked before DB connect | maintainability | Fixed in `456064a` |
| 18 | No `--help`/`-h` | maintainability | Fixed in `456064a` |
| 19 | Surviving `as string` cast in `parseArgs` | maintainability | Fixed in `456064a` |

Two additional items surfaced during this review with no code fix in this PR, both cross-repo:
- **Google Sheets boolean CSV-export case** (whether a boolean-looking cell renders lowercase or uppercase) — this script already defends both ways; the open question is whether the already-merged gas-lifting-logbook validator has the same gap. Filed: [gas-lifting-logbook#29](https://github.com/brownm09/gas-lifting-logbook/issues/29) (needs live Sheets access this session doesn't have).
- `describe.skip` on `OWNER_DATABASE_URL` absence (`seed-tm-history.db.e2e.spec.ts`) and the two `?? null` occurrences in `parseArgs`/`runSeed` were checked against this repo's suppression/test-integrity policies and confirmed **not violations** — the former matches the identical, already-established `rls.db.e2e.spec.ts`/`programs.db.e2e.spec.ts` convention; the latter normalize `string | undefined` to a declared `string | null` field type, not silence a type error.

**Documentation / ADR checks:** no `Documentation Maintenance` table exists in this repo's `CLAUDE.md` (checked at both `.claude/CLAUDE.md` and root `CLAUDE.md`) — doc-reconciliation not applicable. No README exists at any ancestor directory of the changed files (`apps/api/scripts/`, `apps/api/`, `apps/`) — no stale-README blocking finding; a `apps/api/scripts/README.md` would add navigational value given the directory now holds a break-glass production write script alongside `check-globalsetup-failure-paths.js`, but that's a suggestion, not a blocker. None of ADR-011's four criteria apply (no `claude/`-equivalent config, hook, or cross-project workflow rule touched) — ADR not warranted.

---

PR #889, opened 2026-08-10
[chore] Add typecheck and lint coverage for apps/api/scripts/

### LINKED ISSUE #887: apps/api/scripts/ has no typecheck or lint coverage
## Summary

Surfaced during `/review` of [#886](https://github.com/merickvaughn/lifting-logbook/pull/886) (the Phase 2B seed script, `apps/api/scripts/seed-tm-history.ts`).

`apps/api/tsconfig.json` sets `"rootDir": "src"` / `"include": ["src/**/*.ts"]`, and `apps/api/package.json`'s `"lint": "eslint src"`. Neither covers `apps/api/scripts/` — `nest build` never type-checks anything there, and `eslint` never lints it. The directory currently holds two files: the pre-existing `check-globalsetup-failure-paths.js` and the new `seed-tm-history.ts` (plus its two spec files, which *do* get incidental type-checking via ts-jest compiling them — but only while the DB e2e half isn't skipped, e.g. via `LIFTING_SKIP_DB_E2E=1` or no Docker).

`seed-tm-history.ts` imports across the package boundary into `../src/ports/...` and `../src/adapters/prisma/...` — interfaces CI actively refactors, with no signal today if a change there breaks the script. Given it runs during a break-glass production procedure, the first sign of a break would otherwise be an operator hitting it live.

## Suggested fix

Add `scripts/**/*.ts` to a type-check target — a dedicated `tsconfig.scripts.json` referenced by a new `typecheck` script is cleanest given the existing `rootDir: "src"` constrains build output — and widen `"lint"` to `eslint src scripts`.

## Scope

Verified manually for #886 (`nest build` after building `packages/core`/`packages/types`, plus `npx eslint scripts/...` run directly) — this issue is about making that coverage permanent/automatic, not about #886 itself being unverified.

brownm09:
## Summary

`apps/api/tsconfig.json` sets `"rootDir": "src"` / `"include": ["src/**/*.ts"]`, and `apps/api/package.json`'s `"lint": "eslint src"` — so `apps/api/scripts/` was invisible to both `nest build`'s type-checking and ESLint. Surfaced during `/review` of #886, which added `apps/api/scripts/seed-tm-history.ts` — a break-glass production seeding script that imports directly across the package boundary into `apps/api/src/adapters/prisma/*` (code CI actively refactors), with no automated signal if a refactor there silently breaks the script.

## What's in this PR

- **New `apps/api/tsconfig.typecheck.json`** + `"typecheck": "tsc --noEmit -p tsconfig.typecheck.json"` script — `apps/api` was the only workspace besides `apps/mobile` with no `typecheck` script at all. Resolves `@lifting-logbook/core`/`@lifting-logbook/types` (plus their own internal `@src/core`/`@src/core/*` aliases) to source, not `dist/`, so the task stays buildless — mirroring the existing `packages/core/tsconfig.spec.json` / `packages/api-client/tsconfig.typecheck.json` / `apps/web/tsconfig.typecheck.json` convention (issues #651/#656).
- **`apps/api/package.json`**: `"lint": "eslint src"` → `"lint": "eslint src scripts"`.
- **`eslint.config.js`**: two changes —
  - Exclude the one pre-existing `.js` file in `scripts/` (`check-globalsetup-failure-paths.js`, a standalone CommonJS `require.cache`-clearing test harness never wired into any pipeline) from lint scanning. `@typescript-eslint/no-require-imports` fires on `require()` regardless of file extension — confirmed empirically before this PR (3 errors on its 3 `require()` calls). Rewriting it to ESM was considered and rejected: nothing in the chain sets `"type": "module"`, `tsconfig.base.json` pins `module: "commonjs"`, and the file's `require.cache`-clearing technique for dynamic test-harness reloading has no clean ESM equivalent — it would risk breaking the script's actual runtime execution (`node apps/api/scripts/check-globalsetup-failure-paths.js`), which is out of scope here. Matches the repo's own existing precedent for this exact class of problem: `tools/eslint-rules/package.json`'s `"lint": "echo \"(no lint for eslint-rules workspace)\" && exit 0"`. The ignore is a directory glob, not the exact filename, so a future plain-`.js` utility dropped in the same directory doesn't silently reproduce the same break.
  - Widen the `no-direct-prisma-transaction` custom rule's `files` glob to also cover `apps/api/scripts/**/*.ts`. Today it's scoped to `apps/api/src/adapters/prisma/**/*.ts` only, which structurally can't match anything under `scripts/` — restoring this closes the exact regression-safety gap #886's own manual review already had to catch by hand once (an earlier version of `seed-tm-history.ts` called `.$transaction()` directly before being fixed to use the allowlisted `runInteractive` helper). Zero violations today; verified live by temporarily reintroducing a raw `.$transaction()` call and confirming lint now catches it (then reverted).
- **`turbo.json`**: added `scripts/**` to the `lint` task's `inputs` allowlist. This one wasn't in the issue text — `lint.inputs` is an explicit allowlist (unlike `typecheck`, which has no `inputs` override), so without this addition Turbo's lint cache would never invalidate on `scripts/`-only changes even after the `lint` script itself is fixed.

### Deviations from the issue's literal suggested fix

The issue suggested `tsconfig.scripts.json` with a **whole-workspace** include (`src/**` + `scripts/**`), matching every other workspace's typecheck precedent. Two changes from that, both discovered during implementation:

1. **Renamed to `tsconfig.typecheck.json`** — matches the *dominant* convention (2 of 3 precedents: `apps/web`, `packages/api-client`); `packages/core`'s `tsconfig.spec.json` name is justified there by genuine dual-purpose use (it's also that package's ts-jest transform config) — a condition that doesn't hold for `apps/api` (`apps/api/jest.config.js` already pins ts-jest's transform to plain `tsconfig.json`).
2. **Scoped to `scripts/**/*.ts` only, not whole-workspace** — a whole-workspace include surfaced **31 pre-existing, unrelated type errors across 12 files** in `src/**/*.spec.ts` (test-mock shapes trailing behind interface changes, `exactOptionalPropertyTypes` friction, Fastify `.inject()` typing drift) that `nest build` has always silently skipped via `tsconfig.build.json`'s `**/*.spec.ts` exclusion. Fixing those would have been a large, unrelated scope expansion well past this issue's actual ask — tracked instead in #890, with the full list. TypeScript still fully checks whatever `scripts/` actually imports from `src/`/`packages/*`, transitively — this file's `rootDir`/`paths` are unchanged from the original design, only `include` narrowed.

One error **did** land inside the narrowed scope and needed fixing: `seed-tm-history.db.e2e.spec.ts`'s `new PrismaClient({ datasources: { db: { url: OWNER_DATABASE_URL } } })` — `OWNER_DATABASE_URL: string | undefined`, and `exactOptionalPropertyTypes` rejects it. `describeOrSkip` already guarantees this code only runs when `OWNER_DATABASE_URL` is truthy, a runtime invariant TypeScript can't see across the closure boundary — fixed with an explicit runtime guard (`if (!OWNER_DATABASE_URL) throw new Error(...)`) rather than a non-null assertion, so a future accidental break of that invariant fails with a clear message instead of a generic Prisma error. The identical unaddressed pattern (a bare, unchecked `string | undefined`) still exists in the already-merged `apps/api/src/programs/programs.db.e2e.spec.ts:154` — confirming this predates this branch; now also covered by #890.

## Testing

```
Tests: 599 passed, 0 skipped, 0 failed (apps/api, full monorepo `npm test` run — Docker available, real Testcontainers-backed DB e2e included)
Tests: 27 passed (api-client), 322 passed (core), 295 passed (web) — full monorepo run, all clean
```

- `cd apps/api && npx tsc --noEmit -p tsconfig.typecheck.json` — clean.
- `cd apps/api && npm run lint` — clean (confirmed via `--debug` that the 3 `.ts` files are actually parsed/linted, not silently skipped, while the 1 `.js` file is correctly excluded).
- `npm run typecheck` / `npm run lint` from repo root (Turbo fan-out, all 7 workspaces) — `@lifting-logbook/api` now participates in `typecheck`; all workspaces clean, no regressions.
- `npm test` from repo root — full clean pass, all 12 turbo tasks successful (summary above).
- After addressing review findings: `apps/api` typecheck re-verified clean, root lint fan-out re-verified 7/7 clean, `scripts/` tests re-run against real Postgres (Docker) — 2/2 suites, 88/88 tests passed, confirming the new explicit runtime guard (replacing the non-null assertion) actually executes correctly, not just compiles.
- Live-fire check: temporarily reintroduced a raw `.$transaction()` call in `scripts/seed-tm-history.ts`, confirmed `lifting-logbook/no-direct-prisma-transaction` now fires, reverted.
- Suppression grep (repo policy Rule 3) + manual scan of every `!` in the diff — **zero suppressions** (the review-findings fix replaced the one non-null assertion with an explicit guard; the only remaining `!` is a plain boolean negation, not a type assertion).
- Test-integrity grep (skip markers, deleted tests, lowered thresholds) — clean.
- Lockfile sync check (`npm install --package-lock-only --ignore-scripts` + `git diff --exit-code package-lock.json`) — clean, no dependency changes in this PR.
- `node scripts/check-board-id-sync.mjs` — clean (unrelated to this change).

No `## Testing` section gap — this repo's CLAUDE.md documents `npm test` as the command; ran in full.

Full review: [PR comment](https://github.com/merickvaughn/lifting-logbook/pull/889#issuecomment-5243721723) (`blocking=0 non_blocking=3`). All 3 findings fixed in-PR (commit `eda3e81`):

| Finding | Disposition |
|---|---|
| Pre-existing `src/**/*.spec.ts` type errors (discovered but deliberately out-of-scope) had no tracking issue | Fixed `eda3e81` — filed [#890](https://github.com/merickvaughn/lifting-logbook/issues/890) with the full enumerated list (31 errors, 12 files), referenced from the tsconfig comment |
| Non-null assertion (`OWNER_DATABASE_URL!`) is a silent failure mode if the `describeOrSkip` invariant is ever broken | Fixed `eda3e81` — replaced with an explicit `if (!OWNER_DATABASE_URL) throw new Error(...)` guard; diff now has zero suppressions |
| `eslint.config.js` comment didn't explain "directory glob, not exact filename" rationale | Fixed `eda3e81` — folded into the code comment |

Closes #887


brownm09:
## Review: [chore] Add typecheck and lint coverage for apps/api/scripts/

### Summary
Ready to merge with minor changes. This is a well-scoped, thoroughly-verified tooling PR — no blocking findings. The highest-value finding is non-blocking: the PR deliberately excludes ~10 pre-existing type errors it discovered in `src/**/*.spec.ts` from its new typecheck gate (a well-justified scope decision), but doesn't leave a trackable record of what those errors are, so that discovery risks being lost. Diff is small (66 lines) and appropriately scoped to a single concern.

---

### Non-Blocking Findings

**[maintainability]** `apps/api/tsconfig.typecheck.json:1-9` (comment)
This PR discovered roughly a dozen pre-existing TypeScript errors in `src/**/*.spec.ts` (test-mock shapes trailing behind interface changes) that a whole-workspace typecheck would have surfaced, and deliberately narrowed `include` to `scripts/**/*.ts` to avoid fixing them here. The only record of this discovery is prose in a tsconfig comment and the PR body — neither is searchable or actionable once this PR merges and the description scrolls out of recent activity. Nobody will know this debt exists to schedule fixing it.
**Suggestion:** File a tracking issue enumerating the specific errors (file:line + error code) found during this PR's investigation, and reference it from the tsconfig comment in place of (or alongside) "see the PR body."

**[maintainability]** `apps/api/scripts/seed-tm-history.db.e2e.spec.ts:99-105`
The non-null assertion (`OWNER_DATABASE_URL!`) is correct and safe — `describeOrSkip` guarantees the surrounding code only runs when the value is truthy, and `OWNER_DATABASE_URL` is an immutable `const` so that guarantee holds across the closure boundary. But if that invariant were ever accidentally broken by a future refactor of `describeOrSkip`, the failure mode is a bare assertion telling TypeScript "trust me," so at runtime it degrades to whatever generic error Prisma throws for a `datasources.db.url: undefined` (a `PrismaClientInitializationError` with no context pointing back at the real cause), rather than a clear, on-point message.
**Suggestion:** An explicit guard (`if (!OWNER_DATABASE_URL) throw new Error('OWNER_DATABASE_URL must be set — describeOrSkip should have skipped this suite otherwise')`) would fail with a much clearer diagnostic if the invariant is ever violated, at the cost of one more line. Not required — the current fix is correct, and mirrors the existing (currently unchecked) pattern in `programs.db.e2e.spec.ts:154`, so consistency with that file's implicit style is also a reasonable reason to leave it as-is.

**[maintainability]** `eslint.config.js:14`
The comment explaining the `.js` ignore entry documents *why* `check-globalsetup-failure-paths.js` needs excluding, but not why the ignore is a directory glob (`apps/api/scripts/**/*.js`) rather than the exact filename — that reasoning ("so a future plain-.js utility dropped in the same directory doesn't silently reproduce the same break") only appears in the PR body, which won't be visible to someone reading this file in six months.
**Suggestion:** Fold one clause of that rationale into the code comment itself.

---

### Style Notes
No issues beyond what's captured above — comment density and JSON formatting in the new `tsconfig.typecheck.json` match the established style of the precedent files it mirrors (`apps/web/tsconfig.typecheck.json`, `packages/core/tsconfig.spec.json`).

---

### Findings Disposition
All three findings above will be addressed in this PR (fix comments + file a tracking issue) before merge, per this repo's "fix everything, always" merge discipline — disposition to be recorded in the PR body once applied.

---

brownm09:
## Review: [chore] Add typecheck and lint coverage for apps/api/scripts/ (re-review after findings fix)

### Summary
Ready to merge. This re-review covers commit `eda3e81`, which addresses all three non-blocking findings from the prior review pass ([earlier comment](https://github.com/merickvaughn/lifting-logbook/pull/889#issuecomment-5243721723)). All three fixes are correct and verified: the tracking issue (#890) is filed with the full enumerated error list, the non-null assertion is now an explicit guard whose control-flow narrowing was re-verified to compile cleanly, and the eslint comment gap is closed. No new findings from this pass — the incremental diff introduces nothing beyond the three targeted fixes.

---

### Findings Disposition
No outstanding findings — all three from the prior pass are resolved in this PR (commit `eda3e81`) and confirmed on re-review. Nothing further to disposition.

---

